PR #102, opened 2015-10-05
Typical IDEA flotsam and jetsam

elharo:
@patflynn @loosebazooka Not intended for checkin, but just as an FYI, this is what IDEA routinely adds to my working directory simply by importing build.gradle and running a unit test.

Any suggestions for managing this? Could we add any of these files to .gitignore? 
Is there something I should change abiout my IntelliJ config?
Should any of this be checked in?


loosebazooka:
can we close this?

PR #103, opened 2015-10-06
Add functionality to display and allow the user to accept the usage tracking for Cloud Tools plugin (enabled only in IntelliJ, asked only once when plugin is first used; create a tab in Settings dialog that users can change configuration)  

nbashirbello:
@patflynn @elharo @loosebazooka 
Will add functionality to pop up notification the first time plugin is installed.

### DIFF
--- a/google-login-plugin/src/com/google/gct/login/stats/UsageTrackerManager.java
+++ b/google-login-plugin/src/com/google/gct/login/stats/UsageTrackerManager.java
@@ -18,4 +18,11 @@ public class UsageTrackerManager {
         return PropertiesComponent.getInstance().getBoolean(USAGE_TRACKER_KEY, false);
     }
 
+    /**
+     * @return true if the user has opted in/out of usage tracking; false otherwise
+     */
+    public static boolean isTrackingConfigured() {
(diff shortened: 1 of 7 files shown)


google-login-plugin/src/com/google/gct/login/stats/UsageTrackerManager.java
```diff
@@ -0,0 +1,28 @@
+package com.google.gct.login.stats;
+
+import com.intellij.ide.util.PropertiesComponent;
+
+/**
+ * Stores the users choice to opt in/out of sending usage metrics via the Google Usage Tracker.
```
patflynn: This TOS story is actually going to have to be general to all Google plugins in IntelliJ. As it's architected it can't be specific to our cloud tools plugin

patflynn:
All of our TOS stuff needs to be disabled in Android Studio since we don't collect metrics directly there.

patflynn:
Is this PR an old version of https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/124?

can we abandon this one?

nbashirbello:
Yes, this is the one on my personal branch before we went public. I'll abandon it.

PR #107, opened 2015-10-07
turn on GoogleLoginPrefsTest

elharo:
@patflynn @nbashirbello

### DIFF
--- a/google-login-plugin/testSrc/com/google/gct/login/GoogleLoginPrefsTest.java
+++ b/google-login-plugin/testSrc/com/google/gct/login/GoogleLoginPrefsTest.java
@@ -15,42 +15,34 @@
  */
 package com.google.gct.login;
 
-import junit.framework.TestCase;
-
-import java.io.*;
-
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
 
 /**
- * Tests for {@link GoogleLoginPrefs}
+ * Tests for {@link GoogleLoginPrefs}.
  */
-
-//public class GoogleLoginPrefsTest extends TestCase {
 public class GoogleLoginPrefsTest {
-  // The required permission for the preference file
-  private static final String PREFERENCE_FILE_PERMISSION = "-rw-r-----";
 
-  /**
-   * Tests that the Google Login credentials file is only accessible by the user.
-   */
-  // TODO: when upgraded to Java7, use java.nio.file.Files.getPosixFilePermissions
-  public void testPreferenceFileAccessibility() throws IOException {
-    String preferencesPath = GoogleLoginPrefs.getPreferencesPath();
-    File userRootDir = new File(System.getProperty("java.util.prefs.userRoot",
-      System.getProperty("user.home")), ".java/.userPrefs");
-    String absolutePrefPath = userRootDir.getAbsolutePath() + preferencesPath + "/prefs.xml";
+  private String oldPreferencesPath;
 
-    // Get the preference file's permissions
-    Process process = Runtime.getRuntime().exec("ls -al " + absolutePrefPath);
+  @Before
+  public void setUp() {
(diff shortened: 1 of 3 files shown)


elharo:
@patflynn @loosebazooka The more I look at this test the more broken it seems.

1. It's really testing the umask on the user's system, nothing in our code at all.
2. It's Unix specific.
3. It assumes the Preferences class is file based, which is not necessarily so. 

I'd like to repolace it with a very different test that doesn't assume any of these things. What do you think?

elharo:
Ping

nbashirbello:
You can replace the test. Like I mentioned earlier the Privacy group just wanted to have a test for login to check that only the user has access to the preferences.

elharo:
Is than an LGTM?

google-login-plugin/testSrc/com/google/gct/login/GoogleLoginPrefsTest.java
```diff
@@ -17,40 +17,16 @@
...
-    BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream));
-    String line = reader.readLine();
-    inputStream.close();
-    String[] splitOutput = line.split(" ");
-
-    // Check the output
-    //assertTrue(splitOutput.length > 1);
-    //assertEquals(PREFERENCE_FILE_PERMISSION, splitOutput[0]);
+  public void testActiveUser() {
+    GoogleLoginPrefs.saveActiveUser("cheese");
```
nbashirbello: I think we need  to override the GoogleLoginPrefs.PREFERENCES_PATH so you can have a different preference file during testing.

nbashirbello: @elharo Why do you think preferences are not stored in a file? I remember I verified this for Unix, Windows and Mac.

elharo: I can do this. It's non-trival but may be worthwhile. I think it requires a major refactoring of GoogleLoginPrefs so it's non-static and can take an injected Preferences object. Should we consider this?

Also, what, is anything, is our pledge on API compatiblity here? That is, do I have to worry about breaking thrid party clients?

elharo: http://docs.oracle.com/javase/7/docs/api/java/util/prefs/Preferences.html "A node in a hierarchical collection of preference data. This class allows applications to store and retrieve user and system preference and configuration data. This data is stored persistently in an implementation-dependent backing store. Typical implementations include flat files, OS-specific registries, directory servers and SQL databases. The user of this class needn't be concerned with details of the backing store."

nbashirbello: Would this work:

PREFERENCES_PATH = ApplicationManager.getApplication().isUnitTestMode() ? "/com/google/gct/login" : "/com/google/gct/login/test";

and we can delete test directory after.

nbashirbello: This is from back when I researched the preferences, " In Linux, the user’s preferences file is stored in their home directory and in Windows, it uses the registry and in Mac, the user’s preferences file is stored in their home directory."

nbashirbello:
I am fine with just having testActiveUser if we can't test the accessibility of the user preference file.

elharo:
I think 


PREFERENCES_PATH = ApplicationManager.getApplication().isUnitTestMode() ? "/com/google/gct/login" : "/com/google/gct/login/test";

would work. However it's a little icky because it mixes test code into the model code and depeneds on global static state. An alternative is to add a VisibleForTesting setprefernecesPath method, although this is still all static.

nbashirbello:
ok

elharo:
done

nbashirbello:
LGTM

PR #110, opened 2015-10-07
Remove google feedback for android studio

patflynn:
I got a hit on the forums for a way to dynamically add the Google Feedback error reporter extension.  
https://devnet.jetbrains.com/thread/473190?tstart=30

Tested it out and it seems legit.

### DIFF
--- a/common-lib/src/main/java/com/google/gct/idea/feedback/FeedbackUtil.java
+++ b/common-lib/src/main/java/com/google/gct/idea/feedback/FeedbackUtil.java
@@ -1,0 +1,22 @@
+package com.google.gct.idea.feedback;
+
+import com.intellij.ExtensionPoints;
+import com.intellij.ide.plugins.PluginManager;
+import com.intellij.openapi.extensions.Extensions;
+import com.intellij.openapi.extensions.PluginId;
+
+/**
+ * Handy plugin related methods for Google Feedback integration.
+ */
+public final class FeedbackUtil {
+
+  private FeedbackUtil() {}
+
+  public static void enableGoogleFeedbackErrorReporting(String pluginId) {
+    GoogleFeedbackErrorReporter errorReporter = new GoogleFeedbackErrorReporter();
+    errorReporter
+        .setPluginDescriptor(PluginManager.getPlugin(PluginId.getId(pluginId)));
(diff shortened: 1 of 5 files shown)


common-lib/src/main/java/com/google/gct/idea/feedback/FeedbackUtil.java
```diff
@@ -0,0 +1,23 @@
...
+import com.intellij.openapi.extensions.PluginId;
+
+/**
+ * Handy plugin related methods for Google Feedback integration.
+ */
+public final class FeedbackUtil {
+
+  private FeedbackUtil() {}
+
+  public static void enableGoogleFeedbackErrorReporting(String pluginId) {
```
elharo: I don't know if this will work, but I would try to write a test that calls this and see what happens. You might further be able to check that the error reporter is registered. May not work, but worth a shot.

patflynn: Yeah sadly it doesn't work.  I'll come back round to this in a few days as soon as I've synced core and login into AOSP.  I want to get my head around testing and have a clear strategy for managing collaborators and making our code more testable and tested, but I don't want to block on that to get this in, since it's needed for the sync to AOSP, and I've manually verified that it works.

google-login-plugin/src/com/google/gct/login/idea/PluginInitializationComponent.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.intellij.openapi.components.ApplicationComponent;
+import com.intellij.util.PlatformUtils;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Performs runtime initialization for the Google Login plugin.
+ */
+public class PluginInitializationComponent implements ApplicationComponent {
```
patflynn: Do you see value in testing the dumb getters? Google Java test practices is not to test logic-less methods.

patflynn:
I hereby do swear to fix the testing for this. https://github.com/GoogleCloudPlatform/gcloud-intellij/issues/115

@loosebazooka can I get an LGTM?

loosebazooka:
lgtm

PR #112, opened 2015-10-08
Test commit testing

patflynn:
Testing to see if travis identifies bad intermediary commits


patflynn:
@loosebazooka @elharo Travis integration seems to have broken when we went public.

PR #116, opened 2015-10-09
Update travis ci widget link

patflynn:
This is to fix the fact the Travis URI changed when we went public.


loosebazooka:
Would it make sense to add in the reason to the commit message?

PR #118, opened 2015-10-12
remove a bit of TMI from test output

elharo:
@loosebazooka  @patflynn not everything we need to get rid of, but a small start in the right direction


elharo:
ping

loosebazooka:
lgtm

PR #119, opened 2015-10-12
turn on more tests

elharo:
@nbashirbello @patflynn @loosebazooka turns out IntelliJ loads a different, fake JDK in tests than it does in a real inspection and this breaks some of our tests.

### DIFF
--- a/core-plugin/testSrc/com/google/gct/idea/appengine/validation/MethodParameterTypeInspectionTest.java
+++ b/core-plugin/testSrc/com/google/gct/idea/appengine/validation/MethodParameterTypeInspectionTest.java
@@ -24,6 +24,14 @@ import com.intellij.codeInspection.ex.LocalInspectionToolWrapper;
  */
 public class MethodParameterTypeInspectionTest extends EndpointTestBase {
 
+  @Override
+  protected void setUp() throws Exception {
+    super.setUp();
+
+    myFixture.addClass("package java.util; public interface Collection {}");
+    myFixture.addClass("package java.util; public interface List extends Collection {}");
+  }
+
   /**
    * Tests that method parameters of arrays of arrays cause MethodParameterTypeInspection errors.
    */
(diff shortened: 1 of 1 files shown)


elharo:
ping

core-plugin/testSrc/com/google/gct/idea/appengine/validation/MethodParameterTypeInspectionTest.java
```diff
@@ -24,6 +24,14 @@
  */
 public class MethodParameterTypeInspectionTest extends EndpointTestBase {
 
+  @Override
```
loosebazooka: The fake jdk doesn't define these? I'm confused what's going on.

elharo: The fake JDK is not included uin our download, so not, it doesn't. :-(

nbashirbello:
LGTM

core-plugin/testSrc/com/google/gct/idea/appengine/validation/MethodParameterTypeInspectionTest.java
patflynn: so the fake JDK is not available in tests? Do you mean the fake JDK is not in the binary intellij distribution so it's missing when configured as the IDEA SK?

elharo: It is not available in the version we download in fetchIdea.sh.

patflynn: So does that mean it's not part of their open API?  Perhaps we should file a bug against them, since it sounds like a core part of their testing framework.

elharo: This needs more investigation. It's not part of some (not all) of their downloads. Not yet clear why.

PR #120, opened 2015-10-13
remove most of the TMI; replace with link to report

elharo:
@loosebazooka @patflynn

### DIFF
--- a/core-plugin/build.gradle
+++ b/core-plugin/build.gradle
@@ -46,7 +46,10 @@ sourceSets.test.java.srcDirs = ['testSrc']
 
 test {
     testLogging {
-        events "passed", "skipped", "failed", "standardOut", "standardError"
+        events "skipped", "failed"
+    }
+    doLast{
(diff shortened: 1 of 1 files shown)


core-plugin/build.gradle
```diff
@@ -46,7 +46,10 @@ sourceSets.test.java.srcDirs = ['testSrc']
 
 test {
     testLogging {
-        events "passed", "skipped", "failed", "standardOut", "standardError"
+        events "skipped", "failed"
+    }
+    doLast{
```
loosebazooka: I'm not sure a doLast block will get executed if tests are failing. http://stackoverflow.com/questions/20973365/gradle-execute-task-after-test-phase-even-if-test-has-failed

On the other hand, the test result link is printed to screen on failure by the system, so maybe it doesn't matter?

elharo: correct. My problem was only that the test report disappeared on success.

patflynn:
LGTM

PR #128, opened 2015-10-14
use bundled version of mockito

elharo:
@patflynn @loosebazooka This change ensures that gradkle and IDEA use the same mockito jar file. An alternative would be to remove the bundled mockiot jar, and figure out how to tell IDEA load the jar from maven central instead.


patflynn:
LGTM

PR #133, opened 2015-10-15
simple test for plugin.xml

elharo:
@patflynn @loosebazooka


patflynn:
LGTM. I guess we should include this in the login plugin as well.

PR #134, opened 2015-10-15
prefer bundled libraries in third_party to those downloaded from maven

elharo:
@patflynn @loosebazooka


core-plugin/build.gradle
```diff
@@ -32,8 +32,9 @@ dependencies {
 
     // todo(appu): Should we be using + values here? Will that lead to a mismatch with what intellij is configuring?
     testCompile 'junit:junit:4.+'
```
patflynn: what about junit4?

elharo: At the moment we don't bundle it in third_party. I'll look at syncing this in a future cl.

patflynn:
LGTM

PR #137, opened 2015-10-16
test plugin.xml for login plugin; closes #82

### LINKED ISSUE #82: Verify plugin.xml
There is no test which verifies META-INF/plugin.xml

either that is well-formed, valid, or that the names used map to actual classes. 

E.g. removing AppEngineUpdateAction class only causes a problem when the plugin is loaded into IntelliJ.

elharo:
@patflynn @nbashirbello


nbashirbello:
LGTM

PR #147, opened 2015-10-20
cure NullPointerException

elharo:
@loosebazooka


core-plugin/src/com/google/gct/idea/debugger/ui/CloudDebugHistoricalSnapshots.java
```diff
@@ -389,8 +390,11 @@ public void run() {
           if (bp.getIsFinalState() != Boolean.TRUE) {
             continue;
           }
-          if (bp.getStatus().getIsError() == Boolean.TRUE) {
-            continue;
+          StatusMessage status = bp.getStatus();
```
loosebazooka: Is status message == null something we should log? Or is that normal behavior?

elharo: It seems to be normal behavior in the non-error case

patflynn:
LGTM

PR #155, opened 2015-10-21
Fix external lib scopes for plugin dependencies

loosebazooka:
@patflynn @elharo I think we misconfigured these scopes, when running we get exceptions because classes are coming in from multiple places


patflynn:
LGTM

PR #156, opened 2015-10-21
remove red X

elharo:
@patflynn

### DIFF
--- a/core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
+++ b/core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
@@ -22,6 +22,7 @@ import com.google.gct.idea.debugger.ui.CloudDebugHistoricalSnapshots;
 import com.google.gct.idea.util.GctBundle;
 import com.intellij.debugger.ui.DebuggerContentInfo;
 import com.intellij.debugger.ui.breakpoints.BreakpointManager;
+import com.intellij.execution.ExecutionBundle;
 import com.intellij.execution.configurations.RunProfile;
 import com.intellij.execution.process.ProcessHandler;
 import com.intellij.execution.ui.RunnerLayoutUi;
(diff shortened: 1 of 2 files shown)


elharo:
ping

loosebazooka:
what's this red x?

loosebazooka:
I found the bug but something else was merged to do the same thing?

elharo:
first fix didn't work :-(

core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
```diff
@@ -347,7 +348,18 @@ public void registerAdditionalActions(@NotNull DefaultActionGroup leftToolbar,
 
     leftToolbar.remove(manager.getAction(IdeActions.ACTION_RERUN));
     leftToolbar.remove(manager.getAction(IdeActions.ACTION_STOP_PROGRAM));
-    leftToolbar.remove(manager.getAction(IdeActions.ACTION_CLOSE));
+
+    // XDebugSessionTab puts this action second from end.
+    // Is there a predictable ID we could use instead? ACTION_CLOSE isn't it.
+    AnAction[] actions = leftToolbar.getChildActionsOrStubs();
```
loosebazooka: This does not appear to be working for me when I patch it in... actions[actions.length - 3] == pin tab, the correct value in my environment was actions[actions.length - 2]

I was digging around the code and the type of the action CloseAction is com.intellij.execution.ui.actions.CloseAction which as it turns out is never registered with the ActionManager for some reason, but clones its presentation from whatever action is IdeActions.ACTION_CLOSE

Could we just search through the actions list to find the type or text instead of guessing the position?

elharo: good idea; done

loosebazooka:
Actually wondering about the reasoning for this CL, if we remove the X button, users can't close a debugger tab they don't want open without going to the right-click context menu for the tab.

patflynn:
They can close it via the 'stop' button.  I'm a bit confused about the lifecycle of a cloud debug session myself. I guess there's no way to fully stop listening for snapshot events without closing and restarting the IDE.

elharo:
PTAL. Based on the discussion this morning we definitely want to get rid of the red X, and this PR does that. We still need to change the behavior of the red square stop button in the next PR.

elharo:
ping @patflynn @loosebazooka

core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
```diff
@@ -347,7 +348,18 @@ public void registerAdditionalActions(@NotNull DefaultActionGroup leftToolbar,
 
     leftToolbar.remove(manager.getAction(IdeActions.ACTION_RERUN));
     leftToolbar.remove(manager.getAction(IdeActions.ACTION_STOP_PROGRAM));
-    leftToolbar.remove(manager.getAction(IdeActions.ACTION_CLOSE));
+
+    // XDebugSessionTab puts this action second from end.
```
loosebazooka: Is this comment up to date?

loosebazooka:
One thing about a comment being out of date but otherwise lgtm...

patflynn:
LGTM

core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
elharo: no, done

PR #157, opened 2015-10-21
Cloud Debugger usage tracking

loosebazooka:
@patflynn I'm not super sure about some of this stuff, but it covers most cases I think...

### DIFF
--- a/core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
+++ b/core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
@@ -20,6 +20,9 @@ import com.google.api.services.clouddebugger.model.Breakpoint;
 import com.google.gct.idea.debugger.CloudDebugProcessStateController.ResolveBreakpointHandler;
 import com.google.gct.idea.debugger.ui.CloudDebugHistoricalSnapshots;
 import com.google.gct.idea.util.GctBundle;
+import com.google.gct.idea.util.GctTracking;
+import com.google.gct.login.stats.UsageTrackerService;
+
 import com.intellij.debugger.ui.DebuggerContentInfo;
 import com.intellij.debugger.ui.breakpoints.BreakpointManager;
 import com.intellij.execution.ExecutionBundle;
@@ -469,6 +472,8 @@ public class CloudDebugProcess extends XDebugProcess implements CloudBreakpointL
     @Override
     public void actionPerformed(AnActionEvent e) {
       ActionManager.getInstance().getAction(IdeActions.ACTION_STOP_PROGRAM).actionPerformed(e);
+      UsageTrackerService.getInstance()
+          .trackEvent(GctTracking.CATEGORY, GctTracking.CLOUD_DEBUGGER, "stop.session", null);
     }
 
     @Override
--- a/core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessWatcher.java
+++ b/core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessWatcher.java
@@ -16,12 +16,14 @@
 package com.google.gct.idea.debugger;
 
 import com.google.gct.idea.util.GctBundle;
+import com.google.gct.idea.util.GctTracking;
+import com.google.gct.login.stats.UsageTrackerService;
+
 import com.intellij.execution.Executor;
 import com.intellij.execution.ProgramRunnerUtil;
 import com.intellij.execution.RunManager;
 import com.intellij.execution.RunnerAndConfigurationSettings;
 import com.intellij.execution.executors.DefaultDebugExecutor;
-import com.intellij.execution.executors.DefaultRunExecutor;
 import com.intellij.notification.Notification;
 import com.intellij.notification.NotificationGroup;
 import com.intellij.notification.NotificationListener;
@@ -70,30 +72,28 @@ public class CloudDebugProcessWatcher implements CloudBreakpointListener {
       public void run() {
         NotificationGroup notificationGroup = NotificationGroup.balloonGroup("Cloud Debugger watcher");
         String message = GctBundle.getString("clouddebug.balloonnotification.message", state.getProjectName());
+        UsageTrackerService.getInstance()
+            .trackEvent(GctTracking.CATEGORY, GctTracking.CLOUD_DEBUGGER, "notify.background.bubble", null);
         notificationGroup.createNotification("", message, NotificationType.INFORMATION, new NotificationListener() {
-                                               @Override
-                                               public void hyperlinkUpdate(@NotNull Notification notification,
-                                                                           @NotNull HyperlinkEvent event) {
-                                                 notification.expire();
-                                                 RunnerAndConfigurationSettings targetConfig = null;
-                                                 RunManager manager = RunManager.getInstance(state.getProject());
-                                                 for (final RunnerAndConfigurationSettings config : manager
-                                                   .getAllSettings()) {
-                                                   if (config
-                                                         .getConfiguration() instanceof CloudDebugRunConfiguration &&
-                                                       ((CloudDebugRunConfiguration)config.getConfiguration())
-                                                         .getProcessState() == state) {
-                                                     targetConfig = config;
-                                                   }
-                                                 }
+          @Override
+          public void hyperlinkUpdate(@NotNull Notification notification,
+              @NotNull HyperlinkEvent event) {
+            notification.expire();
+            RunnerAndConfigurationSettings targetConfig = null;
+            RunManager manager = RunManager.getInstance(state.getProject());
+            for (final RunnerAndConfigurationSettings config : manager.getAllSettings()) {
+              if (config.getConfiguration() instanceof CloudDebugRunConfiguration &&
+                  ((CloudDebugRunConfiguration) config.getConfiguration()).getProcessState() == state) {
+                targetConfig = config;
+              }
+            }
 
-                                                 if (targetConfig != null) {
-                                                   Executor executor = DefaultDebugExecutor.getDebugExecutorInstance();
-                                                   ProgramRunnerUtil
-                                                     .executeConfiguration(state.getProject(), targetConfig, executor);
-                                                 }
-                                               }
-                                             }).notify(state.getProject());
+            if (targetConfig != null) {
+              Executor executor = DefaultDebugExecutor.getDebugExecutorInstance();
+              ProgramRunnerUtil.executeConfiguration(state.getProject(), targetConfig, executor);
+            }
+          }
+        }).notify(state.getProject());
       }
     });
   }
--- a/core-plugin/src/com/google/gct/idea/debugger/CloudDebugRunConfiguration.java
+++ b/core-plugin/src/com/google/gct/idea/debugger/CloudDebugRunConfiguration.java
@@ -16,8 +16,12 @@
 package com.google.gct.idea.debugger;
 
 import com.google.api.client.repackaged.com.google.common.base.Strings;
+import com.google.gct.idea.util.GctTracking;
+import com.google.gct.login.stats.UsageTrackerService;
+
 import com.intellij.execution.ExecutionException;
 import com.intellij.execution.Executor;
+import com.intellij.execution.RunManager;
 import com.intellij.execution.configurations.*;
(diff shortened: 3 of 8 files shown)


patflynn:
There's a crapload of random formatting changes in here. Makes it a bit difficult to see the meat.

loosebazooka:
Yeah some stuff really annoyed me. I'll split it into two commits

elharo:
Thanks, please ping me when that's done

core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessWatcher.java
```diff
@@ -70,30 +72,28 @@ public void onBreakpointListChanged(final CloudDebugProcessState state) {
...
-                                                 for (final RunnerAndConfigurationSettings config : manager
-                                                   .getAllSettings()) {
-                                                   if (config
-                                                         .getConfiguration() instanceof CloudDebugRunConfiguration &&
-                                                       ((CloudDebugRunConfiguration)config.getConfiguration())
-                                                         .getProcessState() == state) {
-                                                     targetConfig = config;
-                                                   }
-                                                 }
+          @Override
```
loosebazooka: this is the only formatting change in this CL

core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessWatcher.java
```diff
@@ -70,30 +72,28 @@ public void onBreakpointListChanged(final CloudDebugProcessState state) {
...
+          public void hyperlinkUpdate(@NotNull Notification notification,
+              @NotNull HyperlinkEvent event) {
+            notification.expire();
+            RunnerAndConfigurationSettings targetConfig = null;
+            RunManager manager = RunManager.getInstance(state.getProject());
+            for (final RunnerAndConfigurationSettings config : manager.getAllSettings()) {
+              if (config.getConfiguration() instanceof CloudDebugRunConfiguration &&
+                  ((CloudDebugRunConfiguration) config.getConfiguration()).getProcessState() == state) {
```
elharo: .equals

core-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
```diff
@@ -583,4 +587,29 @@ public String getVersion() {
...
+        UsageTrackerService.getInstance()
+            .trackEvent(GctTracking.CATEGORY, GctTracking.CLOUD_DEBUGGER, "run.sourceContext.stash.and.sync", null);
+      } else if (mySyncResult.needsStash()) {
+        UsageTrackerService.getInstance()
+            .trackEvent(GctTracking.CATEGORY, GctTracking.CLOUD_DEBUGGER, "run.sourceContext.stash", null);
+      } else if (mySyncResult.needsSync()) {
+        UsageTrackerService.getInstance()
+            .trackEvent(GctTracking.CATEGORY, GctTracking.CLOUD_DEBUGGER, "run.sourceContext.sync", null);
+      }
+    } else if (mySyncStashCheckbox.isVisible()) { // && !mySyncStashCheckbox.isSelected()
```
elharo: why the comment?

loosebazooka: Oh it was a redundant condition, forgot it was in there, taking it out.

core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessWatcher.java
loosebazooka: I'd rather not change the code here, merely reformatting.

loosebazooka:
updated

core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
```diff
@@ -458,6 +461,8 @@ public SaveAndExitAction() {
     @Override
     public void actionPerformed(AnActionEvent e) {
       ActionManager.getInstance().getAction(IdeActions.ACTION_STOP_PROGRAM).actionPerformed(e);
+      UsageTrackerService.getInstance()
```
elharo: can you inject this object instead?

core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessWatcher.java
```diff
@@ -70,30 +72,28 @@ public void onBreakpointListChanged(final CloudDebugProcessState state) {
       public void run() {
         NotificationGroup notificationGroup = NotificationGroup.balloonGroup("Cloud Debugger watcher");
         String message = GctBundle.getString("clouddebug.balloonnotification.message", state.getProjectName());
+        UsageTrackerService.getInstance()
```
elharo: Is it possible to inject this object in the constructor instead?

core-plugin/src/com/google/gct/idea/debugger/CloudLineBreakpointType.java
```diff
@@ -166,6 +169,8 @@ public CloudLineBreakpointProperties createBreakpointProperties(@NotNull Virtual
   @NotNull
   @Override
   public Breakpoint createJavaBreakpoint(Project project, XBreakpoint breakpoint) {
+    UsageTrackerService.getInstance()
```
elharo: Could this be injected?

core-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
```diff
@@ -583,4 +587,28 @@ public String getVersion() {
     }
   }
 
+  private void doTracking() {
+    if (mySyncResult.isDeterminable()) {
+      UsageTrackerService.getInstance()
```
elharo: Could this be injected?

patflynn:
@nbashirbello @loosebazooka @elharo 
It seems like the right thing to do is probably store the usage tracker service as an instance and retrieve it using the IntelliJ service lookup API
then have a setter so that you can mock it for unit testing.

To do that the service needs to actually be implemented as an intellij service.  What do you guys think?

All this obviously should be in a separate PR and not block this.

elharo:
Is there a reason not to simply always inject it in the constructor? Code
that wants the real thing can look it up and pass it in. I.e. is there some
code path in which IntelliJ is instantiating this from a no-args
constructor via reflection, in which case that wouldn't work.

OTOH, even if we do need to look it up internally I'd still prefer to keep
the instance final and immutable by passing it into a @VisibleForTesting
constructor rather than a setter method.



On Tue, Oct 27, 2015 at 11:50 AM, Patrick Flynn <notifications@github.com>
wrote:

> @nbashirbello <https://github.com/nbashirbello> @loosebazooka
> <https://github.com/loosebazooka> @elharo <https://github.com/elharo>
> It seems like the right thing to do is probably store the usage tracker
> service as an instance and retrieve it using the IntelliJ service lookup API
> then have a setter so that you can mock it for unit testing.
>
> To do that the service needs to actually be implemented as an intellij
> service. What do you guys think?
>
> All this obviously should be in a separate PR and not block this.
>
> —
> <https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/157#issuecomment-151608102>
> .
>



-- 
Elliotte Rusty Harold
elharo@macfaq.com

patflynn:
Agreed that constructor injection is better.

I don't think that the client should be constructing and providing the dependencies of the services they need. 

I've filed an issue here: https://github.com/GoogleCloudPlatform/gcloud-intellij/issues/165

Let's take the conversation there.

core-plugin/src/com/google/gct/idea/debugger/CloudDebugRunConfiguration.java
```diff
@@ -58,7 +62,12 @@ protected CloudDebugRunConfiguration(Project project, @NotNull ConfigurationFact
 
   @Override
   public final RunConfiguration clone() {
-    final CloudDebugRunConfiguration configuration = (CloudDebugRunConfiguration)super.clone();
+    if (this == RunManager.getInstance(getProject()).getConfigurationTemplate(this.getFactory()).getConfiguration()) {
```
patflynn: why do you need to do this check?

core-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
```diff
@@ -156,6 +159,7 @@ protected JComponent createCenterPanel() {
   @Override
   protected void doOKAction() {
     if (getOKAction().isEnabled()) {
```
patflynn: can doOKAction happen if getOKAction is disabled?

core-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
```diff
@@ -156,6 +159,7 @@ protected JComponent createCenterPanel() {
...
+      doTracking();
```
patflynn: shouldn't all the tracking cases in doTracking be inlined into the isSelected block and syncOrStash?

core-plugin/src/com/google/gct/idea/debugger/ui/CloudDebugHistoricalSnapshots.java
```diff
@@ -405,6 +408,7 @@ public void run() {
           break;
         }
         if (rowForPopup != -1) {
+          UsageTrackerService.getInstance().trackEvent(GctTracking.CATEGORY, GctTracking.CLOUD_DEBUGGER, "snapshot.received", null);
```
patflynn: could you just run this as the beginning of this method?

core-plugin/src/com/google/gct/idea/debugger/CloudDebugRunConfiguration.java
loosebazooka: It uses the same clone method to create instances of existing run configurations for editing, in the run configuration editor.

core-plugin/src/com/google/gct/idea/debugger/ui/CloudDebugHistoricalSnapshots.java
loosebazooka: From what I could tell, a new snapshot received happens in here.  Other changes are still processed by this method.

patflynn:
There's a few unanswered comments in this PR. Let's get this in ASAP.

core-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
loosebazooka: Well, I think initially it can, at least until the validator kicks in. We can explore this in another bug though.

loosebazooka: https://github.com/GoogleCloudPlatform/gcloud-intellij/issues/182

core-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
loosebazooka: It was kind of hard to do that because everything was all over the place. I'll try again though

core-plugin/src/com/google/gct/idea/debugger/CloudDebugRunConfiguration.java
patflynn: could you add a comment explaining this code.

core-plugin/src/com/google/gct/idea/debugger/CloudLineBreakpointType.java
```diff
@@ -166,6 +169,8 @@ public CloudLineBreakpointProperties createBreakpointProperties(@NotNull Virtual
...
+        .trackEvent(GctTracking.CATEGORY, GctTracking.CLOUD_DEBUGGER, "create.breakpoint", null);
```
loosebazooka: turns out this is wrong. This method is called under a variety of conditions, need to look deeper in to it.

patflynn:
Let's get in what we can, and we can added finer grained metrics later.

loosebazooka:
@patflynn @elharo ptal, looking in to making the UsageTracker service more testable in another change.

patflynn:
LGTM

Issue #141, opened 2015-10-20
remove the 'x' button and the 'console' tab from the Cloud Debug UI


patflynn:
1)The cloud debug tool window includes a 'console' tab next to the 'debugger' tab which I don't think displays anything. Ideally it would be removed.

2) in the left margin of the cloud debug tool panel there is a red 'x' button. Clicking it brings up a dialog that says 'do you want to disconnect from the process 'cloud debug'? with the option to terminate the process after disconnect.  I'm not sure what this means in the context of the cloud debugger as their is no local process. Ideally this dialog would not be displayed or the 'x' button would be removed.

3) Also the blue square button that represents stop is kind of confusing.  For some reason it did not register with me as a stop button. I'd be tempted to make it red like the other debugger stop buttons.

elharo:
Going forward it would be really cool if we could get log messages and STDOUT/STDERR from the cloud app to show up here, but for now YAGNI, I suppose.

elharo:
You also get the "Do you want to disconnect" dialog if you just close the IDE

elharo:
I think the blue square is GoogleCloudToolsIcons.CLOUD_DEBUG_SAVE_EXIT
i.e.
  public static final Icon CLOUD_DEBUG_SAVE_EXIT = load("/icons/debugsaveexit.png");

What tools do we use to edit these?

patflynn:
I'm still seeing the red 'x' with the latest code.

elharo:
Yes, I am too. I wonder how ti snuck back in. Still digging...

elharo:
Decision:

1. remove red X
2. Red square will close tab and pop up a dialog that asks whether or not to continue listening for breakpoints.

elharo:
When  the red square is pressed a dialog will pop up asking the user to make a choice:

1. Continue receiving and beinig notified of breakpoints, so that the session can be continued later. (This is the current behavior.)
2. Stop listening for breakpoints.

Language TBD.

Possibly user a Yes-No IntelliJ DialogWrapper or some such.

elharo:
This is in CloudDebugProcess.SaveAndExitAction

PR #159, opened 2015-10-22
change blue square icon to red square icon

elharo:
@patflynn #141


loosebazooka:
lgtm, is there any way we could use the default ide icons? or do we have to have our own copy?

elharo:
Maybe. Do you know where IDEA keeps those?

PR #160, opened 2015-10-22
don't show disconnect confirmation dialog when exiting IDEA

### LINKED ISSUE #151: Don't show disconnect dialog when quitting
Closing/quitting IDEA while running the cloud debugger brings up a dialog that says 'do you want to disconnect from the process 'cloud debug'? with the option to terminate the process after disconnect. This doesn't make sense in the context of the cloud debugger as there is no local process. This dialog should not be displayed.

related to #141

elharo:
@patflynn closes #151


loosebazooka:
lgtm

PR #166, opened 2015-10-28
MockitoTestCase is no longer necessary

elharo:
@patflynn @nbashirbello With JUnit 4.4+ we no longer need MockitoTestCase.


patflynn:
LGTM

PR #172, opened 2015-11-02
support non-GCP git repositories; also differentiate between different reasons for not matching source context

### LINKED ISSUE #145: Cloud Debugger: More precise message for not matching source context
Right now, we do not differentiate between these scenarios:

1. the server sends us a null source context meaning there is no information for matching a git repo.
2. the server sent us a valid source context but it wasn't Git
3. the server sent us a valid source context, but we could not locate a suitable Git Repo that has that SHA

Right now, if we encounter any of these we show a generic message indicating that we "could not verify that the source is correct"
We should give more detailed info so the user knows how to address the problem.

elharo:
@patflynn close #145


patflynn:
I don't think that the repo is a GCP repo is a case that matters to us.  

I think it's basically this:
case 1)
The source context for the debugee is some git repo (uri + hash) whether a GCP repo or not
  a) we look up the local projects repositories on the client and if there is match we stash and switch to it
  b) if there is no match we inform the user. "Sorry, the project you're working on doesn't seem to match the project being debugged, we can't promise this will work." (I guess start the debugger hoping for the best)
case 2)
The source context is a non git repo.
"Sorry we only support git repositories for cloud debugger source contexts for now"
case 3)
There is no source context. "No source context was provided with the binary being Cloud Debugged. We don't promise this will work as we can't know whether your source matches the running code."

obviously my text needs work but you get the idea.

patflynn:
em, for all steps where we don't match we can still proceed optimistically and message accordingly.

elharo:
So my question is how tdo we distinguish these three cases in the code? The lookup happens in ProjectRepositoryValidator.checkSyncStashState. At that point we find out the following things:

1. Is there a cloud repo source context, and if so what is it? If this doesn't exist at all I think we're in case 3.
2. Assuming there is a cloud repo, Is there a local repo that corresponds to it?
3. Assuming there is a local repo, does it need syncing (doesn't have all changes from server yet) or statshing *(has changed the server doesn't yet have). 

Possibly we can look for other things in ways we aren't looking now.

I don't yet see how to tell whether the cloud repo source context is or is not a GIT repo. maybe some URL magic? maybe it just is always a GIT repo for now?

patflynn:
@benwulfe Hey Ben, if you have some time could you chime in on Elliotte's questions?

benwulfe:
response.getSourceContext().getContextCase()  should return to you an enum indicating which source context the server has stored info for.  This should indicate to you whether there is any source info at all versus not having a compatible type.

We should review the list and make sure we are supporting anything that is backed by a git revision Id.
Actually, there does look to be a bug where "Git" (external Git repo) isn't being supported by our code.
To fix it, you need to call "getGit()" in addition to "getCloudRepo()" -- both should support revision ids.

elharo:
ping

benwulfe:
to be clear, response.getSourceContext().getContextCase() tells you whether its a GIT repo, Cloud Repo (which is GIt) or something else.

let me know if you want to sync up in gvc or if you have additional questions.

elharo:
PTAL

core-plugin/src/com/google/gct/idea/debugger/ProjectRepositoryValidator.java
```diff
@@ -105,13 +106,25 @@ public SyncResult checkSyncStashState() {
...
           if (myProcessState.getDebuggeeId() != null && myProcessState.getDebuggeeId().equals(debuggee.getId())) {
             foundDebuggee = true;
+            // TODO: This field is deprecated. Consumers should use
+            // ext_source_contexts if it is not empty.
             List<SourceContext> contexts = debuggee.getSourceContexts();
             if (contexts != null) {
               for (SourceContext sourceContext : contexts) {
                 cloudRepo = sourceContext.getCloudRepo();
                 if (cloudRepo != null) {
+                  repoType = "Google Cloud Repository";
```
patflynn: Localization?

core-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
```diff
@@ -250,12 +254,24 @@ else if (mySyncResult.needsSync()) {
...
       myInfoPanel.setVisible(true);
     }
-    else if (!mySyncResult.isDeterminable()) {
+    else if (mySyncResult.getTargetRepository() != null && mySyncResult.getTargetSyncSHA() == null) {
       setOKButtonText(getIsContinued() ? GctBundle.getString("clouddebug.continueanyway") : GctBundle.getString("clouddebug.attach.anyway"));
       myWarningLabel.setVisible(true);
       myWarningLabel2.setVisible(true);
       myInfoPanel.setVisible(true);
-      myWarningLabel2.setText("Could not verify that current source matches module.");
+      myWarningLabel2.setText("Could not locate a local Git repository with matching SHA");
```
patflynn: localization

core-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
```diff
@@ -250,15 +254,27 @@ else if (mySyncResult.needsSync()) {
...
+    }
+    else if (mySyncResult.getTargetRepository() == null && mySyncResult.getCloudRepository() == null) {
+      setOKButtonText(isContinued() ? GctBundle.getString("clouddebug.continueanyway") : GctBundle.getString("clouddebug.attach.anyway"));
       myWarningLabel.setVisible(true);
       myWarningLabel2.setVisible(true);
       myInfoPanel.setVisible(true);
-      myWarningLabel2.setText("Could not verify that current source matches module.");
+      if (mySyncResult.getRepositoryType() != null) {
+        myWarningLabel2.setText(mySyncResult.getRepositoryType() + " repositories are not supported.");
```
benwulfe: why are you not supporting other git repos when the code is exactly the same?
You should just be getting the SHA from those source contexts as you do from the cloud repo.

core-plugin/src/com/google/gct/idea/debugger/ProjectRepositoryValidator.java
```diff
@@ -105,13 +106,25 @@ public SyncResult checkSyncStashState() {
...
             List<SourceContext> contexts = debuggee.getSourceContexts();
             if (contexts != null) {
               for (SourceContext sourceContext : contexts) {
                 cloudRepo = sourceContext.getCloudRepo();
                 if (cloudRepo != null) {
+                  repoType = GctBundle.getString("clouddebug.cloud.repository");
                   break;
                 }
+                else if (sourceContext.getGerrit() != null) {
+                  repoType = GctBundle.getString("clouddebug.gerrit");
```
benwulfe: grab syncSHA from Gerrit source context.

core-plugin/src/com/google/gct/idea/debugger/ProjectRepositoryValidator.java
```diff
@@ -105,13 +106,25 @@ public SyncResult checkSyncStashState() {
...
+                }
+                else if (sourceContext.getCloudWorkspace() != null) {
+                  repoType = GctBundle.getString("clouddebug.workspace");
+                }
+                else if (sourceContext.getGit() != null) {
+                  repoType = GctBundle.getString("clouddebug.nongoogle.git");
```
benwulfe: this is probably the most important to support because it will enable github and other scenarios.
Is there a reason why this is any more complicated than getting the target SHA from the "Git" source context and letting the rest of the code execute with it?

core-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
elharo: On Thu, Nov 12, 2015 at 12:27 PM, Patrick Flynn <notifications@github.com>
wrote:

> In core-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
> <https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/172#discussion_r44686879>
> :
>
> >        setOKButtonText(getIsContinued() ? GctBundle.getString("clouddebug.continueanyway") : GctBundle.getString("clouddebug.attach.anyway"));
> >        myWarningLabel.setVisible(true);
> >        myWarningLabel2.setVisible(true);
> >        myInfoPanel.setVisible(true);
> > -      myWarningLabel2.setText("Could not verify that current source matches module.");
> > +      myWarningLabel2.setText("Could not locate a local Git repository with matching SHA");
>
> localization
>
>
>

done


-- 
Elliotte Rusty Harold
elharo@macfaq.com

core-plugin/src/com/google/gct/idea/debugger/ProjectRepositoryValidator.java
elharo: The code's a bit of a mess, especially since we have to parse through a lot of optional protobuf fields. This is going to need some refactoring but we should probably do that anyway.

elharo:
Does anyone know what a ClousWorkspaceSourceContext is? Is this also some form of Git or something else like CITC?

elharo:
PTAL.  I've added support for Gerrit and other non-Cloud repository Git workspaces as Ben suggested. This is a new feature. I don't think this worked before.

benwulfe:
That's great!
I agree with you -- I'm pretty sure it didn't work before because the other
repo types weren't even in the discovery document.

core-plugin/src/com/google/gct/idea/debugger/ProjectRepositoryValidator.java
```diff
@@ -123,19 +145,36 @@ public SyncResult checkSyncStashState() {
...
+      revisionId = cloudRepo.getRevisionId();
+      repoType = GctBundle.getString("clouddebug.cloud.repository");
+    }
+    else if (gerritRepo != null) {
+      revisionId = gerritRepo.getRevisionId();
+      repoType = GctBundle.getString("clouddebug.gerrit");
+    }
+    else if (otherGitRepo != null) {
+      revisionId = otherGitRepo.getRevisionId();
+      repoType = GctBundle.getString("clouddebug.nongoogle.git");
```
patflynn: does this have to be highlighted as 'non-google' and 'cloud'?  Really this seems like the default git case.

core-plugin/src/com/google/gct/idea/debugger/ProjectRepositoryValidator.java
```diff
@@ -123,19 +145,36 @@ public SyncResult checkSyncStashState() {
...
-      return new SyncResult(/*isinvalid*/true,
+      return new SyncResult(/*isinvalid*/ true,
                             /*needsstash*/ false,
                             /*needssync*/ false,
-                            /*isdeterminable*/ true,
                             /*target SHA*/ null,
-                            /*target repo*/ null);
+                            /*target repo*/ null,
+                            /* hasCloudRepository */ false,
+                            /* repoType */ "Unknown");
```
patflynn: does Unkown get displayed to the user?

core-plugin/src/com/google/gct/idea/debugger/SyncResult.java
```diff
@@ -0,0 +1,80 @@
...
+  public String getTargetSyncSHA() {
+    return myTargetSyncSHA;
+  }
+
+  public boolean isValidDebuggee() {
+    return !myInvalidDebuggee;
+  }
+
+  /**
+   * @return whether the local repository has uncommitted changes we need to stash
```
patflynn: let's stick to style and always have a method javadoc comment (when a comment is required), but leave the @ javadoc s optional.

core-plugin/src/com/google/gct/idea/debugger/ProjectRepositoryValidator.java
elharo: Yes, I think it is along one path. You want it internationalized?

core-plugin/src/com/google/gct/idea/debugger/SyncResult.java
elharo: done

elharo:
Ping. Changed description since this PR is now a little more ambitious iun response to Ben's comments.

core-plugin/src/com/google/gct/idea/debugger/ProjectRepositoryValidator.java
patflynn: do you mean along a url path?  Otherwise anything that's displayed as language to the user should be read from resource bundles.

core-plugin/src/com/google/gct/idea/debugger/ProjectRepositoryValidator.java
elharo: Removed "Cloud". We should probably distinguish this from the other cases though.

core-plugin/src/com/google/gct/idea/debugger/ProjectRepositoryValidator.java
elharo: On Mon, Nov 16, 2015 at 9:49 AM, Patrick Flynn <notifications@github.com>
wrote:

> do you mean along a url path? Otherwise anything that's displayed as
> language to the user should be read from resource bundles.


done


-- 
Elliotte Rusty Harold
elharo@macfaq.com

patflynn:
squash and merge. LGTM.

PR #184, opened 2015-11-04
Disable Attach button until project is configured

### LINKED ISSUE #182: Ok (Attach) button for Cloud Attach Dialog should validate at startup
A user can press it even if the state of the window is not valid to allow an OK action to progress. It currently displays an error and disables the button if a user clicks it before selecting valid field values.

elharo:
@loosebazooka close #182


elharo:
ping @loosebazooka

patflynn:
LGTM
Squash and merge.

loosebazooka:
lgtm, what patrick said

PR #185, opened 2015-11-04
do not use "nor" without "neither"

elharo:
@nbashirbello Minor grammar fix.


elharo:
@GoogleCloudPlatform/spinnaker-google

patflynn:
LGTM

Issue #187, opened 2015-11-05
CloudDebugHistoricalSnapshotsTest.testOnBreakpointListChanged_twoBreakPoints hangs during gradlew test


patflynn:
CloudDebugHistoricalSnapshotsTest.testOnBreakpointListChanged_twoBreakPoints has been @Ignored  
until we figure out what's going on here.

Note: This only happens when you run all the tests, not when you run the test on its own.

PR #190, opened 2015-11-05
turn off CloudDebugHistoricalSnapshotsTest.java

patflynn:
Working on a new CL I introduced more tests and random test methods in
this class started failing.  This test class needs to be carefully
looked at before it's re-enabled.


patflynn:
related to #187

PR #192, opened 2015-11-05
ask user whether to continue listening in background

### LINKED ISSUE #141: remove the 'x' button and the 'console' tab from the Cloud Debug UI

elharo:
@patflynn @loosebazooka close #141


core-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
```diff
@@ -73,13 +73,23 @@ public void run() {
...
-  private static List<CloudDebugProcessState> getStates() {
+  public synchronized void stopBackgroundListening() {
+    if (myWatchTimer != null) {
+      myWatchTimer.cancel();
+      myWatchTimer = null;
+    }
+  }
+
+
+    private static List<CloudDebugProcessState> getStates() {
```
patflynn: is the indent off here?

core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
```diff
@@ -467,8 +468,10 @@ public SaveAndExitAction() {
     }
 
     @Override
-    public void actionPerformed(AnActionEvent e) {
-      ActionManager.getInstance().getAction(IdeActions.ACTION_STOP_PROGRAM).actionPerformed(e);
+    public void actionPerformed(AnActionEvent event) {
+      ExitDialog exitDialog = new ExitDialog(getXDebugSession().getProject());
+      exitDialog.showAndGetOk();
+      ActionManager.getInstance().getAction(IdeActions.ACTION_STOP_PROGRAM).actionPerformed(event);
```
patflynn: could you add usage tracking here?

patflynn:
Should we be removing the run configuration listening setting as part of this?

core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
elharo: done

elharo:
done

core-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
elharo: done

elharo:
Regarding "Should we be removing the run configuration listening setting as part of this?" I added this, but now I'm not so. If I read it right this setting controls whether we notify when a new snapshot is received both when we're listening and when we're not. If so it still applies.

If I'm wrong about that, then yes, we should take it out.

elharo:
PTAL. Does not yet close the debugger tab. This PR is getting big, so I'm goign to file a sepaarte bug for that.

patflynn:
looks good. Is there a way we could track close and keep watching vs close and stop watching?

elharo:
With Gogole Analytics? sure. we just add

          UsageTrackerService.getInstance()
              .trackEvent(GctTracking.CATEGORY, GctTracking.CLOUD_DEBUGGER,
"close.stoplistening", null);

and

          UsageTrackerService.getInstance()
              .trackEvent(GctTracking.CATEGORY, GctTracking.CLOUD_DEBUGGER,
"close.continuelistening", null);

at the relevant locations. Do we have any particular standards for
labelling these events?




On Fri, Nov 6, 2015 at 3:32 PM, Patrick Flynn <notifications@github.com>
wrote:

> looks good. Is there a way we could track close and keep watching vs close
> and stop watching?
>
> —
> <https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/192#issuecomment-154524234>
> .
>



-- 
Elliotte Rusty Harold
elharo@macfaq.com

elharo:
ping

core-plugin/src/com/google/gct/idea/debugger/ui/ExitDialog.java
```diff
@@ -0,0 +1,44 @@
...
+  private final JLabel myLabel = new JLabel(GctBundle.getString("clouddebug.continuelistening"));
+
+  public ExitDialog(@Nullable Project project) {
+    super(project, false, IdeModalityType.MODELESS);
+
+    init();
+
+    setOKButtonText(GctBundle.getString("clouddebug.continue"));
+    setCancelButtonText(GctBundle.getString("clouddebug.stoplistening"));
+  }
```
patflynn: can you add the tracking to differentiate when a user opts to stop vs continue listening?

elharo:
are the labels I suggested above OK or is there some other convention we use?

patflynn:
labels work for me. No conventions as of now that I'm aware of.

patflynn:
+nabila

elharo:
done

patflynn:
LGTM

PR #197, opened 2015-11-06
Add base test case

patflynn:
Let's talk about these base test classes offline.

### DIFF
--- a/core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
+++ b/core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
@@ -475,6 +475,7 @@ public class CloudDebugProcess extends XDebugProcess implements CloudBreakpointL
       ExitDialog exitDialog = new ExitDialog(getXDebugSession().getProject());
       exitDialog.showAndGetOk();
       ActionManager.getInstance().getAction(IdeActions.ACTION_STOP_PROGRAM).actionPerformed(event);
+      ActionManager.getInstance().getAction(IdeActions.ACTION_CLOSE).actionPerformed(event);
       UsageTrackerService.getInstance()
           .trackEvent(GctTracking.CATEGORY, GctTracking.CLOUD_DEBUGGER, "stop.session", null);
     }


patflynn:
@loosebazooka

common-lib/build.gradle
```diff
@@ -23,6 +23,8 @@ dependencies {
     compile files(ideaLib + 'jdom.jar')
 
     compile files('third_party/guava/guava-18.0.jar')
+    compile files(ideaLib + 'junit-4.11.jar')
```
elharo: we should try to avoid adding dependencies on JUnit and Mockito to our non-test code

common-lib/src/main/java/com/google/gct/idea/testing/BasePluginTestCase.java
```diff
@@ -0,0 +1,76 @@
...
+import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.project.Project;
+
+import org.jetbrains.annotations.NotNull;
+import org.junit.After;
+import org.junit.Before;
+import org.picocontainer.MutablePicoContainer;
+
+/**
+ * Test base class. <p/> <p>Provides a mock application and a mock project.
```
elharo: probably don't need the <p> tags

common-lib/build.gradle
patflynn: the only way to avoid this would be to introduce a new module.. core-test-lib?

elharo: I'd argue it's worth it. Keep our dependencies light.

common-lib/src/main/java/com/google/gct/idea/testing/BasePluginTestCase.java
```diff
@@ -0,0 +1,76 @@
...
+
+  public static class Container {
+
+    private final MutablePicoContainer container;
+
+    Container(@NotNull MutablePicoContainer container) {
+      this.container = container;
+    }
+
+    public <T> Container register(Class<T> klass, T instance) {
```
elharo: nit: clazz is commonly used for this

PR #198, opened 2015-11-06
Add a user-agent when deploying with appcfg

nbashirbello:
@loosebazooka @patflynn @elharo


core-plugin/src/com/google/gct/idea/appengine/deploy/UserAgentStrings.java
```diff
@@ -0,0 +1,16 @@
...
+ * it should be reflected in the name of the user-agent e.g 'MAVEN_VIA_CLOUD_TOOLS_FOR_ANDROID_STUDIO'
+ */
+public class UserAgentStrings {
+    public static String USER_AGENT_KEY = "appengine.useragent";
+
+    private UserAgentStrings () {
+    }
+
+    // Sources for appcfg update/deploy calls
+    public static String CT4AS = "CLOUD_TOOLS_FOR_ANDROID_STUDIO";
```
elharo: ANDROID_STUDIO?

patflynn:
@nbashirbello @elharo @loosebazooka what do you think about moving the AE1 deploy codebase into our Android Studio plugin?

core-plugin/src/com/google/gct/idea/appengine/deploy/UserAgentStrings.java
nbashirbello: Done

elharo:
If we do thhat can it also be deleted from whereever it lives now?

patflynn:
yes. that would be the idea.

elharo:
I don't have any objections then. 

The only issue I can imagine is that it might be a tad confusing to external devs since we don't support deploying from inside IntelliJ, but that's not major.

nbashirbello:
Created an issue for moving the AE1 code to Andriod Studio: https://github.com/GoogleCloudPlatform/gcloud-intellij/issues/199

elharo:
lgtm

PR #204, opened 2015-11-09
Removes the Google Settings section in Settings if there is no usage tracker configured.

### LINKED ISSUE #203: Settings -> Other Settings -> Google should not display tracking info when analytics is not configured

patflynn:
Also removes the opt-in setting from our Google settings panel for when
we have multiple settings and removing the whole section will be
inappropriate.  Disabling the control is not the right thing to do in
this case as it will give the user the impression that this a setting
that has a real effect but they can't change.

closes #203


elharo:
lgtm

PR #221, opened 2015-11-11
simplify base test case usage

patflynn:
The current implementation an intTest template method for subclass configuration.   Since the method is implemented as the last step, and junit runs @Before superclass first, it seems simpler to just have subclasses call register directly during their setup.

### DIFF
--- a/common-test-lib/src/main/java/com/google/gct/idea/testing/BasePluginTestCase.java
+++ b/common-test-lib/src/main/java/com/google/gct/idea/testing/BasePluginTestCase.java
@@ -15,11 +15,9 @@ package com.google.gct.idea.testing;
  * limitations under the License.
  */
 
-import com.intellij.mock.MockProject;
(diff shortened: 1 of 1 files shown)


patflynn:
@loosebazooka

common-test-lib/src/main/java/com/google/gct/idea/testing/BasePluginTestCase.java
```diff
@@ -30,34 +28,21 @@
...
-    }
-  }
+  protected MutablePicoContainer applicationContainer;
 
   @Before
   public final void setup() {
     TestUtils.createMockApplication();
-    MutablePicoContainer applicationContainer = (MutablePicoContainer)
+    applicationContainer = (MutablePicoContainer)
         ApplicationManager.getApplication().getPicoContainer();
```
loosebazooka: nit: make it one line?

loosebazooka:
lgtm

Issue #167, opened 2015-10-29
Can't build on Windows because fetchIdea.sh is Linux specific


Adam-Origamiiris:
Not sure this is the right place but I've been unable to build the plugin after pulling from git. 

What version of IDEA are you using?
IntelliJ community Edition 14.1.5
What version of Java are you using?
1.8.0_65 64 bit on Windows 7 and ran cmd as Admin
What did you do?
Cloned then then went into the folder and typed >gradlew plugin
What did you expect to see?
To see build complete or similar
What did you see instead?
with --debug I saw 
:loadIdea Failed 
> A problem occurred starting process 'comman ' .fetchIdea.sh

loosebazooka:
Yeah it's a Linux script. You need to build on Linux at the moment

Adam-Origamiiris:
Ahhh thank you,

elharo:
I'm reopening this one. Low priority but we should figure out how to make it build on Windows, and at the very least mention this in CONTRIBUTING.md

patflynn:
Before we fix this we should experiment with the gradle intellij plugin to
see if it solves this for us.

patflynn:
I have a working build and am fixing tests on a PR that may fix this, so I recommend nobody else pick this up for now.

Issue #161, opened 2015-10-27
Clean plugin artifacts


elharo:
This is for the binary zip files created in IntelliJ, which are currently placed  in the core-plugin and google-login-plugin directories by default gradle clean should remove these.

We can add a delete command to the clean task.

Issue #131, opened 2015-10-14
investigate using the intellij gradle plugin for our builds


patflynn:
https://plugins.gradle.org/plugin/org.jetbrains.intellij

here's a project that's using it and according to the contributors doc gradle import works: https://github.com/go-lang-plugin-org/go-lang-idea-plugin

Here's the project in github https://github.com/JetBrains/gradle-intellij-plugin

patflynn:
@elharo you may find this interesting.

Issue #90, opened 2015-09-30
Deps are duplicated across lib and gradle repo based imports


patflynn:
We have some deps configured in third_party and imported via our intellij modules that are duplicated in our Gradle builds as maven central or jcenter repo deps.

patflynn:
closed by https://github.com/GoogleCloudPlatform/gcloud-intellij/commit/2417d8f5dae210911c336f260deba15599dcb346

PR #227, opened 2015-11-18
build with gradle-intellij-plugin

### LINKED ISSUE #175: Need to clear plugin-sandbox when making changes to login plugin
<idea-config>/system/plugins-sandbox doesn't reflect changes to the plugin sandbox when the "Core-IJ-Community" run configuration is run unless it is missing the login dependency.

It's not really clear to me why, perhaps because the login plugin dependency is "provided"?

patflynn:
I've updated how to use this new setup for importing into IntelliJ. Due
to an issue with how intellij does import you unfortunately have to
revert the imports changes to .idea after import to get the
preconfigured runconfiguration. There's also a couple ugly hacks to support
multi-plugin projects which will be fixed when
 https://github.com/JetBrains/gradle-intellij-plugin/issues/17 is fixed
in a few days.

closes #175, #167, #161, #131, #90


patflynn:
@elharo @nbashirbello @loosebazooka

.idea/compiler.xml
```diff
@@ -18,6 +18,5 @@
         <processorPath useClasspath="true" />
       </profile>
     </annotationProcessing>
-    <bytecodeTargetLevel target="1.6" />
   </component>
```
loosebazooka: Don't we need to keep this requirement for 1.6 target level?

CONTRIBUTING.md
```diff
@@ -41,37 +41,17 @@ Other useful targets while developing include:
...
-
-### Optional:  Downloading source for IDEA to debug.
-
-[Download the IDEA source tarball from JetBrains](http://www.jetbrains.org/display/IJOS/Download) 
-into your home (or any other convenient) directory. Extract it and 
-add the resulting directory (usually something like ideaIC-141.1532.4) 
-in the Sourcepath tab of Project Structure > SDKs for the IntelliJ CE 141.1532 SDK.
-
+1. 'New project from existing sources'
+1. Select the root build.gradle file to import
```
loosebazooka: all numbered one? Is this something markdown takes care of automatically?

core-plugin/build.gradle
```diff
@@ -1,64 +1,53 @@
-apply from: '../scripts/javac2.gradle'
-def ideaHome = '../idea-IC/'
-def pluginName = 'google-cloud-tools-core'
+plugins {
+    id "org.jetbrains.intellij" version "0.0.29"
+}
+
+// **replace** default sourceSets with idea layout
```
loosebazooka: Do you think we can standardize around the gradle layout eventually? Or is that a bad idea?

.idea/compiler.xml
patflynn: not really sure what's going on here. language level is specified in the gradle file, I'm suprised it's removed here.

CONTRIBUTING.md
patflynn: yeah, that's a numbered bullet markdown feature.

google-account-plugin/build.gradle
```diff
@@ -1,84 +1,36 @@
-// see https://github.com/JetBrains/gradle-intellij-plugin
+plugins {
+    id "org.jetbrains.intellij" version "0.0.29"
```
loosebazooka: Are we gonna use 28 or 29 across plugins? Can we do this in the root project and apply it in the subprojects {} closure so we don't have to keep this in 4 different files

core-plugin/build.gradle
patflynn: I would like to do that. Not in this PR though.

patflynn:
I removed the non-used property from gradle.properties

common-test-lib/build.gradle
```diff
@@ -1,10 +1,9 @@
-apply plugin: 'java'
-
-repositories {
-    mavenCentral()
+// Not an IJ plugin project, but included to add IntelliJ platform dependencies to the classpath
+plugins {
+    id "org.jetbrains.intellij" version "0.0.28"
```
loosebazooka: 0.0.29

common-lib/build.gradle
```diff
@@ -1,27 +1,8 @@
-apply plugin: 'java'
-
-repositories {
-    mavenCentral()
+// Not an IJ plugin project, but included to add IntelliJ platform dependencies to the classpath
+plugins {
+    id "org.jetbrains.intellij" version "0.0.28"
```
loosebazooka: 0.0.29

build.gradle
```diff
@@ -1,36 +1,44 @@
...
+        classpath 'org.springframework.build.gradle:propdeps-plugin:0.0.7'
+        classpath 'gradle.plugin.org.jetbrains:gradle-intellij-plugin:0.0.29'
+    }
 }
 
-apply plugin: 'base'
-
-task loadIdea(type:Exec) {
-    commandLine './fetchIdea.sh'
+configure(allprojects) {
```
loosebazooka: do these have to be applied at the "allprojects" level? or can be at the "subprojects" level?

patflynn:
all fixed up.  propdeps only applied to subprojects.

PR #230, opened 2015-11-23
Removing the help button from the Debug toolbox.

### LINKED ISSUE #149: Help button goes to IDEA debugger help
On the left hand toolbar of the cloud debugger pane, the blue question mark icon links to 

https://www.jetbrains.com/idea/help/debug-tool-window.html

I.e. the IDEA debugger help. We should instead link it to Cloud Debugger specific help.

joaoandremartins:
Removing the help button from the Debug toolbox, since it's pointing to the IDEA help page and we don't have any help or our own.

fixes #149.


elharo:
I'm not sure about this. I was hoping we could point it to a page of our own instead.

patflynn:
@elharo @joaoandremartins perhaps a todo an an issue filed to include the help once we have cloud debugger user docs?

elharo:
Wha if we parameterized this so it shows iff a URL for help docs is explicitly provided?

joaoandremartins:
TODO would be a better use of time, I think. I've been looking into how to change the URL for the help button, but the API offers no easy way to specify one.

Looking at the action (http://grepcode.com/file/repository.grepcode.com/java/ext/com.jetbrains/intellij-idea/10.5/com/intellij/ide/actions/ContextHelpAction.java), it appears a lot of hacking would be required to call the update() method.

joaoandremartins:
Actually, there seems to be a way.

joaoandremartins:
I implemented this relying on CloudToolsBundle.properties, where I store a clouddebug.helpurl value that can either be empty or not. I show the button based on its emptiness and, if not empty, I open the browser on the configured URL.

However, mocking GctBundle/ResourceBundle/Properties is proving hard because of the static/final methods, not resorting to anything other than Mockito/GctBundle.

Any suggestions for anywhere else to configure the URL or to mock the properties for testing?

joaoandremartins:
@elharo @patflynn can one of you take a look into this?

Am I parameterising the URL correctly? If so, can I submit unit tests just for one scenario, since it's quite troublesome to mock static methods using only Mockito (and I'd like to avoid adding extra dependencies just for this temporary use case).

core-plugin/src/com/google/gct/idea/debugger/actions/CloudDebugHelpAction.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+public class CloudDebugHelpAction extends AnAction {
+  private static final Icon icon = IconLoader.getIcon("/actions/help.png");
+  private String url;
+
+  public CloudDebugHelpAction(String url) {
+    this.url = url;
+  }
+
+  public void actionPerformed(AnActionEvent event) {
```
elharo: To answer your question about testing, you could right a unit test for this class, with the exception of this method, fairly easily.

core-plugin/testSrc/com/google/gct/idea/debugger/CloudDebugProcessTest.java
```diff
@@ -72,7 +74,7 @@ public void testRegisterAdditionalActions_close() {
 
         assertEquals(3, leftToolbar.getChildrenCount());
         assertEquals(action0, leftToolbar.getChildActionsOrStubs()[0]);
-        assertEquals(action2, leftToolbar.getChildActionsOrStubs()[1]);
+        assertEquals(action3, leftToolbar.getChildActionsOrStubs()[1]);
```
elharo: I think this test is enough for now. If I read it right, this covers the case where the URL is empty. Later we'll make the uRL non-empty, this test will fail, and we'll update it. We don't expect to have both empty and non-empty caes at the same time. Is that correct?

core-plugin/src/com/google/gct/idea/debugger/actions/CloudDebugHelpAction.java
joaoandremartins: Yes, but that doesn't test both scenarios when the URL is set in the parameters and when it isn't.

I should probably write a unit test for this class anyways.

core-plugin/testSrc/com/google/gct/idea/debugger/CloudDebugProcessTest.java
joaoandremartins: Indeed. So, besides a unit test for CloudDebugHelpAction, is there anything else left in this change?

Do I have to notify you to accept this pull?

patflynn: Once you get 'LGTM' from @elharo you can merge the PR.

core-plugin/src/com/google/gct/idea/debugger/actions/CloudDebugHelpAction.java
elharo: Again, I'm OK with just a unit test for the current case (URL not set) that will fail as soon as we set that parameter (likely next week).

core-plugin/testSrc/com/google/gct/idea/debugger/CloudDebugProcessTest.java
elharo: I think we're good if you just add a test for the parts of CloudDebugHelpAction that can be reasonably easily tested.

joaoandremartins: @elharo PTAL.

core-plugin/testSrc/com/google/gct/idea/debugger/actions/CloudDebugHelpActionTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+import java.lang.Override;
+
+public class CloudDebugHelpActionTest {
```
elharo: @RunWith(MockitoJUnitRunner.class)

T^hen you don't need the setUp method below

core-plugin/testSrc/com/google/gct/idea/debugger/actions/CloudDebugHelpActionTest.java
```diff
@@ -0,0 +1,54 @@
...
+  protected static final String URL = "http://helps!";
```
elharo: why protected instread of private? 

I'd be tempted to inline this in either case.

core-plugin/testSrc/com/google/gct/idea/debugger/actions/CloudDebugHelpActionTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @Test
+  public void testUpdate() {
+    Presentation presentation = new Presentation();
+    when(event.getPresentation()).thenReturn(presentation);
+
+    CloudDebugHelpAction action = new CloudDebugHelpAction(URL);
+    action.update(event);
+
+    assertEquals(presentation.getIcon(), IconLoader.getIcon("/actions/help.png"));
```
elharo: expect6ed value shoudl come first so JUNit reports errors correctly

core-plugin/testSrc/com/google/gct/idea/debugger/actions/CloudDebugHelpActionTest.java
```diff
@@ -0,0 +1,54 @@
...
+    assertEquals(presentation.getText(), CommonBundle.getHelpButtonText());
```
elharo: 1. expected value first
2. Using literal strings in tests is preferred to using constants (i.e. it also tests that we've got the correct string in the constant)

joaoandremartins: @elharo PTAL.

elharo: do you need to push? I don't see any new changes.

joaoandremartins:
There was an issue pushing. It is fixed now, the changes are visible.

core-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
```diff
@@ -363,6 +364,26 @@ public void registerAdditionalActions(@NotNull DefaultActionGroup leftToolbar,
...
+    // https://github.com/GoogleCloudPlatform/gcloud-intellij/issues/149
+    for (AnAction child : leftToolbar.getChildActionsOrStubs()) {
+      if (child.getClass().getCanonicalName().equalsIgnoreCase(
+          "com.intellij.ide.actions.ContextHelpAction")) {
+        // we never want to show IDEA's help.
+        leftToolbar.remove(child);
+
+        // show our help if we have it.
+        String helpUrl = GctBundle.getString("clouddebug.helpurl");
+        if (!helpUrl.equalsIgnoreCase("")) {
```
elharo: !"".equals(helpUrl)

to avoid NullPointerExceptions and because we don't need to check for case in an empty string

core-plugin/testSrc/com/google/gct/idea/debugger/actions/CloudDebugHelpActionTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.lang.Override;
+
+@RunWith(MockitoJUnitRunner.class)
+public class CloudDebugHelpActionTest {
+  @Mock protected AnActionEvent event;
+
+  @Test
+  public void testActionPerformed() {
+    CloudDebugHelpAction action = Mockito.spy(new CloudDebugHelpAction("http://helps!"));
```
elharo: wouldn't hurt to use a syntactically valid URL here; e.g. http://www.example.com

core-plugin/testSrc/com/google/gct/idea/debugger/actions/CloudDebugHelpActionTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import java.lang.Override;
+
+@RunWith(MockitoJUnitRunner.class)
+public class CloudDebugHelpActionTest {
+  @Mock protected AnActionEvent event;
```
elharo: can this be private? (not sure. Maybe it needs to be portected for @Mock to work?)

elharo:
lgtm modulo my last comments

patflynn:
Hey everyone,

We forgot to instruct Joao to squash commits before merging into master.  

Joao for future reference, our current practice is to squash all commits into one before merging a PR. Also don't merge master into your branch, rebase your branch on master. There should only be one merge from your PR into master.

PR #233, opened 2015-11-23
Add a run configuration that runs IJ with our plugin

patflynn:


loosebazooka:
lgtm

loosebazooka:
Does this work with a new import? If there's no gradle project imported does it look funny?

PR #235, opened 2015-11-23
Force runIdea to always run even when 'up-to-date' for the core plugin

patflynn:
If nothing had changed, runIdea would do nothing.  Which is not the right behavior.


elharo:
lgtm

PR #238, opened 2015-11-23
Remove redundant Travis config. 

patflynn:
Travis will automatically run gradlew check


patflynn:
@elharo

loosebazooka:
Lgtm. Now that we have a bunch of violations, can you file a bug for us to look at the violations.

PR #242, opened 2015-11-25
Transition to idea15

loosebazooka:
Preliminary, compiles and tests seem to pass.


loosebazooka:
@elharo @patflynn Take a look, I didn't much deep testing here, but I think is what we need.

Did not manually run through all our features
Have not yet checked if this is still compatible with idea14

loosebazooka:
more notes : 

Is NOT compatible with idea14.

patflynn:
Does our Cloud Repo feature in Android Studio master still work?

patflynn:
LGTM

Issue #121, opened 2015-10-13
End to end test


elharo:
We need one end-to-end test that builds and installs the plugin, runs IDEA, and verifies that the relevant menu items show up and can be selected and do what we expect them to do.

This can be built in pieces. E.g. a first step might be to verify that IDEA starts up and doesn't crash and burn when our plugin is installed.

elharo:
Request for assistance at JetBrains forums: https://devnet.jetbrains.com/thread/473640?tstart=0

devBuild task in master build.gradle may be helpful here.

elharo:
Some options: Marathon: http://marathontesting.com/marathon/
JFCUnit (abandoned? does this still work?)
UISpec4J https://github.com/UISpec4J/UISpec4J

http://stackoverflow.com/questions/79891/what-is-the-best-testing-tool-for-swing-based-applications

elharo:
The gradle application plugin adds a run task to run an application:

https://docs.gradle.org/current/userguide/application_plugin.html

elharo:
Linux_Desktop_Testing_Project http://ldtp.freedesktop.org/wiki/  looks interesting. requires accessibility enabled app. Is IDEA such?

elharo:
Option 3: Launch IntelliJ from the test (perhaps in different classloader) then use
IdeFrame[] frames = WindowManager.getInstance().getAllProjectFrames();
and start mucking with the API. For once all the global static state might actually help us.

elharo:
Possibly JMX, JVMTI, and/or JDI might help here. Explore.

patflynn:
Definitely take a look at what Android Studio is doing if you haven't
already.

etanshaul:
closing in favor a new issue to track this work. this is very old

PR #256, opened 2015-12-02
MInimal GUI testing framework. 

elharo:
@patflynn  @loosebazooka @nbashirbello @alexruiz

This is the start of a FEST based testing framework for for GUI tests in IDEA. Notes:

1. Flaky.
2. Runs only in IDEA. Will not run at the command line or headless.

This is all blatantly copied from Alex Ruiz's work in android studio, with minor modifications to make it run in IDEA. These limitations are present today in Android Studio GUI tests. If we want something better, I'm going to have to go back to the drawing board.

I had to go back to his original cls. The current tests shipped with android studio proved too tied to Android Studio details. If this approach seems sound, I can start recreating the history of that work, piece by piece.

#121


common-test-lib/src/main/java/com/google/gct/idea/testing/gui/IdeTestApplication.java
```diff
@@ -0,0 +1,174 @@
...
+  private static final Logger LOG = Logger.getInstance(IdeTestApplication.class);
+
+  private static IdeTestApplication ourInstance;
+
+  public static synchronized IdeTestApplication getInstance() throws Exception {
+    return getInstance(null);
+  }
+
+  public static synchronized IdeTestApplication getInstance(@Nullable String configPath) throws Exception {
+    System.setProperty(PlatformUtils.PLATFORM_PREFIX_KEY, "Idea");
```
patflynn: Could we test against Android Studio releases?

PR #262, opened 2015-12-02
remove third party libs that are no longer used

patflynn:


elharo:
so going forward the plan is to pull everything we can from maven central, right? 

lgtm

patflynn:
yes. everything that we can.

PR #263, opened 2015-12-02
Use transitive dependencies between project modules

patflynn:
@loosebazooka 

All the classloading issues between modules are solved this way. It turned out to be a pain to require explicit deps between our plugin modules.


core-plugin/build.gradle
```diff
@@ -11,25 +11,15 @@ intellij {
...
-    }
-    compile(project(':common-lib')) {
-        transitive = true
-    }
+    compile(project(':google-account-plugin'))
+    compile(project(':common-lib'))
 
     compile 'com.google.appengine:gradle-appengine-builder-model:0.1.0'
     compile 'org.eclipse.jgit:org.eclipse.jgit:3.3.0.201403021825-r'
-    // todo(elharo): we shouldn't need Apache commons; Guava will suffice
```
elharo: I'd like to keep this if we can

elharo:
lgtm

core-plugin/build.gradle
patflynn: done

PR #294, opened 2015-12-07
Changes the GoogleLoginAction to be DumbAware

### LINKED ISSUE #117: Google Login widget is disabled during indexing
I don't think there's anything about the widget that requires the project to be out of dumb mode. We shouldn't disable the widget during indexing.

patflynn:
This prevents it from being disabled during indexing.

Closes #117


loosebazooka:
LGTM

PR #295, opened 2015-12-07
remove irrelevant, inactive Drop Frame button

### LINKED ISSUE #282: Remove drop frame button
The drop frame button in the cloud debugger  is deactivated and doesn't do anything:

http://stackoverflow.com/questions/2367816/when-using-the-java-debugger-in-intellij-what-does-drop-frame-mean

This is something we inherit from the local IDEA debugger. But in CDB it's meaningless. Take it out.

elharo:
@patflynn The button I'm removing is inherited from the regular IntelliJ debugger, but is always inactive in the Cloud Debugger. In the local debugger it reverses the execution by one stack frame (neat trick) but this is not something Cloud Debugger can do. 

close #282


patflynn:
LGTM

Issue #234, opened 2015-11-23
Display project ID and module version during cloud debug session


patflynn:
The user should be able to identify which project and version they are 'connected' to during a debug session.

elharo:
Good idea. Where should we display this information? Maybe we can add a JLabel to the title bar of the debugger tab since we've removed almost everything else that used ton live there.

patflynn:
SGTM.

elharo:
decided to limit IDEA to running a single CDB session at a time, which makes this low priority.

If we do pick it up, it would be nice to figure out how to add this to the tab title for the run/debug configuration tab.

PR #298, opened 2015-12-08
display project name in top toolbar

elharo:
@patflynn  Adds the project name to the top toolbar as discussed. 

Module version will be next. However that's going to require deeper surgery since the module is not currently avbailable in CloudDebugProcessState, so I want to make sure you're happy with this approach before doing that. 

#234


patflynn:
So the header only accepts toolbar buttons?

Could you include a screenshot?

elharo:
The toolbar only accepts actions, more specifically com.intellij.openapi.actionSystem.AnAction

Issue #228, opened 2015-11-19
If no debuggable module is available for a project the module drop down is just empty


patflynn:
Some kind of message should be displayed explaining to the user why they can't attach.

joaoandremartins:
Yes, I meant to file this issue. On top of that, I found out that CDB also doesn't work if you have no instance alive, which makes sense, but isn't documented anywhere I could find.

We should probably add a hint that CDB won't work if you have no instance alive.

loosebazooka:
![image](https://cloud.githubusercontent.com/assets/1304826/11667863/38daa726-9dc2-11e5-9070-cc674310fbb1.png)

how's this @elharo @patflynn

elharo:
Not sure of the GUI standard here, but can we put an italic gray text in
the combobox itself saying woth the warning? and maybe deactivate the
popup?

On Tue, Dec 8, 2015 at 3:42 PM, Appu Goundan <notifications@github.com>
wrote:

> [image: image]
> <https://cloud.githubusercontent.com/assets/1304826/11667863/38daa726-9dc2-11e5-9070-cc674310fbb1.png>
>
> how's this @elharo <https://github.com/elharo> @patflynn
> <https://github.com/patflynn>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/GoogleCloudPlatform/gcloud-intellij/issues/228#issuecomment-163012115>
> .
>



-- 
Elliotte Rusty Harold
elharo@macfaq.com

patflynn:
+1. That's the right way to handle it. The select box should be disabled and the contents should explain that there are no modules available. 

BTW @loosebazooka why is the 'Warning: unkown repositories are not supported' displayed in this case?

loosebazooka:
Sorry that screenshot was outdated.

It's possible, but I realized that I'd probably have to make the code really weird to put in a non-module item.

Anyway uploading that as an alternative patch.

PR #305, opened 2015-12-09
Disable debugger module selector when no modules available

loosebazooka:


loosebazooka:
image

@elharo @patflynn this is what this patch makes it look like.

loosebazooka:
closes #228

patflynn:
Cool. Can we hide the ' please select a module' when there are none to select?

loosebazooka:
The validation system kind of drops that in (and it toggles the Ok/Attach button enable/disable). Is there a message you would rather have? It'll start to get real hacky if we take it out.

elharo:
Do we need the PLease select a mopdule label at all? Or maybe it shoudln't be where it is. Instead it should be the default unselectable item in the popup, and is shown only when we're not showing "No debuggable modules found" If you want to do that in a later PR, I'm OK with this PR; but if not I agree withy Patrick that we should hide the label in the case where no module can be selected.

core-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -77,7 +77,8 @@ clouddebug.stashbuttontext=Stash local changes
 clouddebug.attach=Attach
 clouddebug.nologin=You must be logged in to perform this action.
 clouddebug.noprojectid=Please enter a Project ID.
-clouddebug.nomodule=Please select a Module.
+clouddebug.nomoduleselected=Please select a Module.
```
elharo: Anyone know the UI standards for capitalization? seems weird to have "Module" here and "module" in the very next line.

patflynn: +katie.

patflynn:
I guess if it's a validation error it should be disabled when the field is
disabled. If it's a label I agree with elliotte that it's redundant and can
be removed.

loosebazooka:
Yeah, it's a strange message, but it's part of the Validation system built into Dialogs into intellij.  You can either 1. have a message and disable the ok button or 2. return null and enable the ok button.

We don't have direct control over it.  I mean we could write our own validation system, but right now we use it control the attach button's enabled-state.

patflynn:
In that case for this particular case maybe it's better to have reinforce
the message that instructs the user what to do in the case of no debuggable
modules.

something like "Please select a project with debuggable modules"

patflynn:
Might be a good place to link to Cloud Debugger docs on deploying
debuggable modules..

loosebazooka:
image

here's what it looks like now

patflynn:
ship it!

loosebazooka:
is that an lgtm?

patflynn:
lgtm.

PR #306, opened 2015-12-09
show more errors to the end user

elharo:
@patflynn @loosebazooka This is a possible way to address this issue. For now this is incomplete and just for discussion. That is, does this general approach make sense?

### DIFF
--- a/core-plugin/src/com/google/gct/idea/debugger/CloudDebuggerClient.java
+++ b/core-plugin/src/com/google/gct/idea/debugger/CloudDebuggerClient.java
@@ -117,10 +117,10 @@ public class CloudDebuggerClient {
         }
       }
       catch (IOException ex) {
-        LOG.warn("Error connecting to Cloud Debugger API", ex);
+        LOG.warn("Could not load Google root certificates", ex);
       }
       catch (GeneralSecurityException ex) {
-        LOG.warn("Error connecting to Cloud Debugger API", ex);
+        LOG.warn("Could not load Google root certificates", ex);
       }
 
       if (cloudDebuggerClient != null) {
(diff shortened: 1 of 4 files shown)


patflynn:
It seems like the debugger should be listening for login status change using the GoogleLoginListener and disabling itself upon logout.

elharo:
What would this mean in UI terms? Disable the launch configurations that
use the cloud debugger?

Also, what if the user logs out while a cloud debugger session is open? Do
we close the frame?






On Wed, Dec 9, 2015 at 3:29 PM, Patrick Flynn <notifications@github.com>
wrote:

> It seems like the debugger should be listening for login status change
> using the GoogleLoginListener and disabling itself upon logout.
>
> —
> <https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/306#issuecomment-163383260>
> .
>



-- 
Elliotte Rusty Harold
elharo@macfaq.com

patflynn:
I don't think that you need to disable the launch configs. You just will
see the login widget when your prompted to select a project on launch.

An open cloud debug sessions should definitely be closed (with a prompt)
upon logout.

elharo:
On Wed, Dec 9, 2015 at 4:03 PM, Patrick Flynn <notifications@github.com>
wrote:

>
> An open cloud debug sessions should definitely be closed (with a prompt)
> upon logout.
>

Closing open cloud debug sessions seems to risk losing data. I.e. the user
has received snapshots and may not want them to just disappear.


-- 
Elliotte Rusty Harold
elharo@macfaq.com

patflynn:
That's not great, but not awful considering the user has to explicitly
logout to trigger it.

The alternative is to introduce a disconnected state to the debugger
session, so that the session is clearly disconnected, but still available
to view. (I fear that would mean undoing the work we did earlier to remove
the red x)

PR #312, opened 2015-12-14
Update our IDEA SDK to 15.0.2 to match Android Studio

patflynn:


elharo:
We might want to wait on this. Android studio has been having some troubles with the latest release.

patflynn:
Have they reverted back from 15.0.2?

patflynn:
@elharo  Looking at the Android Studio thread it looks like they're moving forward with 15.0.2.

elharo:
lgtm then

PR #320, opened 2015-12-15
enable selection of pending snapshots

elharo:
@patflynn Last PR had side effect of unselecting pending snapshots so no snapshot was selected. This PR fixes that.

### DIFF
--- a/core-plugin/src/com/google/gct/idea/debugger/ui/CloudDebugHistoricalSnapshots.java
+++ b/core-plugin/src/com/google/gct/idea/debugger/ui/CloudDebugHistoricalSnapshots.java
@@ -262,9 +262,6 @@ public class CloudDebugHistoricalSnapshots extends AdditionalTabComponent
 
     if (breakpointList != null) {
       for (int i = 0; i < breakpointList.size(); i++) {
-        if (breakpointList.get(i).getFinalTime() == null) { // todo: what does this mean?
-          continue;
-        }
         Breakpoint snapshot = myProcess.getCurrentSnapshot();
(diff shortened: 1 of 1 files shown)


core-plugin/src/com/google/gct/idea/debugger/ui/CloudDebugHistoricalSnapshots.java
```diff
@@ -262,9 +262,6 @@ private void onBreakpointsChanged() {
 
     if (breakpointList != null) {
       for (int i = 0; i < breakpointList.size(); i++) {
-        if (breakpointList.get(i).getFinalTime() == null) { // todo: what does this mean?
```
patflynn: why remove this?  It looks like a guard against NPEs , which I think Ben may have seen during testing.

patflynn:
Have @loosebazooka look at these as well since he's been looking at the CDB code.

core-plugin/src/com/google/gct/idea/debugger/ui/CloudDebugHistoricalSnapshots.java
elharo: This was here because the pending snapshots didn't have times. As a side effect of my previous PR, now they do. 

There is an alternate approach I could take here. I could sue a separate code path for pending snapshots that deos not add a time. However that would increase the cyclomatic complexity of the code rather than decrease it as this fix does. epicycles on epicycles.

patflynn:
LGTM

loosebazooka:
I'm not sure I understand what's going on in this CL? Is there a bad associated with it?

loosebazooka:
lgtm

Issue #174, opened 2015-11-03
user-agent needs to always be passed as part of any RPC to GCP


patflynn:
Cloud Repo, Cloud Debugger, AE deployments, etc should always include a useragent specific to our plugin/platform combo.

elharo:
Any preference for what the user agent should be here?

patflynn:
+markpell

Could we mirror browser user-agent structure and construct a composite user
agent from  $platform/$version $plugin/$version?  e.g. Android Studio/1.5
(Windows 10)  GoogleCloudTools/0.9-alpha.

Mark, are you guys using the user-agent for server-side analytics?  Do you
know what the recommended way to get this data into concord is?

patflynn:
You can see how we construct our UA here:
https://cs.corp.google.com/#piper///depot/google3/third_party/py/googlecloudsdk/core/cli.py&l=50

Unfortunately concord does not provide direct access to the UA in the user
facing tables because they are worried about PII.  So we need to get them
to add a column to the table for each field in the UA we want to look at.
We want to look at all of them so we are in the process of getting them to
add them all, but it's a little slow going.  +Vilas Jagannath
<vilasj@google.com> is working on that.

patflynn:
Adding to what Mark said; the current process for getting this data into
concord is:

1) Include it in the UA for service calls (after privacy/security approval).
2) Request concord team to parse this data and include it as a column in
the usage tables.

The turnaround is slow.

Cheers,

Vilas

patflynn:
Is there a standard UA agent format for separating dimensions?

patflynn:
We have been using space separated '/' separated pairs
<https://cs.corp.google.com/#piper///depot/google3/third_party/py/googlecloudsdk/core/cli.py&l=50>
which
conforms to the product token specification
<http://www.ietf.org/rfc/rfc2616.txt>.

patflynn:
+elharo

elharo:
I thiunk we're now done with this unless someone knows of another place where we're connecting over HTTP.

PR #327, opened 2015-12-22
set User-Agent for cloud debugger client

elharo:
@loosebazooka #174


core-plugin/src/com/google/gct/idea/debugger/CloudDebuggerClient.java
```diff
@@ -49,14 +51,15 @@
   private static final Logger LOG = Logger.getInstance(CloudDebuggerClient.class);
   private static final String ROOT_URL = "https://clouddebugger.googleapis.com";
   private static final ConcurrentHashMap<String, Debugger> myDebuggerClientsFromUserEmail =
-    new ConcurrentHashMap<String, Debugger>();
+      new ConcurrentHashMap<String, Debugger>();
+  private static final String USER_AGENT = "GCP Plugin for IntelliJ";
```
loosebazooka: Are we set on this name for this user agent?

google-account-plugin/src/com/google/gct/login/GoogleLogin.java
```diff
@@ -98,7 +98,7 @@ public static synchronized GoogleLogin getInstance() {
   }
 
   @VisibleForTesting
-  static void setInstance(GoogleLogin newInstance) {
+  public static void setInstance(GoogleLogin newInstance) {
```
loosebazooka: Is there a reason this had to be made public for this code change?

core-plugin/src/com/google/gct/idea/debugger/CloudDebuggerClient.java
elharo: Not at all. I just had to pick something. I asked apiary/one platform folks if they had a preferred format, but they haven't responded.

google-account-plugin/src/com/google/gct/login/GoogleLogin.java
elharo: It is now used in CloudDebuggerClientTest in a different package.

core-plugin/testSrc/com/google/gct/idea/debugger/CloudDebuggerClientTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.mockito.runners.MockitoJUnitRunner;
+
+import java.util.LinkedHashMap;
+
+@RunWith(MockitoJUnitRunner.class)
+public class CloudDebuggerClientTest {
+
+  @Before
+  public void setUp() {
+    GoogleLogin mockLogin = Mockito.mock(GoogleLogin.class);
```
loosebazooka: Can we use the MockGoogleLogin class here? So we don't have to make "setInstance" public. I'd rather that not be visible as an option for non-test development.

elharo: I think I tried that. I forget why it didn't work.

elharo:
PTAL @loosebazooka

elharo:
PIng @loosebazooka @patflynn

core-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
```diff
@@ -119,12 +129,19 @@ public synchronized void stopBackgroundListening() {
     return states;
   }
 
-  private static void queryServerForBreakpoints(CloudDebugProcessState state, Debugger client) throws IOException {
```
patflynn: is this just a reformat?

core-plugin/src/com/google/gct/idea/debugger/CloudDebuggerClient.java
```diff
@@ -49,14 +51,16 @@
   private static final Logger LOG = Logger.getInstance(CloudDebuggerClient.class);
   private static final String ROOT_URL = "https://clouddebugger.googleapis.com";
   private static final ConcurrentHashMap<String, Debugger> myDebuggerClientsFromUserEmail =
-    new ConcurrentHashMap<String, Debugger>();
+      new ConcurrentHashMap<String, Debugger>();
+  private static final String USER_AGENT
```
patflynn: I think we should at least include plugin version as well as IDEA platform type and version in the user-agent, and use this spec for formatting http://www.ietf.org/rfc/rfc2616.txt

core-plugin/src/com/google/gct/idea/debugger/CloudDebuggerClient.java
```diff
@@ -105,15 +109,20 @@ public void statusChanged(boolean login) {
           HttpRequestInitializer initializer = new HttpRequestInitializer() {
             @Override
             public void initialize(HttpRequest httpRequest) throws IOException {
+              HttpHeaders headers = new HttpHeaders();
+              headers.setUserAgent(USER_AGENT);
```
patflynn: We want this for all server-side actions including cloud repo syncing and eventually deployment.

core-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
elharo: Looks like it now. I think there was a genuine change in this class at some point, but it went away during code review.

core-plugin/src/com/google/gct/idea/debugger/CloudDebuggerClient.java
elharo: Cloud repo syncing happens, I think, in com.google.gct.idea.git.GcpCheckoutProvider (correct me if I'm wrong about that.) This uses completely different classes so we'll need different code to set the user agent though we can reuse the calculation of the user agent string.

elharo:
PTAL, format of user agent had changed

core-plugin/src/com/google/gct/idea/debugger/CloudDebuggerClient.java
```diff
@@ -105,15 +110,27 @@ public void statusChanged(boolean login) {
...
               credential.initialize(httpRequest);
             }
           };
 
           HttpTransport httpTransport = GoogleNetHttpTransport.newTrustedTransport();
-          cloudDebuggerClient = new Builder(httpTransport, JSON_FACTORY, initializer).setRootUrl(ROOT_URL)
-            .setApplicationName(GoogleLoginUtils.getCurrentPlatformName()).build().debugger();
+          String userAgent = USER_AGENT;
+          try {
+            userAgent += GoogleLoginUtils.getCurrentPlatformName()
```
patflynn: What about putting the user-agent string definition somewhere common?

elharo: Sounds good, though I do wish we could inject it rather than making it static. Where would you suggest? Somewhere in the login package?

patflynn: Maybe common-lib?

I guess it doesn't have to be static. I wouldn't object to that.

elharo: Destaticfying it would requrie making this method non-static first, and then everything that depends on it, and so on. :-(

elharo: com.google.gct.idea.util.PlatformInfo maybe?

patflynn: com.google.gct.idea.util.UserAgent..  Although I'm not too picky.

elharo: Probably don't want to move it to common-lib because that would make common-lib depend on the login package.

patflynn: Why? Can't we move the platform name stuff out of login?  Doesn't seem to
belong there anyway.  Also I'm not convinced it's doing much that's useful.

elharo: Maybe. Let me polay wiht that and see. Who else besides us depends on the login plugin if I move things? (I notice there's a lot of code in there we don't use. I assume someone does.)

patflynn: You can check the android repo but I would hope that people are not using
this util class. Only the login specific API is expected to be in use.

elharo:
@patflynn @loosebazooka PTAL. As suggested I have moved several things into the common-lib module so they can be shared more broadly.

common-lib/src/main/java/com/google/gct/idea/util/PlatformInfo.java
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  @NotNull
+  public static String getCurrentPlatformName() {
+    return getCurrentPlatform().getName();
+  }
+
+  @NotNull
+  public static String getUserAgent() {
+    String userAgent = "GCP-Plugin-for-IntelliJ/0.91";
```
patflynn: could you add a todo + issue to make the version derived by the actual plugin version?

common-lib/src/main/java/com/google/gct/idea/util/LandingPages.java
```diff
@@ -1,4 +1,4 @@
-package com.google.gct.login;
```
patflynn: my hesitation here is that the auth LandingPages really feel specific to the Account plugin. 

I think I introduced the IntelliJPlatform enum but I'm wondering if it's punching its weight here.  Maybe we'd simplify the dependencies a bit by just using PlatformUtils directly.

common-lib/src/main/java/com/google/gct/idea/util/PlatformInfo.java
elharo: #342

common-lib/src/main/java/com/google/gct/idea/util/PlatformInfo.java
```diff
@@ -0,0 +1,32 @@
...
+  public static String getCurrentPlatformName() {
+    return getCurrentPlatform().getName();
+  }
+
+  @NotNull
+  public static String getUserAgent() {
+    // todo read this from build.gradle
+    String userAgent = "GCP-Plugin-for-IntelliJ/0.91";
+    try {
+      userAgent += " " + getCurrentPlatformName()
```
patflynn: shouldn't we cache this? I guess it doesn't change during the lifetime of the application.

elharo: done

patflynn:
Squash and submit!

Thanks for your patience.  Is there a filed issue for adding the useragent to other RPC calls}?

PR #331, opened 2015-12-30
Master 14 merger

loosebazooka:
merge in changes from master


loosebazooka:
@patflynn @elharo

elharo:
I'm confused. What is this? It doesn't look like a merge into master that would require a review.

patflynn:
Auto-merging would be cool. Like that we'll know early when we create a conflict.  I'll file an issue.

Otherwise LGTM

PR #332, opened 2015-12-30
Testing : Fix vfs pointer leak and use automatic mocking

loosebazooka:


loosebazooka:
@elharo @patflynn 

One of these mocks wasn't working with idea14, so updated it in a way we could have common code in both branches.

Also fixed that pesky vfs pointer not disposed thing.

core-plugin/testSrc/com/google/gct/idea/appengine/validation/EndpointTestBase.java
```diff
@@ -75,4 +86,24 @@ public void run() {
       }
     }
   }
+
+  private void removeEndpointsSdkFromProject() {
+    ApplicationManager.getApplication().runWriteAction(new Runnable() {
+      @Override
+      public void run() {
+        final LibraryTable table =
```
elharo: why final? Google style is not to make local varibles final unless required. Is IntelliJ style different?

core-plugin/testSrc/com/google/gct/idea/debugger/ui/CloudDebugHistoricalSnapshotsTest.java
```diff
@@ -61,17 +28,10 @@
...
-import java.util.Comparator;
 import java.util.List;
-import java.util.Set;
 
-import javax.swing.Icon;
 import javax.swing.JComponent;
-import javax.swing.event.HyperlinkListener;
 
 
 // TODO: This test hangs when run as part of the full suite.
```
elharo: Is this still true?

elharo:
lgtm

core-plugin/testSrc/com/google/gct/idea/appengine/validation/EndpointTestBase.java
loosebazooka: There's no reason, I must have just copied it from the "addEndpointsSdk..." code above, will change it back to be consistent.

core-plugin/testSrc/com/google/gct/idea/debugger/ui/CloudDebugHistoricalSnapshotsTest.java
loosebazooka: I don't think so.

loosebazooka: oh I missed the @ignore here, let me see what's going on.

loosebazooka: Still broken, so leaving the ignore in. I think we need to use IdeaTestCase or one of the other built ins so they all the state is managed correctly.

PR #341, opened 2016-01-07
Add a Gradle release plugin for automation of version updating and tag creation

patflynn:
I've added and configured the [researchgate gradle release plugin](https://github.com/researchgate/gradle-release) to our build.gradle.  It should allow those with push access to our master branch to be able to create a release using `gradlew release` from the command line.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,3 @@
-
 buildscript {
     repositories {
         maven { url "https://plugins.gradle.org/m2/" }
@@ -6,6 +5,10 @@ buildscript {
     dependencies {
         classpath 'gradle.plugin.org.jetbrains:gradle-intellij-plugin:0.0.35'
     }
+}
+
+plugins {
+    id 'net.researchgate.release' version '2.3.4'
 }
 
 subprojects {
@@ -18,7 +21,6 @@ subprojects {
     targetCompatibility = javaVersion
     tasks.withType(JavaCompile) { options.encoding = 'UTF-8' }
     group = 'com.google.gct'
-    version = '0.9.1-alpha-SNAPSHOT'
 
     apply plugin: 'org.jetbrains.intellij'
     intellij {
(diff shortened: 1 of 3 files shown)


patflynn:
@elharo

build.gradle
```diff
@@ -80,4 +82,21 @@ subprojects {
...
+release {
+    buildTasks = ['doRelease']
+    // Do not change the tagTemplate value to double quotes or the version will be evaluated before
+    // the SNAPSHOT version is updated to the release version.
+    tagTemplate = 'v$version'
+}
+
+// We aren't building or doing anything interesting for release.
+// We just update the version and generate the tag as CI will handle deployment.
+task doRelease {
```
loosebazooka: maybe we can include this in a different script, that has to be added manually, so there can be less of an accidental release?

gradle.properties
```diff
@@ -1,3 +1,4 @@
 ideaVersion = IC-15.0.2
 javaVersion = 1.6
 ijPluginRepoChannel = alpha
+version = 0.9.1-alpha-SNAPSHOT
```
elharo: Any obvious way to reference this from our Java code? e.g. for User-Agent string?

patflynn: Not yet!

We need to package into the plugin resources similarly to how we include
the usageTrackerProperty.

build.gradle
loosebazooka: https://docs.gradle.org/current/userguide/init_scripts.html

build.gradle
```diff
@@ -80,4 +82,21 @@ subprojects {
...
+    println '===============================!!PLEASE READ!!=================================\n\n' +
+            'IMPORTANT:  The release command will trigger the creation of a new release ' +
+            'in Github, the uploading of binaries to github, and the publishing of our plugin' +
+            'to the IntelliJ plugin repository.\n' +
+            'Do not proceed to execute this command if that is not your intention.\n'
```
elharo: delete "proceed to"

build.gradle
elharo: @patflynn Do you want to do this?

patflynn: Yes. I'm distracted with some other stuff but will revisit this afternoon.

patflynn: I did not use the init_scripts suggestion. Instead I'm only allowing a user to call this if they're on a branch named in a very specific format.

elharo:
lgtm

PR #346, opened 2016-01-11
more detail in user agent

elharo:
@patflynn


patflynn:
And now I've just noticed that we're using the client's 'applicationName' model to set the user-agent.  Do we know this is safe?  How does One Platform treat application name?

elharo:
https://developers.google.com/api-client-library/java/google-api-java-client/reference/1.19.1/com/google/api/client/googleapis/services/json/AbstractGoogleJsonClient.Builder#setApplicationName(java.lang.String)

plus we have a unti test

patflynn:
LGTM

PR #348, opened 2016-01-11
set user agent on oauth request

elharo:
@patflynn #174


patflynn:
LGTM

loosebazooka:
It'd be nice if the actual commit messages were updated before committing.

PR #349, opened 2016-01-11
remove unnecessary dependencies

loosebazooka:


loosebazooka:
This is a cleanup of the build. I think we were adding too much to our build. This partially removes dependencies we can get from intellij and some that we just don't need anymore. When we then add binaries to android studio things will be more compact.

@patflynn

patflynn:
have you tested this change in the UI? Try logging in/out and checking out a GCP repo.

google-account-plugin/build.gradle
```diff
@@ -16,18 +16,26 @@ dependencies {
     compile(project(':common-lib'))
     compile files('lib/google.gdt.eclipse.login.common.jar')
     compile files('lib/google-gct-login-context-ij-pg.jar')
-    compile fileTree(dir: 'third_party/repackaged', include: '*.jar')
+    compile files('third_party/repackaged/google-api-java-client-min-repackaged-1.20.0.jar')
 
-    compile 'com.google.apis:google-api-services-oauth2:v2-rev70-1.18.0-rc'
-    compile 'com.google.http-client:google-http-client-jackson:1.18.0-rc'
+    // TODO : remove jackson1 when we update com.google.gdt.eclipse.login.common.GoogleLoginState
```
elharo: file a bug for this

loosebazooka: file a bug and remove that line? or keep both?

loosebazooka:
@patflynn did login testing, have not tried gcp repo test, will do that now.

loosebazooka:
@patflyn gcp checkin/checkout works, login works.
@elharo Added bug for #350 , not sure if a different comment in code might be more meaningful? perhaps link to the bug instead?

elharo:
At this point, the bug is looking like it's going to lead to a lot of work, so we probably shouldn't include a todo at all, IMHO. But up to you.

build.gradle
```diff
@@ -38,14 +38,10 @@ subprojects {
     configurations.all {
         resolutionStrategy {
             failOnVersionConflict()
-            // Force the IDEA SDK provided version of these dependencies to avoid classloading issues.
-            force 'org.apache.httpcomponents:httpclient:4.4.1', 'com.google.guava:guava:17.0'
```
patflynn: I'd like us to fail if we ever include these provided dependencies.  Is that possible?

loosebazooka: it looks like it would have to be something like this inside the `resolutionStrategy` block, but this feels weird... :             
```
eachDependency { DependencyResolveDetails dependency ->
  if (dependency.requested.group == "org.apache.httpcomponents" && dependency.requested.name == "httpclient" ||
      dependency.requested.group == "com.google.guava" && dependency.requested.name == "guava") {
    throw new GradleException('Dependency should not be included in your build')
  }
}
```

patflynn: feels like an improvement to me.  Are there any other deps that we should exclude?

loosebazooka: I worry we will depend on it too much to make sure we don't add unnecessary additions. Perhaps a dependency count check is better? if total dependencies != X, then are you sure you need this dependency?

patflynn: I'm actually not worried about the dep count (yet).  My concern is that the
'force.. ' statements I put in place were to fix bugs in class version
conflict between local versions of the libraries and IJs.  Removing them
allows developers to inadvertently reintroduce these bugs that are tricky
and only visible at runtime.  So I think we should do one of two things:
1) completely disallow the inclusion of any version of a dependency
provided by IJ
2) force any dependency that gets duplicated to be the same version as IJ

option 1 seems like the better solution since we don't run the risk of
going out of sync with the actual versions provided by the IJ SDK.

loosebazooka: The force was to allow developers to include those dependencies but automatically changing the version. By removing that line the current result of introducing conflicting versions of libraries is a build failure (because failOnVersionConflict() is still in there)

My main concern (and my reason for writing this CL) was to reduce the total number of dependencies, so we would have less resistance including ourselves as a binary.

Anyway, I think we can 1. Not sure if my sample is doing enough checks though.

patflynn: Can you verify that it causes a build failure?  I thought the
gradle-intellij-plugin just manually adds all the IJ jars to the classpath,
and wasn't sure that gradle could detect the version conflict.

loosebazooka: Oh I see. Will do.

loosebazooka:
@patflynn ptal

build.gradle
patflynn: so i guess you were able to verify that the failOnVersionConflict wasn't catching the idea provided libs?

If so, we should file an issue with the gradle-intellij-plugin

patflynn:
LGTM

build.gradle
loosebazooka: It's probably something they'd have to do manually, but I'll file the bug.

loosebazooka:
OOPS forgot to merge down

PR #353, opened 2016-01-12
Fix issue with server breakpoints not being added to the IDE.

### LINKED ISSUE #338: Adding a breakpoint in the Developers Console adds a pending snapshot to the list, but no actual breakpoint on the text editor
Should we support this?

joaoandremartins:
This happens when a breakpoint is created in the server and a debug session is started after. The breakpoint is added to the CDB plugin list of breakpoints, but not to the IDE text editor.

Suggestions for alternatives to cropCloudPath are welcome.

@elharo @patflynn fixes #338


core-plugin/src/com/google/gct/idea/debugger/CloudBreakpointHandler.java
```diff
@@ -274,6 +274,24 @@ public void run() {
...
   /**
+   * Returns a path in the form of <package>/<class_name>.
+   *
+   * <p>This is so that the intermediate map JavaUtil.getFileFromCloudPath() method uses recognises
+   * the path key.
+   *
+   * <p>https://github.com/GoogleCloudPlatform/gcloud-intellij/issues/338
+   */
+  protected String cropCloudPath(String originalCloudPath) {
+    int start;
```
elharo: Let's initialize this here instead of in the if for clarity

core-plugin/src/com/google/gct/idea/debugger/CloudBreakpointHandler.java
```diff
@@ -274,6 +274,24 @@ public void run() {
   }
 
   /**
+   * Returns a path in the form of <package>/<class_name>.
+   *
+   * <p>This is so that the intermediate map JavaUtil.getFileFromCloudPath() method uses recognises
```
elharo: delete "uses"

core-plugin/src/com/google/gct/idea/debugger/CloudBreakpointHandler.java
```diff
@@ -274,6 +274,24 @@ public void run() {
...
+   * the path key.
+   *
+   * <p>https://github.com/GoogleCloudPlatform/gcloud-intellij/issues/338
+   */
+  protected String cropCloudPath(String originalCloudPath) {
```
elharo: why protected?

core-plugin/src/com/google/gct/idea/debugger/JavaUtil.java
```diff
@@ -85,4 +87,12 @@ private static String getCloudPathFromFile(@Nullable Project project, @Nullable
     }
     return getCloudPathFromJavaFile((PsiJavaFile)javaFile);
   }
+
+  // Used to fill ourLocationMaps with test data.
+  protected static void initializeLocationsOnUnitTest(
```
elharo: why protected?
@VisibleForTesting

joaoandremartins: Protected so I can unit test. I added @VisibleForTesting to indicate visibility was relaxed for testing purposes.

core-plugin/src/com/google/gct/idea/debugger/CloudBreakpointHandler.java
```diff
@@ -274,6 +275,25 @@ public void run() {
...
+   * Returns a path in the form of <package>/<class_name>.
+   *
+   * <p>This is so that the intermediate map in JavaUtil.getFileFromCloudPath() method recognises
+   * the path key.
+   *
+   * <p>https://github.com/GoogleCloudPlatform/gcloud-intellij/issues/338
+   */
+  @VisibleForTesting
+  protected String cropCloudPath(String originalCloudPath) {
+    int start = originalCloudPath.indexOf("main/java");
```
patflynn: what do the originalCloudPaths and the JavaUtil map keys look like?

joaoandremartins: JavaUtil: package/class

ninja/joao/guestbook/SignGuestbookServlet.java

originalCloudPath:

jamjavatest/guestbook/src/main/java/ninja/joao/guestbook/SignGuestbookServlet.java

Looking around ProjectFileIndex, I couldn't find any method that would give us anything different and more useful than getPackage() and getClass() joined.

core-plugin/src/com/google/gct/idea/debugger/JavaUtil.java
elharo: would simply making it package access work?

joaoandremartins: It does work. I'll go ahead and remove VisibleForTesting too, unless you have any strong objections.

joaoandremartins:
PTAL. Much as I'd like to completely annihilate JavaUtil, we still need a mapping of every Java file we have. This is explained on the message I sent to the CDB team and cc'ed our team.

elharo:
lgtm

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudBreakpointHandler.java
```diff
@@ -366,7 +373,10 @@ public void registerBreakpoint(
       return;
     }
     SourceLocation location = new SourceLocation();
-    location.setPath(JavaUtil.getCloudPathFromJavaFile((PsiJavaFile)javaFile));
+    // Sending the file as com/package/example/Class.java to Cloud Debugger because it's the
+    // recommended format.
+    location.setPath(
+        ((PsiJavaFile) javaFile).getPackageName().replace(".", "/") + "/" + javaFile.getName());
```
patflynn: Are we sure this cast will succeed?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudBreakpointHandler.java
```diff
@@ -75,6 +79,8 @@
   private final Map<String, XBreakpoint> myIdeBreakpoints = new ConcurrentHashMap<String, XBreakpoint>();
   private final CloudDebugProcess myProcess;
   private PsiManager myPsiManager;
+  private VirtualFileSystem fileSystem =
```
patflynn: wouldn't you prefer to initialize this in the constructor and use constructor injection for tests?  Not sure what's involved in the VFM initialization and what the deps are.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudBreakpointHandler.java
joaoandremartins: It was already being done previously.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudBreakpointHandler.java
joaoandremartins: Given that we will need a RepoToLocalFile map (JavaUtil) regardless, I think I will roll back all of this and try to fix my original problem, which is to somehow extract the package name from jamjavatest/guestbook/src/main/java/ninja/joao/guestbook/SignGuestbookServlet.java and get the corresponding VirtualFile from the map.

joaoandremartins:
@elharo @patflynn @loosebazooka sorry for the repeated reviews. This is closer to my ideal solution, PTAL.

There is an issue with a unit test with mismatching temp directories, that I gotten to the bottom of yet. e.g.,

/private/var/folders/00/162_8000h01000cxqpysvccm004r99/T/unitTest_getFileFromPath_fullPath_525/unitTest0.tmp/path/to/prj/src/main/com/java/package/Class.java

/private/var/folders/00/162_8000h01000cxqpysvccm004r99/T/unitTest_getFileFromPath_fullPath_525/testGetFileFromPath_fullPath0/path/to/prj/src/main/com/java/package/Class.java

I left it as a TODO, since unit testing has drained far too much time already. If any of you has any ideas on how to fix it, please let me know, otherwise I'm hoping for a response at https://intellij-support.jetbrains.com/hc/en-us/community/posts/206101899-Getting-all-java-files-within-a-project?page=1#community_comment_207049975.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudBreakpointHandler.java
```diff
@@ -75,11 +78,15 @@
...
   private PsiManager myPsiManager;
+  private ServerToIDEFileResolver fileResolver;
 
-  public CloudBreakpointHandler(@NotNull CloudDebugProcess process) {
+  public CloudBreakpointHandler(@NotNull CloudDebugProcess process,
+      @NotNull ServerToIDEFileResolver fileResolver) {
     super(CloudLineBreakpointType.class);
     myProcess = process;
-    setPsiManager(PsiManager.getInstance(myProcess.getXDebugSession().getProject()));
+    Project currentProject = myProcess.getXDebugSession().getProject();
```
joaoandremartins: Just noticed this local variable is extra. I'll remove it.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudExecutionStack.java
```diff
@@ -28,17 +27,19 @@
...
- * CloudExecutionStack represents an entire stack for a {@link com.google.api.services.debugger.model.BreakPoint} It
- * stores the individual frames, and also the variables and custom watch expressions.
+ * CloudExecutionStack represents an entire stack for a
+ * {@link com.google.api.services.clouddebugger.model.Breakpoint}. It stores the individual frames,
+ * and also the variables and custom watch expressions.
  */
 public class CloudExecutionStack extends XExecutionStack {
   private final List<CloudStackFrame> myFrames = new ArrayList<CloudStackFrame>();
 
   public CloudExecutionStack(@NotNull Project project,
```
patflynn: If you're stacking the params I'd rather have the first one aligned with the rest.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
```diff
@@ -0,0 +1,124 @@
...
+   * re-attaching to a target because the project state may change between debug sessions.
+   */
+  public void initializeLocationsForProject(@NotNull final Project project, boolean forceRefresh) {
+    if (forceRefresh || !ourLocationMaps.containsKey(project)) {
+      final Map<String, VirtualFile> locationMap =
+          Maps.newHashMap(); // The inner does not have to be concurrent.
+      Collection<VirtualFile> projectJavaFiles = FileBasedIndex.getInstance().getContainingFiles(
+          FileTypeIndex.NAME, JavaFileType.INSTANCE, GlobalSearchScope.projectScope(project));
+      for (VirtualFile virtualFile : projectJavaFiles) {
+        PsiFile javaFile = PsiManager.getInstance(project).findFile(virtualFile);
```
patflynn: why keep a map of this rather than just compute it at request time?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * /home/user/workspace/repo/path/com/my/package/Class.java
+ *
+ * <p>These are methods which are Java specific for the cloud debugger. When we add other languages,
+ * some of this may need to be extracted to an extensionpoint.
+ */
+public class ServerToIDEFileResolver {
+  private final Map<Project, Map<String, VirtualFile>> ourLocationMaps =
+      new ConcurrentHashMap<Project, Map<String, VirtualFile>>();
+  private VirtualFileSystem fileSystem =
```
patflynn: did you consider making this injectable?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
```diff
@@ -0,0 +1,124 @@
...
+    VirtualFile file = fileSystem.findFileByPath(project.getBasePath() + "/" + path);
+    // Try class name with package and class file name.
+    if (file == null) {
+      file = ourLocationMaps.get(project).get(path);
+    }
+    // If we still couldn't find the file, search for possible file name matches and return the
+    // first.
+    // We might want to improve string matching and return more than one possible match.
+    if (file == null) {
+      Collection<VirtualFile> projectJavaFiles = FileBasedIndex.getInstance().getContainingFiles(
```
patflynn: here too it would be nice to be able to inject a bock FileBasedIndex for testing.

google-cloud-tools-plugin/testSrc/com/google/gct/idea/debugger/ServerToIDEFileResolverTest.java
```diff
@@ -0,0 +1,80 @@
...
+    PsiJavaFile psiJavaFile = mock(PsiJavaFile.class);
+    when(psiJavaFile.getPackageName()).thenReturn("com.java.package");
+    when(psiJavaFile.getName()).thenReturn("Class.java");
+    assertEquals(ServerToIDEFileResolver.getCloudPathFromJavaFile(psiJavaFile),
+        "com/java/package/Class.java");
+  }
+
+  // When searching for full file system path.
+  @Test
+  public void testGetFileFromPath_fullPath() {
```
patflynn: I'd rather we @Ignore the test rather than comment out.  You should file an issue to fix.

patflynn:
This work is looking really good BTW.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
joaoandremartins: I did that at first, but then understood that I'd have to inject a PsiManager, etc.

It turns out that the recommended way of testing a project with Psi files is by using the mocks provided by IDEA, that I'm using in my tests.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
joaoandremartins: Good question. Because we'd have to replicate this logic in multiple places, so I decided to factor it out here.

We also don't know when the paths CDB sends us are going to change. If they do, we just have to make changes in one place.

That way, this class is responsible for matching a provided path with a file, using a number of different heuristics. This is similar to what the CDB team is doing. They're searching for the best match for a file, we're also doing so but in O(1) most of the times.

patflynn: Couldn't you centralize the transform in one method and call that from everywhere you check the map?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
patflynn: I think this is fine for now. We'll need to come up with a principle when our lightweight mocking vs IJ heavy-weight mocking applies. I'll need to spend more time writing tests before I'm strongly opinionated on that.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
joaoandremartins: Do you mean initializeLocationsForProject shouldn't be called explicitly, but rather only when a user queries the map and the map for a project is empty?

I didn't want to change that too much in order not to introduce bugs, but I can come up with something so we don't have to explicitly call initialize.

patflynn: hold on. let's chat in person to make sure I'm not missing something.

joaoandremartins:
@patflynn PTAL. I decided to call the resolver non-statically so that I didn't have to spend a huge amount of time fixing unit tests (in particular, I'd have to completely re-do CloudBreakpointHandlerTest). I think it's better for testing, too.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
```diff
@@ -0,0 +1,109 @@
...
+   * file system. Then, it tries the full class name form (com/google/gct/idea/debugger/
+   * CloudDebugProcess.java). Finally, it searches for possible file matches within the project.
+   */
+  public VirtualFile getFileFromPath(@NotNull Project project, @NotNull String path) {
+    // Try the relative full project path.
+    VirtualFile file = FILE_SYSTEM.findFileByPath(project.getBasePath() + "/" + path);
+    // Try class name with package and class file name.
+    if (file == null) {
+      // Produce "com.java.package" from "com/java/package/Class.java".
+      String[] tokens = path.split("/");
```
patflynn: This logic looks like a nice candidate to extract as a method and test.

joaoandremartins: PTAL.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+    return file;
+  }
+
+  /**
+   * Produces a Java package name from a file path.
+   *
+   * <p>Example: returns "com.java.package" from "com/java/package/Class.java".
+   */
+  static String getPackageFromPath(String path) {
```
patflynn: @VisibleForTesting

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
```diff
@@ -0,0 +1,134 @@
...
+    String[] tokens = path.split("/");
+    StringBuilder packageBuilder = new StringBuilder();
+    if (tokens.length > 1) {
+      if(tokens[0].length() > 0) {
+        packageBuilder.append(tokens[0]);
+      }
+      for (int nToken = 1; nToken < tokens.length - 1; nToken++) {
```
patflynn: do you know about Guava Joiner?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
```diff
@@ -0,0 +1,134 @@
...
+    return packageBuilder.toString();
+  }
+
+  /**
+   * Produces a class name from a file path.
+   *
+   * <p>Example: returns "Class" from "com/java/package/Class.java".
+   */
+  static String getClassNameFromPath(String path) {
+    String[] tokens = path.split("/");
```
patflynn: if you just want the last bit of the path why not substring based off of lastIndexOf(/)?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
joaoandremartins: Didn't know about it, thanks! Unfortunately, Joiner doesn't cope well with empty strings, it only offers explicit support for nulls.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
joaoandremartins: Good point. PTAL

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * Produces a class name from a file path.
+   *
+   * <p>Example: returns "Class" from "com/java/package/Class.java".
+   */
+  @VisibleForTesting
+  static String getClassNameFromPath(String path) {
+    if (path.indexOf('.') != -1 && path.lastIndexOf("/") < path.lastIndexOf(".")) {
```
patflynn: if this is false should you throw an exception?

patflynn:
LGTM. Merge once you decide what to do about error handling of an invalid path.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
joaoandremartins: Because the path can't be a class? I'm trying to be as flexible as possible, to accept anything that CDB sends. For breakpoints, CDB tries to match whatever path the client sends and returns that same path. It might be possible that a path is defined as /foo/bar/Class and CDB matches it to a file, in which case this support would be necessary.

Issue #354, opened 2016-01-12
rename the 'Google Cloud Tools Core'  plugin name to something sensible.


patflynn:
Both the 'core-plugin' and the 'Google Cloud Tools Core' plugin name are bad names.

1) Users looking at our plugin in the IJ repo aren't going to understand what distinction 'core' makes, since there are no other Google Cloud Tools plugins.
2) If we eventually refactor our plugins and break them out into pieces, for instance 'Cloud Debugger' plugin, 'Cloud Repo' plugin, 'App Engine' plugin, we won't want to have to make the user reinstall all these new plugins.  We'd rather the first install to imply the super-set of GCP tooling plugins, so that we can refactor it to depend on these new factored out plugins, and hopefully things will still work.
3) If we never do componentize our plugin, it's likely to stay a monolith and support multiple IJ SKUs through conditional plugin.xml includes. 

All that to say that we should probably just rename the plugin to 'Google Cloud Tools'..

patflynn:
@GoogleCloudPlatform/cloud-tools-for-intellij, a penny for your thoughts.

elharo:
Makes sense. "Google Cloud Tools", "Google Cloud Platform Tools",
or "Google Cloud Platform Plugin"

On Tue, Jan 12, 2016 at 5:12 PM, Patrick Flynn <notifications@github.com>
wrote:

> Both the 'core-plugin' and the 'Google Cloud Tools Core' plugin name are
> bad names.
>
> 1) Users looking at our plugin in the IJ repo aren't going to understand
> what distinction 'core' makes, since there are no other Google Cloud Tools
> plugins.
> 2) If we eventually refactor our plugins and break them out into pieces,
> for instance 'Cloud Debugger' plugin, 'Cloud Repo' plugin, 'App Engine'
> plugin, we won't want to have to make the user reinstall all these new
> plugins. We'd rather the first install to imply the super-set of GCP
> tooling plugins, so that we can refactor it to depend on these new factored
> out plugins, and hopefully things will still work.
> 3) If we never do componentize our plugin, it's likely to stay a monolith
> and support multiple IJ SKUs through conditional plugin.xml includes.
>
> All that to say that we should probably just rename the plugin to 'Google
> Cloud Tools'..
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/GoogleCloudPlatform/gcloud-intellij/issues/354>.
>



-- 
Elliotte Rusty Harold
[...]

PR #355, opened 2016-01-13
Rename cloud tools plugin

patflynn:
Renaming our plugin from 'Google Cloud Tools Core' to 'Gogle Cloud Tools'.

Relates to #354


loosebazooka:
were you gonna change the directory name?

patflynn:
not in this PR

loosebazooka:
how about core-plugin/build.gradle : pluginName = 'google-cloud-tools-core'?

patflynn:
good catch. fixed.

loosebazooka:
I think actually as a consequence of that change, the module name changes in intellij, which might have some consequences that I'm not aware of :

like : core-plugin/.idea/runConfigurations/GctCorePlugin.xml <--- what is this runconfig anyway?

loosebazooka:
if it still works after removal, I'd be onboard with that

patflynn:
that should be a separate PR though.

loosebazooka:
sure, #357

lgtm

Issue #309, opened 2015-12-10
Avoid flashing in Attach dialog


elharo:
When the below dialog pops up it often flashes through several variations with different error labels before settling down to the final state shown here.

![image](https://cloud.githubusercontent.com/assets/1005544/11720779/da15e6c4-9f2e-11e5-907a-eef7549924ba.png)

In particular:

1. "Please select a module" (in red) is shown and then vanishes.
2. The left button is initially a disabled "Attach" button instead of the "Continue Session" button shown here.

This is distracting to the user. I suspect we're loading the modules asynchronously, and once they come in we change the widgets we show.

We should try to figure out what the state will be sooner so the dialog doesn't flash like that.

patflynn:
There should be a loading state with a clear visual indicator.

etanshaul:
It seems elharo is correct that the modules are being loaded asynchronously leading to this issue:

The combobox is currently being populated by a separate thread (see ProjectDebuggeeBinding.refreshDebugTargetList). Furthermore, the since dialog validation happens via a timer, it *may* happen that validation occurs on a fully displayed dialog with the combobox still unpopulated - in this case the error message will display until the data gets loaded and the next validation cycle occurs.  This explains the current non-deterministic behavior where the flashing message only occurs sometimes.

Looking into a fix.

patflynn:
awesome. Thanks for the update.

etanshaul:
The dialog class (and our GUI classes in general) is proving hard to unit test without significant refactoring outside the scope of this issue. Either we undertake the necessary refactoring as we encounter these issues to improve test coverage, or we separate this effort into separate issue(s). 

Any thoughts on approaches to handle these scenarios?

patflynn:
I don't think there's a hard and fast rule.  You have some luxury of time
right now, so if you want to undertake the refactoring work, please do so.
Although send out a doc or mail explaining the issue, and how you want to
address it first.  Otherwise do what you can, don't make things worse, and
file a refactoring issue, that we can prioritize against our other work.

PR #362, opened 2016-01-14
sync error messages for no module selected

### LINKED ISSUE #361: Inconsistent error messaging in Attach Dialog for module selection
The error messaging in the Attach Dialog for the same state (no debuggable modules) varies depending on the sequence of events the user takes to arrive at that state.

1. Create a new Cloud Debug run configuration. Give it a name. Select a Project with no debuggable modules.
2. Run -> Debug. Notice the error message in dialog says "Please select a project with debuggable modules.":

image

3. In the same dialog, select a different Project that *does* have debuggable modules.

4. With the dialog still open, again select the Project you chose in step 1. Notice the error message now says "Please select a module":

image

This dialog should display messaging consistently for a given state regardless of previous selections. The message that appears should probably be "Please select a project with debuggable modules." for all steps used because, in this example, the module dropdown is empty.

elharo:
fix #361

### DIFF
--- a/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
+++ b/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
@@ -79,7 +79,6 @@ clouddebug.nologin=You must be logged in to perform this action.
 clouddebug.noprojectid=Please enter a Project ID.
 clouddebug.selectvalidproject=Please select a project with debuggable modules.
 clouddebug.nomodulesfound=No debuggable modules found.
-clouddebug.nomoduleselected=Please select a module.
 clouddebug.readonly=<html><b>Note:</b><i>The Cloud Debugger is read-only and won't suspend or modify your application.</html>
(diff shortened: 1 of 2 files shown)


patflynn:
So the way I understand this change is that the 'please select a module' validation error, is not needed, since if there are valid modules, one will be preselected.  Is that right?

etanshaul:
@patflynn yes, with some caveats: in the "normal" case where modules are retrieved (or not found), without exceptions, then one of two things will happen: either the appropriate debuggable module will be selected, *or* the module dropdown will be populated with an item "No debuggable modules found" but the dropdown will be disabled:

image

In the case I demonstrated for #361, the module dropdown is not populated at all (not even in a disabled state with an error item) because of the module retrieval logic is throwing an exception and the error handling isn't handled the same way for this case. 

As part of #309 I am working on cleaning up these things. Regarding this specific issue, I think @elharo change makes sense.

patflynn:
LGTM

PR #367, opened 2016-01-14
Fix to allow checking for local repo existence to display appropriate…

### LINKED ISSUE #366: CloudAttachDialog not properly checking if local repo exists
Attempting to attach a debugger to a module with no corresponding local repository allows user to attach with no warning.

To reproduce set up a project with a module on appengine with no local version of the repository, then select it from the IJ dialog.

etanshaul:
Fixes #366

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
@@ -273,7 +273,7 @@ public class CloudAttachDialog extends DialogWrapper {
       myWarningLabel2.setVisible(false);
       myInfoPanel.setVisible(true);
     }
-    else if (mySyncResult.needsSync() && mySyncResult.getTargetSyncSHA() == null) {
+    else if (mySyncResult.getTargetSyncSHA() == null) {
       setOKButtonText(isContinued()
           ? GctBundle.getString("clouddebug.continueanyway")
           : GctBundle.getString("clouddebug.attach.anyway"));


etanshaul:
I signed it!

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
```diff
@@ -273,7 +273,7 @@ else if (mySyncResult.needsSync()) {
       myWarningLabel2.setVisible(false);
       myInfoPanel.setVisible(true);
     }
-    else if (mySyncResult.needsSync() && mySyncResult.getTargetSyncSHA() == null) {
+    else if (mySyncResult.getTargetSyncSHA() == null) {
```
loosebazooka: so if getTargetSyncSHA is null, needsSync is not set?

etanshaul: Check out the `else if` clause right about the one I modified:

`else if (mySyncResult.needsSync())`

If needsSync is true, the clause I modified would never have been reached in the first place.

That aside, from debugging through the execution I noticed that for the module with the missing local repo, the SHA was null and needsSync was false causing no warning to be shown.  Looking at ProjectRepositoryValidator.checkSyncStashState() needsSync will always be false if the SHA is null.

loosebazooka:
cool, lgtm

patflynn:
You shouldn't have needed to sign it.. @loosebazooka, did we ever get
everybody added to the GCP org?

loosebazooka:
We fixed this, he didn't actually sign it. It appears it was because his
corp email address wasn't added as one of his github email addresses.

PR #368, opened 2016-01-14
Add sourcejar create task

loosebazooka:


loosebazooka:
@patflynn

source : https://discuss.gradle.org/t/how-to-configure-gradle-to-build-a-source-jar/3209

build.gradle
```diff
@@ -22,6 +22,12 @@ subprojects {
     tasks.withType(JavaCompile) { options.encoding = 'UTF-8' }
     group = 'com.google.gct'
 
+    task jarSources(type:Jar) {
```
patflynn: isn't src jar generation available by default?

loosebazooka: I can't really find any other way to do this. But by default no sources jar is generated.

patflynn: LGTM

PR #372, opened 2016-01-19
only warn user about local repo SHA mismatch if sync is needed

etanshaul:
We only want to warn the user about a missing local repo with matching SHA if a sync is needed.  Issue #366


elharo:
lgtm

PR #375, opened 2016-01-19
Fixes #374

### LINKED ISSUE #374: Fix breakpoint UI presentation
When a breakpoint cannot be set on the server side (e.g. the user has revoked IDEA's access or the user has been logged out), the breakpont's icon remains in unconfirmed state (solid blue circle) instead of invalid state (blue circle with a red X). 
Also the error message giving the reason for the failure is not visible only in the IDEA log instead of in the tooltip of the breakpoint.

akerekes:
This is just an initial version to validate the solution, I'm going to add tests if this is accepted.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudBreakpointHandler.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudBreakpointHandler.java
@@ -70,7 +70,8 @@ import javax.swing.SwingUtilities;
  */
 public class CloudBreakpointHandler
     extends XBreakpointHandler<XLineBreakpoint<CloudLineBreakpointProperties>> {
-  private static final Key<String> CLOUD_ID = Key.create("CloudId");
+  public static final Key<String> CLOUD_ID = Key.create("CloudId");
+
   private static final Logger LOG = Logger.getInstance(CloudBreakpointHandler.class);
   private final Map<String, XBreakpoint> myIdeBreakpoints = new ConcurrentHashMap<String, XBreakpoint>();
   private final CloudDebugProcess myProcess;
@@ -187,7 +188,7 @@ public class CloudBreakpointHandler
               (CloudLineBreakpointType.CloudLineBreakpoint) cloudIdeBreakpoint;
           cloudIdeLineBreakpoint.setVerified(true);
           cloudIdeLineBreakpoint.setErrorMessage(null);
-          cloudIdeLineBreakpoint.updateUI();
+          myProcess.updateBreakpointPresentation(cloudIdeLineBreakpoint);
         }
         continue;
       }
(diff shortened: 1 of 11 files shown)


google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudBreakpointHandler.java
```diff
@@ -341,7 +341,7 @@ public void registerBreakpoint(
       BreakpointManager.getJavaBreakpoint(xIdeBreakpoint);
     if (!(cloudIdeBreakpoint instanceof CloudLineBreakpointType.CloudLineBreakpoint)) {
       LOG.error("breakpoint was not of the correct type to create on the cloud.  It was not a " +
-                "CloudLineBreakpoint");
+                "CloudLineBreakpoint"); //TODO akerekes: use class.getName() instead of hard coded name
```
elharo: I'm not sure this todo is necessary.

akerekes: If the class is renamed, then I'd rather not rely on the refactoring tool to take care of it, regardless how much I trust it.

elharo: Yoda rule: Do or do not. There is no TODO.

loosebazooka: The refactoring tool in Intellij is pretty great.

patflynn: I think you can use your judgement on this one.  Also agree with @elharo that there's no need for a TODO, this is small enough that you should just do it, or not do it.

akerekes: I know, I like it, but still seems fragile depending on the tool to take care of it. Anyway, I'll remove the TODO, and submit a separate PR for the change, I don't want to mix it into this one.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
```diff
@@ -341,15 +343,29 @@ else if (breakpoint.getIsFinalState() == Boolean.TRUE) {
...
-            cloudBreakpoint.updateUI();
+            updateBreakpointPresentation(cloudLineBreakpoint);
           }
         }
       }
     }
   }
 
+  void updateBreakpointPresentation(CloudLineBreakpoint cloudLineBreakpoint) {
+    final XBreakpointManager manager = XDebuggerManager.getInstance(
```
patflynn: break at the higher semantic level.

akerekes:
ptal @elharo @loosebazooka @patflynn

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/UpdateIdeWithBreakpointAction.java
```diff
@@ -0,0 +1,73 @@
...
+
+import com.intellij.debugger.ui.breakpoints.BreakpointManager;
+import com.intellij.openapi.vfs.VirtualFile;
+import com.intellij.xdebugger.breakpoints.XBreakpoint;
+import com.intellij.xdebugger.breakpoints.XBreakpointManager;
+import com.intellij.xdebugger.breakpoints.XLineBreakpoint;
+
+import java.util.Map;
+
+class UpdateIdeWithBreakpointAction implements Runnable {
```
patflynn: I wouldn't call this ..Action since that's a naming pattern to indicate AnAction subclasses.

akerekes: Yeah, I was contemplating on it as well, but calling it Runnable would have been redundant, Action seemed a bit better. Maybe `UpdateIdeWithBreakpointTask`?

patflynn: It looks like Task is also a jetbrains concept
(com.intellij.openapi.progress).  I'm not full of great idea about this..
maybe DoUpdateIdeWithBreakpointAction?

patflynn: corrected suggestion to be 'DoUpdateIdeWithBreakpoint'

akerekes:
PTAL @patflynn @elharo @loosebazooka

patflynn:
make sure to squash your commits before merging.

LGTM

Issue #378, opened 2016-01-20
Improve naming in the cloud debug dialog


etanshaul:
Currently we have names like:

myWarningLabel
myWarningLabel2
etc.

Refactor these to make them more meaningful.

PR #379, opened 2016-01-20
variable name refactoring

etanshaul:
See #378.  Simple name changes.

- Removed "my" prefix from instance variables
- Changed the name altogether when a more meaningful one was possible

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
@@ -84,38 +84,38 @@ import git4idea.repo.GitRepository;
 public class CloudAttachDialog extends DialogWrapper {
   private static final Logger LOG = Logger.getInstance(CloudAttachDialog.class);
 
-  private final Project myProject;
-  private final ProjectDebuggeeBinding myWireup;
-  private JComboBox myDebuggeeTarget; // Module selector
-  private ProjectSelector myElysiumProjectId; // Project combo box
-  private JBLabel myInfoPanel;
-  private String myOriginalBranchName;
-  private JPanel myPanel;
-  private CloudDebugProcessState myProcessResultState;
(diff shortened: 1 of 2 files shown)


elharo:
lgtm

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
```diff
@@ -129,19 +129,19 @@ public void actionPerformed(ActionEvent e) {
...
     });
 
-    myWarningLabel.setVisible(false);
-    myWarningLabel.setFont(new Font(myWarningLabel.getFont().getName(), Font.BOLD,
-                                    myWarningLabel.getFont().getSize() - 1));
-    myWarningLabel.setForeground(JBColor.RED);
-    myWarningLabel2.setVisible(false);
-    myWarningLabel2.setFont(new Font(myWarningLabel2.getFont().getName(), Font.PLAIN,
-                                     myWarningLabel.getFont().getSize() - 1));
-    myWarningLabel2.setText(GctBundle.getString("clouddebug.sourcedoesnotmatch"));
```
patflynn: since these labels always have the same error string in them, I'm thinking they should be named after the error they produce rather than the more generic name.

etanshaul: actually, the warning label is overridden for several different contexts. So its not just 1 string that will always appear in it. Also, myWarningLabel was just the "title" (e.g. "Warning: ") - I changed it to warningHeader / warningMessage to make it clearer

etanshaul: to be more concise, it can have things like "..source does not match..", "..Could not locate a local / remote Git repository.." and so on. I guess we could call it something like "repositoryWarningMessage" or something like that, but that's a bit of a stretch.

patflynn:
Aha. I missed that. In that case LGTM

PR #380, opened 2016-01-20
remove unused methods

elharo:
@etanshaul


etanshaul:
It looks like TestUtils is used only by BasePluginTestCase, which itself is never used. Can we just remove both classes?

loosebazooka:
@patflynn wrote this as a framework for testing, probably should ask him what's going on here.

patflynn:
@etanshaul @elharo please don't remove this code. It exists to provide a lightweight mocking ability for tests that depend on IJ injected services.

PR #384, opened 2016-01-20
build rules to create archives for android studio

loosebazooka:


loosebazooka:
@patflynn

scripts/android-studio-build.gradle
```diff
@@ -0,0 +1,27 @@
+task buildAndroidStudioRelease(type: Zip) {
+    archiveName 'androidStudioRelease.zip'
```
patflynn: can we include our version number in the release zip name?

loosebazooka: well really this is a zip for us to transfer it to their codebase, versions are still in the jars BUT, yes I can do that.

patflynn: Thanks!!

scripts/android-studio-build.gradle
```diff
@@ -0,0 +1,27 @@
...
+    destinationDir = buildDir
+    def pluginDirs = [":google-cloud-tools-plugin", ":google-account-plugin"]
```
patflynn: I'm noticing when I try out this script that the google-cloud-tools folder has the plugin sources.jar but not the actual code.jar.  google-account-plugin looks good though.

One other thing is that common-lib is missing the sources jar, and most of the lib jars don't have srcjars. Not sure if AS want that.

loosebazooka:
oh yeah I mucked up the task dependencies, lemme check.

loosebazooka:
This is a particularly gnarly looking build script, I'm sorry. @patflynn

scripts/android-studio-build.gradle
```diff
@@ -0,0 +1,140 @@
+def jarLicenseMap = [
+        "GoogleFeedback.jar" : "apache2",
```
patflynn: so there's no way to query this from maven central?

scripts/android-studio-build.gradle
```diff
@@ -0,0 +1,140 @@
...
+            // includes jars
+            project.tasks.jar.outputs.files.each { outputFile ->
+                from (outputFile) {
+                    into "${project.intellij.pluginName}/lib"
+                }
+            }
+            // include sources jar
+            project.tasks.jarSources.outputs.files.each { outputFile ->
+                from (outputFile) {
+                    into "${project.intellij.pluginName}/lib/src"
```
patflynn: typically don't src jars just get included along side the binaries?

PR #386, opened 2016-01-22
standardize on term snapshot location for user messages

elharo:
@patflynn


akerekes:
Shouldn't we also change the message keys to use snapshot instead of breakpoint? E.g.: https://github.com/GoogleCloudPlatform/gcloud-intellij/blob/messages/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties#L99

elharo:
No, they're not user visible and various classes in the IDEA codebase we inheit from use the word "breakpoint"

elharo:
ping

patflynn:
LGTM

PR #387, opened 2016-01-22
skip validation until asynchronous module loading has completed to av…

### LINKED ISSUE #309: Avoid flashing in Attach dialog

etanshaul:
…oid intermittent error messages from being displayed

fixes #309 

- Refactored CloudAttachDialog so that more components are testable
- Update CloudAttachDialog to skip validation until async module loading has completed
- Updated ProjectDebuggeeBinding to include a default disabled module with a warning when non are found + trivial variable renaming
- Tests

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
@@ -74,7 +74,8 @@ public class CloudAttachDialog extends DialogWrapper {
 
   private final Project project;
   private final ProjectDebuggeeBinding wireup;
-  private JComboBox debuggeeTarget; // Module selector
+  private ProjectRepositoryValidator projectRepositoryValidator;
+  private JComboBox moduleSelector;
   private ProjectSelector elysiumProjectSelector; // Project combo box
   private JBLabel infoMessage;
   private String originalBranchName;
@@ -87,7 +88,8 @@ public class CloudAttachDialog extends DialogWrapper {
   private JBLabel warningHeader;
   private JBLabel warningMessage;
 
-  public CloudAttachDialog(@NotNull Project project) {
+  public CloudAttachDialog(@NotNull Project project,
+                           @VisibleForTesting ProjectDebuggeeBinding wireup) {
     super(project, true);
 
     this.project = project;
@@ -137,12 +139,13 @@ public class CloudAttachDialog extends DialogWrapper {
     }
     BasicAction.saveAll();
 
-    wireup = new ProjectDebuggeeBinding(elysiumProjectSelector, debuggeeTarget);
-    debuggeeTarget.setEnabled(false);
-    debuggeeTarget.addActionListener(new ActionListener() {
+    this.wireup = wireup == null ?
+        new ProjectDebuggeeBinding(elysiumProjectSelector, moduleSelector, getOKAction()) : wireup;
+    moduleSelector.setEnabled(false);
+    moduleSelector.addActionListener(new ActionListener() {
       @Override
       public void actionPerformed(ActionEvent e) {
-        if(debuggeeTarget.isEnabled()) {
+        if(moduleSelector.isEnabled()) {
           buildResult();
           checkSyncStashState();
         }
@@ -191,14 +194,14 @@ public class CloudAttachDialog extends DialogWrapper {
           elysiumProjectSelector);
     }
 
-    if (!debuggeeTarget.isEnabled()) {
+    if (!moduleSelector.isEnabled() && moduleSelector.getItemCount() > 0) {
       return new ValidationInfo(GctBundle.getString("clouddebug.selectvalidproject"),
           elysiumProjectSelector);
     }
 
-    if (debuggeeTarget.getSelectedItem() == null) {
+    if (moduleSelector.getSelectedItem() == null && moduleSelector.getItemCount() > 0) {
       return new ValidationInfo(GctBundle.getString("clouddebug.selectvalidproject"),
-          debuggeeTarget);
+          moduleSelector);
     }
 
     return null;
@@ -219,8 +222,8 @@ public class CloudAttachDialog extends DialogWrapper {
   }
 
   @VisibleForTesting
-  JComboBox getDebuggeeTarget() {
-    return debuggeeTarget;
+  JComboBox getModuleSelector() {
+    return moduleSelector;
   }
 
   @VisibleForTesting
@@ -231,6 +234,11 @@ public class CloudAttachDialog extends DialogWrapper {
   @VisibleForTesting
   JLabel getWarningMessage() {
     return warningMessage;
+  }
+
+  @VisibleForTesting
+  void setProjectRepositoryValidator(ProjectRepositoryValidator projectRepositoryValidator) {
+    this.projectRepositoryValidator = projectRepositoryValidator;
   }
 
   private void buildResult() {
@@ -250,7 +258,10 @@ public class CloudAttachDialog extends DialogWrapper {
       LOG.error("unexpected result state during a check sync stash state");
       return;
     }
-    syncResult = new ProjectRepositoryValidator(processResultState).checkSyncStashState();
+
+    syncResult = projectRepositoryValidator == null ?
+        new ProjectRepositoryValidator(processResultState).checkSyncStashState() :
+          projectRepositoryValidator.checkSyncStashState();
 
     if (syncResult.needsStash() && syncResult.needsSync()) {
       setOKButtonText(isContinued()
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/ProjectDebuggeeBinding.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/ProjectDebuggeeBinding.java
@@ -37,6 +37,7 @@ import org.jetbrains.annotations.Nullable;
 import java.io.IOException;
 import java.util.Map;
 
+import javax.swing.Action;
 import javax.swing.JComboBox;
 import javax.swing.SwingUtilities;
 import javax.swing.event.DocumentEvent;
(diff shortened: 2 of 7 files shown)


google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
```diff
@@ -137,7 +139,8 @@ public void actionPerformed(ActionEvent e) {
     }
     BasicAction.saveAll();
 
-    wireup = new ProjectDebuggeeBinding(elysiumProjectSelector, debuggeeTarget);
+    this.wireup = wireup == null ?
```
elharo: Personal pref, but I find if-else much easier to read than a ternary operator

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
```diff
@@ -250,7 +258,10 @@ private void checkSyncStashState() {
       LOG.error("unexpected result state during a check sync stash state");
       return;
     }
-    syncResult = new ProjectRepositoryValidator(processResultState).checkSyncStashState();
+
+    syncResult = projectRepositoryValidator == null ?
```
elharo: ditto

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/ProjectDebuggeeBinding.java
```diff
@@ -87,13 +91,13 @@ public void treeStructureChanged(TreeModelEvent e) {
 
   @NotNull
   public CloudDebugProcessState buildResult(Project project) {
-    Long number = myElysiumProjectId.getProjectNumber();
+    Long number = elysiumProjectId.getProjectNumber();
     String projectNumberString = number != null ? number.toString() : null;
-    DebugTarget selectedItem = (DebugTarget)myDebugeeTarget.getSelectedItem();
+    DebugTarget selectedItem = (DebugTarget) debugeeTarget.getSelectedItem();
```
elharo: debuggeeTarget? I'm not sure.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/ProjectDebuggeeBinding.java
```diff
@@ -148,15 +153,14 @@ public void run() {
                 DebugTarget targetSelection = null;
 
                 if (debuggees == null || debuggees.getDebuggees() == null || debuggees.getDebuggees().isEmpty()) {
-                  myDebugeeTarget.setEnabled(false);
-                  myDebugeeTarget.addItem(GctBundle.getString("clouddebug.nomodulesfound"));
+                  disableModuleSelector();
                 }
                 else {
-                  myDebugeeTarget.setEnabled(true);
+                  debugeeTarget.setEnabled(true);
```
elharo: debuggeeTarget

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/ProjectDebuggeeBinding.java
```diff
@@ -167,29 +171,47 @@ public void run() {
...
                       }
                     }
-                    myDebugeeTarget.addItem(item);
+                    debugeeTarget.addItem(item);
+                    okAction.setEnabled(true);
                   }
                 }
                 if (targetSelection != null) {
-                  myDebugeeTarget.setSelectedItem(targetSelection);
+                  debugeeTarget.setSelectedItem(targetSelection);
```
elharo: debuggeeTarget

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/ProjectDebuggeeBinding.java
```diff
@@ -167,29 +171,47 @@ public void run() {
...
-                        myDebugeeTarget.removeItem(existing);
+                        debugeeTarget.removeItem(existing);
                       }
                       perModuleCache.put(key, item);
                     }
-                    if (myInputState != null && !Strings.isNullOrEmpty(myInputState.getDebuggeeId())) {
-                      assert myInputState.getDebuggeeId() != null;
-                      if (myInputState.getDebuggeeId().equals(item.getId())) {
+                    if (inputState != null && !Strings.isNullOrEmpty(inputState.getDebuggeeId())) {
+                      assert inputState.getDebuggeeId() != null;
```
elharo: This assert seems meaningless.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/ProjectDebuggeeBinding.java
```diff
@@ -167,29 +171,47 @@ public void run() {
...
+
           LOG.error("Error listing debuggees from Cloud Debugger API", ex);
         }
       }
     });
   }
+
+  @SuppressWarnings("unchecked")
+  private void disableModuleSelector() {
+    debugeeTarget.setEnabled(false);
```
elharo: debuggeeTarget

google-cloud-tools-plugin/testSrc/com/google/gct/idea/debugger/ui/CloudAttachDialogTest.java
```diff
@@ -80,37 +81,67 @@ public void testErrorWhenNoProjectSelected() {
...
+
+    assertNull(error);
+    assertFalse(moduleSelector.isEnabled());
+    assertNull(moduleSelector.getSelectedItem());
+  }
 
-    // Module selector
-//    assertTrue(moduleSelector.isEnabled());
+  private void initDialog() {
+    dialog = new CloudAttachDialog(this.getProject(), binding);
```
elharo: seems like reassigning the field might have unintended consequences between setUp and tearDown; e.g. one or the other not getting closed

etanshaul: In this PR I actually moved the instantiation of the dialog out of setUp so its happening on a per-test basis. So there shouldn't be an reassignment here unless I misunderstood your comment.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/ProjectDebuggeeBinding.java
etanshaul: You mean the name? Yea, I wasn't sure myself either - this came from preexisting code. I'll change it to moduleSelector which is what it really is.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudAttachDialog.java
etanshaul: Unless there is a strong opinion on this, I'm going to leave this specific case with the ternary operator just because the conditional is being used for assignment and the ternary (in my opinion) makes it slightly cleaner since it returns a value.

etanshaul:
Updated based on initial comments from @elharo.
- Renamed debuggeeTarget
- Got rid of pointless assert

google-cloud-tools-plugin/testSrc/com/google/gct/idea/debugger/ui/CloudAttachDialogTest.java
elharo: In that case, make it a local variable, not a field.

etanshaul: I considered that, but given that tearDown() needs to clean up any instantiated dialog I didn't see a clean way to do this without having the dialog be a field. If you see a better pattern for this please let me know.

etanshaul: Actually, each test itself could be responsible for the cleanup since each test is already creating the dialog. Does this make more sense to you?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/ProjectDebuggeeBinding.java
```diff
@@ -48,24 +49,28 @@
...
-  private CredentialedUser myCredentialedUser = null;
-  private CloudDebugProcessState myInputState;
-
-  public ProjectDebuggeeBinding(@NotNull ProjectSelector elysiumProjectId, @NotNull JComboBox debugeeTarget) {
-    myElysiumProjectId = elysiumProjectId;
-    myDebugeeTarget = debugeeTarget;
-
-    myElysiumProjectId.getDocument().addDocumentListener(new DocumentAdapter() {
+  private final JComboBox moduleSelector;
+  private final ProjectSelector elysiumProjectId;
```
patflynn: Shouldn't we change this name  as well to something like projectSelector?

etanshaul: Definitely. Missed that one. Its been changed as well as minor naming cleanup in DebugTarget to make things more consistent (removing the "my" prefix from fields).

patflynn:
otherwise LGTM.

etanshaul:
@elharo - code updated based on latest comments

etanshaul:
@elharo ptal (in case this slipped under your radar)

elharo:
lgtm

PR #388, opened 2016-01-22
Reactivates working unit tests.

joaoandremartins:
@elharo these tests were left as "ignored" by you due to a NullPointerException, but I found they're working. Let me know if there's anything I'm missing.


elharo:
lgtm

loosebazooka:
Just making sure they work in the IDE and the command line from gradle?

joaoandremartins:
Recently I've only been running unit tests from the IDE (using the Gradle config "test").

Has it happened that tests succeed in one place and fail in the other?

patflynn:
As long as you use the Gradle config I don't think they typically diverge.
What has happened is that it's passed on Linux, but not on Travis. Or it
passed on Mac, but not on Linux.

PR #395, opened 2016-01-26
Fixes #394

### LINKED ISSUE #394: If user logs out, detach any running debug sessions
If the user logs out, she won't be able to set/unset snapshot locations, receive snapshots from the Cloud Debugger, so it's better to detach any running Cloud Debugger sessions.

akerekes:
update logout warning message, detach Cloud Debugger session

### DIFF
--- a/google-account-plugin/src/com/google/gct/login/GoogleLogin.java
+++ b/google-account-plugin/src/com/google/gct/login/GoogleLogin.java
@@ -22,11 +22,11 @@ import com.google.api.client.googleapis.auth.oauth2.GoogleOAuthConstants;
 import com.google.api.client.http.HttpRequestFactory;
 import com.google.api.client.repackaged.com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
-import com.google.gct.stats.LoginTracking;
-import com.google.gct.stats.UsageTrackerProvider;
 import com.google.gct.login.ui.GoogleLoginActionButton;
 import com.google.gct.login.ui.GoogleLoginCopyAndPasteDialog;
 import com.google.gct.login.ui.GoogleLoginIcons;
+import com.google.gct.stats.LoginTracking;
+import com.google.gct.stats.UsageTrackerProvider;
 import com.google.gdt.eclipse.login.common.GoogleLoginState;
 import com.google.gdt.eclipse.login.common.LoggerFacade;
 import com.google.gdt.eclipse.login.common.OAuthData;
@@ -51,6 +51,7 @@ import com.intellij.openapi.wm.ex.WindowManagerEx;
 
 import net.jcip.annotations.Immutable;
 
+import org.apache.commons.lang.WordUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -565,8 +566,18 @@ public class GoogleLogin {
    * An implementation of {@link UiFacade} using Swing dialogs and external browsers.
    */
   private class AndroidUiFacade implements UiFacade {
+
+    /**
+     * number of characters to wrap lines after in the logout message
+     */
+    public static final int WRAP_LENGTH = 60;
     private GoogleLoginActionButton myButton;
     private volatile CancellableServerReceiver receiver = null;
+    private GoogleLoginMessageExtender[] messageExtenders;
+
+    public AndroidUiFacade() {
+      this.messageExtenders = Extensions.getExtensions(GoogleLoginMessageExtender.EP_NAME);
+    }
 
     @Override
     public String obtainVerificationCodeFromUserInteraction(String title, GoogleAuthorizationCodeRequestUrl authCodeRequestUrl) {
@@ -640,7 +651,14 @@ public class GoogleLogin {
       String updatedMessage = message;
       if (message.equals("Are you sure you want to sign out?")) {
         updatedMessage += " This will sign out all logged in users.";
+        for (GoogleLoginMessageExtender messageExtender : messageExtenders) {
+          String additionalLogoutMessage = messageExtender.additionalLogoutMessage();
+          if (!Strings.isNullOrEmpty(additionalLogoutMessage)) {
+            updatedMessage += " " + additionalLogoutMessage;
+          }
+        }
       }
+      updatedMessage = WordUtils.wrap(updatedMessage, WRAP_LENGTH, /* newLinestr */ null, /* wrapLongWords */ false);
       return (Messages.showYesNoDialog(updatedMessage, title, GoogleLoginIcons.GOOGLE_FAVICON) == Messages.YES);
     }
 
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessHandler.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessHandler.java
@@ -15,7 +15,12 @@
  */
 package com.google.gct.idea.debugger;
 
+import com.google.gct.idea.debugger.ui.LogoutDebugProcessDetacher;
+import com.google.gct.login.CredentialedUser;
+import com.google.gct.login.GoogleLogin;
+
 import com.intellij.execution.process.ProcessHandler;
+import com.intellij.openapi.diagnostic.Logger;
 
 import java.io.OutputStream;
 
@@ -24,10 +29,29 @@ import java.io.OutputStream;
  * returns its output stream.
  */
 public class CloudDebugProcessHandler extends ProcessHandler {
+
+  private static final Logger LOG = Logger.getInstance(CloudDebugProcessHandler.class);
+
   private final CloudDebugProcess myProcess;
 
   public CloudDebugProcessHandler(CloudDebugProcess process) {
     myProcess = process;
+    if (myProcess != null && myProcess.getProcessState() != null) {
+      String userEmail = myProcess.getProcessState().getUserEmail();
+      if (userEmail != null) {
+        final CredentialedUser user = GoogleLogin.getInstance().getAllUsers().get(userEmail);
+        if (user.getGoogleLoginState() != null) {
+          user.getGoogleLoginState()
+              .addLoginListener(new LogoutDebugProcessDetacher<CloudDebugProcessHandler>(this));
+        } else {
+          LOG.error("GoogleLoginState is null. To launch a debug session user needs to be logged in");
+        }
+      } else {
+        LOG.error("userEmail is null. To launch a debug session user needs to be logged in");
+      }
+    } else {
+      LOG.error("process or process state is null. This should only happen in tests");
+    }
   }
 
   @Override
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebuggerClient.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebuggerClient.java
@@ -101,8 +101,12 @@ public class CloudDebuggerClient {
           user.getGoogleLoginState().addLoginListener(new LoginListener() {
             @Override
             public void statusChanged(boolean login) {
-              //aggressively remove the cached item on any status change.
-              myDebuggerClientsFromUserEmail.remove(hashkey);
+              if (!login) {
+                // aggressively remove the cached item on any status change.
+                myDebuggerClientsFromUserEmail.remove(hashkey);
+              } else {
+                // user logged in, should we do something?
+              }
             }
           });
           HttpRequestInitializer initializer = new HttpRequestInitializer() {
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/LogoutDebugProcessDetacher.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/LogoutDebugProcessDetacher.java
@@ -1,0 +1,28 @@
+package com.google.gct.idea.debugger.ui;
+
+import com.google.gdt.eclipse.login.common.LoginListener;
+
+import com.intellij.execution.process.ProcessHandler;
+
+/**
+ * Detaches a Cloud Debug sessions represented by an instance of ProcessHandler on a logout event
+ */
+public class LogoutDebugProcessDetacher<H extends ProcessHandler> implements LoginListener {
+
+  private H processHandler;
+
+  public LogoutDebugProcessDetacher(H processHandler) {
+    this.processHandler = processHandler;
+  }
+
+  @Override
+  public void statusChanged(boolean loggedIn) {
+    if (!loggedIn) {
+      detachCloudDebugProcess();
+    }
+  }
+
+  private void detachCloudDebugProcess() {
+    processHandler.detachProcess();
+  }
+}
--- a/google-cloud-tools-plugin/src/com/google/gct/login/IntelliJGoogleLoginMessageExtender.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/login/IntelliJGoogleLoginMessageExtender.java
@@ -1,0 +1,11 @@
+package com.google.gct.login;
+
+import com.google.gct.idea.util.GctBundle;
+
+public class IntelliJGoogleLoginMessageExtender implements GoogleLoginMessageExtender {
+
+  @Override
+  public String additionalLogoutMessage() {
+    return GctBundle.message("clouddebug.logout.additional.message");
+  }
+}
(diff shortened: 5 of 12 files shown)


google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebuggerClient.java
```diff
@@ -101,8 +102,13 @@ private static Debugger getClient(final @Nullable String userEmail, final int ti
           user.getGoogleLoginState().addLoginListener(new LoginListener() {
             @Override
             public void statusChanged(boolean login) {
-              //aggressively remove the cached item on any status change.
-              myDebuggerClientsFromUserEmail.remove(hashkey);
+              if (!login) {
+                //aggressively remove the cached item on any status change.
```
elharo: nit: space after //

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudDebugProcessDetacher.java
```diff
@@ -0,0 +1,22 @@
...
+import com.intellij.xdebugger.XDebuggerManager;
+
+public class CloudDebugProcessDetacher {
+  public void detachCloudDebugProcesses() {
+    for (Project project : ProjectManager.getInstance().getOpenProjects()) {
+      XDebuggerManager debuggerManager = project
+          .getComponent(XDebuggerManager.class);
+      if (debuggerManager != null) {
+        for (CloudDebugProcess cloudDebugProcess : debuggerManager
+            .getDebugProcesses(CloudDebugProcess.class)) {
```
elharo: I'd prefer not to break before the . if we can help it. Maybe after the colon?

elharo:
lgtm

google-account-plugin/src/com/google/gct/login/GoogleLogin.java
```diff
@@ -639,7 +639,8 @@ public void showErrorDialog(String title, String message) {
     public boolean askYesOrNo(String title, String message) {
       String updatedMessage = message;
       if (message.equals("Are you sure you want to sign out?")) {
-        updatedMessage += " This will sign out all logged in users.";
+        updatedMessage += " This will sign out all logged in users "
```
patflynn: these messages should be in a resource bundle.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebuggerClient.java
```diff
@@ -101,8 +102,13 @@ private static Debugger getClient(final @Nullable String userEmail, final int ti
...
+                myDebuggerClientsFromUserEmail.remove(hashkey);
+                detachCloudDebugProcesses();
+              } else {
+                // user logged in, should we do something?
```
patflynn: are we adding multiple login listeners for each client session?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudDebugProcessDetacher.java
```diff
@@ -0,0 +1,22 @@
+package com.google.gct.idea.debugger.ui;
+
+import com.google.gct.idea.debugger.CloudDebugProcess;
+
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.project.ProjectManager;
+import com.intellij.xdebugger.XDebuggerManager;
+
+public class CloudDebugProcessDetacher {
```
patflynn: classes require javadoc

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebuggerClient.java
akerekes: Whenever a new connection is requested, a new login listener is added to the user's login state which will remove the connection from the `myDebuggerClientsFromUserEmail` map. In this case `detachCloudDebugProcesses()` will be called multiple times, however once the first one finishes, any subsequent call will not find any active debug sessions, and will not do anything. Nevertheless, it looks like a bad design, so I move the detach action to the process handler class which will also register as a LoginListener, and this way all process will be detached by itself upon notification.

google-account-plugin/src/com/google/gct/login/GoogleLogin.java
patflynn: I've added a bug #397. Don't fix it here.

akerekes:
PTAL @elharo @loosebazooka @patflynn

google-account-plugin/src/com/google/gct/login/GoogleLogin.java
```diff
@@ -639,7 +639,8 @@ public void showErrorDialog(String title, String message) {
...
+            + "and detach all Cloud Debugger session.";
```
elharo: session --> sessions

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/LogoutDebugProcessDetacher.java
```diff
@@ -0,0 +1,28 @@
+package com.google.gct.idea.debugger.ui;
+
+import com.google.gdt.eclipse.login.common.LoginListener;
+
+import com.intellij.execution.process.ProcessHandler;
+
+/**
+ * Detaches a Cloud Debug sessions represented by an instance of ProcessHandler on a logout event
```
elharo: singular or plural?

google-account-plugin/src/com/google/gct/login/GoogleLogin.java
akerekes: fixed

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/LogoutDebugProcessDetacher.java
akerekes: Singular, fixed.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessHandler.java
```diff
@@ -28,6 +32,16 @@
 
   public CloudDebugProcessHandler(CloudDebugProcess process) {
     myProcess = process;
+    if (myProcess != null && myProcess.getProcessState() != null) {
+      String userEmail = myProcess.getProcessState().getUserEmail();
+      if (userEmail != null) {
+        final CredentialedUser user = GoogleLogin.getInstance().getAllUsers().get(userEmail);
+        if (user.getGoogleLoginState() != null) {
```
patflynn: is it even possible for this to be null?  If it is, should it raise an error?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/LogoutDebugProcessDetacher.java
```diff
@@ -0,0 +1,28 @@
...
+  private H processHandler;
+
+  public LogoutDebugProcessDetacher(H processHandler) {
+    this.processHandler = processHandler;
+  }
+
+  @Override
+  public void statusChanged(boolean loggedIn) {
+    if (!loggedIn) {
+      detachCloudDebugProcess();
```
patflynn: does this notify the user why the process was detached?

akerekes: No it does not. The notification happens through another channel (i.e. the user clicking on the signout button a confirmation message is shown: [GoogleLogin.java#L643](https://github.com/GoogleCloudPlatform/gcloud-intellij/blob/62aa76b773d82fe84addb02e20122b3f3b6690f9/google-account-plugin/src/com/google/gct/login/GoogleLogin.java#L643)  )

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessHandler.java
akerekes: It can be null in case of incorrect API usage. We could log and error in IDEA that would show up as in internal error and if the user wants, he could submit it as a bug report.

patflynn: SGTM.

elharo:
lgtm

patflynn:
We still need to remove references to the cloud debugger form the login flow.

akerekes:
@patflynn ptal

google-account-plugin/src/com/google/gct/login/GoogleLogin.java
```diff
@@ -567,6 +567,16 @@ public String getInfo() {
...
     private GoogleLoginActionButton myButton;
     private volatile CancellableServerReceiver receiver = null;
+    private GoogleLoginMessageExtender messageExtender;
+
+    public AndroidUiFacade() {
+      GoogleLoginMessageExtender[] messageExtenders = Extensions.getExtensions(GoogleLoginMessageExtender.EP_NAME);
+      if (messageExtenders.length == 0) {
+        messageExtender = new DummyGoogleLoginMessageExtender();
+      } else {
+        messageExtender = messageExtenders[0];
```
patflynn: why only accept the first message?

patflynn:
Just curious,  was it not workable to just pop up a debugger dialog or notification on logout?

akerekes:
When the user clicks on logout, there is already a dialog shown to confirm the logout. It feels a better experience for me to have the information in the same dialog than showing another popup or notification.

google-account-plugin/src/com/google/gct/login/GoogleLogin.java
akerekes: At that time it seemed reasonable to expect only one extension to provide messages. We can however allow any number, though their order of execution will be arbitrary initially (the order IJ/AS returns  the corresponding extensions) as I see no point in implementing some kind of priority logic currently.

patflynn: I don't think that's a safe assumption.  If you include an EP in the
account plugin many of the plugin's clients in Android Studio (such as
Google Play, Cloud Testing, Firebase), could conceivably try to use it.

google-cloud-tools-plugin/src/com/google/gct/login/IntelliJGoogleLoginMessageExtender.java
```diff
@@ -0,0 +1,11 @@
+package com.google.gct.login;
+
+import com.google.gct.idea.util.GctBundle;
+
+public class IntelliJGoogleLoginMessageExtender implements GoogleLoginMessageExtender {
+
+  @Override
+  public String additionalLogoutMessage() {
+    return GctBundle.message("clouddebug.logout.additional.message");
```
loosebazooka: Is this only activated if we have debugger sessions active? It seems odd that all users will see this message?

loosebazooka: Preferably, there could be a "has additional message" or something?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessHandler.java
```diff
@@ -24,10 +29,29 @@
...
+          user.getGoogleLoginState()
+              .addLoginListener(new LogoutDebugProcessDetacher<CloudDebugProcessHandler>(this));
+        } else {
+          LOG.error("GoogleLoginState is null, to launch a debug session user needs to be logged in");
```
elharo: . To launch

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessHandler.java
```diff
@@ -24,10 +29,29 @@
...
+        }
+      } else {
+        LOG.error("userEmail is null, to launch a debug session user needs to be logged in");
```
elharo: . To launch

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessHandler.java
```diff
@@ -24,10 +29,29 @@
...
+      }
+    } else {
+      LOG.error("process or process state is null, this should only happen in tests");
```
elharo: . This should

google-account-plugin/src/com/google/gct/login/GoogleLogin.java
```diff
@@ -565,8 +566,23 @@ public String getInfo() {
...
+     */
+    public static final int WRAP_LENGTH = 60;
     private GoogleLoginActionButton myButton;
     private volatile CancellableServerReceiver receiver = null;
+    private GoogleLoginMessageExtender[] messageExtenders;
+
+    public AndroidUiFacade() {
+      GoogleLoginMessageExtender[] messageExtenders = Extensions.getExtensions(GoogleLoginMessageExtender.EP_NAME);
+      if (messageExtenders.length == 0) {
+        this.messageExtenders = new GoogleLoginMessageExtender[0];
```
patflynn: how is this different from what's already assigned to messageExtenders?

google-account-plugin/src/com/google/gct/login/GoogleLogin.java
```diff
@@ -640,7 +656,14 @@ public boolean askYesOrNo(String title, String message) {
       String updatedMessage = message;
       if (message.equals("Are you sure you want to sign out?")) {
```
patflynn: not related to your change, but this code is crazy. (just saying. I'm not asking you to fix it)

google-cloud-tools-plugin/src/com/google/gct/login/IntelliJGoogleLoginMessageExtender.java
akerekes: There's a method to obtain all states listening in the background that is being renamed in an upcoming PR (not yet submitted). I'd rather merge that PR, then extract the method and make it available for this message extender to check if there are any debug sessions running in the background and display the message only if there's any.

google-account-plugin/src/com/google/gct/login/GoogleLogin.java
```diff
@@ -640,7 +656,14 @@ public boolean askYesOrNo(String title, String message) {
...
         updatedMessage += " This will sign out all logged in users.";
+        for (GoogleLoginMessageExtender messageExtender : messageExtenders) {
+          String additionalLogoutMessage = messageExtender.additionalLogoutMessage();
+          if (!Strings.isNullOrEmpty(additionalLogoutMessage)) {
+            updatedMessage += " " + additionalLogoutMessage;
+          }
+        }
       }
+      updatedMessage = WordUtils.wrap(updatedMessage, WRAP_LENGTH, null, false);
```
patflynn: it helps readability here if you add param docs /* paramName */ as part of each null or boolean flag.

google-account-plugin/src/com/google/gct/login/GoogleLogin.java
```diff
@@ -753,4 +776,12 @@ public void logWarning(String msg) {
       LOG.warn(msg);
     }
   }
+
+  private class DummyGoogleLoginMessageExtender implements GoogleLoginMessageExtender {
```
patflynn: I'm not seeing this in use anywhere.

google-account-plugin/src/com/google/gct/login/GoogleLogin.java
akerekes: this.messageExtenders is null by the field declaration, this saves a null check whenever we want to iterate over the array. I can move the initialization to the field declaration though.

patflynn: Extensions.getExtensions returns a @NotNull though.

google-account-plugin/src/com/google/gct/login/GoogleLogin.java
akerekes: I agree!

patflynn:
Make sure you ping this thread when you've addressed all comments and are ready for another look.

LGTM. Make sure you squash commits.

akerekes:
PTAL @patflynn

patflynn:
@akerekes is there a recent change I should look at? You already have my LGTM.

patflynn:
you should make sure that @elharo and @loosebazooka have resolved their concerns.

akerekes:
PTAL @elharo @loosebazooka

google-cloud-tools-plugin/src/com/google/gct/login/IntelliJGoogleLoginMessageExtender.java
loosebazooka: Sure, maybe file a issue against additional message functionality? Perhaps additionalLogoutMessage can return null (with a @nullable annotation).

loosebazooka:
LGTM, probably still want @elharo to look again.

PR #400, opened 2016-01-28
Fixes #392 - remove Android Studio specific contents from login panel

### LINKED ISSUE #392: Google account plugin contains Android Studio specific messages and link
If you have not signed into IJ the copy and link in the Google Account widget displays 
screenshot from 2016-01-26 11 42 23

which is android studio specific content and an android studio specific link. That content should be generified and the help link should go out to the IDE specific help page.  Talk to me about what those link urls should be.

akerekes:
change text to be product neutral, change "learn more" url to point to GCP/Development and Administration Tools page

### DIFF
--- a/google-account-plugin/src/com/google/gct/login/ui/UsersListCellRenderer.java
+++ b/google-account-plugin/src/com/google/gct/login/ui/UsersListCellRenderer.java
@@ -56,8 +56,8 @@ public class UsersListCellRenderer extends JComponent implements ListCellRendere
   private final static String CLOUD_LABEL_TEXT = "Open Google Developers Console";
   private final static String PLAY_LABEL_TEXT = "Open Play Developer Console";
   private final static String DEFAULT_AVATAR = "/icons/loginAvatar@2x.png";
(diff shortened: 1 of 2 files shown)


google-account-plugin/src/com/google/gct/login/ui/UsersListCellRenderer.java
```diff
@@ -189,9 +189,9 @@ public boolean inCloudConsoleUrl(Point point, int activeIndex) {
   }
 
   public boolean inLearnMoreUrl(Point point) {
-    // 3 is for the number of labels and row of texts
+    // 2 is for the number of labels and row of texts
```
patflynn: wow that code was fragile. Thanks for spotting this.

patflynn:
LGTM

google-account-plugin/src/com/google/gct/login/ui/UsersListCellRenderer.java
akerekes: I've made a note to myself to fix it with a flexible solution.

PR #401, opened 2016-01-28
Fixes #398 - Stopping background debug session and notify user if user is logged out

### LINKED ISSUE #398: Stopping background debug session and notify user if user is logged out
This is part of #229 

If user decides to keep listening for snapshots in the background when the actual debug session is stopped and the user logs out, any background session should be stopped and the user notified about it.

akerekes:


google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -128,3 +128,5 @@ clouddebug.snapshots=Cloud Debugger Snapshots
 clouddebug.couldnotconnect=Could not connect to Google. Please make sure you have a working internet connection.
 clouddebug.stop.and.create.new.session=A cloud debugger session is already running.\nWould you like to stop it and start a new session?
 clouddebug.stop.background.and.create.new.session=A cloud debugger session is listening in the background.\nWould you like to stop it and start a new session?
+clouddebug.background.listener.error.title=Error while connecting to Cloud Debug backend
```
elharo: Cloud Debugger backend

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -128,3 +128,5 @@ clouddebug.snapshots=Cloud Debugger Snapshots
...
+clouddebug.background.listener.error.message=A cloud debug process is listening in the background but cannot access the backend, listening will stop now. Verify that there is a Google account signed in with access permissions to the project
```
elharo: cloud debugger process

google-cloud-tools-plugin/testSrc/com/google/gct/idea/debugger/CloudDebugGlobalPollerTest.java
```diff
@@ -0,0 +1,47 @@
...
+  public void setUp() throws Exception {
+    cloudDebugProcessState = new CloudDebugProcessState();
+  }
+
+  @Test
+  public void testPollForChangesFiresNotificationIfBackgroundListeningFails() throws Exception {
+    cloudDebugProcessState.setListenInBackground(true);
+    Application application = ApplicationManager.getApplication();
+    Notifications handler = Mockito.mock(Notifications.class);
+    // sending out notifications rely on several static method calls in
```
elharo: relies

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
```diff
@@ -166,13 +170,27 @@ private void fireBreakpointsChanged(@NotNull CloudDebugProcessState state) {
...
+  void pollForChanges(@NotNull final CloudDebugProcessState state) {
     final Debugger client = CloudDebuggerClient.getShortTimeoutClient(state);
     if (client == null) {
-      // It is ok if there is no client.  We may want to consider notifying the user that
-      // background polling is disabled until they login, but for now, we can just doc it.
-      LOG.info("no client available attempting to checkForChanges");
-      return;
+      if (state.isListenInBackground()) {
+        // state is supposed to listen, but does not have access to the backend
+        LOG.warn("CloudDebugProcessState is listening in the background but no debugger client "
```
patflynn: remind me, does 'warn' trigger an intellij error visible to the user?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPollerTimerTask.java
```diff
@@ -0,0 +1,22 @@
+package com.google.gct.idea.debugger;
+
+import java.util.TimerTask;
+
+/**
+ * Polls changes from the Cloud Debug backend executed by a timer in CloudDebugGlobalPoller
```
patflynn: Polls for changes from the Cloud Debugger.  You don't want to refer to your client classes in your Javadoc.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
```diff
@@ -44,15 +48,22 @@
...
 public class CloudDebugGlobalPoller {
+
   private static final int DELAY_MS = 5000;
   private static final Logger LOG = Logger.getInstance(CloudDebugGlobalPoller.class);
+
+  /**
+   * Display group used to display notifications in the IDE
+   */
+  public static final String CLOUD_DEBUGGER_ERROR_NOTIFICATIONS_DISPLAY_GROUP =
+      "Cloud Debugger Error Notifications";
```
patflynn: I guess this text is never seen by the user?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
```diff
@@ -166,13 +170,27 @@ private void fireBreakpointsChanged(@NotNull CloudDebugProcessState state) {
...
    */
-  private void pollForChanges(@NotNull final CloudDebugProcessState state) {
+  void pollForChanges(@NotNull final CloudDebugProcessState state) {
     final Debugger client = CloudDebuggerClient.getShortTimeoutClient(state);
     if (client == null) {
-      // It is ok if there is no client.  We may want to consider notifying the user that
-      // background polling is disabled until they login, but for now, we can just doc it.
-      LOG.info("no client available attempting to checkForChanges");
-      return;
+      if (state.isListenInBackground()) {
```
patflynn: I'm a bit confused why pollForChanges while !state.isListenInBackground means that there is an access issue.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
etanshaul: no it doesn't. I tested this myself previously.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
akerekes: The idea for "listening will stop now" was to notify the user that the plugin will stop listening in the background for events from the backend. If we omit it, it might not be obvious to the user that this is not only a warning, but also something else have changed.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
akerekes: That's right, it is used only internally by IntelliJ to differentiate between different notification groups allowing to set different behavior for them.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
akerekes: On line 174 we failed to obtain a Debugger client, presumably because of access error (or e.g. network failure). If state.isListenInBackground() is true it means that this debug session is listening for events in the background. Thus if it is supposed to listen but cannot get access view the Debugger client, that's an unexpected situation.

patflynn: Is there a way to idenitfy the specific case of no longer being authorized?  What happens in getShortTimeoutClient when the user has lost access?

akerekes: This PR is for the case when the user explicitly logs out. In this case the locally cached user credentials are deleted during logout. When the next time getShortTimeoutClient() is called, it cannot find the credentials and therefore does not attempt a server request, but returns a `null` instead of a Debugger instance.
We can change the return type of the method to be a (status,Debugger) tuple and the status could show the error reason if Debugger is null.

google-cloud-tools-plugin/testSrc/com/google/gct/idea/debugger/CloudDebugGlobalPollerTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+    CloudDebugGlobalPoller cloudDebugGlobalPoller = new CloudDebugGlobalPoller();
+    cloudDebugGlobalPoller.pollForChanges(cloudDebugProcessState);
+
+    assertFalse(cloudDebugProcessState.isListenInBackground());
+    verify(handler)
+        .notify(argThat(HasPropertyWithValue.<Notification>hasProperty("title",
+                                                            is("Error while connecting to Cloud Debugger backend"))));
+  }
+}
```
patflynn: make sure you have a new line at the EOF here and on the CloudDebugGlobalPollerTimerTaskTest

patflynn:
lgtm once you've addressed formatting comments. make sure you squash commits.

PR #405, opened 2016-02-02
auto generating copyright headers for new project files

### LINKED ISSUE #403: Copyright header should be included by default in new project files
I'm pretty sure there's a way to do this in IJ.

etanshaul:
fixes #403 

simple project setting update - requires copyright plugin to be installed (which I believe is the default).


patflynn:
LGTM

PR #406, opened 2016-02-02
Adds the plugin version to the set of crash report values sent for error reporting

### LINKED ISSUE #333: We need to include plugin version in our feedback reports and ideally our user-agent
I don't know that this information is available yet at runtime. Should be packaged into our plugin.

### LINKED ISSUE #342: Derive version in code from plugin version in build.gradle
com.google.gct.idea.util.PlatformInfo#getUserAgent()

patflynn:
It turns out we don't need to package the version ourselves since it's included in the plugin.xml. 

Funny story, I have a CL at work that starts this work from the other end of including the version using resource filtering etc. 

Fixes #333


elharo:
lgtm

loosebazooka:
so ErrorBean is something intellij gives us? That has the plugin data?

patflynn:
There's code on line 154 of GoogleFeedbackErrorReporter that sets the plugin version on the error by reading it from the IdeaPluginDescriptor.

loosebazooka:
cool, lgtm

PR #407, opened 2016-02-03
removing reliance on Thread.sleep from snapshot unit tests, refactori…

### LINKED ISSUE #187: CloudDebugHistoricalSnapshotsTest.testOnBreakpointListChanged_twoBreakPoints hangs during gradlew test

etanshaul:
…ng shared state, and removing @Ignore

Fixes #187 

Revives CloudHistoricalSnapshotsTest which was previously ignored due to unpredictable failures. 

- Un-ignoring the unit test
- Refactoring associated classes for testability
- Removing shared state from the unit test, and removing call to asynchronous method from CloudHistoricalSnapshots (instead moving it to the caller). These were the sources of the problems with this test.


elharo:
lgtm

patflynn:
lgtm

PR #409, opened 2016-02-04
Gets all project pages from CRM and sorts the result alphabetically.

### LINKED ISSUE #408: Project list unsorted alphabetically in attach dialog and debuggable project produces no targets list
The projects list is now not alphabetically sorted in the attach dialog. There also appear to be missing projects from that list.
The test project I'm using to troubleshoot CDB doesn't appear in my projects list. This could be because of the high number of projects I have on my account.
When I type the project ID in the project text box, the target list remains unchanged and I can't start my debug session.

joaoandremartins:
@elharo @patflynn @loosebazooka 
Fixes #408 .

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/elysium/GoogleUserModelItem.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/elysium/GoogleUserModelItem.java
@@ -34,7 +34,10 @@ import java.awt.Image;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
 
 import javax.swing.SwingUtilities;
 import javax.swing.tree.DefaultMutableTreeNode;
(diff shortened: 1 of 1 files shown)


patflynn:
what's the actual page size?  Is there no way to modify it in the request?

joaoandremartins:
Apparently the server changed it from a very big one to 50 recently.

We could specify a page size in the request, but we'd always be interested in getting every project for a user, so it wouldn't make much of a difference.

patflynn:
Is there a max page size, or a way to return all results in one request?

google-cloud-tools-plugin/src/com/google/gct/idea/elysium/GoogleUserModelItem.java
```diff
@@ -98,7 +106,7 @@ public void synchronize() {
       @Override
       public void run() {
         try {
-          loadUserProjects(myUser);
```
joaoandremartins: Notice that loadUserProjects is called with myUser, and then in the method, the user argument is only used to create the Cloud Resource Manager client.

I relocated building the client to the constructor because I wanted to unit test this method, but couldn't find a way to get the projects in GoogleUserModelItem. I still think it's a good idea to cleanup the code, but I'm not sure if passing that argument like that achieves anything. I don't think so, it should be safe to remove the argument, but maybe you can double check this.

joaoandremartins:
It's not documented at https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/list, but Cyrus from the CRM team says he thinks max page size is about 500. I'll go ahead and set it to 300, as I think it's a sensible value.

patflynn:
LGTM

PR #411, opened 2016-02-04
Updating release.md to current workflow

patflynn:


loosebazooka:
Can you pull this down to your computer and recommit so it uses the correct email address?

patflynn:
@loosebazooka done.

RELEASING.md
```diff
@@ -2,45 +2,18 @@
 
 This page describes how to cut releases of our plugin. It's strictly for the owners of the project.
 
-1. Create a new branch as if you were starting a normal bugfix PR
+1. Create a new branch named release_v{$version}.  It's critical that you create the branch according to this schema or the release will fail.
```
loosebazooka: Is there concern local branches wont be up to date?

how about something like?

```
git fetch
git checkout -b release_v1.0 origin/master
```

patflynn: The release will fail if the local branch isn't up to date, but fair point
that the instructions should help the user make sure they are up to date
before they start.

patflynn: done.

patflynn:
Any more questions about this change?

loosebazooka:
lgtm

PR #413, opened 2016-02-04
Fixes #399  - Stop background debug session and notify user if user loses permission to cloud project

### LINKED ISSUE #399: Stopping background debug session and notify user if user loses permission to cloud project
This is part of #229 

If the user's access is revoked either by third party or the user revokes IDEA's OAuth permissions to access the cloud project any background debug sessions listening for snapshots should be stopped and the user notified about it.

akerekes:


akerekes:
PTAL @patflynn @loosebazooka

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -130,6 +130,7 @@ clouddebug.stop.and.create.new.session=A cloud debugger session is already runni
 clouddebug.stop.background.and.create.new.session=A cloud debugger session is listening in the background.\nWould you like to stop it and start a new session?
 clouddebug.logout.additional.message=Any Cloud Debugger sessions listening in the background will be stopped.
 clouddebug.background.listener.error.title=Error while connecting to Cloud Debugger backend
-clouddebug.background.listener.error.message=A cloud debugger process is listening in the background but cannot access the backend. Verify that there is a Google account signed in with access to the project.
+clouddebug.background.listener.access.error.message=A cloud debugger process is listening in the background but cannot access the backend. Verify that there is a Google account signed in with access to the project.
```
patflynn: 'cannot access the backend' sounds like a network error.  If we know it's an auth issue we should just say 'You do not have access to Cloud Debug this project..."

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -130,6 +130,7 @@ clouddebug.stop.and.create.new.session=A cloud debugger session is already runni
...
+clouddebug.background.listener.general.error.message=A cloud debugger process is listening in the background but cannot access the backend, reason: {0}
```
patflynn: Same here.  Be more general: 'There was an unexpected error while listening for Cloud Debugger snapshots. Error detauls:..'

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebuggerClient.java
```diff
@@ -46,8 +46,8 @@
  */
 public class CloudDebuggerClient {
 
-  private static final int LONG_CONNECTION_TIMEOUT_MS = 120 * 1000;
-  private static final int SHORT_CONNECTION_TIMEOUT_MS = 10 * 1000;
+  static final int LONG_CONNECTION_TIMEOUT_MS = 120 * 1000;
```
patflynn: if you change access for tests you should add an @VisibleForTesting annotation.

google-cloud-tools-plugin/testSrc/com/google/gct/idea/debugger/CloudDebugGlobalPollerTest.java
```diff
@@ -13,35 +19,77 @@
...
-    CloudDebugGlobalPoller cloudDebugGlobalPoller = new CloudDebugGlobalPoller();
     cloudDebugGlobalPoller.pollForChanges(cloudDebugProcessState);
 
     assertFalse(cloudDebugProcessState.isListenInBackground());
-    verify(handler)
+    verifyNotificationFired();
+  }
+
+  @Test
+  public void testPollForChangesFiresNotificationOnIOException() throws IOException {
```
patflynn: the Google style naming convention for unit tests of a specific method are testMethodName_conditionInCamelCase(). I think in this case it would be something like testPollForChanges_firesNotificationIfNoDebugClientObtained()

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebuggerClient.java
akerekes: In this case I think these may be used in other parts of the code later, so that's why I have not added the annotation. I actually think these could become configurable eventually by the user.

patflynn: Do you mean used by a subclass or package member?  I don't see any usages
outside of this class.

akerekes: I see what you are saying. If I'd correctly followed what I just said above, it should have been made public. However it is not currently used anywhere, so I'll add the annotation to avoid confusion, and it can be removed later if need be.

akerekes:
@patflynn PTAL

patflynn:
LGTM. squash commits.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -130,6 +130,7 @@ clouddebug.stop.and.create.new.session=A cloud debugger session is already runni
 clouddebug.stop.background.and.create.new.session=A cloud debugger session is listening in the background.\nWould you like to stop it and start a new session?
 clouddebug.logout.additional.message=Any Cloud Debugger sessions listening in the background will be stopped.
 clouddebug.background.listener.error.title=Error while connecting to Cloud Debugger backend
-clouddebug.background.listener.error.message=A cloud debugger process is listening in the background but cannot access the backend. Verify that there is a Google account signed in with access to the project.
+clouddebug.background.listener.access.error.message=Access denied while listening for Cloud Debugger snapshots for project <strong>{0}</strong>. Verify that you are signed in with a Google account that has the appropriate permissions set.
```
loosebazooka: just a question? Is it meaningful to include the type (html) of the message when not standard text

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
```diff
@@ -135,7 +137,9 @@ public synchronized void stopBackgroundListening() {
 
   private void queryServerForBreakpoints(CloudDebugProcessState state, Debugger client)
       throws IOException {
-
+    if (state.getDebuggeeId() == null) {
```
loosebazooka: When can this happen?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
```diff
@@ -213,6 +212,11 @@ void pollForChanges(@NotNull final CloudDebugProcessState state) {
     }
     catch (IOException ex) {
       LOG.warn("exception listing breakpoints", ex);
+      handleBreakpointQueryError(state, ex);
+      return;
+    } catch (Exception e) {
```
loosebazooka: which exception is this catching?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
```diff
@@ -221,4 +225,36 @@ void pollForChanges(@NotNull final CloudDebugProcessState state) {
...
+
+  private void handleBreakpointQueryError(@NotNull CloudDebugProcessState state, String message) {
+    state.setListenInBackground(false);
+    String title = GctBundle.message("clouddebug.background.listener.error.title");
+    Notification notification =
+        new Notification(CLOUD_DEBUGGER_ERROR_NOTIFICATIONS_DISPLAY_GROUP,
+                         title,
+                         message,
+                         NotificationType.ERROR);
+    Notifications.Bus.notify(notification,  state.getProject());
```
loosebazooka: nit: extra space after `notification,  `
Is this simliar to the bubble message that pops up when an ide exception happens?

loosebazooka:
LGTM, only concern is catching runtime exceptions.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
akerekes: @loosebazooka you mean into the message's key?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
akerekes: Normally it should not. However you can create a state object with null fields and since it is not annotated with @notnull IJ gives a warning about it, that was the reason I've added this check.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
akerekes: Right now the IllegalStateException from queryServerForBreakpoints(...), but I assumed if something else happens we'd like to know about it instead of silently ignoring it and not giving the user any feedback.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
akerekes: It appears at the same location, but its design is different.

PR #418, opened 2016-02-05
update attach dialog image to match latest version

etanshaul:
"Module" was changed  to "Target"
and a "Default" copy was prepended to default module versions.

Updating the image to reflect these changes


patflynn:
Is it me or does it look like it's saying 'default version', as in
'default' applies to 'version'.

etanshaul:
No, I agree now that you point it out. 

Adding a comma may be clearer: "Default, Version: 123".

etanshaul:
Or better yet:

"Default (Version: 123)".

patflynn:
Should we identify 'Default' as module name? e.g. "Module: Default,
Version: 123"?

patflynn:
That is better.

etanshaul:
Maybe it is confusing actually to call it Target instead of Module. Just for comparison (with the proposed update):

Target   |  Default (Version: 123)
Module  |  Default (Version: 123)

I think the second one is clearer, especially since they actually called modules in the Web UI and all the documentation. I think it may make sense to revert to using the word Module instead of Target.

etanshaul:
see #424. Depending on the outcome of that PR, the screenshot may need to be redone again.

elharo:
I suggest we merge this one in since it's closer and ready to go. LGTM.

etanshaul:
Closing - migrated to #425

Issue #369, opened 2016-01-15
We need a way to specify and check feature flags


patflynn:
1) Ideally it would support build-time and run-time configuration, but build-time is most important.
2) It should be minimally invasive
3) it should be easy to mock in tests.

PR #421, opened 2016-02-05
Add feature flagging service and move a few static plugin metadata and configuration methods into the same service.

patflynn:
You guys should probably start looking at this craziness as it's pretty much ready.  

It feels a little heavyweight so I'm happy to hear objections to this strategy.  For now this is the simplest way I can think of architecting our components so that they're testable.

There are some code smells in here that I left because the CL was getting enormously large.  

@elharo 
@loosebazooka 
@etanshaul 
@akerekes 
@joaoandremartins


common-lib/src/main/java/com/google/gct/idea/DefaultPluginConfigurationService.java
```diff
@@ -0,0 +1,55 @@
...
+
+  @Override
+  public void enabledGoogleFeedbackErrorReporting(@NotNull String pluginId) {
+    GoogleFeedbackErrorReporter errorReporter = new GoogleFeedbackErrorReporter();
+    IdeaPluginDescriptor plugin = PluginManager.getPlugin(PluginId.getId(pluginId));
+    if (plugin == null) {
+      throw new IllegalArgumentException(pluginId + " is not a valid plugin ID.");
+    }
+    errorReporter
+        .setPluginDescriptor(plugin);
```
joaoandremartins: Can this go on the same line?
errorReporter.setPluginDescriptor(plugin);

common-test-lib/src/main/java/com/google/gct/idea/testing/IntegrationTest.java
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.gct.idea.testing;
+
+/**
+ * TODO: Describe this class well.
```
joaoandremartins: Did you want to do anything else here?

google-account-plugin/src/com/google/gct/idea/AccountPluginInitializationComponent.java
```diff
@@ -63,13 +53,13 @@ void enableFeedbackUtil() {
    * and is therefore not subscribed and listening.
    */
   @VisibleForTesting
-  void enableUsageTracking() {
+  void maybeEnableUsageTracking() {
```
elharo: I don't like the name prefixed with "maybe". Sounds indeterminate. Perhaps enableUsageTrackingIfPossible

elharo:
No major concerns aside from finishing up the unfnished classes.

common-lib/src/main/java/com/google/gct/idea/BasePluginInfoService.java
```diff
@@ -0,0 +1,117 @@
...
+
+  @Override
+  public boolean shouldEnableErrorFeedbackReporting() {
+    return !getCurrentPlatform().equals(IntelliJPlatform.ANDROID_STUDIO);
+  }
+
+  /**
+   * !!!Please update {@link Feature}'s class documentation if you change the algorithm that
+   * determines whether a feature is enabled.!!!
+   */
```
akerekes: Should not the documentation of the algorithm be placed here and provide a link from Feature's class javadoc to this method? I think having the algorithm and the doc at the same place makes it easier keep them consistent.

google-account-plugin/src/com/google/gct/idea/AccountPluginInitializationComponent.java
```diff
@@ -63,13 +53,13 @@ void enableFeedbackUtil() {
...
     UsageTrackerManager usageTrackerManager = UsageTrackerManager.getInstance();
-    NotificationsManager.getNotificationsManager();
-    NotificationsConfiguration.getNotificationsConfiguration().register(
-        TrackerMessageBundle.message("notification.group.display.id"),
-        NotificationDisplayType.STICKY_BALLOON);
     if (usageTrackerManager.isUsageTrackingAvailable() && !usageTrackerManager.hasUserRecordedTrackingPreference()) {
+      NotificationsManager.getNotificationsManager();
+      NotificationsConfiguration.getNotificationsConfiguration().register(
+          TrackerMessageBundle.message("notification.group.display.id"),
```
akerekes: This is used only internally by IJ, not a user visible message. Declaring it as a constant would remove the additional and false information of being visible to users.

akerekes:
Is there a convention for having set the visibility of @Mock-ed fields in test to private or leave it with default visibility?

common-test-lib/src/main/java/com/google/gct/idea/testing/IntegrationTest.java
patflynn: Yes. Deleted until I do.

common-lib/src/main/java/com/google/gct/idea/DefaultPluginConfigurationService.java
patflynn: done.

google-account-plugin/src/com/google/gct/idea/AccountPluginInitializationComponent.java
patflynn: changed it to 'configureUsageTracking' which is less specific about whether it gets turned on or off.

common-lib/src/main/java/com/google/gct/idea/BasePluginInfoService.java
patflynn: That's a good point.  I really want the documentation to be where the developer is defining a Feature and I didn't want to duplicate it..  I'll think about it a bit more.

google-account-plugin/src/com/google/gct/idea/AccountPluginInitializationComponent.java
patflynn: so groupDisplayName is never actually displayed anywhere?

patflynn:
@akerekes 

No convention about visibiltiy. That was just me being lazy. I'll fix.

google-account-plugin/src/com/google/gct/idea/AccountPluginInitializationComponent.java
akerekes: As far as I have seen it in the code it is never displayed. It is used for some internal grouping of the notifications.

loosebazooka: Yeah I think it's just a key in their notification group database.

google-cloud-tools-plugin/src/com/google/gct/idea/GctFeature.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Set;
+
+/**
+ * The set of Google Cloud Tools features that we want to be able to conditionally enable.
+ */
+public enum GctFeature implements Feature {
+  // DEBUGGER is enabled in IDEA Ultime and Community, and disabled everywhere else. Not flaggable.
```
loosebazooka: `Ultime` -> `Ultimate`

common-lib/src/main/java/com/google/gct/idea/BasePluginInfoService.java
patflynn: For now I've decided to leave the docs on the semantics of the Feature configuration on the Feature class. I'm not crazy about the current design, and open to suggestions, but want to get this in, so that I can merge into the MVM deployment PR.

patflynn:
This is ready for another look.

patflynn:
closes #369

patflynn:
@elharo 
@akerekes 
@etanshaul 
@joaoandremartins 

Still waiting on LGTM for this.  Let me know if you have concerns.

akerekes:
If I see it correctly, this will also fix #342, right?

common-lib/src/main/java/com/google/gct/idea/DefaultPluginConfigurationService.java
```diff
@@ -0,0 +1,54 @@
...
+    Extensions.getRootArea().getExtensionPoint(extensionPoint)
+        .registerExtension(extension);
+  }
+
+  @Override
+  public void enabledGoogleFeedbackErrorReporting(@NotNull String pluginId) {
+    GoogleFeedbackErrorReporter errorReporter = new GoogleFeedbackErrorReporter();
+    IdeaPluginDescriptor plugin = PluginManager.getPlugin(PluginId.getId(pluginId));
+    if (plugin == null) {
+      throw new IllegalArgumentException(pluginId + " is not a valid plugin ID.");
```
elharo: Any chance throwing an exception here will cause a problem for an end user? This doesn't seem like critical path code.

patflynn:
Yes!

Fixes #342

common-lib/src/main/java/com/google/gct/idea/Feature.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * The algorithm that determines whether a given {@code feature} configuration is enabled is as
+ * follows:
+ * <ol>
+ *   <li>If the current platform the plugin is running in is contained in the list of
+ * {@link #getSupportedPlatforms()}, then the feature is enabled.</li>
+ *   <li>If the plugin's config.properties file contains a key with {@link #getResourceFlagName()}
+ *   set to {@code true}, then the feature is enabled.</li>
+ *   <li>If there is a system environment variable named {@link #getSystemFlagName()} with a value
+ *   of "true|TRUE". Then it's enabled</li>.
```
elharo: , then it's

common-lib/src/main/java/com/google/gct/idea/DefaultPluginConfigurationService.java
patflynn: I think it will have an impact.  I think failures during component initialization causes the plugin to fail to load. 

You do raise a good point that if this fails, we can't capture the fact it fails, and should the plugin be disabled completely at that point?  We certainly would hear about it sooner if we did.

elharo:
lgtm

etanshaul:
lgtm

PR #422, opened 2016-02-05
Fixes issue when a class isn't in the classpath and NPE is thrown when seeing snapshot.

### LINKED ISSUE #420: NullPointerException opening snapshots when class isn't in classpath
It is possible to start a debug session by checking out a Cloud repo in a weird IDE state which doesn't contain some used classes in the project's classpath (from the IDE point of view). When this happens, the IDE doesn't know the sources and the PsiPackage for the project throws a non-caught NullPointerException.

joaoandremartins:
@loosebazooka @patflynn @elharo fixes #420

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
@@ -27,6 +27,7 @@ import com.intellij.openapi.vfs.VirtualFileSystem;
 import com.intellij.psi.JavaPsiFacade;
 import com.intellij.psi.PsiClass;
 import com.intellij.psi.PsiJavaFile;
+import com.intellij.psi.PsiPackage;
 import com.intellij.psi.search.FileTypeIndex;
 import com.intellij.psi.search.GlobalSearchScope;
 import com.intellij.util.indexing.FileBasedIndex;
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
```diff
@@ -75,11 +76,15 @@ public VirtualFile getFileFromPath(@NotNull Project project, @NotNull String pat
     VirtualFile file = FILE_SYSTEM.findFileByPath(project.getBasePath() + "/" + path);
     // Try class name with package and class file name.
     if (file == null) {
-      PsiClass[] matchingClasses = JavaPsiFacade.getInstance(project)
-          .findPackage(getPackageFromPath(path))
-          .findClassByShortName(getClassNameFromPath(path), GlobalSearchScope.allScope(project));
-      if (matchingClasses.length > 0) {
-        file = matchingClasses[0].getContainingFile().getVirtualFile();
+      PsiPackage psiPackage = JavaPsiFacade.getInstance(project)
```
patflynn: if it is null should we log something? or notify the user that the class can't be found?

joaoandremartins: Right now, the line in the stack with the unrecognized class appears in a grey color and, when you click it, nothing happens. I don't know if implementing a more explicit warning is worth the effort. What did you have in mind?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
```diff
@@ -75,11 +76,15 @@ public VirtualFile getFileFromPath(@NotNull Project project, @NotNull String pat
...
+          .findPackage(getPackageFromPath(path));
+      // If a class isn't in a project's classpath, psiPackage will be null.
+      if (psiPackage != null) {
+        PsiClass[] matchingClasses = psiPackage.findClassByShortName(
```
elharo: any chance this is going to find the wrong class?

joaoandremartins: It gets the wrong class if two classes with the same package and class name exist in the same project. I think this is very unusual and can only happen if there is more than one library, with two or more classes having the same package and name.

If this were to happen, we wouldn't have any way of deciding what the right class would be and would have to ask the user for. Then there would also be a chance the user wouldn't know which class that would be...

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ServerToIDEFileResolver.java
patflynn: That's the right way to deal with it.  LGTM.

patflynn:
I'm going to merge this so I can cut a release this morning.

PR #423, opened 2016-02-05
forcing redraw of snapshot results when snapshot was pre-selected in …

### LINKED ISSUE #416: Pre-selected snapshots do not show stack frames or variables
In IJ..
1) Attach a debugger to a debuggable target
2) Select a line of code and create a snapshot
3) Select the snapshot in the debugger window (e.g. click on the entry so that its selected)
4) Trigger the snapshot by visiting the app

Notice that no frames or variables are display. If you rerun the scenario, but don't pre-select the snapshot (skip Step #3), then the debug data shows up.

etanshaul:
…pending state

Fixes #416 

The issue was that when the user selected a snapshot (by clicking on the row in the IJ debug window) prior to the snapshot being triggered, the results were not being redrawn when the data came in. Previously the stack frames and variables were only redrawn on a mouse click; and the mouse click is ignored if it was a click on an already selected row.

The solution is to force redraw the results when they come in if the row was _already_ selected before they came in.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
@@ -286,9 +286,7 @@ public class CloudDebugProcess extends XDebugProcess implements CloudBreakpointL
                     return;
                   }
 
-                  if (myCurrentSnapshot == null || !myCurrentSnapshot.getId().equals(result.getId())) {
-                    navigateToBreakpoint(result);
-                  }
+                  navigateToBreakpoint(result);
                 }
               }
             });
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudDebugHistoricalSnapshots.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudDebugHistoricalSnapshots.java
@@ -557,10 +557,22 @@ public class CloudDebugHistoricalSnapshots extends AdditionalTabComponent
             .showDialog(myProcess.getBreakpointHandler().getXBreakpoint(breakpoint));
       }
       else if (event.getClickCount() == 1 && breakpoint != null && myTable.getSelectedRows().length == 1) {
-        getModel().unMarkAsNewlyReceived(breakpoint.getId());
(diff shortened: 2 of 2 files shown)


etanshaul:
@patflynn @loosebazooka

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
```diff
@@ -261,7 +261,7 @@ public void run() {
   /**
    * Finds the snapshot associated with the given id and sets it as the active snapshot in the current debug session.
    */
-  public void navigateToSnapshot(@NotNull final String id) {
+  public void navigateToSnapshot(@NotNull final String id, final boolean isSelectedBeforeTrigger) {
```
patflynn: could you add javadoc for isSelectBeforeTrigger?  Typically boolean flags to a method to change its behavior is discouraged for public API.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
```diff
@@ -286,7 +286,7 @@ public void run() {
                     return;
                   }
 
-                  if (myCurrentSnapshot == null || !myCurrentSnapshot.getId().equals(result.getId())) {
+                  if (isSelectedBeforeTrigger || myCurrentSnapshot == null || !myCurrentSnapshot.getId().equals(result.getId())) {
```
elharo: what if we jsut did this unconditionally? i.e. removed the if () completely?

etanshaul: @elharo we could do that. That would mean though that if you already have a snapshot selected, and you click on it again it would do a redundant reloading of the result.

etanshaul:
@patflynn, @elharo - I refactored the code a bit based on both of your comments:

- I moved the extra "override" boolean parameter out of the public API and into the caller.
- I also moved the conditional logic that prevents redundant reloads of snapshot data (when clicking on an already selected row) out to the caller

patflynn:
LGTM

patflynn:
merging to get this into our next release cut.

PR #426, opened 2016-02-08
Fixes #412 - Display logout warning related to cloud debugger sessions only if there is at least one in progress

### LINKED ISSUE #412: Display logout warning related to cloud debugger sessions only if there is at least one in progress
When users click logout, they are presented a confirmation dialog that allows for adding messages to it. Currently the plugin will add a warning about background debug sessions always which would be weird for a user if she never touched the Cloud Debugger functionality, so let's make it conditional on having at least one debug session listening for events in the background.

This issue addresses https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/395#discussion_r51919400

akerekes:


google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -20,6 +20,9 @@ Code inspections for AppEngine Java code.</description>
     <component>
       <implementation-class>com.google.gct.idea.CloudToolsPluginInitializationComponent</implementation-class>
     </component>
+    <component>
+      <implementation-class>com.google.gct.idea.debugger.CloudDebugProcessStateCollector</implementation-class>
```
loosebazooka: So application components are just intellij instantiated singletons? I should probably use this more.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPollerTimerTask.java
```diff
@@ -7,15 +9,18 @@
  */
 class CloudDebugGlobalPollerTimerTask extends TimerTask {
 
+  private final CloudDebugProcessStateCollector stateCollector;
   private CloudDebugGlobalPoller cloudDebugGlobalPoller;
 
   public CloudDebugGlobalPollerTimerTask(CloudDebugGlobalPoller cloudDebugGlobalPoller) {
     this.cloudDebugGlobalPoller = cloudDebugGlobalPoller;
+    stateCollector = ApplicationManager.getApplication()
```
loosebazooka: I feel like intellij uses a style where they define the `getInstance()` method in `CloudDebugProcessStateCollection` class and that just calls the `ApplicationManager.getApplication().getComp....`

akerekes:
@elharo PTAL

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPollerTimerTask.java
akerekes: Yeah, that's a pretty frequent pattern. Luckily this way mocking these services is quite easy just by registering mocks into the application.

google-cloud-tools-plugin/resources/META-INF/plugin.xml
patflynn: @loosebazooka 
The main property of ApplicationComponents is that they hook into the IntelliJ defined plugin lifecycle. 

If you just want a plugin scoped object you should use a service.

patflynn:
Please don't submit this yet.  I'd like to take a closer look but I'm a little short on time.

google-cloud-tools-plugin/resources/META-INF/plugin.xml
akerekes: In light of that I think this is more like a service rather than a component.

loosebazooka: Yeah, I read about this last night. Service seems like the way to go

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessStateCollector.java
```diff
@@ -0,0 +1,96 @@
+/*
```
joaoandremartins: Why do we need this new class? From its name, its functionality sounds similar to CloudDebugGlobalPoller.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessStateCollector.java
```diff
@@ -0,0 +1,96 @@
...
+
+import org.jetbrains.annotations.NotNull;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Retrieves {@link CloudDebugProcessState}s from application run configurations
```
joaoandremartins: nit: {@link CloudDebugProcessState}s doesn't look great in Java doc. "{@link CloudDebugProcessState} objects" has better formatting.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessStateCollector.java
```diff
@@ -0,0 +1,96 @@
...
+    return debuggingProfiles;
+  }
+
+  @VisibleForTesting
+  boolean notStoppedAndHasRunProfile(XDebugSession session) {
+    return !session.isStopped() && session.getRunProfile() != null;
+  }
+
+  @VisibleForTesting
+  boolean notRunningConfiguration(Set<RunProfile> runningConfigurations, RunConfiguration config) {
```
joaoandremartins: What do you think of renaming this method to something like "isConfigurationRunning" and then checking on the if for "!isRunningConfiguration()"? I found this version more readable.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessStateCollector.java
akerekes: Makes testing easier and also it has nothing to do with CloudDebugGlobalPoller's state, it's a completely independent functionality.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessStateCollector.java
akerekes: I'll fix this.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessStateCollector.java
akerekes: Although I agree that the method in itself is not very well named, but in the call context I think it makes the code more readable, and I also tend to miss negations in if conditions, so I prefer to avoid them.

google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -97,6 +97,11 @@ Code inspections for AppEngine Java code.</description>
             implementation="com.google.gct.login.IntelliJGoogleLoginMessageExtender"/>
   </extensions>
 
+  <extensions defaultExtensionNs="com.intellij">
+    <applicationService serviceInterface="com.google.gct.idea.debugger.CloudDebugProcessStateCollector"
```
loosebazooka: I think the serviceInterface param is redundant here and can be skipped?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPollerTimerTask.java
```diff
@@ -7,15 +9,17 @@
  */
 class CloudDebugGlobalPollerTimerTask extends TimerTask {
 
+  private final CloudDebugProcessStateCollector stateCollector;
   private CloudDebugGlobalPoller cloudDebugGlobalPoller;
 
   public CloudDebugGlobalPollerTimerTask(CloudDebugGlobalPoller cloudDebugGlobalPoller) {
     this.cloudDebugGlobalPoller = cloudDebugGlobalPoller;
+    stateCollector = ServiceManager.getService(CloudDebugProcessStateCollector.class);
```
loosebazooka: are we opposed to doing

```
CloudDebuggerProcessStateCollector.getInstance() {
  return ServiceManager.getService(CloudDebugProcessCollector.class)
}
```

and just calling getInstance() here?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessStateCollector.java
joaoandremartins: How is testing made easier?

And I disagree that functionality is independent. I would expect that a CloudDebugGlobalPoller polls everything - snapshots and states in this case. What we could want to do is to rename CloudDebugGlobalPoller to something else (like CloudBreakpointPoller/Collector), but you might want to leave a TODO for another PR.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessStateCollector.java
joaoandremartins: In that case, you're rendering the ! operator useless, or making it more complicated. 

I know that these methods are package-protected, but if I were to use them outside of this class, I would want them to tell me if, e.g., a config is running. Otherwise, to know that, I would need to do !notRunningConfiguration, which isn't very readable.

google-cloud-tools-plugin/testSrc/com/google/gct/idea/debugger/CloudDebugProcessStateCollectorTest.java
```diff
@@ -0,0 +1,292 @@
...
+  private XDebugSession createMockSession(boolean isStopped, RunProfile runProfile) {
+    XDebugSession debugSession = mock(XDebugSession.class);
+    when(debugSession.isStopped()).thenReturn(isStopped);
+    if (runProfile != null) {
+      when(debugSession.getRunProfile()).thenReturn(runProfile);
+    }
+    return debugSession;
+  }
+
+}
```
joaoandremartins: Add carriage return.

google-cloud-tools-plugin/testSrc/com/google/gct/login/IntelliJGoogleLoginMessageExtenderTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Test
+  public void testAdditionalLogoutMessage_returnsMessageWhenAStateListens() throws Exception {
+    when(stateCollector.getBackgroundListeningStates())
+        .thenReturn(Collections.singletonList(mock(CloudDebugProcessState.class)));
+    assertThat(new IntelliJGoogleLoginMessageExtender().additionalLogoutMessage(),
+               is("Any Cloud Debugger sessions listening in the background will be stopped."));
+  }
+
+}
```
joaoandremartins: Add carriage return.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessStateCollector.java
akerekes: Easier: I prefer to test smaller classes resulting in smaller test classes. To test just this method it's about 250 lines and CloudDebugGlobalPoller has way more functionality (which is currently not very well tested) which could lead to a huge test class.

Querying the states that are listening in the background does not depend on any fields/methods of CloudDebugGlobalPoller, hence I think querying them is independent of that class' state.
CloudDebugGlobalPoller indeed queries snapshots and snapshot locations, but how I think about CloudDebugProcessStateCollector is more like a repository functionality which currently has only one use case: returning the debug sessions that listen in the background. Maybe in the future we'll need others like: return all cloud debugger sessions, return all active cloud debugger sessions, etc. I'd put those into this class.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessStateCollector.java
akerekes: I'm still using ! when it does not seem to be more useful to write a method to wrap negation, e.g. !List.isEmpty(). My reason here for wrapping it is that the method calling notRunningConfiguration() is already pretty complex (and it was even more before refactoring it), and not needing the extra step to convert ! into "not" when I'm reading the code helps me.

google-cloud-tools-plugin/resources/META-INF/plugin.xml
akerekes: Fixing it.

PR #439, opened 2016-02-12
clear frames and variables tables after breakpoint delete

### LINKED ISSUE #434: Removing a selected breakpoint does not clear out frames and variables
Before:

image

After clicking the "remove" button:

image

Also happens when clicking "delete all"

etanshaul:
fixes #434

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
@@ -259,6 +259,24 @@ public class CloudDebugProcess extends XDebugProcess implements CloudBreakpointL
   }
 
   /**
+   * Clears out the current execution stack while keeping the debug session alive.
+   * Useful for breakpoint delete operations which result in no selected snapshots, requiring us
+   * to display an empty stack in the UI, while keeping the debug session alive.
+   */
+  public void clearExecutionStack() {
+    SwingUtilities.invokeLater(new Runnable() {
+      @Override
+      public void run() {
+        if (!getXDebugSession().isStopped()) {
+          // Since there is no equivalent metaphor in traditional debug sessions, this simulates the desired
+          // behavior of clearing the current context by setting the current position to an empty context
+          getXDebugSession().positionReached(new XSuspendContext() {});
+        }
+      }
+    });
+  }
+
+  /**
    * Finds the snapshot associated with the given id and sets it as the active snapshot in the current debug session.
    */
   public void navigateToSnapshot(@NotNull final String id) {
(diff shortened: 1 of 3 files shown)


google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/CloudDebugHistoricalSnapshots.java
```diff
@@ -659,6 +659,8 @@ public void run() {
         myBalloon.show(new RelativePoint(myTable, new Point(myTable.getWidth() / 2, rectangle.y)), Position.above);
 
         reloadSnapshot();
+      } else if(oldModel.hasPendingDeletes() && oldModel.getBreakpoints().size() > newModel.getBreakpoints().size()){
```
elharo: space after if(

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
```diff
@@ -259,6 +259,21 @@ public void run() {
   }
 
   /**
+   * Clears out frames and variables
```
elharo: period at end of sentence

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
```diff
@@ -259,6 +259,21 @@ public void run() {
...
+   * Useful for breakpoint delete operations which result in no selected row
+   */
+  public void clearExecutionStack() {
+    SwingUtilities.invokeLater(new Runnable() {
+      @Override
+      public void run() {
+        if (!getXDebugSession().isStopped()) {
+          getXDebugSession().positionReached(new XSuspendContext() {});
```
elharo: not obvious to me how this clears the frame, but I'll take your word for it.

elharo:
lgtm

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcess.java
patflynn: I don't really understand what's going on here.  How does this clear the stack?  Is this what Jetbrains does when you stop a debug session and it clears the debugger context?

etanshaul: I agree its a little unclear:

Essentially what is going is that its telling IJ the a position was reached with an empty Context. This has the effect of clearing out the variables and the frames from the UI as well (since we are now in an empty context). 

I was trying to find an alternative mechanism to achieve this but wasn't able to (open to suggestions here). 

When you stop a "normal" debug session in IJ, I believe that it calls the stop() method of the XDebugSession which as the effect of clearing out ALL of our snapshot frames and ending the debug session altogether. In our case, we only want to clear out an individual debug location and display an empty stack _while keeping the debug session alive_ (there's no equivalent of this as far as I can tell in normal IJ debug sessions).

patflynn: can you add an explanatory comment to that line?

LGTM.

etanshaul: Yes, good point.

PR #441, opened 2016-02-12
manually setting plugin version range so that we can cover versions 1…

### LINKED ISSUE #438: We should make our plugin compatible with the IJ 16 EAP.
This should hopefully be as simple as changing the gradle-intellij-plugin config.

etanshaul:
…5+ (including 16EAP) - temporary workaround

fixes #438 

Temporary workaround for inability to individually control "since" and "until" version attributes in plugin.xml via the gradle-intellij-plugin. Setting the range manually to allow us to cover all versions 15+ in a single release.

updateSinceUntilBuild = false - this prevents the plugin from adding version ranges of its own. Instead it will leave the manually entered ranges as is.

NOTE: this is not a permanent solution. This is temporary until the gradle-intellij-plugin issue referenced in #438 is resolved.

### DIFF
--- a/google-account-plugin/resources/META-INF/plugin.xml
+++ b/google-account-plugin/resources/META-INF/plugin.xml
@@ -19,6 +19,9 @@
   <vendor>Google</vendor>
   <!-- "version" set by gradle-intellij-plugin -->
   <!-- "idea-version since-build" set by gradle-intellij-plugin -->
+  <!-- Workaround for gradle-intellij-plugin limitation: https://github.com/JetBrains/gradle-intellij-plugin/issues/66 -->
+  <!-- Setting version to 15.0.3 and up. https://confluence.jetbrains.com/display/IDEADEV/IntelliJ+IDEA+15+143.1821.5+Release+Notes -->
+  <idea-version since-build="143.1821.5" />
 
   <description>
     Provides Google account setting and authentication for IntelliJ plugins that need it.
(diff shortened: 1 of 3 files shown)


google-account-plugin/resources/META-INF/plugin.xml
```diff
@@ -19,6 +19,8 @@
   <vendor>Google</vendor>
   <!-- "version" set by gradle-intellij-plugin -->
   <!-- "idea-version since-build" set by gradle-intellij-plugin -->
+  <!-- Workaround for gradle-intellij-plugin limitation: https://github.com/JetBrains/gradle-intellij-plugin/issues/66 -->
+  <idea-version since-build="143.381.42" />
```
patflynn: which version of IJ does that correspond to?  Can you include a link to the mapping table?

patflynn:
LGTM. I'd like to cut a new release based off of this change this morning.

etanshaul:
The version I hardcoded corresponds to the earliest non-EAP version of IJ 15. I found it experimentally (and then confirmed it) by giving the IJ gradle plugin version 15.0 and then looking at the raw plugin.xml that it produces.

I'll find a table and include it in the comments.

patflynn:
After offline conversation with Etan we've decided to make the since version whatever we're building against.  

The rationale is that we don't want to support older point releases, as we're always going to ask the user to upgrade their IDE rather than fix deprecated versions.

PR #443, opened 2016-02-16
warning user to ensure that app has running instances

### LINKED ISSUE #345: Add warning mentioning users must have running instances on their app to be able to take snapshots
The CDB API only returns debuggable modules which have alive instances. If no module does, the IntelliJ attach dialog only shows a generic error message.

I noticed there's no mention of the alive instances message anywhere in our docs. Is this on purpose, or should we say in the attach dialog something like "Ensure your application has alive instances so you can take snapshots"? I spent some time wondering why I couldn't attach and such a message would have saved me some time.

etanshaul:
Fixes #345

When the CDB API returns 0 modules without throwing exceptions we warn the user that no modules were found and suggest to ensure that there are running instances. 

My assumption here is that the message is generic enough to accurately cover all scenarios where the API returns 0 modules without throwing an exception - which happens when either the project has no modules at all, or no instances that are awake.

@akerekes @joaoandremartins @patflynn ptal


elharo:
lgtm

akerekes:
lgtm

patflynn:
We could actually provide them with a uri to ping their live instances couldn't we?  I think Erez included it an older thread, although I could be misremembering.

etanshaul:
Assuming that we can identify the sleeping instances, as well as get a uri for each module to ping (which I need to verify), then there would be non-trivial ux changes that would need to accompany this fix. Here's one route:

- The target dropdown lists _all_ modules, both awake _and_ sleeping. 
- If an awake module is selected, the flow is as usual - click 'Attach' to debug.
- If a sleeping module is selected, instead of showing the user an 'Attach' button, the user is shown a 'Wake Instance' button (or some equivalent). 
- After clicking to wake the instance and showing a loading icon, the button now becomes a regular 'Attach' button (assuming the wakeup was successful).

patflynn:
Let's check in this change and add an enhancement issue to look into that.

PR #460, opened 2016-02-22
Fixes #459 - When backend server has changes it causes error on the client side

### LINKED ISSUE #459: When backend server has changes it causes error on the client side
When a breakpoint is set, and the user keeps a debug session listening in the background the client periodically checks the Cloud Debugger backend for changes with a hanging get using a wait token.
If the breakpoint is hit, the backend will return an HTTP 409 to the client to notify it about results being ready to query. When this 409 is received due to recent changes in how we handle errors in requests, it was also handled as error this showing an IDE error notification to the user, instead of simply suppressing it.

akerekes:
@etanshaul @joaoandremartins @elharo PTAL

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
@@ -16,7 +16,6 @@
 package com.google.gct.idea.debugger;
 
 import com.google.api.client.googleapis.json.GoogleJsonResponseException;
-import com.google.api.client.http.HttpStatusCodes;
 import com.google.api.client.repackaged.com.google.common.base.Strings;
 import com.google.api.services.clouddebugger.Clouddebugger.Debugger;
 import com.google.api.services.clouddebugger.Clouddebugger.Debugger.Debuggees;
@@ -36,6 +35,7 @@ import com.intellij.util.containers.ContainerUtil;
 import org.jetbrains.annotations.NotNull;
 
 import java.io.IOException;
+import java.net.HttpURLConnection;
 import java.net.SocketTimeoutException;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -129,8 +129,9 @@ public class CloudDebugGlobalPoller {
   }
 
   /**
-   * pollForChanges sends a synchronous, non-hanging query to the server and compares the result to
-   * see if there are changes from the current state.
+   * pollForChanges sends a synchronous, hanging query to the server and compares the result to
+   * see if there are changes from the current state. Explanation of
+   * <a href="https://en.wikipedia.org/wiki/Push_technology#Long_polling">hanging query (a.k.a. long poll)</a>
    *
    * @param state represents the target debuggee to query
    */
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
```diff
@@ -169,6 +169,12 @@ void pollForChanges(@NotNull final CloudDebugProcessState state) {
     }
     catch(SocketTimeoutException sto) {
       //noop, this is expected behavior.
+    } catch (GoogleJsonResponseException e) {
+      // HTTP 409 is expected when backend responds to "hanging get" either for timeout or because
+      // a result is available (which will be retrieved via the subsequent query)
+      if (e.getStatusCode() != 409) {
+        handleBreakpointQueryError(state, e);
```
joaoandremartins: Should we log something on a debug/info level just to inform a 409 was received?

elharo: Let's not. Logs are hard enough to parse without a lot fo messages from code that ios doing exactly what we expect it to do.

etanshaul: Also, just so it's clearer to whoever is reading this, maybe add more to the comments explaining the "hanging get" as well as what is meant by "query".

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
```diff
@@ -169,6 +169,12 @@ void pollForChanges(@NotNull final CloudDebugProcessState state) {
     }
     catch(SocketTimeoutException sto) {
       //noop, this is expected behavior.
+    } catch (GoogleJsonResponseException e) {
+      // HTTP 409 is expected when backend responds to "hanging get" either for timeout or because
+      // a result is available (which will be retrieved via the subsequent query)
+      if (e.getStatusCode() != 409) {
```
elharo: named constant HttpURLConnection.HTTP_CONFLICT

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugGlobalPoller.java
joaoandremartins: @elharo that's why I suggested to log it on debug -- it doesn't clutter >=info logs.

We expect 409's to be returned -- however, there might be an issue with the server not returning any 409. Having 409's logged on the client would help debug that scenario.

elharo: YAGNI

akerekes:
@etanshaul @joaoandremartins @elharo PTAL

etanshaul:
lgtm

joaoandremartins:
Aside from the logging issue, LGTM.

elharo:
lgtm

PR #464, opened 2016-02-23
fixes #417 - Re-enabling snapshot location does not re-enable it on the backend

### LINKED ISSUE #417: Re-enabling snapshot location does not re-enable it on the backend
If a snapshot location is set, and the snapshot is taken, the location becomes disabled. If the user right-clicks on it, he can re-enable it in the popup to re-enable snapshot capture for that line (even a message is indicating so). The snapshot location appears to be enabled (blue disc with checkmark), but there is no new item in the debug pane, and indeed when the application hits the snapshot location again, it will not take a new snapshot.
Deleting the disabled maker and re-adding it works as expected, snapshot will be captured at the next execution.

akerekes:


elharo:
lgtm

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudBreakpointHandler.java
```diff
@@ -331,10 +331,10 @@ public void registerBreakpoint(
       // gets called during construction.
       return;
     }
-
-    if (xIdeBreakpoint.getProperties().isAddedOnServer()) {
+    if (xIdeBreakpoint.getProperties().isAddedOnServer() && !xIdeBreakpoint.getProperties().isDisabledByServer()) {
       // If this breakpoint was already added, but was hit offline, let's not add it again once we
-      // resume a debug session.
+      // resume a debug session. unless this reigster request comes from re-enabling a previously disabled
```
joaoandremartins: Nit: "Unless the register request"

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudBreakpointHandler.java
```diff
@@ -440,16 +440,16 @@ public void unregisterBreakpoint(
...
+      String breakpointId = xIdeBreakpoint.getUserData(CLOUD_ID);
+      if (!Strings.isNullOrEmpty(breakpointId)) {
+        myProcess.getStateController().deleteBreakpointAsync(breakpointId);
+      } else {
+        LOG.warn("could not delete breakpoint because it was not added through the cloud handler.");
+      }
     }
+    // reset this flag: either it has been disabled by the server or the client has deleted it, in
+    // both cases we need to add it again, if it is re-enabled
+    xIdeBreakpoint.getProperties().setAddedOnServer(false);
```
joaoandremartins: What happens here when you add a breakpoint, stop the session, resume the session, stop the session, hit the breakpoint offline and resume the session again? Does the original bug resurface?

joaoandremartins:
lgtm

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudBreakpointHandler.java
akerekes: Works as expected (apart from the refresh bug #465)

PR #474, opened 2016-03-07
Fixes #472 - Filter older debuggees when offering debug targets to user

### LINKED ISSUE #472: Filter older debuggees when offering debug targets to user
If a user changes the code that is being debugged, and updates the webapp on AppEngine, when the next time he launches a new debug session (not too long after the update happened), the backend will send back the debuggee for both the latest and the previous revision. If the user does several changes and updates in a short time the backend will send back several debuggees. Currently we select the version that was previously selected (i.e. the not the latest).
We should filter the debuggees received from the backend and based on the minor version, we should show only the latest to the user.

The current workaround for the user is just to wait until the old versions are cleaned up in the backend and only a single debuggee is returned.

akerekes:


elharo:
lgtm

etanshaul:
Maybe add a simple comment describing the situation - that "default" modules do not have module names coming back from the API hence the reason for setting it manually.  lgtm.

patflynn:
I think I'm missing something.  How does this filter older debugees?

etanshaul:
@patflynn it's an indirect effect: `ProjectDebugeeBinding` (around line 176) uses a hashmap to filter out dups by module name. @akerekes's fix ensures that even default modules (that have no module names from the API) get that filtering logic applied to them.

patflynn:
Then definitely +1 to a comment explaining how this works.  I don't think
somebody would understand the behavior intended just by looking at this
code.

patflynn:
LGTM

Issue #241, opened 2015-11-23
Existing Code Analysis issues should be addressed or ignored 


patflynn:
Findbugs and PMD are finding issues in our modules.  We should go through these and address them.

PR #481, opened 2016-03-08
Suppress benign FindBugs warning DM_STRING_CTOR

chanseokoh:
Fixes a subset of #241. Suppresses FindBugs complaining about calling the String() constructor. All the warnings are due to strings being concatenated with "+".

Adds a FindBugs dependency in build.gradle too (using 'compile' instead of 'testCompile' as the latter doesn't seem to work). Uses version 3.0.1.

No DM_STRING_CTOR at all from now on for all the projects. One minor concern is that an annotation applies to an entire method. (No finer granularity is possible when using annotations.)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -63,6 +63,7 @@ subprojects {
             exclude group: 'org.hamcrest', module: 'hamcrest-core'
         }
         testCompile 'org.hamcrest:hamcrest-all:1.3'
+        compile 'com.google.code.findbugs:annotations:3.0.1'
     }
 
     apply plugin: 'findbugs'
@@ -70,7 +71,7 @@ subprojects {
         ignoreFailures = true
         sourceSets = [sourceSets.main]
         effort = "max"
-        toolVersion = '2.0.3'
+        toolVersion = '3.0.1'
     }
 
     tasks.withType(FindBugs) {
--- a/common-test-lib/src/main/java/com/google/gct/idea/testing/TestUtils.java
+++ b/common-test-lib/src/main/java/com/google/gct/idea/testing/TestUtils.java
@@ -19,6 +19,8 @@ import com.intellij.openapi.vfs.encoding.EncodingManager;
 import com.intellij.openapi.vfs.encoding.EncodingManagerImpl;
 import com.intellij.util.pico.DefaultPicoContainer;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.mockito.Mockito;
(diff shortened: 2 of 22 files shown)


chanseokoh:
The test fails because of the 3.0.1 FindBugs version.

build.gradle
```diff
@@ -70,7 +71,7 @@ subprojects {
         ignoreFailures = true
         sourceSets = [sourceSets.main]
         effort = "max"
-        toolVersion = '2.0.3'
+        toolVersion = '3.0.1'
```
patflynn: yeah you'll have trouble here since our continuous build environment uses Java 6.  We may want to reconsider that.

common-test-lib/src/main/java/com/google/gct/idea/testing/TestUtils.java
```diff
@@ -127,6 +129,7 @@ public static MockProject mockProject(@Nullable PicoContainer container) {
     return new MockProject(container, getParentDisposableForCleanup());
   }
 
+  @SuppressFBWarnings(value="DM_STRING_CTOR", justification="Warning due to string concatenation")
```
patflynn: What's the warning/bug in this case?  Is this something we want to blanket disable? Surely findbugs has a way to do that without having to annotate every instance.

chanseokoh: FindBugs warns about calling a String constructor like in a case of String s = new String("str") or SomeFnCall(String("str")). FindBugs suggests to use the string primitive directly. Unfortunately, this warning comes up where strings are concat'ed, e.g., a few lines below, you will see two occurrences of string concats.

It's about negligible performance impact. Perhaps we may blanket disable and forget about this warning forever instead of annotating each method. What's your thought?

patflynn: Something seems off, since the findbug is explicitly about using the String
constructor when the code examples use string concatenation, which I assume
is fine.

If we can't figure out why there's a mismatch we should definitely blanket
disable. I don't want to ban or explicitly mute every instance of String +
concatenation.

chanseokoh: AFAK, FindBugs sees actual bytecode instead of source, so I guess it's explicit that things like new String("str") are happening when "+" are compiled into bytecode. We could check the bytecode directly using a debugger if this is really happening, but I don't know how to do so. Perhaps with some compiler optimization options, this could be avoided?

chanseokoh: Also, there are other warnings in a similar line, e.g., complaining checking for not-null unnecessarily whenever a reference is returned from a method, so being able to look into bytecode somehow would be useful for determining the root cause of other similar warnings.

chanseokoh:
I have inspected a few bytecode files. For the reported cases, "+" eventually calls String.concat(), and a fresh String is "new"ed as a result. I think this is exactly where FindBugs is complaining. Weird enough, I wasn't able to find documents or discussion online about this correlation between string concat and the FindBugs warning.

I believe we don't have any control over how javac generates bytecode and don't have much option left: we either blank disable this warning once and for all, or suppress it by annotating each method for finer granularity, or just ignore and live with it.

patflynn:
let's blanket disable this.

chanseokoh:
I will close this PR and put some configuration in build.gradle (possibly with an addition of an XML file) to disable this String()-constructor-calling warning.

patflynn:
thanks!

PR #482, opened 2016-03-08
update re-open debugger dialog wording when listening in the background

### LINKED ISSUE #480: Misleading wording for continuing a background debug session
1) Start a debug session, and add a breakpoint
2) Click the stop icon and choose to listen in the background
3) Click the debug icon again and notice the following popup:

image

4) Click 'OK' on the above dialog. Notice the the state of the debug dialog now:

image

The wording on these dialogs is inconsistent because the first one asks the user to stop the debug session, and after doing so, the next dialog asks the user to continue the session. The logic is working properly: when reopening the debug session it actually attaches a new session and uses the already existing breakpoints from the server. 

I propose that this should be a wording fix: for instance instead of "start a new session", the wording (in both cases) should could be something like "re-attach your session".

etanshaul:
fixes #480


patflynn:
Is that what it's doing if you hit OK? Just resuming the session?

elharo:
lgtm

etanshaul:
closing: obviated by #488 (this dialog was removed)

PR #487, opened 2016-03-08
Fix several issues reported by FindBugs

chanseokoh:
Addresses several issues reported by FindBugs. (#241)

1. Remove unused field and local variable
No one references it. (Anonymous classes may look like they would reference it later, but they are actually referencing another 'appEngineHelper' qualified with 'final'.
 
2. Copy an array when storing it (i.e., not a ref to a mutable outside object)
Security issue. It is currently storing a reference to a mutable external array. I've checked all usages of 'setWatchExpression' function in the codebase, and it is safe to make a copy. I reckon that performance impact of copying is negligible.

3. Start method names with lower case letters
Close() -> close(), Create() -> create(). Not sure if they were intended though.

4. Always close ZipFile as soon as possible
Not sure if it is tolerable to let it open for a while in this situation.

### DIFF
--- a/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
+++ b/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
@@ -152,3 +152,4 @@ clouddebug.background.listener.general.error.message=There was an unexpected err
 clouddebug.debug.targets.error=Error accessing debug targets: {0}
 clouddebug.debug.targets.accessdenied=Access denied. You may not have the necessary permissions to run Cloud Debugger with this project.
 
+settings.error.closing.file=Error closing settings file {0}.\nError details: {1}
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmDeploymentRunConfigurationEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmDeploymentRunConfigurationEditor.java
@@ -58,13 +58,11 @@ public class ManagedVmDeploymentRunConfigurationEditor extends
   private TextFieldWithBrowseButton dockerFilePathField;
   private JButton generateAppYamlButton;
   private JButton generateDockerfileButton;
-  private AppEngineHelper appEngineHelper;
 
 
   public ManagedVmDeploymentRunConfigurationEditor(final Project project,
       final DeploymentSource source, final AppEngineHelper appEngineHelper) {
     this.project = project;
-    this.appEngineHelper = appEngineHelper;
     configTypeComboBox.setModel(new DefaultComboBoxModel(ConfigType.values()));
     configTypeComboBox.setSelectedItem(ConfigType.AUTO);
     mvmConfigFilesPanel.setVisible(false);
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudLineBreakpointProperties.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudLineBreakpointProperties.java
@@ -20,6 +20,8 @@ import com.intellij.util.xmlb.annotations.Tag;
 import com.intellij.xdebugger.breakpoints.XBreakpointProperties;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.Arrays;
+
 /**
  * CloudLineBreakpointProperties holds custom properties not normally set on a java line breakpoint. Right now, this is
  * just watch expressions.  Custom conditions are supported by default as is the enabled state and other attributes such
@@ -68,7 +70,9 @@ public class CloudLineBreakpointProperties extends XBreakpointProperties<CloudLi
 
   public final boolean setWatchExpressions(@Nullable String[] watchExpressions) {
     boolean changed = !arrayEqual(myWatchExpressions, watchExpressions);
-    myWatchExpressions = watchExpressions;
+    if (changed) {
+      myWatchExpressions = watchExpressions == null ? null : Arrays.copyOf(watchExpressions, watchExpressions.length);
+    }
     return changed;
   }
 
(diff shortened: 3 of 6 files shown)


google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmDeploymentRunConfigurationEditor.java
```diff
@@ -58,13 +58,11 @@
   private TextFieldWithBrowseButton dockerFilePathField;
   private JButton generateAppYamlButton;
   private JButton generateDockerfileButton;
-  private AppEngineHelper appEngineHelper;
 
 
   public ManagedVmDeploymentRunConfigurationEditor(final Project project,
       final DeploymentSource source, final AppEngineHelper appEngineHelper) {
     this.project = project;
-    this.appEngineHelper = appEngineHelper;
```
patflynn: we should also remove the constructor parameter.

chanseokoh: That constructor parameter is actually required since it is this 'final' parameter that anonymous classes below will reference.

google-cloud-tools-plugin/src/com/google/gct/idea/settings/ImportSettings.java
```diff
@@ -122,6 +124,17 @@ public static void doImport(String path) {
...
                                DIALOG_TITLE);
     }
+    finally {
+      try {
+        if (saveZipFile != null) {
+          saveZipFile.close();
+        }
+      } catch (IOException e1) {
+        Messages.showErrorDialog(
+          "Error closing file " + presentableFileName(saveFile) + ".\\nThere was " + e1.getMessage(),
```
akerekes: This should come from a message bundle like the one a few lines above.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudLineBreakpointProperties.java
```diff
@@ -68,7 +70,7 @@ public void loadState(CloudLineBreakpointProperties state) {
 
   public final boolean setWatchExpressions(@Nullable String[] watchExpressions) {
     boolean changed = !arrayEqual(myWatchExpressions, watchExpressions);
-    myWatchExpressions = watchExpressions;
+    myWatchExpressions = watchExpressions == null ? null : Arrays.copyOf(watchExpressions, watchExpressions.length);
```
patflynn: it seems kind of weird to me to do a copy even when nothing has changed.

google-cloud-tools-plugin/src/com/google/gct/idea/settings/ImportSettings.java
```diff
@@ -55,6 +55,7 @@
    */
   public static void doImport(String path) {
     final File saveFile = new File(path);
+    ZipFile saveZipFile = null;
```
patflynn: why move the declaration up here?

google-cloud-tools-plugin/src/com/google/gct/idea/settings/ImportSettings.java
```diff
@@ -122,6 +124,17 @@ public static void doImport(String path) {
       Messages.showErrorDialog(IdeBundle.message("error.reading.settings.file.2", presentableFileName(saveFile), e1.getMessage()),
                                DIALOG_TITLE);
     }
```
patflynn: 'finally' is placed on the same line as the closing curly brace it follows.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmDeploymentRunConfigurationEditor.java
patflynn: aha! Got it.

google-cloud-tools-plugin/src/com/google/gct/idea/settings/ImportSettings.java
chanseokoh: I added a message to 'CloudToolsBundle.properties' and now use 'com.google.gct.idea.util.GctBundle'.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudLineBreakpointProperties.java
chanseokoh: Good point. Now it copies only when 'changed' is true.

google-cloud-tools-plugin/src/com/google/gct/idea/settings/ImportSettings.java
chanseokoh: Fixed.

google-cloud-tools-plugin/src/com/google/gct/idea/settings/ImportSettings.java
chanseokoh: Seems like I have to put it there so that I can use it inside 'finally' to close the file. Is there some other elegant way?

patflynn: doh. yeah you're doing it right. Once we move away from java 6 we can use try with resources.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -152,3 +152,4 @@ clouddebug.background.listener.general.error.message=There was an unexpected err
 clouddebug.debug.targets.error=Error accessing debug targets: {0}
 clouddebug.debug.targets.accessdenied=Access denied. You may not have the necessary permissions to run Cloud Debugger with this project.
 
+settings.error.closing.file=Error closing settings file {0}.\nThere was {1}
```
patflynn: This message is a bit weird.  Maybe just replace 'There was ..' with 'Error details:.. '

Typically for these kind of errors we probably just want to log error rather than show a dialog, but I don't want to waste time on this class since it's not actually in use.

patflynn:
LGTM once you've changed the message text and squashed the commits.

google-cloud-tools-plugin/src/com/google/gct/idea/settings/ImportSettings.java
chanseokoh: Interesting. Good to know about try with resources.

PR #488, opened 2016-03-08
ui updates for re-attaching to debugger when previously put in the ba…

### LINKED ISSUE #483: When continuing a debug background session the "ok" button text should change if a different module is selected
Currently, if the user puts the debug session for Module A in the background, then opens the debug dialog to continue debugging but then selects Module B (which was not in the background), the ok action button text says "Continue Session" when it should instead say "Attach". The background listener was already killed off when the debug dialog was re-opened so this is only a wording / UX bug.

etanshaul:
…ckground

fixes #483 

Setup: The user puts a debug session in the background, then attempts to open the debugger again - either by clicking the debug icon, or by triggering a snapshot then clicking on the popup to open the debugger.

The following dialog is now gone:

image

The user is sent directly to the debug dialog:

image

If the user switches to a different module (not the one that was being debugged in the background), the button text will be changed and a warning will be shown:

image


patflynn:
can you attach a screenshot of the new UI?

patflynn:
LGTM

PR #489, opened 2016-03-09
Avoid bad practice of comparing Boolean

chanseokoh:
Fix a subset of #241 (FindBugs issue: RC_REF_COMPARISON_BAD_PRACTICE_BOOLEAN).

Comparing Booleans using == or != might pose a problem, e.g., if Boolean objects are created using the 'new Boolean(b)' constructor (although creating such objects would be an unusual practice). In such cases, == or != can give different results than '.equals()' would do.

However, this PR hurts readability slightly IMHO. If you think it is just ok to use == or !=, feel free to ditch this PR.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/util/PsiUtils.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/util/PsiUtils.java
@@ -201,12 +201,13 @@ public class PsiUtils {
       return false;
     }
 
-    return Boolean.TRUE == type.accept(new PsiTypeVisitor<Boolean>() {
+    Boolean accepted = type.accept(new PsiTypeVisitor<Boolean>() {
       @Nullable
       @Override
       public Boolean visitClassType(PsiClassType classType) {
         return classType.getParameterCount() > 0;
       }
     });
+    return Boolean.TRUE.equals(accepted);
   }
 }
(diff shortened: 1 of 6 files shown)


google-cloud-tools-plugin/src/com/google/gct/idea/appengine/util/PsiUtils.java
```diff
@@ -201,12 +201,12 @@ public static boolean isParameterizedType(PsiType type) {
       return false;
     }
 
-    return Boolean.TRUE == type.accept(new PsiTypeVisitor<Boolean>() {
+    return Boolean.TRUE.equals(type.accept(new PsiTypeVisitor<Boolean>() {
```
elharo: for readability I would suggest simply if (...) return true; else return false.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/BreakpointUtil.java
```diff
@@ -41,7 +41,7 @@
    */
   @Nullable
   public static String getUserErrorMessage(@Nullable StatusMessage statusMessage) {
-    if (statusMessage == null || statusMessage.getIsError() != Boolean.TRUE) {
+    if (statusMessage == null || !Boolean.TRUE.equals(statusMessage.getIsError())) {
```
elharo: Is there something going on with null here that prevents us from doing a simple || statusMessage.getIsError()?

elharo:
Why are we not just using simple booleans here?

chanseokoh:
That's not something I can determine as this point. Perhaps Cloud Debugger may not explicitly assign (truth) values for fields such as 'isError', and that may be the reason for using Boolean.

elharo:
I think we should fiogure that out before proceeding. The code, as is, is pretty strange.

patflynn:
LGTM.  Let's not try to fix readability in the same CLs. Otherwise Chanseok will not make any progress.  Let's just do the suggested fix for each findbug warning/error (assuming it makes sense).  We can file bugs or add todos to fix other questionable things we come across along the way.

PR #490, opened 2016-03-09
Removing 'my..' prefix from field names.

patflynn:
Prefixing Java field names with 'my' is not Google style. This change
removes all 'my' prefixes from class fields in our code-base.

Good luck keeping your eyes open reviewing this.

@loosebazooka 
@chanseokoh 
@etanshaul

### DIFF
--- a/google-account-plugin/src/com/google/gct/login/ui/GoogleLoginUsersPanel.java
+++ b/google-account-plugin/src/com/google/gct/login/ui/GoogleLoginUsersPanel.java
@@ -56,11 +56,11 @@ public class GoogleLoginUsersPanel extends JPanel implements ListSelectionListen
   private static final String SIGN_OUT = "Sign Out";
   private static final int MAX_VISIBLE_ROW_COUNT = 3;
 
-  private JBList myList;
-  private DefaultListModel myListModel;
-  private JButton mySignOutButton;
-  private boolean myValueChanged;
-  private boolean myIgnoreSelection;
+  private JBList list;
+  private DefaultListModel listModel;
+  private JButton signOutButton;
+  private boolean valueChanged;
+  private boolean ignoreSelection;
 
   public GoogleLoginUsersPanel() {
     super(new BorderLayout());
@@ -69,10 +69,10 @@ public class GoogleLoginUsersPanel extends JPanel implements ListSelectionListen
     final UsersListCellRenderer usersListCellRenderer = new UsersListCellRenderer();
 
     //Create the list that displays the users and put it in a scroll pane.
-    myList = new JBList(myListModel) {
+    list = new JBList(listModel) {
       @Override
       public Dimension getPreferredScrollableViewportSize() {
-        int numUsers = myListModel.size();
+        int numUsers = listModel.size();
         Dimension superPreferredSize = super.getPreferredScrollableViewportSize();
         if(numUsers <= 1) {
           return superPreferredSize;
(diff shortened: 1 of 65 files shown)


google-account-plugin/src/com/google/gct/login/ui/GoogleLoginUsersPanel.java
```diff
@@ -56,11 +56,11 @@
   private static final String SIGN_OUT = "Sign Out";
   private static final int MAX_VISIBLE_ROW_COUNT = 3;
 
-  private JBList myList;
-  private DefaultListModel myListModel;
-  private JButton mySignOutButton;
-  private boolean myValueChanged;
-  private boolean myIgnoreSelection;
+  private JBList list;
```
loosebazooka: not sure I like this name "list"

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/BreakpointErrorStatusPanel.form
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
loosebazooka: did we do right by the form editor? Do they still sync up?

google-account-plugin/src/com/google/gct/login/ui/GoogleLoginUsersPanel.java
patflynn: I agree. But I want this to be a straight 'my' strip. We can improve naming
in another CL.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/ui/BreakpointErrorStatusPanel.form
patflynn: they appear to.

patflynn: if they didn't I don't think it would build, which is nice.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessStateController.java
```diff
@@ -122,8 +122,8 @@ public void fireBreakpointsChanged() {
    * @param state the {@link CloudDebugProcessState} the controller will be bound to
    */
   public void initialize(@NotNull CloudDebugProcessState state) {
-    myState = state;
-    myState.setWaitToken(null);
+    this.state = state;
+    state.setWaitToken(null);
```
loosebazooka: what happens here?

loosebazooka: actually this seems fine, but looks weird.

patflynn:
For prosperity this was how I did the rename: 
find . -type f -name "*.java" -o -name "*.form" -exec sed -i -r 's/my([[:upper:]][[:lower:]]*)([[:upper:]]\w*\>)?/\l\1\2/g' {} +

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudStackFrame.java
```diff
@@ -126,21 +126,21 @@ public Object getEqualityObject() {
 
   @Override
   public XSourcePosition getSourcePosition() {
-    return myXSourcePosition;
+    return xSourcePosition;
   }
 
   private static class MyValue extends XValue {
```
loosebazooka: what about this MyValue inner class?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/SyncResult.java
```diff
@@ -24,42 +24,42 @@
...
-    myNeedsSync = needsSync;
-    myTargetSyncSHA = targetSyncSHA;
-    myTargetLocalRepository = targetRepository;
-    myHasRemoteRepository = hasRemoteRepository;
-    myRepoType = repoType;
+    this.invalidDebuggee = invalidDebuggee;
+    this.needsStash = needsStash;
+    this.needsSync = needsSync;
+    this.targetSyncSHA = targetSyncSHA;
+    targetLocalRepository = targetRepository;
```
loosebazooka: this?

loosebazooka: opop never mind

google-cloud-tools-plugin/testSrc/com/google/gct/login/MockGoogleLogin.java
```diff
@@ -24,14 +24,14 @@
  */
```
loosebazooka: why do we have two MockGoogleLogin.java files?

loosebazooka:
lgtm for the most part, someone might want to go bottom to top on this, my concentration seemed to have waned a little past 70%

chanseokoh:
After renaming the 'myRootPanel' field to 'rootPanel' in 'elysium/SelectUserDialog.java' and also setting 'custom-create="true"' in 'SelectUserDialog.form', now FindBugs suddenly complains that 'rootPanel' is being read without ever written. Why do you think this is happening? I still think 'rootPanel' is somehow created and initialized outside java code, but I wonder why this warning comes up suddenly.

patflynn:
custom-create should be false. I thought I had fixed that.  I'll change it
back.

I think the reason that findbugs found it is that it works on byte code and
without custom create the assignment is autogenerated.

PR #491, opened 2016-03-09
Address potential NullPointerExceptions

chanseokoh:
Fix a subset of #241 (FindBugs issue: NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE).

Addresses points of a potential NullPointerException. Some warnings are suppressed.

Some explanations:
1. PsiAnnotation.findAttributeValue() is 'Nullable', but comments in our source seems to saying that it will return some default value in those case.
2. PsiParameter.getModifierList() is 'Nullable'. For example, if called by an anonymous inner class, it will return null as it cannot have modifiers. However, since we are calling it against parameters, I think it will not return null in our case.

About suppressing warnings: granularity is per method. Unfortunately, no finer granularity seems possible. That is, we will not be able to further detect this issue for the methods annotated to suppress this warning.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -63,6 +63,7 @@ subprojects {
             exclude group: 'org.hamcrest', module: 'hamcrest-core'
         }
         testCompile 'org.hamcrest:hamcrest-all:1.3'
+        compile 'com.google.code.findbugs:annotations:3.0.1'
     }
 
     apply plugin: 'findbugs'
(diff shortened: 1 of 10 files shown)


chanseokoh:
The test fails because of the 3.0.1 FindBugs version.

build.gradle
```diff
@@ -63,6 +63,7 @@ subprojects {
             exclude group: 'org.hamcrest', module: 'hamcrest-core'
         }
         testCompile 'org.hamcrest:hamcrest-all:1.3'
+        compile 'com.google.code.findbugs:annotations:3.0.1'
```
patflynn: you should probably switch your jdk to 6 so that you can run this stuff with the older version of findbugs which is compatible with our target platform.

chanseokoh:
Closing this PR after opening PR #499. #499 contains only a subset of this PR.

PR #492, opened 2016-03-10
Rename all our.. static fields.

patflynn:
ourBlaBla is not java style, so I'm renaming static fields that use it.


elharo:
lgtm

PR #495, opened 2016-03-10
Address two FindBugs warnings.

chanseokoh:
#241. Surprisingly, after this fix, there will be left only a couple of warnings that need individual attention. All the other remaining issues will be either a NullPointerException case or a redundant null checking.

1. Put 'static' to a constant field.
2. Remove an unused local variable (SettableFuture).

For the second one, please take a look at the code carefully. It is certain that the variable is never used. However, because it is a Future object from the beginning, its presents simply makes me believe that there should have been something going on with this object.


patflynn:
LGTM.

elharo:
lgtm

PR #498, opened 2016-03-14
Address five FindBugs warnings

chanseokoh:
#241. Address the following five FindBugs warnings. Set up FindBugs filter configuration (a new XML file) as well.

1. Concurrent calls may not be atomic (1 instance)
There do exists a concurrency problem because concurrent calls are currently not atomic. However, judging from the nature of using Cloud Debugger, I think this is tolerable. Users may retry if an extremely-rarely-occurring error happens.

2. Inefficient 'new String("plain sring")' constructor (21 instances)
As discussed in a previous pull request, this disables the warning once and for all.

3. Redundant null checks of known non-null values (66 instances)
As discussed, this disables the warning once and for all.

4. String concat("+") in a loop

5. Address NPE (2 instances)

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/validation/RestSignatureInspection.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/validation/RestSignatureInspection.java
@@ -494,7 +494,7 @@ public class RestSignatureInspection extends EndpointInspectionBase {
    * that does not have @Nullable/@Default.
    */
   private String getPathParameter(PsiMethod method) {
-    String path = "";
+    StringBuilder path = new StringBuilder();
     EndpointPsiElementVisitor elementVisitor = new EndpointPsiElementVisitor();
     List<String> annotions =
       Arrays.asList(GctConstants.APP_ENGINE_ANNOTATION_NULLABLE, "javax.annotation.Nullable", GctConstants.APP_ENGINE_ANNOTATION_DEFAULT_VALUE);
@@ -512,11 +512,11 @@ public class RestSignatureInspection extends EndpointInspectionBase {
 
       PsiAnnotationMemberValue namedValue = elementVisitor.getNamedAnnotationValue(aParameter);
       if(namedValue != null) {
-        path += "/{}";
+        path.append("/{}");
       }
     }
 
-    return path;
+    return path.toString();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


google-cloud-tools-plugin/src/com/google/gct/idea/appengine/validation/RestSignatureInspection.java
```diff
@@ -488,7 +488,7 @@ private String guessResourceName(PsiMethod method) {
    * that does not have @Nullable/@Default.
    */
   private String getPathParameter(PsiMethod method) {
-    String path = "";
+    StringBuffer path = new StringBuffer();
```
elharo: prefer StringBuilder unless you require thread safety (and here I don't think you do)

patflynn:
what line is the concurrent find bug issue on?

chanseokoh:
In com.google.gct.idea.debugger.CloudDebuggerClient.getClient():

Line 99 and 144 both access a ConccurentHashMap.

Theoretically, two separate getClient() calls with a same key can reach line 99, and both may see the map return null. The two callers then build their own cloudDebuggerClient instance (line 130). At line 144, only one of the two cloudDebuggerClient instances may survive. In this case, one caller may get (as a return value) a ref to a cloudDebuggerClient instance that is not stored in the cache.

Likewise, it is possible that a caller may retain and use a ref not present in the cache due to .remove(hashkey) on line 110.

patflynn:
Can we  use debuggerClientsFromUserEmail.computeIfAbsent() to initialize
the client safely?

chanseokoh:
computeIfAbsent() doesn't seem to exist prior to Java 8.

patflynn:
that's a bummer.  How about just synchronizing on a lock?

chanseokoh:
I could use a lock. However, because .remove(hashkey) on line 110 can be executed anytime by the Listener mechanism (i.e., I believe the key can be removed concurrently by another thread), it is still theoretically possible that a caller may get a return reference to a Debugger instance that is not in the cache, which I don't think is a problem.

Moreover, I think it is largely harmless to have and use different client instances of Debugger for a user, The chances that an actual user (unless it is a bot) accesses a Debugger client concurrently at the same time will be very slim. So, I am not so certain if it is really worth achieving perfect synchronization by paying some synchronization cost.

chanseokoh:
Now it's using an annotation for the concurrency warning.

Also, I integrated another commit into this PR to address two instances of NPE.

patflynn:
LGTM

PR #499, opened 2016-03-14
Prevent NullPointerException

chanseokoh:
FindBugs issues #241. Address some (not all) of possible NullPointerExceptions. All the cases in this PR can produce NPE according to method contracts (i.e., @Nullable). It also seems that they can actually return null from their contexts.

After this PR, all other remaining 38 instances of the possible NPE warning can be suppressed (although there isn't an optimal way to suppress them).

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/validation/ApiNamespaceInspection.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/validation/ApiNamespaceInspection.java
@@ -75,7 +75,7 @@ public class ApiNamespaceInspection extends EndpointInspectionBase{
           return;
         }
 
-        if(!annotation.getQualifiedName().equals(GctConstants.APP_ENGINE_ANNOTATION_API_NAMESPACE)) {
+        if(!GctConstants.APP_ENGINE_ANNOTATION_API_NAMESPACE.equals(annotation.getQualifiedName())) {
           return;
         }
 
@@ -155,8 +155,7 @@ public class ApiNamespaceInspection extends EndpointInspectionBase{
       }
 
       PsiAnnotation annotation = (PsiAnnotation)psiElement;
-
-      if(!annotation.getQualifiedName().equals(GctConstants.APP_ENGINE_ANNOTATION_API_NAMESPACE)) {
+      if(!GctConstants.APP_ENGINE_ANNOTATION_API_NAMESPACE.equals(annotation.getQualifiedName())) {
         return;
       }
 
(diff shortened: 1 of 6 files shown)


google-cloud-tools-plugin/src/com/google/gct/idea/appengine/validation/ApiNamespaceInspection.java
```diff
@@ -75,7 +75,12 @@ public void visitAnnotation(PsiAnnotation annotation) {
           return;
         }
 
-        if(!annotation.getQualifiedName().equals(GctConstants.APP_ENGINE_ANNOTATION_API_NAMESPACE)) {
+        String annotationQualifiedName = annotation.getQualifiedName();
+        if (annotationQualifiedName == null) {
```
elharo: Optional: if you switch the order of the comparison so the constant comes first, then you don't need the null check. I.e. 
 if (!GctConstants.APP_ENGINE_ANNOTATION_API_NAMESPACE.equals(annotation.getQualifiedName()))

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/validation/FullJavaNameInspection.java
```diff
@@ -97,15 +97,23 @@ private void validateBackendMethodNameUnique(PsiMethod psiMethod, Map<String, Ps
           return;
         }
 
-        String javaName = psiMethod.getContainingClass().getQualifiedName() + "." + psiMethod.getName();
+        PsiClass containingClass = psiMethod.getContainingClass();
```
elharo: Are you sure this can be null? Logically it seems that shouldn't be possible.

elharo:
lgtm

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/validation/FullJavaNameInspection.java
chanseokoh: Actually I am not sure about this one and was hoping to get feedback. If it seems logically impossible, then I'll leave it as-is.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/validation/ApiNamespaceInspection.java
chanseokoh: I think you suggestion is by far a better approach. I simply switched the order of comparison for all the cases.

patflynn:
LGTM

PR #500, opened 2016-03-14
interim login service refactor commit

patflynn:
As part of my mvm work, I need to modify the account plugin to generate credential files. To improve testability and structure I implemented a todo to refactor this api out as a IJ application service. 

Note: There are plenty of things that are still not how I'd like them to be, but I'm intentionally restricting the change to just making the API a service.


common-test-lib/src/main/java/com/google/gct/idea/testing/TestUtils.java
```diff
@@ -119,6 +120,16 @@ private static Disposable getParentDisposableForCleanup() {
   }
 
   @NotNull
+  public static <T> T installMockService(@NotNull Class<T> serviceInterface) {
```
etanshaul: I'm seeing test failures for `CloudDebugProcessStateTest` and `CloudDebugProcessTest` due to a duplicated key error from the Pico Container:

Looks like the mock is being registered during @setUp. I suspect we need either unregister it first, or check for the component before registering it with the container.

etanshaul: Ignore the last comment - I must have checked out the branch in an earlier state - all tests pass.

etanshaul:
lgtm

elharo:
lgtm

PR #501, opened 2016-03-14
Address Findbugs and PMD warnings

chanseokoh:
#241 (FindBugs and PMD)

- FindBugs: use StringBuilder in a loop for string concat (2 instances)
- PMD: mostly about combining nested if-statements and ignoring empty catch-blocks, plus some minor things

### DIFF
--- a/common-test-lib/src/main/java/com/google/gct/idea/testing/TestUtils.java
+++ b/common-test-lib/src/main/java/com/google/gct/idea/testing/TestUtils.java
@@ -152,7 +152,7 @@ public class TestUtils {
       if (out != null) {
         try {
           out.close();
-        } catch (IOException e) {
+        } catch (IOException e) { // NOPMD
           // ignore
         }
       }
--- a/google-account-plugin/src/com/google/gct/stats/GoogleUsageTracker.java
+++ b/google-account-plugin/src/com/google/gct/stats/GoogleUsageTracker.java
@@ -72,7 +72,7 @@ public class GoogleUsageTracker implements UsageTracker {
                            @Nullable String eventLabel,
                            @Nullable Integer eventValue) {
         if (!ApplicationManager.getApplication().isUnitTestMode()) {
-            if (UsageTrackerManager.getInstance().isTrackingEnabled()) {
+            if (UsageTrackerManager.getInstance().isTrackingEnabled()) { // NOPMD
                 ArrayList postData = Lists.newArrayList(analyticsBaseData);
                 postData.add(new BasicNameValuePair("tid", analyticsId));
                 postData.add(new BasicNameValuePair(INTELLIJ_EDITION, PlatformUtils.isCommunityEdition() ? "community" : "ultimate"));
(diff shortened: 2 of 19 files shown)


common-test-lib/src/main/java/com/google/gct/idea/testing/TestUtils.java
```diff
@@ -153,7 +153,7 @@ public static void assertIsSerializable(@NotNull Serializable object) {
         try {
           out.close();
         } catch (IOException e) {
-          // ignore
+          fail("Cannot close output stream: " + e.getMessage());
```
elharo: I don't think we do want to fail here.

patflynn:
I'd like to leave the breaking up of conditionals into nests up to the discretion of the author as much as possible. Sometimes it's more readable to nest.. In these particular cases I'm not strongly opionionated.

patflynn:
LGTM once elharo's comment is addressed.

google-account-plugin/src/com/google/gct/stats/GoogleUsageTracker.java
```diff
@@ -71,27 +71,26 @@ public void trackEvent(@NotNull String eventCategory,
...
-            if (UsageTrackerManager.getInstance().isTrackingEnabled()) {
-                ArrayList postData = Lists.newArrayList(analyticsBaseData);
-                postData.add(new BasicNameValuePair("tid", analyticsId));
-                postData.add(new BasicNameValuePair(INTELLIJ_EDITION, PlatformUtils.isCommunityEdition() ? "community" : "ultimate"));
-                postData.add(new BasicNameValuePair("ec", eventCategory));
-                postData.add(new BasicNameValuePair("ea", eventAction));
-                if (!Strings.isNullOrEmpty(eventLabel)) {
-                    postData.add(new BasicNameValuePair("el", eventLabel));
-                }
+        if ((!ApplicationManager.getApplication().isUnitTestMode()) &&
```
elharo: I'm not sure this one pulls its weight. It seems marginally harder to read. Maybe of you removed the extra parentheses?

patflynn:
(not that I have an opinion)

google-cloud-tools-plugin/src/com/google/gct/idea/ui/CustomizableComboBox.java
```diff
@@ -190,10 +190,9 @@ private static boolean isUsingDarculaUIFlavor() {
   @Override
   public void paint(Graphics g) {
     super.paint(g);
-    if (textField.isFocusOwner() || (getPopup() != null && getPopup().isPopupVisible())) {
-      if (isUsingDarculaUIFlavor()) {
-        DarculaUIUtil.paintFocusRing(g, 3, 3, getWidth() - 4, getHeight() - 4);
-      }
+    if ((textField.isFocusOwner() || (getPopup() != null && getPopup().isPopupVisible())) &&
```
elharo: Ditto. IMHO this was easier to read before.

chanseokoh:
I agree about the nested if's. I'll revert most of the changes back (this would introduce // NOPMD comments). I'll combine only if they are very short.

common-test-lib/src/main/java/com/google/gct/idea/testing/TestUtils.java
chanseokoh: Reverted back.

chanseokoh:
Reverted most of the combined if's. I'll merge it soon.

PR #502, opened 2016-03-15
Use custom PMD ruleset / Follow PMD suggestions

chanseokoh:
#241.

1. First add 'custom-pmd-rule.xml' to use a custom PMD rules instead of the Gradle PMD plugin default. This was to exclude the UselessParentheses rule which somewhat conflicts with the Google style guide where parentheses are often encouraged.

2. Add one more redundant null check exclusion in the FindBugs configuration.

3. The rest are PMD suggestions.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -86,6 +86,9 @@ subprojects {
     pmd {
         ignoreFailures = true
         sourceSets = [sourceSets.main]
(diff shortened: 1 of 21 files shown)


patflynn:
Changes look good.

patflynn:
LGTM

Issue #410, opened 2016-02-04
Memory leak when running test suite via command line


etanshaul:
When running ./gradlew clean test, a memory leak is reported. Reproducible on linux, but not on mac.

The source test of this error appears to be CloudDebugGlobalPollerTest. See exception / stack trace below:

java.lang.RuntimeException: Memory leak detected: com.google.gct.idea.testing.TestUtils$PluginMockApplication@63828d7a of class class com.google.gct.idea.testing.TestUtils$PluginMockApplication
See the cause for the corresponding Disposer.register() stacktrace:
at com.intellij.openapi.util.objectTree.ObjectTree.assertIsEmpty(ObjectTree.java:209)
	at com.intellij.openapi.util.Disposer.assertIsEmpty(Disposer.java:124)
	at com.intellij.openapi.util.Disposer.assertIsEmpty(Disposer.java:120)
	at com.intellij.openapi.application.impl.ApplicationImpl.disposeSelf(ApplicationImpl.java:308)
	at com.intellij.openapi.application.impl.ApplicationImpl.access$400(ApplicationImpl.java:91)
	at com.intellij.openapi.application.impl.ApplicationImpl$5$1.run(ApplicationImpl.java:268)
	at com.intellij.openapi.util.ShutDownTracker$2.run(ShutDownTracker.java:172)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:756)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:76)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:726)
	at com.intellij.ide.IdeEventQueue.defaultDispatchEvent(IdeEventQueue.java:866)
	at com.intellij.ide.IdeEventQueue._dispatchEvent(IdeEventQueue.java:654)
	at com.intellij.ide.IdeEventQueue.dispatchEvent(IdeEventQueue.java:381)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:201)
[...]

PR #503, opened 2016-03-15
fix memory leak error

meltsufin:
@patflynn As discussed in #410...

When a BasePluginTestCase executes followed by a PlatformTestCase, a memory leak error is reported.

The problem goes away when TestRunnerUtil.replaceIdeEventQueueSafely() is called in the setup() method of BasePluginTestCase. This is also something that IntelliJ's UsefulTestCase does in its
setUp() method.


patflynn:
awesome. clean build logs. LGTM.

loosebazooka:
you might need to do a rebase

PR #504, opened 2016-03-15
Address NullPointerException FindBugs warning instances

chanseokoh:
#241.

Here is my take on addressing NPE warnings that FindBugs reports.

1. I learned that I can use 'assert' to mute a warning case. Some cases were muted simply this way.
2. Some warnings can be avoided by simple syntactic restructuring of code.
3. For a couple of cases, I really inserted a null check.

For 1 and 2, it is my own logical conclusion that those cases don't need an explicit null check. I can't say I am 100% confident though. I also believe that the simple syntactic substitution doesn't change the semantics (i.e., I assume function calls are 'const').

After this PR, we would have exactly one FindBug warning instance and one PMD warning instance, which we need to look into together.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/util/EndpointUtilities.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/util/EndpointUtilities.java
@@ -75,6 +75,7 @@ public class EndpointUtilities {
    */
   public static boolean isApiMethod(@NonNls PsiMethod psiMethod) {
     PsiModifierList psiModifierList =  psiMethod.getModifierList();
+    assert psiModifierList != null;
     if(psiModifierList.hasModifierProperty(PsiModifier.PUBLIC) &&
        !psiModifierList.hasModifierProperty(PsiModifier.STATIC)) {
       return true;
@@ -99,6 +100,7 @@ public class EndpointUtilities {
     }
 
     PsiModifierList modifierList = method.getModifierList();
+    assert modifierList != null;
     if(modifierList.hasModifierProperty(PsiModifier.PUBLIC)) {
       return true;
     }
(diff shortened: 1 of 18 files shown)


chanseokoh:
Note: some method calls are @NotNull, so those are easy cases to confirm that they don't return null. However, some other methods are @Nullable.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/util/EndpointUtilities.java
```diff
@@ -75,6 +75,7 @@ public static boolean isEndpointClass(PsiElement psiElement) {
    */
   public static boolean isApiMethod(@NonNls PsiMethod psiMethod) {
     PsiModifierList psiModifierList =  psiMethod.getModifierList();
+    assert psiModifierList != null;
```
elharo: Google policy is not to use the assert keyword

chanseokoh:
Disappointing to hear that asserts are not allowed. I guess I have to fall back to add bunch of elements into the FindBugs XML config file to mute many warnings.

patflynn:
assertions are allowed
https://engdoc.corp.google.com/eng/doc/devguide/java/practices/assertions.md?cl=head

But they should really be seen as simply documentation. We have a habbit of
using asserts in our code today because it gets rid of some IJ warnings
about null.

I haven't looked at this PR yet, but will shortly.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/util/EndpointUtilities.java
patflynn: for this case Preconditions.chechNotNull is more appropriate.  asserts should only be used for things that can't possibly be untrue. In this case the client could conceivably pass a psiMethod that returns a null psiModifierList so an NPE is more appropriate which is what the Preconditions check will throw.

chanseokoh:
Hmm... using those Preconditions and Verify runtime checks in these cases makes sense. Unfortunately, muting FindBugs warnings is an orthogonal matter, and those Google checks cannot mute the warnings.

In this PR, I think I would focus on knocking out the FindBugs warnings. We are back to ground zero.

patflynn:
That's odd. If it works on bytecode I would expect it to see the null
checks in the preconditions.

chanseokoh:
My guess is that it's because those checks are not macro expansion but a function call. The null check will be done outside of the current context (i.e., inside checkNotNull), and FindBugs simply cannot assume that an outside function call has done a null check for a caller.

chanseokoh:
Closing this per our discussion. #508 opened.

PR #505, opened 2016-03-16
adding support for user selection of jar or war files for mvm deployment

### LINKED ISSUE #449: Managed VM deployment should have an easier path to selecting a deployable artifact
IntelliJ's primary binary deployment model elements are 'Artifacts'.  Out of the box IJ CE comes with Jar artifacts and UE has Jar and Wars.  These are native artifacts and not generated automatically from a Maven or Gradle import.  This means the user has to manually create a corresponding IJ Artifact for whatever binaries their project is producing to be able to deploy to MVM. 

Solutions to this problem are:
1) In the MVM deployment configuration allow the user to select an arbitrary war or jar from the project files without needing an 'Artifact' in the project model.
2) Add new Jar and War artifact types available through our plugin that somehow detect the binary artifacts of a Maven or Gradle build and add them automatically to the project model.

etanshaul:
fixes #449 

A possible approach to allowing the user to manually select jar or war files for mvm deployment. Gets around forcing the user to have preconfigured IJ Artifacts.

The other approach discussed was to use IJ Artifact system to create a custom Artifact pointing to a file on the filesystem for deployment. 

todo and better documentation pending

Updated UI and placeholder deployment source selected:
screen shot 2016-03-15 at 11 16 03 pm

A custom chosen jar file:
screen shot 2016-03-15 at 11 19 56 pm

Deployment source dropdown:
screen shot 2016-03-15 at 11 20 12 pm


patflynn:
Thanks for sending this out. The screenshots answer all my questions. Super useful.

My comments are:
- 'Managed VM Deployment' is a confusing deployment name since whatever you select will be a managed vm deployment. 
- If an artifact is selected, is the 'use custom deployment file' still visible?
- 'use custom deployment file' should be something like 'Use alternate deployment archive', and filter by jars and wars.

elharo:
looks good overall. I suggest trying to align the left side of the name input box with the dropdowns below it.

etanshaul:
@patflynn Addressing your points below:

**"'Managed VM Deployment' is a confusing deployment name since whatever you select will be a managed vm deployment."**
Yes this is true, I forgot to include one screenshot that shows why I chose this (I originally had something like "Select a deployment source"):

screen shot 2016-03-16 at 11 34 03 am

Notice that the name that is used for the dropdown item also shows up under the Application Servers panel if the user keeps this placeholder option selected together with a manually browsed to jar/war file. The reason I made this placeholder in the first place is that our extra settings are only visible if there is a deployment source in the dropdown. Therefore I had to create a "fake" deployment source (I updated the icon to make it look more informational) to act as a placeholder.

**"If an artifact is selected, is the 'use custom deployment file' still visible?"**
Yes. One thing I was going to try and do is disable the text area in this case - I prefer this to hiding and showing the field which messes with vertical flow of the dialog.

**'use custom deployment file' should be something like 'Use alternate deployment archive', and filter by jars and wars.**
Yes, that is better.

etanshaul:
One other caveat I forgot to mention:

The elements in the deployment dropdown menu are sorted (by the IJ framework code) alphabetically - its a `SortedComboBoxModel` using a `String.CASE_INSENSITIVE_ORDER` comparator. 

What this means to us is that depending on the label we choose for that placeholder I discussed, it could appear _in the middle_ of the dropdown list. This is not ideal, but it is a consequence of the fact that we are "forcing" our workflow on the IJ ui.

etanshaul:
@elharo those UI elements you referenced are part of the IJ framework not exposed via an extension point. The custom part of that UI is the manual deployment file selection as well as the custom yaml docker file selection.

etanshaul:
I've pushed a couple updates including wording improvements. Before I finalize all the polish on the PR I wanted to make sure we were on board with the approach? My opinion is that it is a lot better than the other artifact based approach I showed (using an IJ artifact to act as a pointer to a filesystem archive).

patflynn:
I think I'm on board. My hesitation is around the way the artifact is displayed in that application servers screenshot you included.  I need to spend a bit of time playing with the capabilties there to make sure we don't break that model.

etanshaul:
It just hit me that there is a way around what is displayed in that Application Server window you mentioned. Its not pretty - involved making a field mutable that wasn't intended to be, as well as instanceof and a cast - but the UI looks a lot better with it:

image

When the user manually browses to an archive, I am setting the displayed name to be the file name.

patflynn:
sweet.

etanshaul:
Minor update:

Mutating the name of the deployment source like I was doing cause the placeholder dropdown element (see previous screenshot) to immediately show the new name - this is not what we want: it doesn't make sense for it to display the name of the browsed to archive.  As a compromise I settled with a hybrid display:

image

This way the name still makes sense when viewed from the context of the dropdown.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmDeploymentConfiguration.java
```diff
@@ -48,8 +50,20 @@ public String toString() {
 
   private String dockerFilePath;
   private String appYamlPath;
+  private boolean manualDeployment;
```
patflynn: how about userSpecifiedArtifactPath?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmDeploymentRunConfigurationEditor.java
```diff
@@ -58,11 +61,20 @@
   private TextFieldWithBrowseButton dockerFilePathField;
   private JButton generateAppYamlButton;
   private JButton generateDockerfileButton;
-
+  private JCheckBox manualDeploymentFileCheckbox;
```
patflynn: I prefer userSpecifiedArtifactPath. LMKWYT

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmCloudType.java
```diff
@@ -148,6 +159,36 @@ public ManagedVmDeploymentConfiguration createDefaultConfiguration(
           new CloudSdkAppEngineHelper(new File(server.getConfiguration().getCloudSdkExecutablePath()),
               server.getConfiguration().getCloudProjectName()));
     }
+
+    protected static class PlaceholderDeploymentSource extends ModuleDeploymentSourceImpl {
```
patflynn: if we go the UserSpecified.. naming route you could also rename this class to UserSpecfiedPathDeploymentSource or some shit like that.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmDeploymentConfiguration.java
etanshaul: I was actually just trying to come up with better naming myself - yea thats much better. Also, I've iterated on the UI again improving some things. Will update screenshots.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmDeploymentRunConfigurationEditor.java
```diff
@@ -76,6 +88,12 @@ public void actionPerformed(ActionEvent e) {
         }
       }
     });
+    manualDeploymentFileSelector.addBrowseFolderListener(
```
patflynn: is there a way to filter the list of files in the browser by some predicate?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmCloudType.java
etanshaul: In my current iteration (still not pushed) I had updated this to FileSystemDeploymentSource. 

But I like your wording better.

etanshaul:
New screenshots reflecting updated UX:

_Deploy window with a preconfigured IJ Artifact Selected:_
screen shot 2016-03-20 at 5 04 24 pm

_Deploy window with option to browse to user selected archive:_
screen shot 2016-03-20 at 5 04 34 pm

_Closeup of deployment options:_
screen shot 2016-03-20 at 5 04 41 pm

_Application Servers window after selecting to deploy a user chosen archive:_
screen shot 2016-03-20 at 5 05 06 pm

patflynn:
Amazing.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmDeploymentRunConfigurationEditor.java
```diff
@@ -58,11 +64,18 @@
   private TextFieldWithBrowseButton dockerFilePathField;
   private JButton generateAppYamlButton;
   private JButton generateDockerfileButton;
-
+  private JPanel userSpecifiedArtifactPanel;
+  private TextFieldWithBrowseButton userSpecifiedArtifactFileSelector;
+  DeploymentSource deploymentSource;
```
patflynn: why isn't this private?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmDeploymentRunConfigurationEditor.java
```diff
@@ -114,9 +140,56 @@ protected void resetEditorFrom(ManagedVmDeploymentConfiguration configuration) {
...
     configuration.setDockerFilePath(dockerFilePathField.getText());
     configuration.setAppYamlPath(appYamlPathField.getText());
     configuration.setConfigType(getConfigType());
+
+    setDeploymentSourceName(configuration.getUserSpecifiedArtifactPath());
+    toggleJarWarSelector();
+    validateConfiguration();
+  }
+
+  private void toggleJarWarSelector() {
```
patflynn: toggle to me means flipping the bit, but this doesn't necessarily do that.  Maybe just 'updateJarWarSelector()'?

etanshaul: That works. To add another option: `toggleJarWarSelector(boolean visible)`. "toggle" seems to work a little better this way.

etanshaul: I like your suggestion better actually. Updated the branch with both of your comments.

patflynn:
lgtm

PR #507, opened 2016-03-16
update missing copyrights

patflynn:


patflynn:
crap, it should be 2016.

patflynn:
fixed.

etanshaul:
`findbugs-excludefilter.xml `has its copyright after the opening tag. Was this on purpose?

patflynn:
you mean after the  <?xml version="1.0" encoding="UTF-8"?>? I think that's
standard for XML.

etanshaul:
Yea you're right, the type declaration has to come first for it to be well formed.

lgtm

PR #508, opened 2016-03-16
Disable FindBugs NPE checking / Fix possible NPE

chanseokoh:
#241.

1. Ignore and disable FindBugs' possible NPE checking.

2. Add // NOPMD to suppress empty if-statement.

3. Fix one point where NPE seems possible.


patflynn:
LGTM

Issue #240, opened 2015-11-23
Code analysis should send results to team for each PR.


patflynn:
Currently findbugs and PMD are just storing the results on local travis and being ignored.  They should store the results somewhere accessible and possibly even break the build. Ideally we could get a nice setup where the PR would list the issues found during code analysis.  A presubmit would also be nice so that people can fix these issues prior to creating a PR.

etanshaul:
obsolete. has been implemented in kokoro / findbugs

PR #509, opened 2016-03-17
Fail bulid on Travis when FindBugs/PMD complains

chanseokoh:
#240

Build will fail on Travis if there are any FindBugs/PMD violations. Local builds will not be affected.


patflynn:
LGTM

PR #513, opened 2016-03-21
disable findbugs missing equals override

### LINKED ISSUE #512: Disable pmd force override equals, hashcode
Currently PMD forces us to override equals and hashcode even if we don't want to modify the behavior of the superclass. It suggests to have a stub override that just calls super. (Whats worse is that if you follow its suggestion it then complains that you are overriding a method that does nothing but call super). We should disable this PMD check altogether.

etanshaul:
fixes #512

### DIFF
--- a/findbugs-excludefilter.xml
+++ b/findbugs-excludefilter.xml
@@ -21,9 +21,13 @@
       1. Inefficient 'new String("plain string")' constructor
       2. Redundant null checking of known non-null value
       3. Possible null pointer deref
+      4. Missing equals override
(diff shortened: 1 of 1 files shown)


etanshaul:
@patflynn @chanseokoh ptal

chanseokoh:
I can tell at least it is the right place to disable the warning globally.

etanshaul:
cool, so lgtmty? (looks good to me, to you?)

chanseokoh:
LGTM

patflynn:
LGTM

Issue #496, opened 2016-03-10
We need to update to use the latest GCP resource API


patflynn:
https://cloud.google.com/resource-manager/

The current client we use for project lookup is unsupported. We need to move to the API linked above.

meltsufin:
@patflynn Which version of the API should we be using? The one currently posted for download on the official site is v1beta1-rev12-1.21.0. I'm trying to use it, but get the error below, even after enabling it for my project. The web site does mention that "Access to this feature is currently invite-only." So, I don't know if it needs to be enabled on some higher level.

```
{
  "code" : 403,
  "errors" : [ {
    "domain" : "global",
    "message" : "Project has not enabled the API. Please use Google Developers Console to activate the 'cloudresourcemanager' API for your project.",
    "reason" : "forbidden"
  } ],
  "message" : "Project has not enabled the API. Please use Google Developers Console to activate the 'cloudresourcemanager' API for your project.",
  "status" : "PERMISSION_DENIED"
}
```

Issue #463, opened 2016-02-23
Pull Google API dependencies from Maven


etanshaul:
Related to #223 

Currently these jars are manually added to the project. We should probably pull them maven; the clouddebugger API jar exists in maven central (1 minor version ahead of ours) - those that aren't in the repository should be added.

http://search.maven.org/#search%7Cga%7C1%7Cclouddebugger

meltsufin:
Fixed by #534.

Issue #223, opened 2015-11-12
Make sure we're using the latest libraries for Google APIs we consume


patflynn:

meltsufin:
Fixed by #534

PR #514, opened 2016-03-21
Replacing the Developer Projects API with the latest Cloud Resource Manager API

meltsufin:
#496 

Developer Projects API is deprecated now, and this update upgrades the project to the latest [Cloud Resource Manager API](https://cloud.google.com/resource-manager/)

We use this API to fetch the list of projects.


google-cloud-tools-plugin/build.gradle
```diff
@@ -38,7 +38,7 @@ dependencies {
     compile(project(':common-lib'))
 
     compile files('lib/google-api-services-clouddebugger-v2-rev4-1.20.0.jar')
-    compile files('lib/google-api-services-developerprojects-v1-rev20140815212553-1.19.0.jar')
+    compile files('lib/google-api-services-cloudresourcemanager-v1beta1-rev12-1.21.0.jar')
```
elharo: any chance they'll put this in maven central?

elharo:
LGTM, squash and commit

google-cloud-tools-plugin/build.gradle
meltsufin: I actually now see they have version v1-rev2-1.21.0 in Maven Central. Should I try using that?

patflynn:
For now the workflow is to squash branch commits to just one commit before
merging.

loosebazooka:
we gotta have a version of this CL in the android studio code so they can continue to work.

google-cloud-tools-plugin/build.gradle
elharo: yes, we'd rather not bundle the jar if we can avoid it.

patflynn: +1 please use maven central artifact.

etanshaul:
just noting that I think this is a dup of #463 (which hasn't been worked yet) so feel free to close that too once this is merged.

google-cloud-tools-plugin/build.gradle
meltsufin: I tried quickly, and the version in Maven Central didn't work out of the box. Need to look more into it...

meltsufin:
@etanshaul I think we'll handle the switch to Maven Central for cloudebugger and cloudresourcemanager separately as proposed by #463, #223, and #516.

PR #517, opened 2016-03-21
Show project name in mvm deploy dialog

### LINKED ISSUE #476: Default Managed VM Cloud name to include project name
When creating a Managed VM cloud config it should include the selected project's name in the title by default.  This will help the user identify which project he's targeting during deployment.

etanshaul:
fixes #476

### DIFF
--- a/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
+++ b/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
@@ -31,6 +31,7 @@ appengine.managedvm.dockerfile.location.label=Dockerfile path\:
 appengine.managedvm.config.destination.folder.label=Choose destination folder\:
 appengine.managedvm.config.choose.destination.folder.window.title=Choose destination folder
 appengine.managedvm.config.destination.chooser.title=Choose Generated Configuration Destination Folder
+appengine.managedvm.config.project.panel.title=Project: {0}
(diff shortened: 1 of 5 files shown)


patflynn:
Is there no way to change the actual cloud name?  The user can edit it, so I was hoping we could set it to default to the project name.

etanshaul:
by cloud name, you mean the name of the element in the "Server:" dropdown? In the screenshot below its "Cloud 1", so you're asking if this could be something like "Cloud 1 - MyProject"?

image

patflynn:
yeah.

etanshaul:
I checked on both ends for any ability to customize the name: at the point of creating the cloud server, and at the point of displaying in the deployment window. It doesn't look like there's a way to do this. I can discuss details if interested - basically run through the web of framework code.

etanshaul:
Updated look:

image

patflynn:
I like it, although I'm not loving the hyphen. It looks kind of weird in
the context of the rest of the border being blue.

etanshaul:
Looks ok to me with a colon. Here's one more look at it (better margins too to make it less cramped):

image

patflynn:
What happens if I click to edit the cloud config from this UI and change
the project?

etanshaul:
fixed. used the google commons EventBus to publish/subscribe to project name changes in the configuration so that we can update the UI appropriately.

patflynn:
I'm not sure that the event bus is the right ool for this. But we can look into patterns for handling this later. LGTM

PR #518, opened 2016-03-22
passing plugin creds to gcloud for mvm deploy

### LINKED ISSUE #454: Managed VM deployment needs to pass our plugin's auth credentials for deployment
We are currently piggy backing on the gcloud CLIs auth context to perform deployments because gcloud does not yet allow a caller to pass a credential for a given command. 

Cloud SDK team is updating gcloud to support this.  Once that is done we should start passing our plugin's user auth context to the command for deployments.

patflynn:
closes #454 

You need to be on the gcloud dogfood for this to work due to a bug which was just recently fixed.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/DoManagedVmDeployment.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/DoManagedVmDeployment.java
@@ -16,6 +16,15 @@
 
 package com.google.gct.idea.appengine.cloud;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.io.Files;
+import com.google.gct.idea.util.GctBundle;
+import com.google.gct.login.CredentialedUser;
+import com.google.gct.login.Services;
+import com.google.gdt.eclipse.login.common.GoogleLoginState;
+import com.google.gson.Gson;
+
 import com.intellij.execution.ExecutionException;
 import com.intellij.execution.configurations.GeneralCommandLine;
 import com.intellij.execution.configurations.GeneralCommandLine.ParentEnvironmentType;
@@ -23,6 +32,7 @@ import com.intellij.execution.process.OSProcessHandler;
 import com.intellij.execution.process.ProcessAdapter;
 import com.intellij.execution.process.ProcessEvent;
 import com.intellij.execution.process.ProcessHandler;
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.remoteServer.runtime.deployment.DeploymentRuntime;
 import com.intellij.remoteServer.runtime.deployment.ServerRuntimeInstance.DeploymentOperationCallback;
@@ -32,11 +42,15 @@ import org.jetbrains.annotations.NotNull;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.charset.Charset;
+import java.util.Map;
 
 /**
  * Performs the deployment of ManagedVM based applications to GCP.
  */
 class DoManagedVmDeployment implements Runnable {
+
+  private static final Logger logger = Logger.getInstance(DoManagedVmDeployment.class);
 
   private LoggingHandler loggingHandler;
   private File deploymentArtifactPath;
@@ -63,6 +77,13 @@ class DoManagedVmDeployment implements Runnable {
   }
 
   public void run() {
+    final File appDefaultCredentialsPath = createApplicationDefaultCredentials();
+    if (appDefaultCredentialsPath == null) {
+      callback.errorOccurred(
+          GctBundle.message("appengine.deployment.credential.not.found",
+              appEngineHelper.getGoogleUsername()));
+      return;
+    }
     File stagingDirectory;
     try {
       stagingDirectory = FileUtil.createTempDirectory(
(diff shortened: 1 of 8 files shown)


google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/DoManagedVmDeployment.java
```diff
@@ -73,7 +93,10 @@ public void run() {
           loggingHandler,
           "Created temporary staging directory: " + stagingDirectory.getAbsolutePath());
     } catch (IOException e) {
-      throw new RuntimeException(e);
+      logger.error(e);
```
etanshaul: Since your are logging the error here, for consistency should we also log it in the other error handling blocks?

patflynn: yeah, and I just noticed that I missed the execution exception. My head is not on straight this morning.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/DoManagedVmDeployment.java
```diff
@@ -117,13 +142,59 @@ public void undeploy(@NotNull UndeploymentTaskCallback callback) {
...
       }
     });
     processHandler.startNotify();
   }
 
+  private static final String CLIENT_ID_LABEL = "client_id";
+  private static final String CLIENT_SECRET_LABEL = "client_secret";
+  private static final String REFRESH_TOKEN_LABEL = "refresh_token";
+  private static final String GCLOUD_USER_TYPE_LABEL = "type";
+  private static final String GCLOUD_USER_TYPE = "authorized_user";
```
etanshaul: Minor comment: I couldn't find a recommendation on placement of member variables in the style guide, but I tend to always put them at the top of classes.

patflynn: I typically do that for class fields as well as constants that figure prominently in a class or are exported. In this case because they are so loca, I'm tempted to leave here or possible even in-line.. What do you think?

etanshaul: Yea makes sense. I'm fine with leaving it above the method (I like that better than inline).

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/DoManagedVmDeployment.java
patflynn: @etanshaul fixed.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/DoManagedVmDeployment.java
```diff
@@ -87,16 +111,18 @@ public void run() {
...
     Process process = null;
     try {
-      consoleLogLn(loggingHandler, "Executing: " + commandLine.getCommandLineString()
-      );
+      consoleLogLn(loggingHandler, "Executing: " + commandLine.getCommandLineString());
       process = commandLine.createProcess();
     } catch (ExecutionException e) {
-      throw new RuntimeException(e);
+      logger.error(e);
+      callback.errorOccurred("appengine.deployment.error");
```
etanshaul: need to use message bundle

etanshaul:
copyFile() in DoManagedVmDeployment throws a RuntimeException. We should also do the same logging / callback.errorOccurred .. here too as this could be a common source of user error.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmCloudType.java
```diff
@@ -210,8 +211,7 @@ public ManagedVmServerConnector(
     public void connect(@NotNull ConnectionCallback<ManagedVmDeploymentConfiguration> callback) {
       if (CloudSdkUtil.isCloudSdkExecutable(configuration.getCloudSdkExecutablePath())) {
         callback.connected(new ManagedVmRuntimeInstance(configuration));
-      }
-      else {
+      } else {
         callback.errorOccurred("Invalid Cloud SDK directory path configured.");
```
etanshaul: maybe extract this to the message bundle too?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/DoManagedVmDeployment.java
```diff
@@ -63,6 +77,13 @@
   }
 
   public void run() {
+    final File appDefaultCredentialsPath = createApplicationDefaultCredentials();
+    if (appDefaultCredentialsPath == null) {
+      callback.errorOccurred(
+              GctBundle.message("appengine.deployment.credential.not.found",
```
etanshaul: need to pass parameter to message bundle here ({0} should get the username).

patflynn: there is a username passed in here.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmCloudType.java
patflynn: I'll file a bug to catch the other instances.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/DoManagedVmDeployment.java
etanshaul: oh yea weird - when I ran the deploy forcing this exception I saw {0} printed on the command line view. I'll check again.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmCloudType.java
```diff
@@ -157,8 +156,10 @@ public ManagedVmDeploymentConfiguration createDefaultConfiguration(
         @NotNull DeploymentSource source,
         @NotNull RemoteServer<ManagedVmServerConfiguration> server) {
       return new ManagedVmDeploymentRunConfigurationEditor(project, source,
-          new CloudSdkAppEngineHelper(new File(server.getConfiguration().getCloudSdkExecutablePath()),
-              server.getConfiguration().getCloudProjectName()));
+          new CloudSdkAppEngineHelper(
+              new File(server.getConfiguration().getCloudSdkExecutablePath()),
+              server.getConfiguration().getCloudProjectName(),
+              server.getConfiguration().getGoogleUserName()));
```
etanshaul: I think I found an edge case here: if the user already has a cloud configured but is not logged in any more for whatever reason, then this line throws an NPE when trying to get back into the deployment editor.

etanshaul: I'll keep an eye out for this, but I can't reproduce this now.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/DoManagedVmDeployment.java
```diff
@@ -73,30 +94,40 @@ public void run() {
...
+          copyFile(stagingDirectory, "target" + artifactType, deploymentArtifactPath);
+      stagedArtifactPath.setReadable(true /* readable */, false /* ownerOnly */);
 
-    copyFile(stagingDirectory, "app.yaml", this.appYamlPath);
-    copyFile(stagingDirectory, "Dockerfile", this.dockerFilePath);
+      copyFile(stagingDirectory, "app.yaml", this.appYamlPath);
+      copyFile(stagingDirectory, "Dockerfile", this.dockerFilePath);
+    } catch (IOException e) {
+      logger.error(e);
+      callback.errorOccurred(GctBundle.message("appengine.deployment.error.during.staging"));
```
etanshaul: missing return

etanshaul: looks like there are a few more missing returns in the class. I assume that after each callback.errorOccurred .. we want to return immediately.

patflynn: Yeah. Thanks for pointing them out. I'd changed these but not pushed them
yet.

patflynn:
@etanshaul PTAL

And thank you for spotting all my mistakes in this time of need.

etanshaul:
lgtm

PR #519, opened 2016-03-22
Use the mockito test runner so we can use @Mock annotations in our pl…

patflynn:
so we can use @Mock annotations in our plugin test cases


etanshaul:
lgtm

PR #520, opened 2016-03-22
I don't want nested ifs to trigger warnings

patflynn:
@etanshaul 
@chanseokoh 

I thought this was turned off. But I'm getting PMD errors on another PR that this change seems to fix.


etanshaul:
lgtm

chanseokoh:
LGTM. This wasn't turned off, and individual cases were suppressed with // NOPMD comments. I will make a separate PR to remove those comments as we have excluded this rule.

PR #521, opened 2016-03-22
Remove nestedif comments

chanseokoh:
Follow up on #520. Remove unnecessary // NOPMD comments for the nested-if warning as #520 turned off it.


patflynn:
LGTM

PR #532, opened 2016-03-23
changing listener mechanism and marking fields transient

### LINKED ISSUE #531: IJ is unable to instantiate ManagedVmServerConfiguration with new attributes
Due to the new attributes I added in #517 IJ is now unable to instantiate this config class causing a downstream null violation and an error when creating a new cloud.

I think it has something to do with the fact that this is a persisted class (part of IJ serialized state) but I need to investigate.

etanshaul:
fixes #531 

Also forces there to be only a single project name change listener (before multiple could have been registered). Also uses a more appropriate listener mechanism.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmServerConfiguration.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmServerConfiguration.java
@@ -16,10 +16,12 @@
 
 package com.google.gct.idea.appengine.cloud;
 
-import com.google.common.eventbus.EventBus;
-
 import com.intellij.remoteServer.configuration.ServerConfigurationBase;
 import com.intellij.util.xmlb.annotations.Attribute;
+import com.intellij.util.xmlb.annotations.Transient;
+
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyChangeListener;
 
 /**
  * Model for the IntelliJ application scoped 'Cloud' configurations.  This is a base configuration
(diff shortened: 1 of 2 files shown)


google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmServerConfiguration.java
```diff
@@ -33,7 +35,8 @@
   private String cloudProjectName;
   private String googleUserName;
 
-  private EventBus projectNameListener = new EventBus();
+  @Transient
```
patflynn: does the IJ serialization mechanism look at this?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmServerConfiguration.java
```diff
@@ -63,7 +67,18 @@ public void setGoogleUserName(String googleUserName) {
...
+  protected void setProjectNameListener(PropertyChangeListener listener) {
+    this.projectNameListener = listener;
+  }
+
+  protected void fireNameChangeEvent(String name) {
+    if(projectNameListener != null) {
+      projectNameListener.propertyChange(new PropertyChangeEvent(
+          this,
+          "cloudProjectName",
+          name, // We don't care about the old name
```
patflynn: we may as well obey the contract, it doesn't seem like a lot of trouble. Maybe I'm OCD.

patflynn:
Thank you for doing this. LGTM.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmServerConfiguration.java
etanshaul: Yes it does. Its actually documented in their plugin docs: http://www.jetbrains.org/intellij/sdk/docs/basics/persisting_state_of_components.html

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmServerConfiguration.java
etanshaul: updated. will merge when travis is done.

PR #533, opened 2016-03-23
updating user facing mvm wording to 'flexible environment'

### LINKED ISSUE #524: update appengine mvm naming to 'flex'

etanshaul:
fixes #524 

This was really the only instance I could find of the old wording in the UI.


patflynn:
LGTM.  Doesn't this get crazy long in the UI?

etanshaul:
Its a little long, but it doesn't get clipped when it matters. I'll post some screenshots so we can make sure.

etanshaul:
Will go ahead and merge. Here is the only place it clips (it was clipping with the old wording too, and I don't think its a big deal):

image

Issue #516, opened 2016-03-21
Use maven central to pull down cloud debugger dependency


elharo:
Instead of bundling the jar.
http://maven-repository.com/artifact/com.google.apis/google-api-services-clouddebugger

meltsufin:
Looks like this is related to #463

PR #534, opened 2016-03-24
Use Maven Central for cloudresourcemanager and clouddebugger APIs

meltsufin:
Fixes: #516, #463, #223 

@patflynn, @loosebazooka PTAL

Transitioning to use use Maven Central for cloudresourcemanager and clouddebugger APIs.

Also updating to use the latest clouddebugger API version which happened to change base package name to include "v2". This forced updates to import statements everywhere clouddebugger API is used.

Notice that we're continuing to bundle the repackaged google-api-java-client in the google-account-plugin module. This forces us to exclude this dependency when pulling cloudresourcemanager and clouddebugger jars.


patflynn:
have you tested out the cloud debugger to verify everything still works as intended?

meltsufin:
Yes, a did a basic test, and it worked.

google-cloud-tools-plugin/build.gradle
```diff
@@ -37,14 +37,19 @@ dependencies {
...
-    compile files('lib/google-api-services-cloudresourcemanager-v1beta1-rev12-1.21.0.jar')
+    compile 'com.google.apis:google-api-services-clouddebugger:v2-rev7-1.21.0'
+    compile 'com.google.apis:google-api-services-cloudresourcemanager:v1beta1-rev12-1.21.0'
     compile files('lib/google-api-services-source.jar')
 
     testCompile(project(':common-test-lib'))
     testRuntime files('../google-account-plugin/lib/google-gct-login-context-ij-pg.jar')
 }
 
+configurations {
```
loosebazooka: does it make sense to add this to our provided checks?

https://github.com/GoogleCloudPlatform/gcloud-intellij/blob/master/build.gradle#L65

meltsufin: Agree, even though the existing provided checks already catches it via the transient dependencies.

meltsufin:
Added google-api-client to provided checks.

patflynn:
LGTM

PR #536, opened 2016-03-24
mvm cost warning label

### LINKED ISSUE #452: Managed VM Cloud configs should clearly state the financial costs of deployment
Unlike regular App Engine deployments Managed VM deployments typically spin up multiple VMs which do not scale down to 0.  Additionally multiple deployments create multiple running versions which can quickly add up to significant costs for the user.  

It's critical that when the user creates a Managed VM Cloud deployment target that we have clear messaging in the UI what the costs are.

etanshaul:
fixes #452 

Seeking input on this wording:

image

### DIFF
--- a/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
+++ b/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
@@ -36,6 +36,7 @@ appengine.flex.config.user.specified.artifact.title=Select a JAR or WAR file
 appengine.flex.config.user.specified.artifact.error=Browse to a JAR or WAR file.
 appengine.flex.config.custom.deploy.title=App Engine Flexible Environment Configuration.
 appengine.config.deployment.source.error=Select a valid deployment source.
+appengine.flex.deployment.cost.warning=<html><font face="sans" size="-1"><i>There is no free quota for App Engine Flexible Environment deployments. Please visit <a href\="https://cloud.google.com/appengine/pricing">GCP Pricing</a> for pricing information.</i></font></html>
 appengine.flex.user.specified.deploymentsource.name=Filesystem JAR or WAR file
 appengine.flex.user.specified.deploymentsource.name.with.filename=Filesystem JAR or WAR file - {0}
 appengine.deployment.credential.not.found=We were unable to find a credential for your cloud project's user. Are you sure {0} is logged in?"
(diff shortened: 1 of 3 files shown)


chanseokoh:
I may be wrong, but is it really true that deploying a new version does not stop old versions? If I deploy an app, the indicators under the "traffic share" column in the App Engine page on the Developers Console change to "N/A" and instances seem to be gone.

etanshaul:
Check out the Compute Engine Instances page and you'll probably see them grow after each deploy.

patflynn:
I think that they're changing the default to stop the old version when you
deploy a new one, the real concern is that the user will expect 1-> 0
scaling so that they can deploy and leave the app running without incurring
any costs.  That's the impression we want to dispel.

etanshaul:
How about something along these lines:

Note: By deploying your application to GCP you will incur additional charges for any instances that are created. See the GCP Pricing page for more information.

patflynn:
how about:

*WARNING!!! The App Engine Flexible Environment does not include a freebie
quota and any live deployments will incur costs to you.  To avoid accruing
significant costs make sure to stop any unused running applications.
(Ideally we'd also include a link on how to stop an application)*

@arouzrokh can you check this wording?

meltsufin:
I just tried deploying twice, and it seems like they are in fact spinning down the old instances.

etanshaul:
Maybe make the font red and wrap it in a <blink> tag? (kidding about the tag, but maybe red is appropriate)

arouzrokh:
Guys apologies for late reply as github emails were being sent to spam for me for some reason. 

We can remove the warning altogether. We have extensive material that we present the user upon billing signup and this does NOT belong in the IDE. Such language should not appear in any tooling anywhere in GCP. 

Thanks,
Amir

patflynn:
So we describe the Flex environment and billing consequences of deployment
in the billing signup?

Another concern is for App Engine Standard users who have had billing
enabled for a long time, who try deploying to Flex but expect to only be
charged if they see traffic.  This is a new entry way into the Flex
platform that can't assume the user has read the billing documentation for
that environment (I think).

arouzrokh:
I think this is not the right forum for this discussion. We also don't have the right people here making the decisions on areas such as pricing and billing practices. But to clarify more the decision to remove such language, yes, users who deploy to Flex are, through documentation and pricing articles, clearly educated on the two models. We can remove this language everywhere, including for both GAE standard and Flex.

etanshaul:
@patflynn ptal.

image

Issue #535, opened 2016-03-24
Cloud SDK path selection should allow selecting root SDK dir


meltsufin:
Currently, when configuring the Google Cloud SDK for Appengine flexible environment, the user must select the bin subdirectory of the SDK installation path. This may be counter to the expectation of the users that selecting the SDK root directory should work.

![screenshot from 2016-03-24 11 53 55](https://cloud.githubusercontent.com/assets/17732492/14022647/1eabcd82-f1b7-11e5-95e5-3e7115a68ac9.png)

patflynn:
+1. This needs to be fixed.

PR #540, opened 2016-03-25
User must select Cloud SDK home path, not /bin

meltsufin:
Addresses #535.

@etanshaul @patflynn PTAL

Prior to this change, when configuring the Google Cloud SDK for App Engine flexible environment, the user must select the bin subdirectory of the SDK installation path. This is counter to the expectation of the users that selecting the SDK home directory should work.

### DIFF
--- a/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
+++ b/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
@@ -20,7 +20,7 @@ cloud.project.label=Project\:
 
 appengine.cloudsdk.location.label=Cloud SDK directory\:
 appengine.cloudsdk.location.browse.window.title=Browse to Cloud SDK directory
-appengine.cloudsdk.location.missing.message=Please select a Cloud SDK directory.
+appengine.cloudsdk.location.missing.message=Please select a Cloud SDK home directory.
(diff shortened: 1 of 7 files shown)


meltsufin:
Selecting /bin is now invalid:

screenshot from 2016-03-25 11 41 42

Selecting SDK home is valid:
screenshot from 2016-03-25 11 42 40

etanshaul:
Just to make it even clearer, what if we also update the label to say "Cloud SDK home directory"?

etanshaul:
otherwise, lgtm

meltsufin:
I was a little unsure of making the label even longer by adding "home", but, yes, it would be clearer.
Other two options I was thinking of:
1. **Cloud SDK home**
-directory is sort of implied
2. **Cloud SDK**
-more vague, but aesthetically pleasing :)

patflynn:
I'm wondering if we shouldn't be storing the home directory rather than the bin path to gcloud.  What if we start invoking other commands or pulling in cloud SDK bundled resources, do we want to resolve the home from the bin path each time we access some other part of the cloud SDK?

meltsufin:
I agree, I think it makes sense to just store the SDK home and figure out the path to the executable dynamically.

patflynn:
LGTM

PR #541, opened 2016-03-25
stop an app deployed to mvm

### LINKED ISSUE #453: The user should be able to stop a running Managed VM.
There is no way in the web console to stop all versions of a Managed VM project.  Because of the financial implications of having MVMs running we should have as part of our default deployment flow, a means for stopping the running application.  'gcloud preview app stop..' can be used for this purpose.

etanshaul:
Fixes #453 

May undergo more refactoring and needs some test coverage. Putting it out there so it can start being looked at.

Adds support to stop a MVM application that was just deployed. Hooks into the IJ undeploy UI. Works by querying gcloud for the list of deployed versions, selects the latest one, then executes a stop command.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineModuleListOutputParser.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineModuleListOutputParser.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright (C) 2016 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineModuleListOutputParser.java
```diff
@@ -0,0 +1,80 @@
...
+   * Given a line from the output of a module listing cli command, parse it into its individual
+   * components.
+   *
+   * Example line: "default  20160325t180009  1.0"
+   *
+   * @param line from module listing cli command
+   * @return @see {@link AppEngineModuleListItem}
+   * @throws RuntimeException
+   */
+  private AppEngineModuleListItem parseLineItem(String line) {
```
patflynn: I believe you can ask gcloud to output the result as json.

etanshaul: Cool, that makes it a lot cleaner. I'll update to the logic to use json instead.

etanshaul:
Migrating to #571 (this branch had a different approach and also diverged too far from master).

PR #546, opened 2016-03-28
enable mvm feature flag

### LINKED ISSUE #539: App Engine Flex deployment should be turned on using resource flags
For now let's just checkin the feature flag.  We can completely remove the flag and checks from the code once the release occurs and we're OK with it.

patflynn:
closes #539


patflynn:
@meltsufin can I get an LGTM?

meltsufin:
LGTM

PR #547, opened 2016-03-28
Externalizing user messages in account plugin

meltsufin:
Addresses Issue #397.

@patflynn PTAL

### DIFF
--- a/google-account-plugin/resources/messages/AccountBundle.properties
+++ b/google-account-plugin/resources/messages/AccountBundle.properties
@@ -1,0 +1,41 @@
+#
+# Copyright (C) 2016 The Android Open Source Project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+login.toolbar.button.sign.in.text="Sign in to Google..."
+
+login.panel.sign.in.button.text=Sign In
+login.panel.sign.out.button.text=Sign Out
+login.panel.add.account.button.text=Add Account
+login.panel.play.console.link.text=Open Google Developers Console
+login.panel.cloud.console.link.text=Open Play Developer Console
+login.panel.sing.in.body.html=<HTML> Sign in with your Google account to start adding <br> Cloud functionality to your applications. </HTML>
+login.panel.learn.more.link.text=Learn more
+
+# do not change this value because it's used as a key
+login.service.are.you.sure.key.text=Are you sure you want to sign out?
+login.service.are.you.sure.append.text=\ This will sign out all logged in users.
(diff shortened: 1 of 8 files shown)


google-account-plugin/resources/messages/AccountBundle.properties
```diff
@@ -0,0 +1,44 @@
...
+#
+
+login.toolbar.button.sign.in.text="Sign in to Google..."
+
+login.panel.sign.in.button.text=Sign In
+login.panel.sign.out.button.text=Sign Out
+login.panel.add.account.button.text=Add Account
+login.panel.play.console.link.text=Open Google Developers Console
+login.panel.cloud.console.link.text=Open Play Developer Console
+login.panel.play.console.link.url=https://play.google.com/apps/publish/#ProfilePlace
```
patflynn: is that #ProfilePlace getting comment stripped?

google-account-plugin/resources/messages/AccountBundle.properties
```diff
@@ -0,0 +1,44 @@
...
+login.panel.cloud.console.link.url=https://console.developers.google.com/
```
patflynn: generally I'm not crazy about having url links in message bundles. If we do eventually get translators, I don't want them to be able to change these.

meltsufin: I was thinking about the translators too and was wondering if they would potentially point to a different URLs. But if that's not a concern, I'll just remove those.

google-account-plugin/resources/messages/AccountBundle.properties
meltsufin: doesn't look like it, but I'm going to undo the inclusion of URLs in the bundle

google-account-plugin/src/com/google/gct/login/ui/UsersListCellRenderer.java
```diff
@@ -53,12 +54,15 @@
  * how each user item in the Google Login panel would be displayed.
  */
 public class UsersListCellRenderer extends JComponent implements ListCellRenderer {
-  private final static String CLOUD_LABEL_TEXT = "Open Google Developers Console";
-  private final static String PLAY_LABEL_TEXT = "Open Play Developer Console";
+  private final static String CLOUD_LABEL_TEXT = AccountMessageBundle.message(
```
patflynn: break after AccountMessageBundle

google-account-plugin/src/com/google/gct/login/util/AccountMessageBundle.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Bundle class to get account plugin messages from resources/messages
+ */
+public class AccountMessageBundle {
+  @NonNls
+  private static final String BUNDLE_NAME = "messages.AccountBundle";
+  private static Reference<ResourceBundle> bundleReference;
+
+  private static ResourceBundle getBundle() {
+    ResourceBundle bundle = com.intellij.reference.SoftReference.dereference(
```
patflynn: same here. Break at the 'higher' syntactic level.

patflynn: this should fit on one line

google-account-plugin/src/com/google/gct/login/ui/UsersListCellRenderer.java
meltsufin: I don't think there's a need for a break at all.

patflynn:
LGTM. Squash and commit.

PR #548, opened 2016-03-28
Restore presentable name of "userSpecifiedSource"

### LINKED ISSUE #526: strange label for an artifact being deployed using mvm
When deploying a filesystem jar/war using mvm the 'Application Server' UI shows the artifact up as 'userSpecfiedSource' instead of the deployment filename. 

screenshot from 2016-03-23 14 53 01

chanseokoh:
Fixes #526. This approach seems like an ideal solution. If you have existing Flex deployment configuration, you should delete it and create a new one for this to work.

There is one very strange compilation problem that I can't figure out why it's happening. Line 117 of 'ManagedVmDeployment.java'. All of a sudden, 'DeploymentConfigurator' requires a fully qualified name although other occurrences of the name are just fine. I just can't understand why. Tried JDKs with both 1.7, and 1.8, but to no avail.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineCloudType.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineCloudType.java
@@ -32,6 +32,7 @@ import com.intellij.openapi.progress.ProgressIndicator;
 import com.intellij.openapi.progress.ProgressManager;
 import com.intellij.openapi.progress.Task;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.packaging.artifacts.Artifact;
 import com.intellij.packaging.artifacts.ArtifactManager;
 import com.intellij.remoteServer.RemoteServerConfigurable;
@@ -39,7 +40,10 @@ import com.intellij.remoteServer.ServerType;
 import com.intellij.remoteServer.configuration.RemoteServer;
 import com.intellij.remoteServer.configuration.deployment.DeploymentConfigurator;
 import com.intellij.remoteServer.configuration.deployment.DeploymentSource;
+import com.intellij.remoteServer.configuration.deployment.DeploymentSourceType;
 import com.intellij.remoteServer.configuration.deployment.JavaDeploymentSourceUtil;
+import com.intellij.remoteServer.configuration.deployment.ModuleDeploymentSource;
+import com.intellij.remoteServer.impl.configuration.deployment.DeployToServerRunConfiguration;
 import com.intellij.remoteServer.impl.configuration.deployment.ModuleDeploymentSourceImpl;
 import com.intellij.remoteServer.runtime.ServerConnector;
 import com.intellij.remoteServer.runtime.ServerTaskExecutor;
@@ -47,6 +51,7 @@ import com.intellij.remoteServer.runtime.deployment.DeploymentLogManager;
 import com.intellij.remoteServer.runtime.deployment.DeploymentTask;
 import com.intellij.remoteServer.runtime.deployment.ServerRuntimeInstance;
 
+import org.jdom.Element;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -196,6 +201,58 @@ public class AppEngineCloudType extends ServerType<AppEngineServerConfiguration>
       public void setName(String name) {
         this.name = name;
       }
+
+      @NotNull
+      @Override
+      public DeploymentSourceType<?> getType() {
+        return DeploymentSourceType.EP_NAME.findExtension(
+            UserSpecifiedPathDeploymentSourceType.class);
+      }
+    }
+  }
+
+  public static class UserSpecifiedPathDeploymentSourceType extends
+      DeploymentSourceType<ModuleDeploymentSource> {
+    private static final String NAME_ATTRIBUTE = "name";
+    private static final String SOURCE_TYPE_ID = "filesystem-war-jar-module";
+
+    public UserSpecifiedPathDeploymentSourceType() { super(SOURCE_TYPE_ID); }
+
+    /**
+     * Restore presentable name (e.g., to be "Filesystem JAR or WAR file - <file path>") of
+     * UserSpecifiedPathDeploymentSource.
+     */
+    @NotNull
+    @Override
+    public ModuleDeploymentSource load(@NotNull Element tag, @NotNull Project project) {
+      AppEngineDeploymentConfigurator.UserSpecifiedPathDeploymentSource userSpecifiedSource =
+          new AppEngineDeploymentConfigurator.UserSpecifiedPathDeploymentSource(ModulePointerManager
+              .getInstance(project).create(tag.getAttributeValue(NAME_ATTRIBUTE)));
+
+      Element settings = tag.getChild(DeployToServerRunConfiguration.SETTINGS_ELEMENT);
+      if (settings != null) {
+        String filePath = settings.getAttributeValue(
+            AppEngineDeploymentConfiguration.USER_SPECIFIED_ARTIFACT_PATH_ATTRIBUTE);
+
+        if (!StringUtil.isEmpty(filePath)) {
+          userSpecifiedSource.setName(
+              GctBundle.message(
+                  "appengine.flex.user.specified.deploymentsource.name.with.filename",
+                  new File(filePath).getName()));
+
+          return userSpecifiedSource;
+        }
+      }
+
+      userSpecifiedSource.setName(
+          GctBundle.message("appengine.flex.user.specified.deploymentsource.name"));
+
+      return userSpecifiedSource;
+    }
+
+    @Override
+    public void save(@NotNull ModuleDeploymentSource source, @NotNull Element tag) {
+      tag.setAttribute(NAME_ATTRIBUTE, source.getModulePointer().getModuleName());
     }
   }
 
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/CloudToolsPluginInitializationComponent.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/CloudToolsPluginInitializationComponent.java
@@ -16,12 +16,14 @@
 package com.google.gct.idea;
 
 import com.google.gct.idea.appengine.cloud.AppEngineCloudType;
+import com.google.gct.idea.appengine.cloud.AppEngineCloudType.UserSpecifiedPathDeploymentSourceType;
 import com.google.gct.idea.debugger.CloudDebugConfigType;
 
 import com.intellij.execution.configurations.ConfigurationType;
 import com.intellij.openapi.components.ApplicationComponent;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.remoteServer.ServerType;
+import com.intellij.remoteServer.configuration.deployment.DeploymentSourceType;
 import com.intellij.remoteServer.impl.configuration.deployment.DeployToServerConfigurationType;
 
 import org.jetbrains.annotations.NotNull;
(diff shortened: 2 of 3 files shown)


chanseokoh:
PMD complains about the unnecessary use of the fully qualified name, so it has to be addressed somehow.

etanshaul:
Regarding the fully qualified path problem: if you make sure that

import com.intellij.remoteServer.configuration.deployment.DeploymentConfigurator;

.. comes before:

import com.google.gct.idea.appengine.cloud.ManagedVmCloudType.ManagedVmDeploymentConfigurator.UserSpecifiedPathDeploymentSource;

Then it works. I haven't had a chance to dig into why yet. If this is the only alternative to a fully qualified name then I don't prefer it because any time a developer does an auto-optimize of the imports the compile errors will return.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmCloudType.java
```diff
@@ -196,6 +204,56 @@ public String getPresentableName() {
...
+      }
+    }
+  }
+
+  public static class UserSpecifiedPathDeploymentSourceType extends
+      DeploymentSourceType<ModuleDeploymentSource> {
+    private static final String NAME_ATTRIBUTE = "name";
+
+    public UserSpecifiedPathDeploymentSourceType() {
+      super("filesystem WAR/JAR module");
```
etanshaul: You can use a property here instead.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmCloudType.java
```diff
@@ -196,6 +204,56 @@ public String getPresentableName() {
...
+     * Restore presentable name (e.g., to be "Filesystem JAR or WAR file - <file path>") of
+     * UserSpecifiedPathDeploymentSource.
+     */
+    public ModuleDeploymentSource load(@NotNull Element tag, @NotNull Project project) {
+      UserSpecifiedPathDeploymentSource userSpecifiedSource = new UserSpecifiedPathDeploymentSource(
+          ModulePointerManager.getInstance(project).create(tag.getAttributeValue(NAME_ATTRIBUTE)));
+
+      Element settings = tag.getChild(DeployToServerRunConfiguration.SETTINGS_ELEMENT);
+      if (settings != null) {
+        String filePath = settings.getAttributeValue("userSpecifiedArtifactPath");
```
etanshaul: Properties for these too since now they actually have some meaning.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmCloudType.java
```diff
@@ -16,6 +16,7 @@
 
 package com.google.gct.idea.appengine.cloud;
 
+import com.google.gct.idea.appengine.cloud.ManagedVmCloudType.ManagedVmDeploymentConfigurator.UserSpecifiedPathDeploymentSource;
```
meltsufin: It looks like this import from "self" is causing the compilation error.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmCloudType.java
chanseokoh: "userSpecifiedArtifactPath" is an attribute you can see in ManagedVmDeploymentConfiguration. It has to match the attribute name defined there. Does it still make sense to use properties?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmCloudType.java
chanseokoh: It's not something that we need translation since it's sort of a unique internal id or key. For example, IntelliJ's SourceType implementations use names like "module" or "artifact". Do we still want to make it a property?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmCloudType.java
chanseokoh: Thanks for catching this. Works fine now without the self import.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmCloudType.java
etanshaul: Yea I think both of those can use the same property.

chanseokoh:
There is one fundamental flaw with this approach of explicitly adding a WAR/JAR file name for display. I'm not going to fix it in this PR, but It's good to think about it. If a user edits an existing deploy configuration to use another file, say 'target.war', and deploy it, the deploy panel will show two different names as below:

flaw

However, both actually point to the same configuration, so for example, if a user selects 'Deploy' in the right-click context menu on 'Filesystem JAR or WAR file - gs-spring-boot-...', it will deploy the current item which is 'target.war', which is clearly misleading.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmCloudType.java
chanseokoh: Seems like I can't do 

@Attribute(GctBundle.message("appengine.flex.config.user.specified.artifact.attribute"))

in 'ManagedVmDeploymentConfiguration.java.' The error is that the attribute value must be constant. I think it makes more sense to make them final constants here actually. PTAL.

patflynn: On Tue, Mar 29, 2016 at 2:23 PM, Chanseok Oh <notifications@github.com>
wrote:

> In
> google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/ManagedVmCloudType.java
> <https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/548#discussion_r57774043>
> :
>
> > +      super("filesystem WAR/JAR module");
> > +    }
> > +
> > +    @NotNull
> > +    @Override
> > +    /**
> > +     * Restore presentable name (e.g., to be "Filesystem JAR or WAR file - <file path>") of
> > +     * UserSpecifiedPathDeploymentSource.
> > +     */
> > +    public ModuleDeploymentSource load(@NotNull Element tag, @NotNull Project project) {
> > +      UserSpecifiedPathDeploymentSource userSpecifiedSource = new UserSpecifiedPathDeploymentSource(
> > +          ModulePointerManager.getInstance(project).create(tag.getAttributeValue(NAME_ATTRIBUTE)));
> > +
> > +      Element settings = tag.getChild(DeployToServerRunConfiguration.SETTINGS_ELEMENT);
> > +      if (settings != null) {
> > +        String filePath = settings.getAttributeValue("userSpecifiedArtifactPath");
>
> Seems like I can't do
>
> @Attribute <https://github.com/Attribute>
> (GctBundle.message("appengine.flex.config.user.specified.artifact.attribute"))
>
woah. Why would we be localizing a persistent property name?

chanseokoh: Reverted the changes and now they just use constants.

chanseokoh: I may be misunderstanding something here: when we say a property, are we referring to things like 'some.long.names.with.dots'?

patflynn: no I'm talking about a persisted field. Which I think is mapped to an xml element using the @Attribute. http://www.jetbrains.org/intellij/sdk/docs/basics/persisting_state_of_components.html

chanseokoh: Yeah. That's right. That string is stored in an xml as an attribute value. So, are we okay at this current state of changes?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineCloudType.java
```diff
@@ -196,6 +201,59 @@ public String getPresentableName() {
...
+  }
+
+  public static class UserSpecifiedPathDeploymentSourceType extends
+      DeploymentSourceType<ModuleDeploymentSource> {
+    private static final String NAME_ATTRIBUTE = "name";
+    private static final String SOURCE_TYPE_ID = "filesystem-war-jar-module";
+
+    public UserSpecifiedPathDeploymentSourceType() { super(SOURCE_TYPE_ID); }
+
+    @NotNull
```
patflynn: this belongs under the javadoc.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineCloudType.java
```diff
@@ -196,6 +201,59 @@ public String getPresentableName() {
...
+    @Override
+    /**
+     * Restore presentable name (e.g., to be "Filesystem JAR or WAR file - <file path>") of
+     * UserSpecifiedPathDeploymentSource.
+     */
+    public ModuleDeploymentSource load(@NotNull Element tag, @NotNull Project project) {
+      AppEngineDeploymentConfigurator.UserSpecifiedPathDeploymentSource userSpecifiedSource =
+          new AppEngineDeploymentConfigurator.UserSpecifiedPathDeploymentSource(
+              ModulePointerManager.getInstance(project).create(
```
patflynn: can tag.getAttribute not fit on the line above?  character limit is 100.

chanseokoh: It overflows by a few characters, unfortunately. It will fit if I break like below:

```
          new AppEngineDeploymentConfigurator.UserSpecifiedPathDeploymentSource(ModulePointerManager
              .getInstance(project).create(tag.getAttributeValue(NAME_ATTRIBUTE)));
```

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineCloudType.java
```diff
@@ -196,6 +201,59 @@ public String getPresentableName() {
...
+                  new File(filePath).getName()));
+
+          return userSpecifiedSource;
+        }
+      }
+
+      userSpecifiedSource.setName(
+          GctBundle.message("appengine.flex.user.specified.deploymentsource.name"));
+
+      return userSpecifiedSource;
```
patflynn: I think it's more readable as:

`       if (!StringUtil.isEmpty(filePath)) {
           userSpecifiedSource.setName(
               GctBundle.message(
                   "appengine.flex.user.specified.deploymentsource.name.with.filename",
                   new File(filePath).getName()));
         } else {
          userSpecifiedSource.setName(
             GctBundle.message("appengine.flex.user.specified.deploymentsource.name")); 
         }
       }
 
 
       return userSpecifiedSource;`

chanseokoh: The default name ('appengine.flex.user.specified.deploymentsource.name') should also be set when 'settings == null', so if we put it inside an else branch, the outer if statement should also have the same duplicate else branch. What do you think?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineCloudType.java
chanseokoh: Fixed.

patflynn:
Hey Chanseok are you still working on this?  I don't think your most recent changes have been pushed.

chanseokoh:
Sorry, I just pushed it. The most recent change is moving annotations.

patflynn:
LGTM. Squash and commit.

PR #553, opened 2016-03-29
Update ManagedVM naming to flex

patflynn:
This change removes 'Managed VM' as a brand and concept for our App
Engine integration.  Where things are clearly Flex only names are
updated accordingly. Otherwise the more generic 'AppEngine' is used.


google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -102,7 +102,7 @@ public void actionPerformed(ActionEvent e) {
       }
     });
     userSpecifiedArtifactFileSelector.addBrowseFolderListener(
-        GctBundle.message("appengine.managedvm.config.user.specified.artifact.title"),
+        GctBundle.message("appengine.flex.config.user.specified.artifact.title"),
```
etanshaul: This class is renamed have 'AppEngine', but is using a property named 'flex'. Isn't this class Flex specific?

patflynn: Some of these classes have flex specific properties or do flex specific things, but conceivably we might unify that in teh future. If we don't we'll break out standard and flex configurations and rename.

etanshaul:
lgtm

patflynn:
@chanseokoh @meltsufin you guys will want to sync this soon so as to limit merge conflicts.

PR #555, opened 2016-03-29
let gcloud validate custom app.yaml / Dockerfile deployments

### LINKED ISSUE #554: Should MVM deployment allow a custom app.yaml without selecting a custom Dockerfile?
Or vice versa.

Currently, if the user attempts this, the deploy will fail with an exception (filenotfound for the config file that was not specified).

etanshaul:
fixes #554


etanshaul:
Updated to not fail during custom deployment when app.yaml and/or Dockerfile aren't specified, but rather let gcloud handle it. Also snuck in a mvm variable name change.

patflynn:
LGTM

Issue #543, opened 2016-03-28
What happens if we try to deploy with an old version of gcloud?


patflynn:
Ideally the user would be notified to update if their version of gcloud does not support credential_file_override flag

meltsufin:
Deployment fails with 

> ERROR: gcloud crashed (AuthPermanentFail): Refresh token is invalid.

**Version used**: Google Cloud SDK 101.0.0

**Console output**:
...
Beginning deployment...
ERROR: gcloud crashed (AuthPermanentFail): Refresh token is invalid.
 
If you would like to report this issue, please run the following command:
   gcloud feedback
Failed to deploy 'userSpecifiedSource': Deployment failed with exit code: 1
...

PR #556, opened 2016-03-29
Notify to upgrade to latest SDK when deploy fails

meltsufin:
Refer to: #543

App Engine deployment currently fails if the user has an outdated Google Cloud SDK. This update adds a notification asking the user to upgrade to the latest SDK when a deployment fails for any reason.

This is what the user will see:
screenshot from 2016-03-29 15 07 39


etanshaul:
This error message could appear even if an outdated gcloud isn't the cause right?

meltsufin:
Yes, but that might be ok.

meltsufin:
@patflynn What do you think?

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -43,7 +43,7 @@ appengine.deployment.error.creating.staging.directory=There was an unexpected er
 appengine.generate.config.button=Generate...
 appengine.appyaml.location.label=app.yaml path\:
 appengine.dockerfile.location.browse.button=Browse for Your Dockerfile Location
-appengine.deployment.error.with.code=Deployment failed with exit code: {0}
+appengine.deployment.error.with.code=Deployment failed with exit code: {0}\nPlease make sure that you are using the latest version of the Google Cloud SDK. (See: https://cloud.google.com/sdk/)
```
etanshaul: Since I have windows on the brain from the other PR I was just working - do we want to be safe here and use a platform independent line break (since windows typically does a crlf)?

etanshaul:
lgtm aside from that last comment

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
meltsufin: I think '\n' should look the same on Windows also. How would you do a device independent line break?

etanshaul: Found this:
http://docs.oracle.com/javase/8/docs/api/java/lang/System.html#lineSeparator--

etanshaul: I think that's good in Java 7 too.

meltsufin:
@etanshaul I took your suggestion to use System.lineSeparator(). It actually makes the bundle cleaner by making the message separate.

etanshaul:
nice. lgtm

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
patflynn: I think the answer to that question is 'no because this is Java so the \n
break in a string is platform independent '.

etanshaul: Yes that's definitely right. @meltsufin sorry..what you had before was fine, I'm not sure what I was thinking really.

meltsufin: I still kind of like the message being its own property.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/DoAppEngineDeployment.java
```diff
@@ -153,7 +153,8 @@ public void undeploy(@NotNull UndeploymentTaskCallback callback) {
         } else {
           logger.error("Deployment process exited with an error. Exit Code:" + event.getExitCode());
           callback.errorOccurred(
-              GctBundle.message("appengine.deployment.error.with.code", event.getExitCode()));
+              GctBundle.message("appengine.deployment.error.with.code", event.getExitCode())
+                  + System.lineSeparator() + GctBundle.message("appengine.deployment.error.old.sdk"));
```
patflynn: I'm skeptical that you should be passing System.lineSeparator() in a Java String for output by a Java program.

etanshaul: Yea I was reading that too. That message just gets displayed within an IJ window - it never gets used in a system dependent way (like doing a system call or something) - so I also \n should be fine.

patflynn:
it seems obvious that they have the cloud SDK already so shouldn't we be asking them to update it using 'gcloud components update' ?

meltsufin:
I wasn't aware of this command and it does seem better. Do you prefer just saying run 'gcloud components update', or point them to the doc on this command: https://cloud.google.com/sdk/gcloud/reference/components/update?

patflynn:
I propose we cutout the middleman and suggest they 'gcloud components
update' on error.

meltsufin:
I went back to using "\n" for line breaks, and now recommending 'gcloud componets update'.

screenshot from 2016-03-30 10 11 33

patflynn:
squash and merge. LGTM.

PR #559, opened 2016-03-29
move deploy auth check to the connect callback

### LINKED ISSUE #552: App Engine server type allows to connect even when logged out
Steps to reproduce:

1. Set up the App Engine Flexible Environment server.
2. Logout from GCP using the account icon on the toolbar.
3. Click "Connect" button on the App Engine Flexible Environment Server and see it succeed.

The same can be said about the disconnect button.

The problem is that after this "successful connection", the deployment action fails with:

> Failed to deploy 'userSpecifiedSource': You must be logged in to deploy."

etanshaul:
fixes #552 

There is no need to duplicate the auth checking in both connect() and deploy() since deploy calls connect. Now, if the user tries to "connect" without being logged in, then an error message will be shown (the only caveat being that it will say "You must be logged in to deploy.".

I think this is a little less confusing than having the connect button (which I couldn't yet figure out how to disable) appear to work but then the deploy fails due to missing auth ("connect" seems to suggest that auth is a prerequisite).

### DIFF
--- a/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
+++ b/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
@@ -46,7 +46,7 @@ appengine.dockerfile.location.browse.button=Browse for Your Dockerfile Location
 appengine.deployment.error.with.code=Deployment failed with exit code: {0}\nPlease make sure that you are using the latest version of the Google Cloud SDK.\nRun ''gcloud components update'' to update the SDK. (See: https://cloud.google.com/sdk/gcloud/reference/components/update.)
 appengine.deployment.error.during.execution=Deployment failed due to an unexpected error during the deployment execution phase.
 appengine.deployment.error.during.staging=Deployment failed due to an unexpected error while creating the staging directory for deployment.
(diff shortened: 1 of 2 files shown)


etanshaul:
Ignore my changes for now. I was wrong about not needing to duplicate the authentication in connect() and deploy():

- Log in, then connect
- Log out
- Deploy -> connect would never get called (since it already happened) and therefore my changes would bypass the necessary auth check

etanshaul:
Updated - added a log in check to the cloud connect() method and showing a connect specific error if the user is not logged in.

etanshaul:
@patflynn @meltsufin ptal

I updated the wording a few times and currently settled on a more generic message. The reason is that this message could be display both from the context of a connect and a deploy, so I wanted it to make sense for both.

patflynn:
LGTM

meltsufin:
LGTM, but see my comment on the issue discussion. #552

patflynn:
- If we want to map connected state to something meaningful such as 'cloud' access, then it's not enough that the user is logged in, the user should have access to the project that the cloud represents.

PR #561, opened 2016-03-30
Set clientVersion on Cloud Debugger Debuggee calls

### LINKED ISSUE #558: Set clientVersion on Cloud Debugger Debuggee requests
client version string should be "google.com/intellij/v{currentPluginVersion}".

meltsufin:
Fixes #558.

This change sets client version string to be "google.com/intellij/v{currentPluginVersion}".

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/CloudToolsPluginInfoService.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/CloudToolsPluginInfoService.java
@@ -21,4 +21,12 @@ package com.google.gct.idea;
  * information service across google-cloud-tools-plugin and google-account-plugin
  */
 public interface CloudToolsPluginInfoService extends PluginInfoService {
+
+  /**
+   * Returns the fully qualified version of the plugin for specifying in Cloud Debugger API
+   * requests.
+   *
+   * @return "google.com/intellij/v{currentPluginVersion}"
+   */
+  String getClientVersionForCloudDebugger();
 }
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessStateController.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessStateController.java
@@ -19,8 +19,10 @@ import com.google.api.client.googleapis.json.GoogleJsonResponseException;
 import com.google.api.client.repackaged.com.google.common.base.Strings;
 import com.google.api.services.clouddebugger.v2.Clouddebugger.Debugger;
 import com.google.api.services.clouddebugger.v2.model.*;
+import com.google.gct.idea.CloudToolsPluginInfoService;
 import com.google.gct.idea.util.GctBundle;
 import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.ui.Messages;
 import com.intellij.util.containers.ContainerUtil;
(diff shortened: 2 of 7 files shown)


meltsufin:
Hold off on reviewing this. I found a few more locations to set clientVersion.

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessStateController.java
```diff
@@ -41,6 +43,8 @@
   private static final int INITIAL_DELAY_MS = 2000;
   private static final Logger LOG = Logger.getInstance(CloudDebugProcessStateController.class);
   private static final int PERIOD_MS = 500;
+  public static final String CLIENT_VERSION = "google.com\\/intellij\\/v" + ServiceManager
```
patflynn: This isn't actually a constant.

I'd prefer public final String clientVersion;  and then initialize it in the constructor.

patflynn:
Do we also need to pass this in the CloudDebugGobalPoller too?

google-cloud-tools-plugin/src/com/google/gct/idea/debugger/CloudDebugProcessStateController.java
meltsufin: Should we then make it private and create a getter for it?

patflynn: ah I just noticed it was public.  Why would you access this client version
from outside of this class instead of using the service?

meltsufin: I didn't want to duplicate the prefix everywhere.

patflynn: Feel free to add a new method to the CloudToolsPluginInfoService

meltsufin: OK, I like that solution.

meltsufin:
@patflynn PTAL

google-cloud-tools-plugin/src/com/google/gct/idea/CloudToolsPluginInfoService.java
```diff
@@ -21,4 +21,5 @@
  * information service across google-cloud-tools-plugin and google-account-plugin
  */
 public interface CloudToolsPluginInfoService extends PluginInfoService {
+  String getClientVersion();
```
patflynn: This name is going to confuse people who see getPluginVersion().  Since this exists really purely for Debugger invocations I wouldn't mind calling it something like getClientVersionForCloudDebugger() or something like that. (I know that's long..)

patflynn:
Fix the name, add javadoc, and then LGTM.

patflynn:
(also squash)

google-cloud-tools-plugin/src/com/google/gct/idea/CloudToolsPluginInfoService.java
meltsufin: I know it's currently only used for Cloud Debugger, but there's nothing really Cloud Debugger specific in the value of the version. Is it possible that other APIs in the future will follow suit, and will also require us to pass the client version?

patflynn: I don't expect that's going to happen, both the API parameter and format of
the string are CDB defined.  In the future the services will use the user
agent provided by setApplicationName, it just seems like the One Platform
infrastructure is not there yet.

Issue #560, opened 2016-03-30
Customer Dockerfile generation ('Generate...' button) error


chanseokoh:
Clicking the 'Generate...' button in the Flex deploy dialog and specifying a directory spits out an error:

[  57172]  ERROR - llij.ide.plugins.PluginManager - Argument for @NotNull parameter 'deployPackage' of com/google/gct/idea/appengine/cloud/DeploymentArtifactType.typeForPath must not be null 
java.lang.IllegalArgumentException: Argument for @NotNull parameter 'deployPackage' of com/google/gct/idea/appengine/cloud/DeploymentArtifactType.typeForPath must not be null
	at com.google.gct.idea.appengine.cloud.DeploymentArtifactType.typeForPath(DeploymentArtifactType.java)

patflynn:
can you fix this Chanseok?

chanseokoh:
I'll do it.

etanshaul:
Ah, this is because its currently possible to generate a Dockerfile before selecting a deployment source. The Dockerfile relies on the deployment source being chosen (e.g. jar or war).

chanseokoh:
Interesting. The app.yaml generation works fine though.

etanshaul:
right, because the app.yaml doesn't care about the deployment source

etanshaul:
I suggest you add a new error message when the user attempts to generate a Dockerfile before having set the deployment source. Something like:

"Please select a deployment source before generating a Dockerfile."

chanseokoh:
I cannot throw a ConfigurationException from an ActionListener. What could be the way?

etanshaul:
What is the error you are seeing?

chanseokoh:
Seems like I need to handle this error inside 'GenerateConfigActionListener.actionPerformed()', but not sure how I can actually propagate the error message to the dialog there.

chanseokoh:
I guess I will create and pop-up an error message dialog rather than using the ConfigurationException mechanism.

patflynn:
Would an option be to disable the app configuration if no artifact has been
selected for user selected jar/war?

etanshaul:
This issue goes a little deeper: if the user selects a filesystem deployment source, the Dockerfile generation doesn't work either. It attempts to call `deploymentSource.getFile()` but there is no file set for this custom deployment source type since the file gets created later (when the user browses to one).

etanshaul:
What I just said may warrant a separate issue.

[...]

PR #565, opened 2016-03-30
Get correct filesystem deployment file for Dockerfile generation

### LINKED ISSUE #564: Dockerfile generation doesn't work for filesystem deployment sources
If the user selects a filesystem deployment source, the Dockerfile generation doesn't work. It attempts to call deploymentSource.getFile() but there is no file set for this custom deployment source type since the file gets created later (when the user browses to one).

etanshaul:
fixes #564

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineCloudType.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineCloudType.java
@@ -176,6 +176,7 @@ public class AppEngineCloudType extends ServerType<AppEngineServerConfiguration>
     protected static class UserSpecifiedPathDeploymentSource extends ModuleDeploymentSourceImpl {
       private String name =
           GctBundle.message("appengine.flex.user.specified.deploymentsource.name");
+      private String userSpecifiedFilePath;
 
       public UserSpecifiedPathDeploymentSource(@NotNull ModulePointer pointer) {
         super(pointer);
@@ -200,6 +201,16 @@ public class AppEngineCloudType extends ServerType<AppEngineServerConfiguration>
 
       public void setName(String name) {
         this.name = name;
+      }
+
+      @Nullable
+      @Override
+      public File getFile() {
+        return userSpecifiedFilePath != null ? new File(userSpecifiedFilePath) : null;
+      }
+
+      public void setFilePath(@NotNull String userSpecifiedFilePath) {
+        this.userSpecifiedFilePath = userSpecifiedFilePath;
       }
 
       @NotNull
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
@@ -34,6 +34,7 @@ import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.openapi.vfs.LocalFileSystem;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.remoteServer.configuration.deployment.DeploymentSource;
+import com.intellij.ui.DocumentAdapter;
 
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -51,6 +52,7 @@ import javax.swing.JComboBox;
 import javax.swing.JComponent;
 import javax.swing.JPanel;
 import javax.swing.border.TitledBorder;
+import javax.swing.event.DocumentEvent;
 
 /**
  * Editor for an App Engine Deployment runtime configuration.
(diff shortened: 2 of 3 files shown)


etanshaul:
@chanseokoh @elharo ptal

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -134,8 +135,14 @@ public File get() {
         new GenerateConfigActionListener(project, "Dockerfile", new Supplier<File>() {
           @Override
           public File get() {
+            File deploymentFile;
+            if(deploymentSource.getType() instanceof UserSpecifiedPathDeploymentSourceType) {
+              deploymentFile = new File(userSpecifiedArtifactFileSelector.getText());
```
patflynn: could we not have the UserSpecifiedPathdDeploymentSourceType implement getFile correctly?

etanshaul: The problem is that we don't know what to return from getFile: this only gets determined when a user manually browses to one. We could manually set it, e.g. add state to the class. But I'm not sure if thats better.

etanshaul: What I mean is, we could override getFile and have it return an internal state that we manually set whenever a user browses to a filesystem jar/war. Do you like that approach better?

patflynn: That is kind of what I expect the DeploymentSource contract is asking our
implementation to do.

chanseokoh:
LGTM.

etanshaul:
@patflynn updated to use DeploymentSource getFile

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineCloudType.java
```diff
@@ -202,6 +203,16 @@ public void setName(String name) {
         this.name = name;
       }
 
+      @Nullable
+      @Override
+      public File getFile() {
+        return new File(userSpecifiedFilePath);
```
patflynn: should this return null if userSpecifiedfilePath is not set?

etanshaul: I think if the userSpecifiedFilePath is not set then the fix that @chanseokoh is working on (#560) should prevent the custom deployment UI from being shown in the first place.

chanseokoh: Regarding #560, even if the userSpecifiedFilePath is set, if that path does not end with either .jar or .war, then Dockerfile auto-generation will still produce an error. For this reason, I am not working on showing/hiding the custom deployment UI, but rather working on displaying an error message dialog.

etanshaul: @chanseokoh Maybe I'm not understanding, but couldn't you only show the custom config if the path is set AND is a jar/war (currently the validation ensures that it has a .jar/.war extension)?

etanshaul: As a side note, I think adding an error dialog into the mix would be a bit odd since we are already showing label error messages. So having the one off dialog may feel strange.

etanshaul:
@patflynn ptal. I updated to better satisfy the @Nullable contract of getFile()

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/DeploymentArtifactType.java
```diff
@@ -31,11 +32,13 @@
    * Returns the right {@code DeploymentArtifactType} for the given {@code deployPackage}.
    */
   @NotNull
-  public static DeploymentArtifactType typeForPath(@NotNull File deployPackage) {
-    if (deployPackage.getPath().endsWith(".jar")) {
-      return JAR;
-    } else if (deployPackage.getPath().endsWith(".war")) {
-      return WAR;
+  public static DeploymentArtifactType typeForPath(@Nullable File deployPackage) {
```
patflynn: apologies for flip flopping, but couldn't we leave this as @NotNull and just have the callers of this method handle the case where deploymenttSource.getFile == null?

patflynn:
LGTM

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/DeploymentArtifactType.java
etanshaul: looking at the usages I think it makes sense to let it handle the nulls rather than the callers. The callers are already handling DeploymentArtifactType.UNKNOWN when needed, so my thinking is to just let null just be treated like another unknown type. Let me know if you disagree.

patflynn: OK. LGTM

PR #568, opened 2016-03-30
Show error pop-up for Dockerfile auto-generation

chanseokoh:
Addresses #560

Users will see a pop-up to select a valid source when clicking a 'Generate...' button for Dockerfile auto-generation.

I'll upload a screenshot shortly.

I am not calling 'setDisposable()' whereas another use case of this balloon pop-up in our codebase sets it. Thought it will be fine without it, but let me know if it has to be set.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
@@ -25,7 +25,12 @@ import com.intellij.openapi.fileChooser.FileChooserDescriptorFactory;
 import com.intellij.openapi.options.ConfigurationException;
 import com.intellij.openapi.options.SettingsEditor;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.ui.MessageType;
 import com.intellij.openapi.ui.TextFieldWithBrowseButton;
+import com.intellij.openapi.ui.popup.Balloon;
+import com.intellij.openapi.ui.popup.Balloon.Position;
+import com.intellij.openapi.ui.popup.BalloonBuilder;
+import com.intellij.openapi.ui.popup.JBPopupFactory;
 import com.intellij.openapi.util.Comparing;
 import com.intellij.openapi.util.Condition;
 import com.intellij.openapi.util.SystemInfo;
@@ -35,10 +40,12 @@ import com.intellij.openapi.vfs.LocalFileSystem;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.remoteServer.configuration.deployment.DeploymentSource;
 import com.intellij.ui.DocumentAdapter;
+import com.intellij.ui.awt.RelativePoint;
 
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.awt.Point;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.beans.PropertyChangeEvent;
@@ -133,7 +140,7 @@ public class AppEngineDeploymentRunConfigurationEditor extends
           public File get() {
             return appEngineHelper.defaultAppYaml();
           }
-        }, appYamlPathField));
+        }, appYamlPathField, userSpecifiedArtifactFileSelector));
     generateDockerfileButton.addActionListener(
         new GenerateConfigActionListener(project, "Dockerfile", new Supplier<File>() {
           @Override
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -129,15 +137,21 @@ public boolean value(VirtualFile file) {
...
     generateDockerfileButton.addActionListener(
         new GenerateConfigActionListener(project, "Dockerfile", new Supplier<File>() {
           @Override
           public File get() {
+            File deploymentFile;
+            if (deploymentSource.getType() instanceof UserSpecifiedPathDeploymentSourceType) {
+              deploymentFile = new File(userSpecifiedArtifactFileSelector.getText());
+            } else {
+              deploymentFile = deploymentSource.getFile();
+            }
```
etanshaul: Did you base this PR off of an intermediate stage of #565? I think I have since removed this code. Or maybe you just ended up doing the same thing (if you look at #565 this checking of the type is no longer needed).

chanseokoh: I will remove this shortly. I copied it manually when you create #565.

chanseokoh:
#565 should be in first to test this.

chanseokoh:
screenshot from 2016-03-30 15 13 42

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -236,20 +243,35 @@ protected JComponent createEditor() {
...
       this.filePicker = filePicker;
+      this.fileSelector = fileSelector;
     }
 
     @Override
     public void actionPerformed(ActionEvent event) {
+      // This can happen only when the deployment source is UserSpecifiedPathDeploymentSource
+      // and the selected file does not have .jar or .war extension. Artifact deployment sources
+      // always have .jar or .war extensions because we filter out others.
+      if (sourceFileProvider.get() == null) {
```
etanshaul: Just a suggestion - if you update the condition to:

`if (isUserSpecifiedPathDeploymentSource() && sourceFileProvider.get() == null)`

Then its sort of self-documenting and you can get rid of the comments. What do you think

chanseokoh: Cannot call that method directly inside this class, unfortunately. I think it's also a bit better to always return when it's null, although we know it cannot happen if it's not UserSpecifiedpathDeploymentSource.

etanshaul: Because the class is static? You should be able to just make it non-static and then call that method - assuming you think my suggestion is a good idea.

chanseokoh: Good idea. Based on you suggestion, how about throwing an AssertionError if it is not UserSpecifiedDeploymentSource? Removed comments and made it static. I also wrapped some lines.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -246,26 +253,48 @@ protected JComponent createEditor() {
...
       this.filePicker = filePicker;
+      this.fileSelector = fileSelector;
     }
 
     @Override
     public void actionPerformed(ActionEvent event) {
+      if (sourceFileProvider.get() == null) {
+        if (!isUserSpecifiedPathDeploymentSource()) {
+          throw new AssertionError("By our design, artifacts should always have .jar or .war; "
+            + " sourceFileProvider will return a correct config file.");
```
etanshaul: minor: you can probably get rid of the second line of the error message (the first one covers it I think). otherwise, lgtm

etanshaul: Oh since its an exception message not a comment, how about something like this:

"Error generating configuration file: artifact deployment source is missing a source .jar or .war file."

chanseokoh: That message is informative. Fixed.

PR #570, opened 2016-03-31
Confirm config file overwrite / check directory creation

### LINKED ISSUE #569: Generating app.yaml and Dockerfile ('Generate...' button) silently overwrites existing files
In the Flex deploy dialog, clicking the 'Generate...' buttons and entering a destination directory will immediately generate (copy) app.yaml or Dockerfile and can overwrite existing files. A small confirmation dialog to confirm overwriting may be desired.

chanseokoh:
Fixes #569 

A confirmation/error dialog will show up if a user
1) selects a directory where another config file already exists (confirm overwrite)
2) enters a file instead of a directory (error/reject)
3) enters a directory that does not yet exist (confirm directory creation)

The dialog shows up after the directory-choosing dialog closes.


chanseokoh:
Also changed the 'Ok' button of the directory select window to 'Generate'.

etanshaul:
Regarding your point (2): how can the user select a file? The generate file browser is a directory browser.

chanseokoh:
You can manually type a file.

etanshaul:
I tried typing a file manually into the Generate dialog (something like "test.jar") and it didn't warn me after clicking generate. It made a directory like ".../test.jar/app.yaml".

chanseokoh:
Ah, pick a directory name that is actually an existing file. For example, type '/tmp' first to create /tmp/app.yaml', and then type '/tmp/app.yaml'. Since you can't generate '/tmp/app.yaml/app.yaml', this leads to an error.

etanshaul:
Ah I see. 

One UI related concern I see here is really long paths being displayed in the dialog causing it to stretch indefinitely. I didn't noticed any max-width or wrap settings in the gui code. Maybe add something like that or don't display the path at all.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -31,6 +31,9 @@ appengine.flex.dockerfile.location.label=Dockerfile path\:
 appengine.flex.config.destination.folder.label=Choose destination folder\:
 appengine.flex.config.choose.destination.folder.window.title=Choose destination folder
 appengine.flex.config.destination.chooser.title=Choose Generated Configuration Destination Folder
+appengine.flex.config.destination.create.dir=Do you want to create the non-existing directory below?
```
etanshaul: Wording suggestion:

"Directory does not exist. Create it now?

chanseokoh:
I'll try experimenting with max-width or wrap settings. The reason I'm showing the path is that this confirm windows pops up _after_ closing the generate window. Whether you click 'Ok' or 'Cancel', you don't see again what you typed in the directory input field.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
chanseokoh: Thanks for the suggestion. I also felt a bit uncomfortable with this wording.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/SelectConfigDestinationFolderDialog.java
```diff
@@ -60,4 +63,15 @@ protected JComponent createCenterPanel() {
   public File getDestinationFolder() {
     return new File(destinationFolderChooser.getText());
   }
+
+  @NotNull
+  protected Action[] createActions() {
+    getOKAction().putValue(Action.NAME, "Generate");
+
+    if (SystemInfo.isMac) {
```
etanshaul: Why not just use the same ordering regardless of platform? We aren't really trying to create a "mac" experience.

etanshaul: On second thought, I see that the superclass does this. Should there be an @Override annotation here?

chanseokoh: Absolutely @Override.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -31,6 +31,9 @@ appengine.flex.dockerfile.location.label=Dockerfile path\:
...
+appengine.flex.config.destination.overwrite=The file below already exists. Do you want to overwrite it?
```
patflynn: I wouldn't use a UI positional relative word like 'below'.  It would be easy to move the UI around and not realize this needed to change.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -31,6 +31,9 @@ appengine.flex.dockerfile.location.label=Dockerfile path\:
 appengine.flex.config.destination.folder.label=Choose destination folder\:
 appengine.flex.config.choose.destination.folder.window.title=Choose destination folder
 appengine.flex.config.destination.chooser.title=Choose Generated Configuration Destination Folder
+appengine.flex.config.destination.create.dir=Directory does not exist. Create it too now?
```
etanshaul: Sorry for nitpicking, but "Create it too now" is a little awkward. Maybe drop the "too"?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/FileConfirmationDialog.java
```diff
@@ -0,0 +1,93 @@
...
+import javax.swing.Action;
+import javax.swing.JComponent;
+import javax.swing.JLabel;
+import javax.swing.JPanel;
+
+/**
+ * A dialog warning for, e.g., overwriting existing app.yaml and Dockerfile configuration files
+ * when user selects a destination folder to auto-generate them.
+ */
+public class FileConfirmationDialog extends DialogWrapper {
```
patflynn: I wonder if Jetbrains doesn't already provide a Dialog like this.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/FileConfirmationDialog.java
```diff
@@ -0,0 +1,93 @@
...
+    return rootPanel;
+  }
+
+  @NotNull
+  @Override
+  protected Action[] createActions() {
+    if (dialogType == DialogType.NOT_DIRECTORY_ERROR) {
+      return new Action[]{ getOKAction() };
+    } else {
+      if (SystemInfo.isMac) {
```
patflynn: Is this a Mac UI thing?  I think we should let the platform handle matching the host OS. If we write point solutions like this we'll create inconsistencies.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/FileConfirmationDialog.java
chanseokoh: It's provided by Jetbrains and I think is a proper choice: http://www.jetbrains.org/intellij/sdk/docs/user_interface_components/dialog_wrapper.html

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
chanseokoh: Fixed.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/FileConfirmationDialog.java
chanseokoh: Gotcha. I should have called a base method - it'll handle the Mac thing.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/SelectConfigDestinationFolderDialog.java
chanseokoh: Changed to call a base method.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
chanseokoh: Fixed.

chanseokoh:
Changes are in. PTAL.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/FileConfirmationDialog.java
etanshaul: Check out com.intellij.openapi.ui.Messages (e.g Messages#showYesNoDialog). It might provide some implementations that you can use.

chanseokoh: Cool. I should have got this Messages class first. I will take a look at.

meltsufin:
I think the dialog needs some text on the title bar. It would also be nice to have a warning icon on it.

chanseokoh:
Ah, yeah, I missed the title. I'm think of simple and short titles: 'Confirm Overwrite', 'Confirm Directory Creation', and 'Error' (for not being a directory).

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/FileConfirmationDialog.java
chanseokoh: I think it's not a bad idea to have an read-only max-size-limited input filed for path display in the current dialog implementation as Etan previously said a dialog may be stretched too long due to a long path.

chanseokoh:
Added titles to the dialog.

meltsufin:
The titles look good.

What do you guys think of disabling the "Generate..." button when there is already a path to a valid file configured? This way as soon as they generate the file, the button will be automatically disabled.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/FileConfirmationDialog.java
etanshaul: I wonder if there's a way to limit the length for built in dialog implementations. Seems like a shame to write our own just for that.

chanseokoh: Yeah, best to avoid it. I'll take a look at the com.intellij.openapi.ui.Messages more closely and thinker with various things in there.

chanseokoh: There is no direct option to limit the length. I could probably do it by subclassing Message.InputDialog or something, but then we're not solving our problem to get away from a new class. I see basically two options: 1) just use one of the Messages functions and forget about the length; 2) live it with the current FileConfirmationDialog implementation.

patflynn:
LGTM.

PR #571, opened 2016-03-31
stop an app deployed to mvm

### LINKED ISSUE #453: The user should be able to stop a running Managed VM.

etanshaul:
Fixes #453

In progress. Functionality works, just needs some cleanup.


etanshaul:
@patflynn @chanseokoh @meltsufin, you guys can start taking a look at this

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -45,6 +45,8 @@ appengine.appyaml.location.label=app.yaml path\:
 appengine.dockerfile.location.browse.button=Browse for Your Dockerfile Location
 appengine.deployment.error.with.code=Deployment failed with exit code: {0}\nPlease make sure that you are using the latest version of the Google Cloud SDK.\nRun ''gcloud components update'' to update the SDK. (See: https://cloud.google.com/sdk/gcloud/reference/components/update.)
 appengine.deployment.error.during.execution=Deployment failed due to an unexpected error during the deployment execution phase.
+appengine.stopapp.error.during.execution=Failed to stop application due to an unexpected error during the execution phase.
```
patflynn: appengine.stop.module.version.execution.error?

I'm just wondering if we might end up supporting module or app level stop.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineAction.java
```diff
@@ -0,0 +1,144 @@
...
+
+  protected LoggingHandler getLoggingHandler() {
+    return loggingHandler;
+  }
+
+  protected void executeProcess(
+      @NotNull GeneralCommandLine commandLine,
+      @NotNull File appDefaultCredentialsPath,
+      @NotNull ProcessListener listener) throws ExecutionException {
+    consoleLogLn("Executing: " + commandLine.getCommandLineString());
```
patflynn: should we be logging the command string before we add the base parameteres?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineAction.java
```diff
@@ -0,0 +1,144 @@
...
+
+    // Common command line settings
+    commandLine.addParameter("--project=" + appEngineHelper.getProjectId());
+    commandLine
+        .addParameter("--credential-file-override=" + appDefaultCredentialsPath.getAbsolutePath());
```
patflynn: why involve the subclass in generating the credentials at all?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/util/AppEngineUtil.java
```diff
@@ -0,0 +1,42 @@
...
+   * Generates a version string in the gcloud version format:
+   *
+   * [year][month][day]t[hours][min][sec]
+   *
+   * For example: 20160331t132711
+   *
+   * @return generated version
+   */
+  public static String generateVersion() {
+    String versionPattern = "yyyyMMdd't'kms";
```
patflynn: I wonder if cloud SDK always uses the local timezone?  Feels like that would be bad for developers deploying to the same project across timezones.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/util/AppEngineUtil.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+package com.google.gct.idea.appengine.util;
+
+import org.joda.time.DateTime;
+import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
+
+public class AppEngineUtil {
```
patflynn: class javadoc is required.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/util/AppEngineUtil.java
etanshaul: Great catch. I'll ask that question on the cloud sdk thread.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: How about: appengine.stop.modules.version.execution.error

All I changed from your suggestion is making module plural.  It reads a little awkwardly but it makes it consistent with the underlying gcloud command. The stop could potentially act on multiple modules for a multimodule deployment.

patflynn: +1

etanshaul:
Updated with comments so far.

Moved all credential checking to the superclass.

meltsufin:
From the UX point of view, I find it a little odd how when you click the undeploy button, the module just instantly disappears and there is no confirmation that the undeployment was successful. Contrast that with the deploy action which takes a while, and you see the process happening in the console. I think it would be nice to bring up a popup confirming the success of the action.

etanshaul:
I agree. Or instead of a popup if we could get the console output to remain visible showing some success message indicating that the app was undeployed.

meltsufin:
Yes, either that, or maybe we can even hook into the IJ Event Log somehow? Right now it only seems to report log.error messages in the Event Log.

etanshaul:
The event log doesn't feel like the right place for a success message. Ideally it would be in the console view but this doesn't look possible because of how IJ deletes the deployment after a successful undeploy (and the console goes away with it). I'm exploring using a dialog here.

etanshaul:
@meltsufin I added a success dialog to the stop process. Let me know what you think.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -45,6 +45,10 @@ appengine.appyaml.location.label=app.yaml path\:
 appengine.dockerfile.location.browse.button=Browse for Your Dockerfile Location
 appengine.deployment.error.with.code=Deployment failed with exit code: {0}\nPlease make sure that you are using the latest version of the Google Cloud SDK.\nRun ''gcloud components update'' to update the SDK. (See: https://cloud.google.com/sdk/gcloud/reference/components/update.)
 appengine.deployment.error.during.execution=Deployment failed due to an unexpected error during the deployment execution phase.
+appengine.stop.modules.version.execution.error=Failed to stop application due to an unexpected error during the execution phase.
+appengine.stop.modules.version.execution.error.with.code=Stop application failed with exit code: {0}
+appengine.stop.modules.version.success.dialog.message=Application successfully undeployed.
```
meltsufin: I guess the correct terminology would be "stopped", not "undeployed". Also, maybe specify which application in the message?
"App Engine application {0} has successfully stopped." or something like that.

meltsufin:
Maybe it then also makes sense to have a similar dialog for completion of deployment as well?

etanshaul:
Yea. I think so.

etanshaul:
- Added successful deployment dialog
- Added an undeployment confirmation dialog. Before, the application stop was immediately executed when clicking the button. Now the user can cancel out of it.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineAction.java
```diff
@@ -0,0 +1,155 @@
...
+  private File credentialsPath;
+
+  public AppEngineAction(
+      @NotNull LoggingHandler loggingHandler,
+      @NotNull AppEngineHelper appEngineHelper,
+      @NotNull RemoteOperationCallback callback) {
+    this.loggingHandler = loggingHandler;
+    this.appEngineHelper = appEngineHelper;
+
+    credentialsPath = createApplicationDefaultCredentials();
```
patflynn: we probably want to minimize the window between when we create the credential file and we execute.  Could we put this this in executeProvess?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineAction.java
```diff
@@ -0,0 +1,155 @@
...
+  private static final String GCLOUD_USER_TYPE = "authorized_user";
+
+  @VisibleForTesting
+  protected File createApplicationDefaultCredentials() {
+    CredentialedUser projectUser = Services.getLoginService().getAllUsers()
+        .get(appEngineHelper.getGoogleUsername());
+
+    GoogleLoginState googleLoginState;
+    if (projectUser != null) {
+      googleLoginState = projectUser
```
patflynn: this can fit on one line.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineAction.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  private static final String CLIENT_ID_LABEL = "client_id";
+  private static final String CLIENT_SECRET_LABEL = "client_secret";
+  private static final String REFRESH_TOKEN_LABEL = "refresh_token";
+  private static final String GCLOUD_USER_TYPE_LABEL = "type";
+  private static final String GCLOUD_USER_TYPE = "authorized_user";
+
+  @VisibleForTesting
+  protected File createApplicationDefaultCredentials() {
```
patflynn: @Nullable

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeployAction.java
```diff
@@ -0,0 +1,248 @@
...
+                int doStop = Messages
+                    .showYesNoDialog(
+                        GctBundle.message("appengine.stop.modules.version.confirmation.message"),
+                        GctBundle.message("appengine.stop.modules.version.confirmation.title"),
+                        "OK", "Cancel", General.Warning);
+
+                if (doStop == Messages.YES) {
+                  stop(callback);
+                } else {
+                  callback.errorOccurred("Undeploy Canceled");
```
patflynn: so this leaves it in a state where you can still click 'undeploy'?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeployAction.java
```diff
@@ -0,0 +1,248 @@
...
+                }
+              }
+            });
+          }
+
+          private void stop(@NotNull UndeploymentTaskCallback callback) {
+            Set<String> modulesToSop;
+            try {
+              modulesToSop = parseDeployOutputToModuleList(deployOutput.toString());
```
patflynn: Can you file a bug to update this code to use the module name configured (once we have module name configuration)?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeployAction.java
etanshaul: yes. specifically invoking errorOccured is necessary for that.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeployAction.java
```diff
@@ -0,0 +1,248 @@
...
+    public void onTextAvailable(ProcessEvent event, Key outputType) {
+      if(outputType.equals(ProcessOutputTypes.STDOUT)) {
+        deployOutput.append(event.getText());
+      }
+    }
+
+    @Override
+    public void processTerminated(final ProcessEvent event) {
+      if (event.getExitCode() == 0) {
+        callback.succeeded(new DeploymentRuntime() {
```
patflynn: can we define this DeploymentRuntime somewhere else?  It distracts from the flow.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeployAction.java
```diff
@@ -0,0 +1,248 @@
...
+                  }
+                });
+          }
+        });
+
+        SwingUtilities.invokeLater(new Runnable() {
+          @Override
+          public void run() {
+            Messages.showMessageDialog(
+                GctBundle.message("appengine.deployment.success.dialog.message"),
```
patflynn: does the user have to dismiss this?  I think it's kind of not IntelliJ style to have a dialog for a successful run configuration

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineStopAction.java
```diff
@@ -0,0 +1,109 @@
...
+                General.Information);
+          }
+        });
+      } else {
+        logger.error("Application stop process exited with an error. Exit Code:" + event.getExitCode());
+        callback.errorOccurred(
+            GctBundle.message("appengine.stop.modules.version.execution.error.with.code", event.getExitCode()));
+      }
+
+      deleteCredentials();
```
patflynn: I'm wondering if we want to include deleteCredentials in a finally clause to ensure it always happens.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeployAction.java
patflynn: ok. 'Undeploy Cancelled' should be localized. 

Also did you look at showOkCancelDialog?

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeployAction.java
etanshaul: Yes, the user has to dismiss it. The reason I created this one is to keep it consistent with the stop flow which now has a "successful stop" dialog. This dialog is not completely necessary since the user can see a message of succesfull deployment in the console..

patflynn: Yeah.  I can see why we added it to the stop flow. But I'd prefer we left
it out of a successful deployment.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeployAction.java
etanshaul: Sure. I think updating this code will be part of whatever PR includes module name configuration.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeployAction.java
etanshaul: Sounds good to me.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineStopAction.java
etanshaul: I think that makes sense

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeployAction.java
meltsufin: Given that we now also have a confirmation before undeploy action, maybe we don't need another one after, if the action is successful. That way neither deploy nor undeploy would have a success confirmation dialog.

etanshaul: That idea also has the nice benefit of getting rid of "useless" informational dialogs.

etanshaul:
Comments addressed.

I left the undeployment success info dialog since I'm on the fence about it. Without it, the user has no indication that the stop was successful.

meltsufin:
LGTM

patflynn:
LGTM>

PR #583, opened 2016-04-04
don't pass app.yaml flag for custom deployments when not specified

### LINKED ISSUE #582: Don't pass app.yaml flag to gcloud when custom with no selected app.yaml

etanshaul:
fixes #582


patflynn:
LGTM

PR #585, opened 2016-04-04
allow user to specify custom version id for flex deployments

### LINKED ISSUE #572: Allow users to enter custom versions and module name for flex deployments

etanshaul:
fixes #572 

Version override feature is implemented:

image

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
@@ -19,6 +19,7 @@ package com.google.gct.idea.appengine.cloud;
 import com.google.common.base.Supplier;
 import com.google.gct.idea.appengine.cloud.AppEngineCloudType.AppEngineDeploymentConfigurator.UserSpecifiedPathDeploymentSource;
 import com.google.gct.idea.appengine.cloud.AppEngineDeploymentConfiguration.ConfigType;
+import com.google.gct.idea.ui.PlaceholderTextField;
 import com.google.gct.idea.ui.BrowserOpeningHyperLinkListener;
 import com.google.gct.idea.appengine.cloud.FileConfirmationDialog.DialogType;
 import com.google.gct.idea.util.GctBundle;
@@ -45,12 +46,15 @@ import com.intellij.remoteServer.configuration.deployment.DeploymentSource;
 import com.intellij.ui.DocumentAdapter;
 import com.intellij.ui.awt.RelativePoint;
 
+import org.apache.commons.lang.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.awt.Point;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.awt.event.ItemEvent;
+import java.awt.event.ItemListener;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.io.File;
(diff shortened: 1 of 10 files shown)


etanshaul:
Just realized, I need to implement the placeholder text feature (something like 'auto' when there is no version override).

patflynn:
yeah I would put something like 'automatically generated', since we have
lots of horizontal space to work with.

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -187,6 +199,7 @@ protected void applyEditorTo(AppEngineDeploymentConfiguration configuration)
     configuration.setDockerFilePath(dockerFilePathField.getText());
     configuration.setAppYamlPath(appYamlPathField.getText());
     configuration.setConfigType(getConfigType());
+    configuration.setVersion(versionOverrideCheckBox.isSelected() ? versionIdField.getText() : null);
```
chanseokoh: Should there be validation to exclude, e.g., white spaces, special characters?

etanshaul: yes, I'll add a regex. The pattern is printed out from gcloud when an invalid version string is sent.

etanshaul:
Updated with placeholder text (I was working on creating a reusable JTextField subclass that adds placeholder functionality. Putting a "hardcoded" version out for review in the meantime):

image

Regarding @chanseokoh's point about validation: wouldn't it be best to just let it pass through to gcloud and let it do the validation? This minimizes client logic that we need to maintain (if the validation pattern is ever changed for instance).

chanseokoh:
About validation: that's a nice idea. That said, I just notice that --version is wrapped in an unusual way in the actual command:

--quiet "--version=some name" --format=json app.yaml

which results in bizarre error message. It will be better to be --version="some name". I guess it's doing automatic escaping then?

etanshaul:
Updated placeholder logic to use a barebones JTextField subclass.

etanshaul:
@chanseokoh what error are you seeing? The placement of the version id hasn't changed. Could it be that you were using an invalid version pattern? I don't think whitespace is allowed.

chanseokoh:
I just tried an unacceptable name with a space for the sake of testing. The message is

Fetching storage object: gs://staging.springboot-maven-project.appspot.com/us.gcr.io/springboot-maven-project/default.some name:latest#1459799851639000
CommandException: No URLs matched: gs://staging.springboot-maven-project.appspot.com/us.gcr.io/springboot-maven-project/default.some
ERROR
ERROR: fetching gcs source: running gsutil: exit status 1
ERROR: (gcloud.preview.app.deploy) Error Response: [2] Build failed; check build logs for details

etanshaul:
Ah. I guess its GeneralCommandLine that's doing that escaping. If you try with another invalid version id, e.g. MyId (caps aren't allowed), you'll get a better error message. I'll think about a solution - ideally we wouldn't duplicate the validation that gcloud already does.

chanseokoh:
I think we are fine with white space. I tested with gcloud with --version="some name" and it returned the same weird message. We don't want to duplicate validation.

etanshaul:
Updated:
- fixed bug discussed offline with @chanseokoh where serialized state was not being restored
- added version validation: if override is selected but version is empty, it warns the user (and prevents deploy)

@chanseokoh @patflynn ptal

google-cloud-tools-plugin/src/com/google/gct/idea/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -168,15 +177,22 @@ public File get() {
                 .defaultDockerfile(DeploymentArtifactType.typeForPath(deploymentSource.getFile()));
           }
         }, dockerFilePathField, userSpecifiedArtifactFileSelector));
+    versionOverrideCheckBox.addItemListener(
+        new CustomFieldOverrideListener( versionOverrideCheckBox, versionIdField));
```
patflynn: there's an errant white space before versionOverrideCheckBox

patflynn:
LGTM

chanseokoh:
The bug is gone. LGTM.

PR #587, opened 2016-04-04
Show IO errors in an error dialog

### LINKED ISSUE #574: Error handling for config file generation ('Generate...' button in Flex deploy dialog)
Errors are currently not well-handled.

- If a user just hit enter or click 'Ok' (will be 'Generate' after #570) in the directory selection window, it will create a file in the following directory. (The confirm dialog below will be after #570.)

empty_dir

- If a user provides a directory with no write permission (e.g., '/' root directory), nothings seems to happen except the following red error pop-up appears at the top-right corner. Seems like this is due to throwing RuntimeError whenever any IOException occurs.

> IDE internal error occurred.
> Click to see details and submit a bug report.

chanseokoh:
Fixes #574.

Since it is IOExceptions, I think it is okay to display any of them in a dialog.

ioerror_dialog

### DIFF
--- a/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
+++ b/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
@@ -38,7 +38,7 @@ appengine.flex.config.destination.overwrite.title=Confirm Overwrite
 appengine.flex.config.destination.not.directory=The path is not a directory. Please choose a directory.
 appengine.flex.config.destination.not.directory.title=Error
 appengine.flex.config.destination.chooser.directory.missing=Please select a directory.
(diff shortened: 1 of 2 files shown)


patflynn:
More info about what context the error is associated with would be good. 

Something like "There was an error generating the app.yaml.  Error details: .."

chanseokoh:
Good suggestion.

ioerror_dialog

patflynn:
thanks for the screenshot!

Looking at it, I think it would look better as:

"
There was an error generating the Dockerfile.

Error reason: /Dockerfile(permission denied)
"

chanseokoh:
You're welcome! I can't figure out how to add a trailing space in .properties values, so no space between 'reason:' and an error message like below. Seems not a big deal though.

ioerror_dialog

patflynn:
LGTM

PR #592, opened 2016-04-04
Rename com.google.gct.idea to com.google.cloud.tools.intellij

patflynn:
@loosebazooka might want to take a look.


etanshaul:
lgtm

PR #603, opened 2016-04-06
fix datetime pattern for version generation

### LINKED ISSUE #596: Generated version time stamp for App Engine deployment does not match the the Cloud SDK
capture

etanshaul:
fixes #596 

Updated the pattern to use 2 digit versions for hours, min, seconds. Thinking about ways to create non-flaky unit tests for this.

### DIFF
--- a/google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/util/AppEngineUtilTest.java
+++ b/google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/util/AppEngineUtilTest.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (C) 2016 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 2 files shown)


patflynn:
lgtm

etanshaul:
added a couple tests

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/util/AppEngineUtilTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.junit.Test;
+
+/**
+ * Tests App Engine Utilities
+ * Uses an external trusted datetime to milli converter to set the datetime to the desired value.
+ */
+public class AppEngineUtilTest {
+
+  @Test
+  public void testDateTimeToVersion() {
```
patflynn: Typically the way we handle this internally is by using a clock class internally that either uses real time or is mocked ofr tests.

etanshaul: Could you explain a bit? If I use a clock internally with real time then the test feels a bit flaky to me because it depends on when the test is run. If the clock is set to a preconfigured datetime then the problem is circular because it would require using a formatter to set it (which is what we are testing in the first place). I tried mocking DateTime but didn't have much luck with that approach.

patflynn: To make this work properly I think you'll have to make an
AppEngineVersionGenerator(Clock)  where for tests you can provide a mock
Clock that returns the instant that specifies some fixed point in time you
can test against.  Or actually I think you can pass a fixed time Clock as
well.

etanshaul: updated

patflynn:
LGTM. We can start using Clock when we move to Java 8, or introduce our own should it become a problem.

PR #604, opened 2016-04-06
Fix issue with storing and retrieving OAuth data

### LINKED ISSUE #601: Corrupted saved credentials cause false logged in state
The error I get in the IntelliJ Event Log is:

> IllegalArgumentException: Argument for @NotNull parameter 'userEmail' of com/google/cloud/tools/intellij/login/CredentialedUserRoster.setActiveUser must not be null

It appears that the Account Plugin saved a corrupted authentication state where everything is null except for the refreshToken. The end result is that the account plugin thinks the user is logged in, but it causes failures because it doesn't have all of the needed user information.

meltsufin:
Fixes #601 

When the user's email address is null, we need to be careful with storing and reading the saved OAuth data because it's keyed on the email address.

I'm still not sure how I ended up with OAuth data stored in Preferences keyed on a null email address in the first place. However, with these changes it should never happen because we will always check that the email address is not null before storing OAuth data.

Also, this change ensures that we only try to read OAuth data when a user is logged in and hence has an associated email address. So, even with the corrupted prefs.xml file, everything should be fine after this change. I tested with the bad prefs.xml file referenced in the issue.

### DIFF
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/login/GoogleLoginPrefs.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/login/GoogleLoginPrefs.java
@@ -58,22 +58,26 @@ public class GoogleLoginPrefs {
    * @param credentials the specified {@code Credentials object}
    */
   public static void saveOAuthData(OAuthData credentials) {
-    Preferences prefs = getPrefs();
     String userEmail = credentials.getStoredEmail();
-    prefs.put(getCustomUserKey(OAUTH_DATA_REFRESH_TOKEN_KEY, userEmail), credentials.getRefreshToken());
 
-    // we save the scopes so that if the user updates the plugin and the
-    // scopes change, we can force the plugin to log out.
-    Joiner joiner = Joiner.on(DELIMITER);
-    prefs.put(getCustomUserKey(OAUTH_SCOPES_KEY, userEmail), joiner.join(credentials.getStoredScopes()));
+    if (userEmail != null) {
+      Preferences prefs = getPrefs();
 
-    String storedEmail = credentials.getStoredEmail();
-    if (storedEmail != null) {
-      prefs.put(getCustomUserKey(OAUTH_DATA_EMAIL_KEY, userEmail), storedEmail);
+      prefs.put(getCustomUserKey(OAUTH_DATA_REFRESH_TOKEN_KEY, userEmail),
+          credentials.getRefreshToken());
+
+      // we save the scopes so that if the user updates the plugin and the
+      // scopes change, we can force the plugin to log out.
+      Joiner joiner = Joiner.on(DELIMITER);
+      prefs.put(getCustomUserKey(OAUTH_SCOPES_KEY, userEmail),
+          joiner.join(credentials.getStoredScopes()));
+
+      prefs.put(getCustomUserKey(OAUTH_DATA_EMAIL_KEY, userEmail), userEmail);
+
+      addUser(credentials.getStoredEmail());
+
+      flushPrefs(prefs);
     }
-    addUser(credentials.getStoredEmail());
-
-    flushPrefs(prefs);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


etanshaul:
Trying to wrap my head around this one: So after these changes, if the user reproduces the original issue, they will be just be logged out with no errors? In other words, if the user is logged in and their auth settings are corrupted (somehow) then what exactly happens?

meltsufin:
After these changes it would not be possible to reproduce the original issue because we're now guarding against saving invalid OAuth data. So, we will never be in a situation where we will be attempting to load invalid OAuth data.

I'm still not entirely sure about the situation which causes the auth tokens to be set while the user email is null. I suspect that happens when the user is logged out for a reason like session timeout and cleanup is not done fully. In that situation the only thing that will be different is that these invalid OAuth session settings will no be saved. The login status will remain unchanged.

google-account-plugin/src/com/google/cloud/tools/intellij/login/GoogleLoginPrefs.java
```diff
@@ -82,17 +86,23 @@ public static void saveOAuthData(OAuthData credentials) {
...
-    String storedEmail = prefs.get(getCustomUserKey(OAUTH_DATA_EMAIL_KEY), null);
-    String storedScopesString = prefs.get(getCustomUserKey(OAUTH_SCOPES_KEY), "");
+    if (Services.getLoginService().getActiveUser() != null) {
+      Preferences prefs = getPrefs();
 
-    // Use a set to ensure uniqueness.
-    SortedSet<String> storedScopes = new TreeSet<String>();
-    for (String scope : storedScopesString.split(DELIMITER)) {
-      storedScopes.add(scope);
+      refreshToken = prefs.get(getCustomUserKey(OAUTH_DATA_REFRESH_TOKEN_KEY), null);
```
patflynn: My guess is a bunch of this code is the way it is because the original credential store did not support multiple users and this tries to support both the schema with just one user, and the one with a list of users and an active user.

google-account-plugin/src/com/google/cloud/tools/intellij/login/GoogleLoginPrefs.java
```diff
@@ -82,17 +86,23 @@ public static void saveOAuthData(OAuthData credentials) {
...
   public static OAuthData loadOAuthData() {
-    Preferences prefs = getPrefs();
+    String refreshToken = null;
+    String storedEmail = null;
+    SortedSet<String> storedScopes = new TreeSet<String>();
 
-    String refreshToken = prefs.get(getCustomUserKey(OAUTH_DATA_REFRESH_TOKEN_KEY), null);
-    String storedEmail = prefs.get(getCustomUserKey(OAUTH_DATA_EMAIL_KEY), null);
-    String storedScopesString = prefs.get(getCustomUserKey(OAUTH_SCOPES_KEY), "");
+    if (Services.getLoginService().getActiveUser() != null) {
```
patflynn: one of the big problems in the design of this library is the two sources of state of active user.

google-account-plugin/src/com/google/cloud/tools/intellij/login/GoogleLoginPrefs.java
```diff
@@ -82,17 +86,23 @@ public static void saveOAuthData(OAuthData credentials) {
    * an {@code OAuthData} object all of whose getters return {@code null} .
    */
   public static OAuthData loadOAuthData() {
-    Preferences prefs = getPrefs();
+    String refreshToken = null;
+    String storedEmail = null;
+    SortedSet<String> storedScopes = new TreeSet<String>();
```
patflynn: looking at the javadoc it seems to indicate that all getters should be null when there is no active user.

patflynn: I guess the old code returns the empty set, so it makes sense not to change  that behavior.

patflynn:
LGTM

Issue #605, opened 2016-04-06
Logger.error() for failures that are not bugs


chanseokoh:
Our plugin often uses Logger.error() when there are failures but not bugs. For example, if 'gcloud deploy' fails with an exit code other than 0, Logger.error() will cause a small red popup saying something like 'IDE internal error occurred. Please file a bug.' The logged error can also be accessible in the event log pane. If we click the logged event, a dialog below appears where there is a button saying 'Report to Google Cloud Platform':

![report_bug](https://cloud.githubusercontent.com/assets/10523105/14328202/ac2ad65c-fc03-11e5-8203-3d172585c8ee.png)

We need to check for all the usage of Logger.error() and use it only when it is truly a bug of our plugin.

patflynn:
Yes this is an issue. It would be nice to have analytics on how often failures occurred. But many errors will be of a transient nature that we can't control and shouldn't be reported as plugin bugs.

chanseokoh:
Adding a note: an non-zero exit code seems to be returned whenever a deployment failed for whatever reason, for example, when a user provided an unacceptable version name (e.g., special characters). Our general policy is to delegate validation to gcloud and have it deal with logic errors, so we need to think about whether we want to have analytics like in this case where there can be a lot of noises.

chanseokoh:
This also happens when throwing RuntimeException for these kinds of failures, which we currently do in many places.

I wonder if we can have a consensus about a nice and uniform way of informing failures to users, although it should depend on contexts. Event logs seem to be a good way for a concurrent task, but it is weird that Logger.error() actually means IDE/plugin bugs.

chanseokoh:
Throwing an exception seems really convenient, because often we want to get out of anywhere. I wonder if there is a suitable exception we can throw that Intellij will handle nicely to show our custom error message to users.

chanseokoh:
Anyway, @etanshaul may have different opinions about these RuntimeExceptions and Logger.error() behaviors, so we should discuss!

patflynn:
logging error is an effective way to trigger error reporting. So IMO it's
sensible to use it if you've anticipated a problem that would be a bug
you'd like reported. Blowing up with an unhandled exception has the same
effect, but is more likely to be the result of something completely
unanticipated.

For errors that are not plugin bugs and that the user can address, we
should definitely not be using log.error or throwing uncaught exceptions.

etanshaul:
[...]

PR #609, opened 2016-04-06
Prevent empty paths to app.yaml and Dockerfile

### LINKED ISSUE #607: 'gcloud' is now auto-generating app.yaml and Dockerfile if not present - require them for custom deployments
Check it out. Run the 'gcloud app deploy' in a directory where there is only one war. If either one or both of them does not exist, it will create them.

We don't need to manually copy our templates.

chanseokoh:
Fixes #607.


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -250,6 +250,12 @@ private void validateConfiguration() throws ConfigurationException {
           GctBundle.message("appengine.config.deployment.source.error"));
     } else if(versionOverrideCheckBox.isSelected() && StringUtils.isBlank(versionIdField.getText())) {
       throw new ConfigurationException(GctBundle.message("appengine.config.version.error"));
+    } else if (getConfigType() == ConfigType.CUSTOM) {
+      if (StringUtils.isBlank(appYamlPathField.getText()) ||
```
etanshaul: We can probably separate out these error messages to show one for a missing Dockerfile and one for a missing app.yaml.

chanseokoh: Expected that feedback. I'll do it.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -44,7 +44,8 @@ appengine.flex.config.generation.io.error=There was an error generating the {0}.
 appengine.config.project.panel.title=Project: {0}
 appengine.flex.config.user.specified.artifact.title=Select a JAR or WAR file
 appengine.flex.config.user.specified.artifact.error=Browse to a JAR or WAR file.
-appengine.flex.config.custom.deploy.title=App Engine Flexible Environment Configuration.
+appengine.flex.config.custom.deploy.title=App Engine Flexible Environment Configuration
+appengine.flex.config.custom.configfiles.error=Provide app.yaml and Dockerfile.
```
etanshaul: To keep it consistent with the jar/war error message, how about:

"Browse to an app.yaml file."
"Browse to a Dockerfile."

etanshaul:
I'd also take a look at this closed PR: #555 

I made some changes there specifically to allow for empty Dockerfile / app.yaml to pass through to gcloud. Given that we are now requiring them, some of that logic might be able to be changed back (making the Dockerfile / app.yaml params @Nullable, for instance, may not make sense anymore).

chanseokoh:
Not sure, but eventually we will allow leaving the fields blank and even remove the 'Generate...' buttons? In that case, it seems good to have the logic in #555. Otherwise, we may forget to do #555 again when we remove this validation.

etanshaul:
I think that even if we later change it to allow blanks we should update the rest of the code to work with the current logic to keep things cohesive. 

Any time we modify that UI we should also be thinking about how it affects the deployment logic, but I take your point: maybe add a small comment in the UI code as a reminder?

chanseokoh:
I looked at #555 more closely. I think it is necessary to restore the logic to check existence of user-provided paths. Currently, even if we add this validation, if a user enters a non-existing path, we silently ignore that error which will make gcloud auto-generate the files. I'll restore the logic.

chanseokoh:
I reverted the changes in #555.

However, throwing RuntimeException does sound quite right (much like #605). There are other places where we throw RuntimeException, and I think they should be handled differently. There should be another PL for these changes.

chanseokoh:
does sound --> doesn't sound

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
```diff
@@ -361,12 +361,12 @@ public void computeDeployments(@NotNull ComputeDeploymentsCallback callback) {
     public void disconnect() {
     }
 
-    @Nullable
+    @NotNull
     private File getFileFromFilePath(String filePath) {
```
etanshaul: I think there's some changes higher up in this class where this method is invoked that also need to be changed alongside this.

etanshaul:
Yea - I think the RuntimeExceptions can be treated on a case by case basis. I'm not entirely sure if its appropriate here or not. I need to look at it more closely.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
chanseokoh: Thanks for catching that!

etanshaul:
lgtm

PR #614, opened 2016-04-08
Present default target dir for app.yaml/Dockerfile

### LINKED ISSUE #575: Generated app.yaml and Dockerfile should have a sensible default target directory
assuming it's a Maven shaped project structure, the app.yaml should go in:
src/main/appengine

and the Dockerfile should go in:
src/main/docker

chanseokoh:
Fixes #575.

Simple logic to check existence of 'src/main/appengine', 'src/main/docker', 'src/main', and 'src'. If none of these directories exists, the base project directory is presented.

I noticed these default paths tend to be long under Unix, so I rearranged the path field and made the file chooser dialog a bit larger.

def_dest_dir


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/SelectConfigDestinationFolderDialog.java
```diff
@@ -53,6 +55,25 @@ public SelectConfigDestinationFolderDialog(@Nullable Project project) {
...
         FileChooserDescriptorFactory.createSingleFolderDescriptor()
     );
+
+    if (project != null && project.getBasePath() != null) {
+      String srcPath = project.getBasePath() + "/src";
+      String srcMainPath = project.getBasePath() + "/src/main";
+      String AppenginePath = project.getBasePath() + "/src/main/appengine";
+      String dockerPath = project.getBasePath() + "/src/main/docker";
+
+      if (fileType == ConfigFileType.APP_YAML && new File(AppenginePath).isDirectory()) {
```
etanshaul: I wonder if it makes sense to default it to the canonical directories even if they don't already exist? In this case it should prompt the user to create them (I think this feature was added previously).

chanseokoh: That's also nice. @patflynn?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/SelectConfigDestinationFolderDialog.java
```diff
@@ -53,6 +55,25 @@ public SelectConfigDestinationFolderDialog(@Nullable Project project) {
         project,
         FileChooserDescriptorFactory.createSingleFolderDescriptor()
     );
+
+    if (project != null && project.getBasePath() != null) {
+      String srcPath = project.getBasePath() + "/src";
+      String srcMainPath = project.getBasePath() + "/src/main";
+      String AppenginePath = project.getBasePath() + "/src/main/appengine";
```
meltsufin: start variable name with a lower-case letter even if the word is normally capitalized (appEnginePath)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -16,6 +16,8 @@
 
 package com.google.cloud.tools.intellij.appengine.cloud;
 
+import com.google.common.base.Supplier;
```
meltsufin: Duplicate import causes PMD failure

chanseokoh: Doh! Made a bad conflict merge. Thanks.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/SelectConfigDestinationFolderDialog.java
chanseokoh: Fixed.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/SelectConfigDestinationFolderDialog.java
meltsufin: :+1: on Etan's comment

chanseokoh:
Closes #616 too.

meltsufin:
I'm wondering if the browse for existing app.yaml and Docker file should also default to existing files if we can find them in the canonical locations. Right now it seems to default to wherever you browsed last.

chanseokoh:
That's a good idea too. However, already selected and existing files should have priority. For this, I will create another PR.

meltsufin:
OK, I'll create another issue to track this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/SelectConfigDestinationFolderDialog.java
patflynn: Just wanted to +1 Etan's suggestion. That's how I saw it working. We want
the user to be guided to placing the files in the standard location.

meltsufin:
Here it is: #617

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/SelectConfigDestinationFolderDialog.java
chanseokoh: This simplifies things. Now presenting canonical paths all the time. Even for a non-existing directories, users will be able to create them.

meltsufin:
:ship:

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/SelectConfigDestinationFolderDialog.java
```diff
@@ -53,6 +55,17 @@ public SelectConfigDestinationFolderDialog(@Nullable Project project) {
         project,
         FileChooserDescriptorFactory.createSingleFolderDescriptor()
     );
+
+    // Present a canonical target folder as default in the path field.
+    if (project != null && project.getBasePath() != null) {
+      if (fileType == ConfigFileType.APP_YAML) {
+        destinationFolderChooser.setText(project.getBasePath() + "/src/main/appengine");
```
patflynn: this is kind of fragile. multi-module projects and projects that don't use the maven directory structure won't like this..This is actually a tricky problem because for the artifact types we support we don't necessarily know what module it's related to so we can't query for the src paths.

I still think it's better than no suggestion, and it gives the user an indication of where they should be. 

:ship:  LGTM

etanshaul: One possible future enhancement would be to detect maven/gradle facets already attached to the project and default the directories based on that. If no information can be learned about the project then we default to the root of the project.

PR #615, opened 2016-04-08
Disconnect on server instance stops deployment [See #630]

### LINKED ISSUE #581: App Engine deploy should support cancelling the deployment

### LINKED ISSUE #588: Closing project during deployment doesn't stop the deployment
When a user tries to close the project while a deployment is in progress, there is no warning, and the project closes, while the deployment process is still executing in the background.

meltsufin:
Fixes #588 and #581.

The process that executes the deployment is tracked and killed when a disconnect is invoked on the App Engine Server Instance.

The disconnect is also called when:

1. The project is closing (Fixes #588)
2. A new deployment is started while another one is in progress
3. The stop action is invoked for the App Engine server in the UI. (Fixes #581)

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
@@ -29,6 +29,8 @@ import com.intellij.execution.ExecutionException;
 import com.intellij.execution.configurations.GeneralCommandLine;
 import com.intellij.execution.configurations.GeneralCommandLine.ParentEnvironmentType;
 import com.intellij.execution.process.OSProcessHandler;
+import com.intellij.execution.process.ProcessAdapter;
+import com.intellij.execution.process.ProcessEvent;
 import com.intellij.execution.process.ProcessHandler;
 import com.intellij.execution.process.ProcessListener;
 import com.intellij.openapi.diagnostic.Logger;
@@ -55,6 +57,7 @@ public abstract class AppEngineAction implements Runnable {
   private AppEngineHelper appEngineHelper;
   private File credentialsPath;
   private RemoteOperationCallback callback;
+  private ProcessHandler processHandler;
 
   public AppEngineAction(
       @NotNull LoggingHandler loggingHandler,
@@ -65,13 +68,28 @@ public abstract class AppEngineAction implements Runnable {
     this.callback = callback;
   }
 
+  /**
+   * Returns the commandline process handler used to execute the action
+   *
+   * @return the process handler, or null, if the action is not executing
+   */
+  public ProcessHandler getProcessHandler() {
+    return processHandler;
+  }
+
   protected LoggingHandler getLoggingHandler() {
     return loggingHandler;
   }
 
-  protected void executeProcess(
+  protected synchronized void executeProcess(
       @NotNull GeneralCommandLine commandLine,
       @NotNull ProcessListener listener) throws ExecutionException {
+
+    // kill action process if one is already executing
+    if (processHandler != null) {
+      processHandler.destroyProcess();
+      processHandler = null;
+    }
 
     credentialsPath = createApplicationDefaultCredentials();
     if (credentialsPath == null) {
@@ -91,11 +109,20 @@ public abstract class AppEngineAction implements Runnable {
 
     consoleLogLn("Executing: " + commandLine.getCommandLineString());
 
-    final Process process = commandLine.createProcess();
-    final ProcessHandler processHandler = new OSProcessHandler(process,
+    Process process = commandLine.createProcess();
+    processHandler = new OSProcessHandler(process,
         commandLine.getCommandLineString());
     loggingHandler.attachToProcess(processHandler);
     processHandler.addProcessListener(listener);
+
+    // mark process as completed by setting handler to null when terminated
+    processHandler.addProcessListener(new ProcessAdapter() {
+      @Override
+      public void processTerminated(ProcessEvent event) {
+        processHandler = null;
+      }
+    });
+
     processHandler.startNotify();
   }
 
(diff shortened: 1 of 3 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
```diff
@@ -65,14 +68,29 @@ public AppEngineAction(
...
+  public ProcessHandler getProcessHandler() {
+    return processHandler;
+  }
+
   protected LoggingHandler getLoggingHandler() {
     return loggingHandler;
   }
 
-  protected void executeProcess(
+  protected synchronized void executeProcess(
```
etanshaul: executeProcess gets called from a Runnable's run() method. Do we need to synchronize this?

meltsufin: I think you're right, this is not needed because we have synchronized on AppEngineRuntimeInstance.deploy() and .disconnect() where the lifecycle of the action is managed.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
```diff
@@ -316,16 +316,21 @@ public void connect(@NotNull ConnectionCallback<AppEngineDeploymentConfiguration
...
     private AppEngineServerConfiguration configuration;
+    private AppEngineAction currentAction;
 
     public AppEngineRuntimeInstance(
         AppEngineServerConfiguration configuration) {
       this.configuration = configuration;
     }
 
     @Override
-    public void deploy(@NotNull final DeploymentTask<AppEngineDeploymentConfiguration> task,
```
patflynn: isn't an AppEngineRuntimeInstance scoped to a deployment? is synchronization necessary?

chanseokoh: That's a hard question. I also wonder what is an answer.

meltsufin: There is the "deploy/redeploy" button always there. So, I guess the answer is 'No'.

patflynn: OK, so if it's not scoped to a specific deployment action, is synchronization still the right thing to do? Won't that just block the next deployment?

meltsufin: It shouldn't block anything for too long because the actual deployment happens in a background process.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
```diff
@@ -377,7 +385,12 @@ public void computeDeployments(@NotNull ComputeDeploymentsCallback callback) {
     }
 
     @Override
-    public void disconnect() {
+    public synchronized void disconnect() {
+      // kill any executing process for the current action
+      if (currentAction != null && currentAction.getProcessHandler() != null) {
+        currentAction.getProcessHandler().destroyProcess();
```
patflynn: could you just add a cancel method to currentAction?

meltsufin: Yes, that would actually be much cleaner!

patflynn: is destroyProcess sigkill?  Don't we want to sigint?

meltsufin: Yes, and yes. I'll look into how to trigger sigint instead and whether it actually will work.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
```diff
@@ -171,6 +171,9 @@ public void processTerminated(final ProcessEvent event) {
       try {
         if (event.getExitCode() == 0) {
           callback.succeeded(new DeploymentRuntimeImpl(deploymentOutput.toString(), version));
+        } else if (event.getExitCode() == 143) {
```
meltsufin: This needs to be tested on Windows to see if exit code is the same.

meltsufin:
I think I've addressed all of the comments, but I want to test on Windows before merging.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
meltsufin: It looks like the exit code on Windows is always 1 when the process terminates. So, we can't rely on the exit code being anything other than 0 or >0.

meltsufin:
Now the soft kill of the deployment process works on Windows also, without showing errors in the IDE.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
```diff
@@ -171,6 +171,9 @@ public void processTerminated(final ProcessEvent event) {
       try {
         if (event.getExitCode() == 0) {
           callback.succeeded(new DeploymentRuntimeImpl(deploymentOutput.toString(), version));
+        } else if (cancelled) {
+          // process cancelled (message should never be seen by user)
+          callback.errorOccurred("Deployment process was cancelled.");
```
patflynn: should this be localized?

meltsufin: The user will never see this message as far as *this* pull request is concerned. However, in #621, it can actually be seen by the user and is localized there.

meltsufin:
Replaced by #630

PR #619, opened 2016-04-11
Use non-capitalized 'flexible environment'

### LINKED ISSUE #608: Flexible Environment should be 'flexible environment'
don't capitalize the App Engine environments.

chanseokoh:
Fixes #608.

After this, there won't be any string starting with 'Flex' in our entire project except for Java method names or alike.


etanshaul:
lgtm

PR #621, opened 2016-04-11
Adds toolbar button for cancelling GAE deployment [See #630]

### LINKED ISSUE #618: Add a toolbar button for cancelling deployment
Currently, when a cloud deployment is selected in the servers tree and is executing, there is no toolbar button to cancel it.

With #581 resolved, the user can cancel the deployment via the server disconnect button, but that require navigating up the tree to the server, away from the deployment node, which is not intuitive.

meltsufin:
Fixes #618.

Note that this builds on #615.

during_deployment

deployment_cancelled

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
@@ -7,6 +7,7 @@
  *
  *      http://www.apache.org/licenses/LICENSE-2.0
  *
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -62,7 +63,9 @@ import org.jetbrains.annotations.Nullable;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import javax.swing.Icon;
 
@@ -72,8 +75,6 @@ import javax.swing.Icon;
  * configurations for infrastructure based deployment flows.
  */
 public class AppEngineCloudType extends ServerType<AppEngineServerConfiguration> {
-
-  private AppEngineAction currentAction;
 
   public AppEngineCloudType() {
     super("gcp-app-engine"); // "google-app-engine" is used by the native IJ app engine support.
@@ -134,10 +135,6 @@ public class AppEngineCloudType extends ServerType<AppEngineServerConfiguration>
   public ServerConnector<?> createConnector(@NotNull AppEngineServerConfiguration configuration,
       @NotNull ServerTaskExecutor asyncTasksExecutor) {
     return new AppEngineServerConnector(configuration);
-  }
-
-  public AppEngineAction getCurrentAction() {
-    return currentAction;
   }
 
   protected static class AppEngineDeploymentConfigurator extends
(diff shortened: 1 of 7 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CancelDeploymentAction.java
```diff
@@ -0,0 +1,60 @@
...
+  @Override
+  protected boolean isEnabled4(DeploymentNode node) {
+    return getDeployment(node).getStatus() == DeploymentStatus.DEPLOYING;
+  }
+
+  @Override
+  protected void doActionPerformed(DeploymentNode node) {
+    ServerType serverType = getDeployment(node).getConnection().getServer().getType();
+    if (serverType instanceof AppEngineCloudType) {
+      ((AppEngineCloudType)serverType).getCurrentAction().cancel();
```
meltsufin: This is not very elegant, but I couldn't find another way to get access to the current action. So, I'm open to other ideas on how to accomplish this.

etanshaul:
Nice! I think it also makes sense to prompt the user for confirmation when clicking 'cancel' like we do with 'undeploy'.

meltsufin:
Good point! I'll add the confirmation dialog.

meltsufin:
Thinking more about the cancel confirmation dialog during deployment...
Should we have it pop up when the user disconnects from the server during deployment?
What about when the project is closing and there is an active deployment in process?

etanshaul:
For the first case (clicking cancel during deployment) - I think it makes sense here. It protects against accidental clicks.

For the second case (closing a project during an active deployment) - is it possible to halt the project from closing? Regardless, in this situation it doesn't feel right to me: if the user closes a project then we should silently cancel the deployment.

meltsufin:
In the second case, you can actually prevent the IDE from closing. I almost feel like the confirmation is more necessary there because the user may not be aware of the fact that the deployment is in progress and will be cancelled.

In general, is it a best practice to protect the user from accidental clicks in the IDE? I can definitely see it as being an annoyance for power users. What are the characteristics of actions that need to be guarded by a confirmation?

I remember encountering a principle somewhere that says that actions that can be undone, should not be confirmed. Do we want to follow that principle?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
```diff
@@ -312,84 +318,89 @@ public void connect(@NotNull ConnectionCallback<AppEngineDeploymentConfiguration
...
-            getFileFromFilePath(deploymentConfig.getDockerFilePath()),
-            deploymentConfig.getVersion(),
-            callback
-        );
-      }
+        // make sure we stop any active deployments
+        disconnect();
 
-      currentAction = deployAction;
+        FileDocumentManager.getInstance().saveAllDocuments();
```
patflynn: is this new indentation right?

meltsufin: It's because this is in a synchronized block now.

meltsufin:
The last commit allows multiple concurrent deployments to be running.

meltsufin:
I had to add AppEngineAction to the AppEngineServerConfiguration which isn't pretty, but that was the only way I was able to get to the action from the handler that's invoked when the "cancel deployment" toolbar button is clicked.

etanshaul:
@meltsufin was this still pending any changes, or were you waiting for lgtm's?

meltsufin:
This is ready for review. I would especially like someone to validate the whole approach with keeping a reference to the AppEngineAction in AppEngineServerConfiguration to allow access to the process from the toolbar action handler.

etanshaul:
can you update the branches so that it shows changes ready to merge into master so that its easier to review?

meltsufin:
Replaced by #630

patflynn:
Looking at this now!

meltsufin:
Look at PR #630, where I merged this in. :eyes:

Issue #450, opened 2016-02-18
Source context not being correctly configured during MVM Deployment


patflynn:
Apparently from a convo with Erez this should work now.

What I've found is you need to call the gcloud command to generate the src context directly, and I still was not able to get CDB to work.

More investigation is required to package the src ctxt correctly for a custom mvm deployment.

If we do a maven or gradle build as part of the deployment (in the 'run before' model), ideally this is when we'd generate the src context, so I can imagine it being a 'run before' task that we auto configure.

random other issue:  If deploy does eventually automatically just work, we still would not work because it it requires the deployment staging directory to be in the git project. Currently we create a staging directory in the user's tmp directory.

patflynn:
IMO, tackle this last as the experience works without it.

chanseokoh:
Does generating the src context files have to be associated with the deployment process? One could think of this outside the context of deployment. For example, our cloud plugin could always generate and add the src context files whenever building a WAR inside the IDE.

If we are going to associate this source-context-generation with deployment, then I think I would go with generating and adding the src context files to an already packaged WAR that is ready to be deployed, i.e., unpacking the WAR, generating the files, and re-packaging them in the staging area.

patflynn:
does the context need to be packaged into the war for it to work?

etanshaul:
https://cloud.google.com/debugger/docs/setting-up-java-on-app-engine explains that the source context file should go in WEB-INF so another question to answer is if there is a way for this to work with jar files.

patflynn:
That's the appcfg.sh flow.  I'm not sure that gcloud does the same.

chanseokoh:
I think it is how it works. For example, you would manually generate them anyway and put them under WEB-INF.

https://cloud.google.com/debugger/docs/setting-up-on-compute-engine#build_test_and_deploy_your_application

chanseokoh:
Okay, we will need to call 'gcloud preview app gen-repo-info-file' to generate the src context files at some point.

Initially, I was thinking I could do that just before calling 'gcloud preview app deploy', as part of our IntelliJ deployment process. @patflynn, you seem to have something else in mind. Can you elaborate on what you said last week, about the 'run before' (IDE/gradle/maven?) task?

Next, the only requirement is that those context files be in class path. One option is to inject those context files into a WAR (WEB-INF/classes) or a JAR (root directory) directly. Another option would be to place them somewhere in a VM and include the directly with CLASSPATH?

[...]

PR #627, opened 2016-04-13
Generate and add repo-info-files to WAR/JAR

chanseokoh:
This PR is meant for feedback/discussion in relation to #450.

Until 'gcloud' supports #450 (I hope it has a plan to support this like what it does for App Engine standard env), we will generate and add repository info files for Cloud Debugger on our side. (Debugger folks often call the files "source context files".)

I was thinking of directly adding the source context files into a WAR/JAR, using java.util.jar facilities. However, per our discussion a while ago, we may consider other ways to include these files in an app. PTAL at the PR, and we can come up with some ideas.


patflynn:
@znewman01 I had assumed that gcloud deploy was doing this for flex env too

znewman01:
No, at this point it does not. We rely on the gcloud-maven-plugin in most cases.

At some point in the future (possibly a long time), we hope to build support for this into the language-specific runtimes in a generic enough way that we don't wind up littering the `gcloud app deploy` source with `if runtime == 'java':` statements, which was the case before when we tried to support Java directly.

PR #628, opened 2016-04-13
Add tools menu link to flex deployment config

### LINKED ISSUE #612: It should be easier to discover our App Engine deployment support
One option would be to have a 'Deploy to App Engine flexible environment' action in the tool bar, that configures a cloud deployment for the user.

etanshaul:
fixes #612 

Using this as a starting point for discussing what features we want to support.

Adds a new menu item to the tools menu which currently links to the settings cloud configuration page:

screen shot 2016-04-13 at 2 38 52 pm

Clicking the link takes the user to the Cloud Configuration where a new flex env configuration can be made.

### DIFF
--- a/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
+++ b/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
@@ -17,7 +17,7 @@
 select.project.signin=Sign in with your Google account to list your Google Developers Console projects.
 
 cloud.project.label=Project\:
-
+appengine.flex.tools.menu.item.label=Deploy to App Engine flexible environment...
 appengine.cloudsdk.location.label=Cloud SDK directory\:
 appengine.cloudsdk.location.browse.window.title=Browse to Cloud SDK directory
 appengine.cloudsdk.location.missing.message=Please select a Cloud SDK home directory.
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/CloudToolsPluginInitializationComponent.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/CloudToolsPluginInitializationComponent.java
@@ -17,9 +17,14 @@ package com.google.cloud.tools.intellij;
 
 import com.google.cloud.tools.intellij.appengine.cloud.AppEngineCloudType;
 import com.google.cloud.tools.intellij.appengine.cloud.AppEngineCloudType.UserSpecifiedPathDeploymentSourceType;
+import com.google.cloud.tools.intellij.appengine.cloud.AppEngineToolsMenuAction;
 import com.google.cloud.tools.intellij.debugger.CloudDebugConfigType;
 
 import com.intellij.execution.configurations.ConfigurationType;
+import com.intellij.openapi.actionSystem.ActionManager;
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.actionSystem.Constraints;
+import com.intellij.openapi.actionSystem.DefaultActionGroup;
 import com.intellij.openapi.components.ApplicationComponent;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.remoteServer.ServerType;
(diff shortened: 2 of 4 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/CloudToolsPluginInitializationComponent.java
```diff
@@ -62,6 +67,13 @@ public void initComponent() {
           new UserSpecifiedPathDeploymentSourceType());
       pluginConfigurationService.registerExtension(ConfigurationType.CONFIGURATION_TYPE_EP,
           new DeployToServerConfigurationType(appEngineCloudType));
+
+      ActionManager actionManager = ActionManager.getInstance();
+      AnAction toolsMenuAction = new AppEngineToolsMenuAction();
+      actionManager.registerAction(AppEngineToolsMenuAction.ID , toolsMenuAction);
+      DefaultActionGroup toolsMenu =
+          (DefaultActionGroup) actionManager.getAction(AppEngineToolsMenuAction.GROUP_ID);
+      toolsMenu.add(toolsMenuAction, Constraints.LAST);
```
meltsufin: Can/should this be done decoratively though plugin.xml?

chanseokoh:
Cool. Can we say 'flex' instead of 'flexible'? It seems also good to put '...' at the end of the menu item for things that pop up new windows.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/CloudToolsPluginInitializationComponent.java
etanshaul: This way its scoped to the feature flag

meltsufin:
When you don't have an App Engine server already added, you don't really see anything App Engine  specific in the Cloud Configuration panel until the [+] button is clicked. Is it possible to automatically add one, if none are configured?

etanshaul:
Yes, I was working on adding that actually. I put this out there first so that we could discuss what is most useful. 

The approach that I was planning on would be that this button always loads the cloud configuration page with a new flex empty flex cloud displayed and ready to be filled out.

patflynn:
I think menu items should be an action phrase.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -17,7 +17,7 @@
 select.project.signin=Sign in with your Google account to list your Google Developers Console projects.
 
 cloud.project.label=Project\:
-
+appengine.flex.tools.menu.item.text=App Engine flex environment deployment
```
patflynn: I would be more precise with this label since we could have multiple menu items.
Also probably menu.tools since the 'tools' refers to a menu specific header.
And .text should be .label.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineToolsMenuAction.java
```diff
@@ -0,0 +1,50 @@
...
+        GctBundle.message("appengine.flex.tools.menu.item.text"),
+        GoogleCloudToolsIcons.APP_ENGINE);
+  }
+
+  @Override
+  public void actionPerformed(AnActionEvent e) {
+    Project project = e.getProject();
+
+    if(e.getProject() != null) {
+      ShowSettingsUtil.getInstance().showSettingsDialog(project, RemoteServerListConfigurable.class);
```
patflynn: what does this show? Don't we want to show an actual AppEngineDeploymentConfigurable?

etanshaul: Yes, about to push an update to load a blank Flex Cloud config.

etanshaul:
Updated:

image

Just noticed that the Cloud Config when opened this way bypasses form validation. Will update again.

etanshaul:
Fixed. When the user clicks the link in the tools menu (see screenshot in my last post), the following pops up:

image

patflynn:
Shouldn't it bring up the cloud deployment config?=

etanshaul:
The user may not have a Cloud configured yet. We could instead bring this up (is this what you are referring to?):

screen shot 2016-04-14 at 9 58 03 am

patflynn:
Yeah, I think this is the right hing to do. They can go to the servers
'...' to create one if they haven't selected a cloud already. I wonder if
there's a way to tie to bring up the last deployment run configuration if
it exists already.

etanshaul:
I will look into our options. I think its difficult to determine what the "latest" actually means to the user as they could have created several clouds each having its own deployment.

etanshaul:
ptal. Now the following deployment config window opens up. Here is what it looks like if no servers have been configured:

image

If a server(s) has already been configured, then it will default to the first one in the list.

patflynn:
LGTM

PR #629, opened 2016-04-13
DBG_ENBALE in app.yaml to enable Cloud Debugger

### LINKED ISSUE #626: DBG_ENABLE needed in app.yaml for Cloud Debugger to work for flex environment
We need the following config in 'app.yaml' for all flex deployments, regardless of WAR or JAR.

env_variables:
  'DBG_ENABLE': 'true'

This is documented in a hard-to-find location: https://cloud.google.com/appengine/docs/flexible/java/dev-jetty9?hl=pt-BR#environment_variables

Does it make sense to put the env variable in our 'app.yaml' template? This may not what 'gcloud' does.

chanseokoh:
Fixes #626.

The decision for now is to do this as a workaround for our story. That is, only Java apps from our IDE plugin will enable Cloud Debugger for flex runtimes (non-compat jetty9 and openjdk-8).

The App Engine team and the Cloud Debugger team may eventually agree on getting rid of this env variable for Java and other languages, but that really seems a long way down.

### DIFF
--- a/google-cloud-tools-plugin/resources/generation/src/appengine/mvm/app.yaml
+++ b/google-cloud-tools-plugin/resources/generation/src/appengine/mvm/app.yaml
@@ -1,2 +1,4 @@
 runtime: custom
 vm: true
+env_variables:
+  'DBG_ENABLE': 'true'


patflynn:
:+1:

PR #630, opened 2016-04-13
Cancel Deployment Process

### LINKED ISSUE #588: Closing project during deployment doesn't stop the deployment

meltsufin:
Fixes #588, #581, and #618.

Combines PRs #615 and #621.


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
```diff
@@ -91,14 +94,31 @@ protected void executeProcess(
 
     consoleLogLn("Executing: " + commandLine.getCommandLineString());
 
-    final Process process = commandLine.createProcess();
-    final ProcessHandler processHandler = new OSProcessHandler(process,
+    Process process = commandLine.createProcess();
+    processHandler = new OSProcessHandler(process,
         commandLine.getCommandLineString());
```
chanseokoh: This now fits in one line.

meltsufin: Fixed

meltsufin:
I removed the code surrounding the cancel button on individual deployments. Please review.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
```diff
@@ -91,14 +94,30 @@ protected void executeProcess(
...
+
     processHandler.startNotify();
   }
 
+  protected void cancel() {
+    // kill any executing process for the action
+    if (processHandler != null && processHandler.getProcess() != null) {
+      cancelled = true;
+      processHandler.getProcess().destroy();
+    }
```
chanseokoh: What if `processhandler` becomes null by the `ProcessAdapter` above right after checking `processHandler != null` but before executing `processHandler.getProcess().destory()`? Do we have a synchronization issue?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
```diff
@@ -91,14 +94,30 @@ protected void executeProcess(
...
+      @Override
+      public void processTerminated(ProcessEvent event) {
+        processHandler = null;
+      }
+    });
+
     processHandler.startNotify();
   }
 
+  protected void cancel() {
```
patflynn: why protected?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
```diff
@@ -316,16 +319,19 @@ public void connect(@NotNull ConnectionCallback<AppEngineDeploymentConfiguration
       ServerRuntimeInstance<AppEngineDeploymentConfiguration> {
 
     private AppEngineServerConfiguration configuration;
+    private Set<AppEngineAction> createdActions;
 
     public AppEngineRuntimeInstance(
         AppEngineServerConfiguration configuration) {
       this.configuration = configuration;
+      this.createdActions = new HashSet<>();
```
patflynn: this is really only AppEngineDeployments isn't it?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
meltsufin: I was just trying to be consistent executeProcess in the same class
`protected void executeProcess() {...}`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
meltsufin: Yes, I will rename to createdDeployments to be more explicit.

patflynn: maybe change the type too?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
meltsufin: Yes, I thought of that too, and I think I had `synchronized` on the method before. However, currently there is no way in the code to execute or cancel an action more than once. The only place `AppEngineAction.cancel()` is called is in `AppEngineCloudType.AppEngineRuntimeInstance.disconnect()`, which is synchronized.

chanseokoh: Hmmmm. I don't think I get it. Not sure if I am missing something. Here's what I thought:

The registered listener that sets `processHandler = null` will be run by a different thread when the gcloud deploy command starts. So that listener can set `processHandler` to null anytime, regardless of `disconnect()` is synchronized or an action can be executed or canceled more than once. There is nothing preventing that listener from setting the handler to null. The listener will do this when the gcloud process exits, and that can coincide with when `disconnect()` is called. I feel like we need to synchronize `processHandler`, since two threads can access the handler concurrently.

What do you think?

meltsufin: I see your point now. That registered listener shouldn't even be there. I put it in there when I mistakenly thought that the action could be executed and cancelled more than once. I'll remove it.

meltsufin: Never mind, it's actually still kind of useful to avoid calling process.destory() unnecessarily. I'll make it thread safe.

patflynn:
:+1: LGTM

PR #632, opened 2016-04-14
Pre-fill path fields if project has usual files

### LINKED ISSUE #617: Browse for app.yaml and Dockerfile should default to files found in project
Browse for existing app.yaml and Dockerfile should default to existing files in the project, if they can be found in the canonical locations (src/main/appengine and src/main/docker).

chanseokoh:
Fixes #617.

Initially, I wanted to point the browse window to config files if they exist in the usual locations (`/src/main/appengine/app.yaml` and `/src/main/docker/Dockerfile`). However, I wasn't able to find a way to direct a file chooser to a certain directory location when it opens. There was no method to achieve it.

As an alternative, this PR pre-fills the app.yaml and Dockerfile fields if they are empty and the project has these files in the usual locations. This way, the browse window will point to the existing files when opened.


etanshaul:
I actually like your approach of pre-filling the paths better regardless. lgtm

patflynn:
:rocket:

PR #633, opened 2016-04-14
Change our logging level to warn for possibly transient errors

patflynn:
file system and gcloud transient errors are going to pollute our crash reports so I'm changing errors that can be caused by transient issues to log.warn.


patflynn:
Fixes #605

etanshaul:
lgtm

PR #634, opened 2016-04-14
Factor out public internal classes of AppEngineCloudType.

patflynn:
AppEngineCloudType is too big, and contains classes that should be
scoped at the package level since they are used by other classes in the package.


patflynn:
@etanshaul 
@meltsufin 
@chanseokoh

etanshaul:
lgtm

Issue #635, opened 2016-04-14
'Deploy to App Engine flexible environment' menu action allows empty archive selection


meltsufin:
When you create a new deployment by using the 'Deploy to App Engine flexible environment' menu action, and do not specify the deployment archive, the dialog allows you to "Run", but then immediately fails with an exception.
```
java.lang.NullPointerException
	at com.intellij.remoteServer.impl.configuration.deployment.DeployToServerRunConfiguration.checkConfiguration(DeployToServerRunConfiguration.java:126)
	at com.intellij.execution.impl.RunnerAndConfigurationSettingsImpl.checkSettings(RunnerAndConfigurationSettingsImpl.java:435)
	at com.intellij.execution.impl.RunManagerImpl.canRunConfiguration(RunManagerImpl.java:535)
	at com.intellij.execution.impl.RunManagerImpl.canRunConfiguration(RunManagerImpl.java:530)
	at com.intellij.execution.ProgramRunnerUtil.executeConfiguration(ProgramRunnerUtil.java:65)
	at com.intellij.execution.ProgramRunnerUtil.executeConfiguration(ProgramRunnerUtil.java:123)
	at com.intellij.remoteServer.impl.configuration.deployment.DeploymentConfigurationManagerImpl.createAndRunConfiguration(DeploymentConfigurationManagerImpl.java:68)
	at com.google.cloud.tools.intellij.appengine.cloud.AppEngineToolsMenuAction.actionPerformed(AppEngineToolsMenuAction.java:56)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.performActionDumbAware(ActionUtil.java:182)
	at com.intellij.openapi.actionSystem.impl.ActionMenuItem$ActionTransmitter$1.run(ActionMenuItem.java:311)
	at com.intellij.openapi.wm.impl.FocusManagerImpl.runOnOwnContext(FocusManagerImpl.java:958)
	at com.intellij.openapi.wm.impl.IdeFocusManagerImpl.runOnOwnContext(IdeFocusManagerImpl.java:124)
	at com.intellij.openapi.actionSystem.impl.ActionMenuItem$ActionTransmitter.actionPerformed(ActionMenuItem.java:281)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at com.intellij.openapi.actionSystem.impl.ActionMenuItem.fireActionPerformed(ActionMenuItem.java:109)
[...]

PR #636, opened 2016-04-15
Catch exceptions when running a deployment config

etanshaul:
This is not my ideal solution, but it does suppress the IDE error when a user attempts to run with an invalid deployment config. 

I attempted to fix this at the source - force validation in the dialog when opening with a programatically created configuration to no avail. The bundled app engine plugin also suffers from this problem (and doesn't even suppress the IDE error). I think this the minimum we should do - suggestions welcome.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineToolsMenuAction.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineToolsMenuAction.java
@@ -21,11 +21,14 @@ import com.google.cloud.tools.intellij.util.GctBundle;
 
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.Project;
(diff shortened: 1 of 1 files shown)


etanshaul:
fixes #635

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineToolsMenuAction.java
```diff
@@ -52,8 +55,12 @@ public void actionPerformed(AnActionEvent e) {
       List<RemoteServer<AppEngineServerConfiguration>> servers =
           RemoteServersManager.getInstance().getServers(serverType);
 
-      DeploymentConfigurationManager.getInstance(project).
-          createAndRunConfiguration(serverType, ContainerUtil.getFirstItem(servers));
+      try {
+        DeploymentConfigurationManager.getInstance(project).
+            createAndRunConfiguration(serverType, ContainerUtil.getFirstItem(servers));
+      } catch (Exception ex) {
```
patflynn: Can you catch NPE and then in the catch block include a comment pointer to the issue?

etanshaul:
good suggestion. done.

patflynn:
LGTM

Issue #600, opened 2016-04-06
It would be cool to configure mvn install or gradle build as a pre deploy step automatically for app engine deployment


patflynn:
If a project is maven or gradle, it could be cool to automatically add producing the project artifacts as a prequesite of deployment.

etanshaul:
👍 

Just adding a note that if IJ already created an artifact (if UE and it created a war artifact) then this isn't needed. This would handle the cases of jar's (CE and UE) and war's (for CE).

One approach could be, when the project has a configured Maven / Grade facet, to create a new custom deployment source type (MavenDeploymentSourceType for example) which automatically adds the pre deploy step and is aware of the target archive.

PR #638, opened 2016-04-15
Make flex deploy build aware

etanshaul:
See #600.

Work in progress. 

Auto detects a maven project (gradle is pending) and pre-configures a new Maven deployment source type. This new type automatically has a pre-deploy step configured to build the project.

Needs a bunch of a cleanup and a solution to a technical problem related to #635

### DIFF
--- a/google-cloud-tools-plugin/build.gradle
+++ b/google-cloud-tools-plugin/build.gradle
@@ -22,7 +22,7 @@ sourceSets.test.resources.srcDirs = ['testResources']
 
 intellij {
     pluginName = 'google-cloud-tools'
-    plugins 'Groovy','gradle', 'git4idea', 'properties', 'junit'
+    plugins 'Groovy', 'gradle', 'git4idea', 'properties', 'junit', 'maven'
 
     publish {
         pluginId '8079'
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/CloudToolsPluginInitializationComponent.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/CloudToolsPluginInitializationComponent.java
@@ -17,6 +17,7 @@ package com.google.cloud.tools.intellij;
 
 import com.google.cloud.tools.intellij.appengine.cloud.AppEngineCloudType;
 import com.google.cloud.tools.intellij.appengine.cloud.AppEngineToolsMenuAction;
+import com.google.cloud.tools.intellij.appengine.cloud.MavenBuildDeploymentSourceType;
 import com.google.cloud.tools.intellij.appengine.cloud.UserSpecifiedPathDeploymentSourceType;
 import com.google.cloud.tools.intellij.debugger.CloudDebugConfigType;
 
@@ -51,33 +52,51 @@ public class CloudToolsPluginInitializationComponent implements ApplicationCompo
 
   @Override
   public void initComponent() {
-    CloudToolsPluginInfoService pluginInfoService = ServiceManager
-        .getService(CloudToolsPluginInfoService.class);
     CloudToolsPluginConfigurationService pluginConfigurationService = ServiceManager
         .getService(CloudToolsPluginConfigurationService.class);
-    if (pluginInfoService.shouldEnable(GctFeature.DEBUGGER)) {
-      pluginConfigurationService
-          .registerExtension(
-              ConfigurationType.CONFIGURATION_TYPE_EP, new CloudDebugConfigType());
-    }
-    if (pluginInfoService.shouldEnable(GctFeature.APPENGINE_FLEX)) {
-      AppEngineCloudType appEngineCloudType = new AppEngineCloudType();
-      pluginConfigurationService.registerExtension(ServerType.EP_NAME, appEngineCloudType);
-      pluginConfigurationService.registerExtension(DeploymentSourceType.EP_NAME,
-          new UserSpecifiedPathDeploymentSourceType());
-      pluginConfigurationService.registerExtension(ConfigurationType.CONFIGURATION_TYPE_EP,
-          new DeployToServerConfigurationType(appEngineCloudType));
+    CloudToolsPluginInfoService pluginInfoService = ServiceManager
+        .getService(CloudToolsPluginInfoService.class);
 
-      ActionManager actionManager = ActionManager.getInstance();
-      AnAction toolsMenuAction = new AppEngineToolsMenuAction();
-      actionManager.registerAction(AppEngineToolsMenuAction.ID , toolsMenuAction);
-      DefaultActionGroup toolsMenu =
-          (DefaultActionGroup) actionManager.getAction(AppEngineToolsMenuAction.GROUP_ID);
-      toolsMenu.add(toolsMenuAction, Constraints.LAST);
+    if (pluginInfoService.shouldEnable(GctFeature.DEBUGGER)) {
+      initDebugger(pluginConfigurationService);
     }
+
+    if (pluginInfoService.shouldEnable(GctFeature.APPENGINE_FLEX)) {
+      initAppEngineFlex(pluginConfigurationService);
+    }
+
     if (pluginInfoService.shouldEnableErrorFeedbackReporting()) {
-      pluginConfigurationService
-          .enabledGoogleFeedbackErrorReporting(pluginInfoService.getPluginId());
+      initErrorReporting(pluginConfigurationService, pluginInfoService);
     }
+  }
+
+  private void initDebugger(CloudToolsPluginConfigurationService pluginConfigurationService) {
+    pluginConfigurationService
+        .registerExtension(
+            ConfigurationType.CONFIGURATION_TYPE_EP, new CloudDebugConfigType());
+  }
+
+  private void initAppEngineFlex(CloudToolsPluginConfigurationService pluginConfigurationService) {
+    AppEngineCloudType appEngineCloudType = new AppEngineCloudType();
+    pluginConfigurationService.registerExtension(ServerType.EP_NAME, appEngineCloudType);
+    pluginConfigurationService.registerExtension(ConfigurationType.CONFIGURATION_TYPE_EP,
+        new DeployToServerConfigurationType(appEngineCloudType));
+    pluginConfigurationService.registerExtension(DeploymentSourceType.EP_NAME,
+        new UserSpecifiedPathDeploymentSourceType());
+    pluginConfigurationService.registerExtension(DeploymentSourceType.EP_NAME,
+        new MavenBuildDeploymentSourceType());
+
+    ActionManager actionManager = ActionManager.getInstance();
+    AnAction toolsMenuAction = new AppEngineToolsMenuAction();
+    actionManager.registerAction(AppEngineToolsMenuAction.ID , toolsMenuAction);
+    DefaultActionGroup toolsMenu =
+        (DefaultActionGroup) actionManager.getAction(AppEngineToolsMenuAction.GROUP_ID);
+    toolsMenu.add(toolsMenuAction, Constraints.LAST);
+  }
+
+  private void initErrorReporting(CloudToolsPluginConfigurationService pluginConfigurationService,
+      CloudToolsPluginInfoService pluginInfoService) {
+    pluginConfigurationService
+        .enabledGoogleFeedbackErrorReporting(pluginInfoService.getPluginId());
   }
 }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
@@ -18,6 +18,8 @@ package com.google.cloud.tools.intellij.appengine.cloud;
 
 import com.google.cloud.tools.intellij.appengine.util.CloudSdkUtil;
 
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.module.ModuleManager;
 import com.intellij.openapi.module.ModulePointer;
 import com.intellij.openapi.module.ModulePointerManager;
 import com.intellij.openapi.options.SettingsEditor;
@@ -31,6 +33,8 @@ import com.intellij.remoteServer.configuration.deployment.JavaDeploymentSourceUt
 
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.idea.maven.project.MavenProject;
+import org.jetbrains.idea.maven.project.MavenProjectsManager;
 
 import java.io.File;
 import java.util.ArrayList;
@@ -52,20 +56,41 @@ class AppEngineDeploymentConfigurator extends
   @NotNull
   @Override
   public List<DeploymentSource> getAvailableDeploymentSources() {
-    List<DeploymentSource> deploymentSources = new ArrayList<DeploymentSource>();
+    List<DeploymentSource> deploymentSources = new ArrayList<>();
+
+    for (Module module : ModuleManager.getInstance(project).getModules()) {
+      if (isJarOrWarMavenBuild(module)) {
+        deploymentSources.add(
+            new MavenBuildDeploymentSource(
+                ModulePointerManager.getInstance(project).create(module), project));
+      }
+    }
 
     ModulePointer modulePointer =
         ModulePointerManager.getInstance(project).create("userSpecifiedSource");
     deploymentSources.add(new UserSpecifiedPathDeploymentSource(modulePointer));
 
     deploymentSources.addAll(JavaDeploymentSourceUtil
-        .getInstance().createArtifactDeploymentSources(project, getJarsAndWars()));
+        .getInstance().createArtifactDeploymentSources(project, getJarAndWarArtifacts()));
 
     return deploymentSources;
   }
 
-  private List<Artifact> getJarsAndWars() {
-    List<Artifact> jarsAndWars = new ArrayList<Artifact>();
+  private boolean isJarOrWarMavenBuild(Module module) {
+    MavenProjectsManager projectsManager = MavenProjectsManager.getInstance(project);
+    MavenProject mavenProject = projectsManager.findProject(module);
+
+    boolean isMavenProject = projectsManager.isMavenizedModule(module)
+        && mavenProject != null;
+
+    return isMavenProject &&
+        ("jar".equalsIgnoreCase(mavenProject.getPackaging())
+            || "war".equalsIgnoreCase(mavenProject.getPackaging()));
+
+  }
+
+  private List<Artifact> getJarAndWarArtifacts() {
+    List<Artifact> jarsAndWars = new ArrayList<>();
     for (Artifact artifact : ArtifactManager.getInstance(project).getArtifacts()) {
       if (artifact.getArtifactType().getId().equalsIgnoreCase("jar")
           || artifact.getArtifactType().getId().equalsIgnoreCase("war")) {
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineRuntimeInstance.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineRuntimeInstance.java
@@ -72,8 +72,11 @@ class AppEngineRuntimeInstance extends
 
     final AppEngineDeployAction deployAction;
     AppEngineDeploymentConfiguration deploymentConfig = task.getConfiguration();
-    File deploymentSource = deploymentConfig.isUserSpecifiedArtifact() ?
-        new File(deploymentConfig.getUserSpecifiedArtifactPath()) : task.getSource().getFile();
+    File deploymentSource = task.getSource().getFile();
+    if (deploymentSource == null) {
+      callback.errorOccurred(GctBundle.message("appengine.deployment.source.not.found.error"));
+      return;
+    }
 
     if (deploymentConfig.getConfigType() == ConfigType.AUTO) {
       deployAction = appEngineHelper.createAutoDeploymentAction(
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/BuildDeploymentSourceType.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/BuildDeploymentSourceType.java
@@ -1,0 +1,168 @@
+/*
+ * Copyright (C) 2016 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.intellij.appengine.cloud;
+
+import com.intellij.execution.BeforeRunTask;
+import com.intellij.execution.RunManagerEx;
+import com.intellij.execution.configurations.RunConfiguration;
+import com.intellij.execution.impl.ConfigurationSettingsEditorWrapper;
+import com.intellij.ide.DataManager;
+import com.intellij.openapi.actionSystem.DataContext;
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.project.Project;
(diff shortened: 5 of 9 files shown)


patflynn:
this is pretty sweet.

chanseokoh:
Wow. Amazing that you figured this out.

etanshaul:
Question for you guys:

Say the project is a Maven project producing a war file and the user is using IJ Ultimate Edition. With this setup, the project will generate a war artifact automatically for the user. As a result, there will be a "native" artifact deployment source available for the user to choose from for cloud deployment. Now, because we are adding a new Maven deployment source I see 4 options in this scenario:

1) Have both of these source types in the dropdown even though they effectively do the same thing
2) Suppress the native artifact source and only show the new Maven source
3) Suppress the Maven source and only show the native artifact source
4) Just dispense of the native artifact sources altogether

I am leaning towards (3). (4) is nice, but has the risk of missing out on IJ Artifacts that could be created from non-maven/gradle build systems.

patflynn:
They aren't actually the same though.  The IJ generated artifact is
actually just an approximation of the Maven war because it's built
natively. That's something that I've found pretty weird and probably not
what end users want to deploy to production. That being said I think it
would be bad form to start ignoring the native IJ artifact types which we
can conceivably deploy, so I'd be tempted to just show both.

Your Maven artifact source type actually seems like something that would be
generally useful to people using IJ.

etanshaul:
Ah yes, good point about how though native Artifacts actually perform the builds. I agree it would be bad form to ignore them so we'll show both.

etanshaul:
@patflynn you can start taking a look. There is currently a bug with the deployment dialog requiring the user to touch one of the configuration modifying UI elements for the configuration to be saved and validation to occur (I reached out to Jetbrains about this and am looking into it - I'll file an issue with them if there is no solution).

etanshaul:
Sample screenshot of deployment dialog with a Gradle project:

image

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/CloudToolsPluginInitializationComponent.java
```diff
@@ -63,10 +65,14 @@ public void initComponent() {
     if (pluginInfoService.shouldEnable(GctFeature.APPENGINE_FLEX)) {
       AppEngineCloudType appEngineCloudType = new AppEngineCloudType();
```
patflynn: maybe extract this block as a method?  Just might help readability a bit.

patflynn:
is 'complete' in the Gradle artifact label the name of the project artifact you're deploying?

etanshaul:
It just happens to be the name of the module. Bad example I guess for a screenshot.

etanshaul:
Update: I've implemented a workaround to the platform issue forcing the user to edit the UI prior to running a new deployment config by tying into the build task setter and manually setting the configuration. This still does not perform validation but this is acceptable for the new gradle/maven build types which are valid by default when the dialog is first opened. Will push after some refactoring.

etanshaul:
Updated and refactored. PTAL.

google-cloud-tools-plugin/build.gradle
```diff
@@ -22,7 +22,7 @@ sourceSets.test.resources.srcDirs = ['testResources']
 
 intellij {
     pluginName = 'google-cloud-tools'
-    plugins 'Groovy','gradle', 'git4idea', 'properties', 'junit'
+    plugins 'Groovy','gradle', 'git4idea', 'properties', 'junit', 'maven'
```
loosebazooka: just cause you already changed this file, any way you can add a space between 'Groovy','gradle'

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
```diff
@@ -52,7 +57,20 @@ public AppEngineDeploymentConfigurator(Project project) {
...
-    List<DeploymentSource> deploymentSources = new ArrayList<DeploymentSource>();
+    List<DeploymentSource> deploymentSources = new ArrayList<>();
+
+    for (Module module : ModuleManager.getInstance(project).getModules()) {
+      if (isMavenBuild(module)) {
+        deploymentSources.add(
+            new MavenBuildDeploymentSource(
+                ModulePointerManager.getInstance(project).create(module), project));
+      }
+      if (isGradleBuild(module)) {
```
loosebazooka: Can it be both gradle and maven, the code is technically allowed to both deployment sources?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/GradleBuildDeploymentSource.java
```diff
@@ -0,0 +1,116 @@
...
+
+  public GradleBuildDeploymentSource(@NotNull ModulePointer pointer, @NotNull Project project) {
+    super(pointer);
+    this.project = project;
+  }
+
+  @NotNull
+  @Override
+  public String getPresentableName() {
+    return String.format("Gradle build: %s", getModulePointer().getModuleName());
```
loosebazooka: Personal preference is for "+" in simple formatting situations like this for readability. But I don't know

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/GradleBuildDeploymentSource.java
```diff
@@ -0,0 +1,116 @@
...
+
+  @NotNull
+  @Override
+  public DeploymentSourceType<?> getType() {
+    return DeploymentSourceType.EP_NAME.findExtension(GradleBuildDeploymentSourceType.class);
+  }
+
+  @Nullable
+  @Override
+  public File getFile() {
```
loosebazooka: I'm not sure how this works in the multi module situation. Why are you going down to the root? What is expected project structure here and what is the expected deployable that would be extracted for a multimodule project?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/GradleBuildDeploymentSourceType.java
```diff
@@ -0,0 +1,87 @@
...
+  @Nullable
+  @Override
+  protected BeforeRunTask createBuildTask(Project project, Module module) {
+    ExternalSystemBeforeRunTask task =
+        new ExternalSystemBeforeRunTask(
+            GradleBeforeRunTaskProvider.ID, GradleConstants.SYSTEM_ID);
+
+    ExternalSystemTaskExecutionSettings taskSettings = task.getTaskExecutionSettings();
+    taskSettings.setExternalProjectPath(getPathToGradleBuildFile(module));
+    taskSettings.setTaskNames(ImmutableList.of(GRADLE_TASK_BUILD));
```
loosebazooka: So generally the build task runs all their check (tests). Would it be possible for them to optionally just run assemble (which runs archive tasks)? Perhaps build is too heavy weight for fast consecutive deployments? Or maybe full checks are what we want before deployment anyway. I don't know

loosebazooka: Actually one successful build should bypass most steps on consecutive builds with no tests...

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
etanshaul: Yes, in which case the user can choose between them. Not a common scenario, but as an example, I tested it with the spring boot sample project which has both gradle and maven configs.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/GradleBuildDeploymentSource.java
etanshaul: I am getting the root project, and using that to get the ExternalProject for the module in question. So this should actually be scoped to a specific project. 

How this actually would work with a multi-module project is an interesting question (which I need to test out), but it is my understanding that flex doesn't support multi-module deployments ( @patflynn ?). Or did you just mean selecting between the multiple modules for deployment? Either way I need to test this out.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/GradleBuildDeploymentSourceType.java
etanshaul: Ah yes. So are you saying you are ok with running successive "build" tasks given that it should bypass most of the steps?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/GradleBuildDeploymentSource.java
etanshaul: For a multi-module maven/gradle project, each module will have its own maven/gradle deployment source:

image

However, currently switching between them doesn't yet properly update the pre-deploy build task. TBD: how do we deal with ear producing modules, such as the one shown in the above screenshot, since these cannot be deployed to flex.

patflynn: Can we only show the modules that produce a war/jar?

etanshaul: Yes. I'll include that.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/GradleBuildDeploymentSource.java
```diff
@@ -0,0 +1,116 @@
...
+    return findGradleBuildArtifact(moduleExternalProject.getBuildDir());
+  }
+
+  /**
+   * Uses a best-effort approach to locate the Gradle build artifact.
+   * Assumes the build artifact is located in the /libs folder under the root build directory.
+   * Also assumes only a single build artifact. Will only return one artifact if multiple are found.
+   */
+  @Nullable
+  private File findGradleBuildArtifact(@NotNull File buildDir) {
```
patflynn: aha.  I thought that we had found a robust way to do this.  So Gradle does not let you determine the build artifact for a given module?

etanshaul: no, not that I could find unfortunately. On the other hand maven does.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/GradleBuildDeploymentSource.java
```diff
@@ -0,0 +1,116 @@
...
+
+    if (moduleExternalProject == null) {
+      return null;
+    }
+
+    return findGradleBuildArtifact(moduleExternalProject.getBuildDir());
+  }
+
+  /**
+   * Uses a best-effort approach to locate the Gradle build artifact.
```
patflynn: This heuristic seems weak.  I really have no idea if these are safe assumptions. I seems like it could be bad to deploy a jar or war other than the one the user had intended.

etanshaul: I agree its not ideal. The scenario I envision that could lead to an unintended deployment is if the build generates multiple artifacts in which case we are just choosing one (curiously the maven code that provides the artifact to us always only returns 1). Additionally it makes the assumption that given the build directory, that the artifacts will be in the "/libs" directory which I guess could be overriden.

etanshaul: Actually, it slipped my mind that I made a note to myself to replace that code as its using a deprecated call (ProjectDataManager#getDataService) which no longer works in 2016. So that logic needs to be replaced with something regardless. i.e. I need to find a different / better way to get a handle on the gradle build artifact - the only other alternative being that we only support maven at first.

etanshaul:
Removed gradle support for now.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
```diff
@@ -52,20 +56,45 @@ public AppEngineDeploymentConfigurator(Project project) {
...
-        .getInstance().createArtifactDeploymentSources(project, getJarsAndWars()));
+        .getInstance().createArtifactDeploymentSources(project, getJarAndWarArtifacts()));
 
     return deploymentSources;
   }
 
-  private List<Artifact> getJarsAndWars() {
-    List<Artifact> jarsAndWars = new ArrayList<Artifact>();
+  private boolean isJarOrWarMavenBuild(Module module) {
+    boolean isMavenProject = MavenProjectsManager.getInstance(project).isMavenizedModule(module)
```
patflynn: it looks like we call MoavenPRojectsManager.getInstance(project) a lot in this method, what do you think about extracting a variable?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
```diff
@@ -52,20 +56,45 @@ public AppEngineDeploymentConfigurator(Project project) {
...
+        && MavenProjectsManager.getInstance(project).findProject(module) != null;
+
+    if(isMavenProject) {
+      MavenProject mavenProject =
+          MavenProjectsManager.getInstance(project).findProject(module);
+
+      if(mavenProject != null) {
```
patflynn: I feel like we should know this already.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
```diff
@@ -52,20 +56,45 @@ public AppEngineDeploymentConfigurator(Project project) {
...
   }
 
-  private List<Artifact> getJarsAndWars() {
-    List<Artifact> jarsAndWars = new ArrayList<Artifact>();
+  private boolean isJarOrWarMavenBuild(Module module) {
+    boolean isMavenProject = MavenProjectsManager.getInstance(project).isMavenizedModule(module)
+        && MavenProjectsManager.getInstance(project).findProject(module) != null;
+
+    if(isMavenProject) {
+      MavenProject mavenProject =
```
patflynn: Not sure how much work findProject is doing, but it feels a little sloppy to retrieve the same object twice.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineRuntimeInstance.java
```diff
@@ -70,8 +72,11 @@ public void deploy(@NotNull final DeploymentTask<AppEngineDeploymentConfiguratio
 
     final AppEngineDeployAction deployAction;
     AppEngineDeploymentConfiguration deploymentConfig = task.getConfiguration();
-    File deploymentSource = deploymentConfig.isUserSpecifiedArtifact() ?
-        new File(deploymentConfig.getUserSpecifiedArtifactPath()) : task.getSource().getFile();
```
patflynn: hadn't we setup task.getSource().getFile() to return the user specified artifact so that we didn't have to special case this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineRuntimeInstance.java
```diff
@@ -123,6 +128,14 @@ public void disconnect() {
     }
   }
 
+  @Nullable
+  private File resolveDeploymentSource(DeploymentTask<AppEngineDeploymentConfiguration> task) {
+    DeploymentSource source = task.getSource();
+
+    return task.getConfiguration().isUserSpecifiedArtifact() ?
```
patflynn: agh. I think I made things more difficult by starting to use Path in our code. Will have to think about that.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/BuildDeploymentSourceType.java
```diff
@@ -0,0 +1,124 @@
...
+        List<BeforeRunTask> beforeRunTasks = editor.getStepsBeforeLaunch();
+        if (select && !hasBuildTaskForModule(beforeRunTasks, module)) {
+          editor.addBeforeLaunchStep(buildTask);
+        }
+      }
+    }
+  }
+
+  @NotNull
+  protected abstract List<? extends BeforeRunTask> getBuildTasks(
```
patflynn: it can probably be safely said that I have a near pathological aversion to inheritance in general, and the template method pattern in particular.   That being said, we live in this big on inheritance IntelliJ platform framework and we have to make do. 

In the cases where this pattern is the most pragmatic, we should definitely document carefully what the contract is  that we expect subclasses to fill. I don't think it's very clear just from looking at these abstract method signatures.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/MavenBuildDeploymentSource.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    MavenProject mavenProject =
+        MavenProjectsManager.getInstance(project).findProject(getModule());
+
+    if (mavenProject == null) {
+      return null;
+    }
+
+    String targetBuild = String.format("%s/%s.%s",
```
patflynn: do we have to worry about system dependent file paths here?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/MavenBuildDeploymentSourceType.java
```diff
@@ -0,0 +1,95 @@
...
+
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * A Maven build deployment source type providing an auto configured pre-deploy build step
+ */
+public class MavenBuildDeploymentSourceType extends BuildDeploymentSourceType {
+
+  private static final String MAVEN_TASK_INSTALL = "install";
```
patflynn: what about 'package', I think that's the goal directly related to creating the target output archive.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/BuildDeploymentSourceType.java
```diff
@@ -0,0 +1,124 @@
...
+      RunConfiguration configuration);
+
+  @Nullable
+  protected abstract BeforeRunTask createBuildTask(Module module);
+
+  protected abstract boolean hasBuildTaskForModule(
+      Collection<? extends BeforeRunTask> beforeRunTasks, Module module);
+
+  /**
+   * Manually set the deployment configuration so that its available immediately in the
```
patflynn: I'm wondering why this breaks us but not other ModuleDeploymentSourceTypes

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/BuildDeploymentSourceType.java
```diff
@@ -0,0 +1,124 @@
...
+      }
+    } else {
+      BeforeRunTask buildTask = createBuildTask(module);
+      List<BeforeRunTask> tasks = runManager.getBeforeRunTasks(configuration);
+      tasks.add(buildTask);
+      runManager.setBeforeRunTasks(configuration, tasks, true);
+    }
+  }
+
+  @Override
```
patflynn: could you add some docs to explain what the difference is between these two overridden methods?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
etanshaul: sorry - know what? if mavenProject is null? (if thats what you were asking, findProject is nullable)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
etanshaul: I'll remove duplication in this method.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineRuntimeInstance.java
etanshaul: Yes we did, and I think this was one place where the refactoring was overlooked. I'll double check and update it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/BuildDeploymentSourceType.java
etanshaul: It does break other ModuleDeploymentSourceTypes. It also breaks ArtifactDeploymentSourceTypes for built in IJ artifacts. Attempting to deploy to a pre-selected IJ artifact will throw an NPE unless a similar fix is in place (or the user edits the UI). Also interesting to note that the existing App Engine plugin has this same issue

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/MavenBuildDeploymentSource.java
etanshaul: the next line wraps it in a new File which is system independent.

etanshaul: But I'll confirm this because on second thought mavenProject.getBuildDirectory() could return a system dependent path which may have unintended effects.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
etanshaul: never mind, ignore my last comment :).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/BuildDeploymentSourceType.java
etanshaul: Yes I agree. The DeploymentSourceType api we have to work with leads to some less than ideal code. I am working on refactoring this and adding documentation.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
patflynn: I also should have be more specific.

etanshaul:
Comments have been addressed. 

I am still using inheritance for the BuildDeploymentSourceType - while I share your aversion, I initially when down the path using something like a helper but it wasn't as nice. So I'm using the pattern of having the abstract base class for common build-system tasks, and delegating to the concrete subclasses to perform the build-system specific tasks. Hopefully the refactoring and documentation helps.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/BuildDeploymentSourceType.java
```diff
@@ -0,0 +1,167 @@
...
+ * sources for specific build-systems should extend this providing the build-system specific
+ * implementations.
+ */
+public abstract class BuildDeploymentSourceType extends ModuleDeploymentSourceType {
+
+  /**
+   * Creates a pre-deploy task ({@link BeforeRunTask}) for the given build-system and attaches it
+   * to this module deployment source type. Invoked when a new deployment configuration is created.
+   *
+   * Provides the common functionality for to creating the build-system packaging task,
```
patflynn: ...common functionality for 'to' creating..

patflynn:
Good job. thanks.

:ship:

PR #640, opened 2016-04-18
Replace deprecated API to eliminate warning

chanseokoh:
Eliminates that irritating warning in our gradle build log that we are using a deprecated API.


patflynn:
LGTM

PR #641, opened 2016-04-18
Make flex more info label shrinkable

chanseokoh:
The problem was that the "more info label" in the server config dialog ("For more information about the App Engine flexible environment, ...") was not shrinkable. The message is very long and prevented the entire window from being resized down from both left and right sides.

This PR just marks the label "shrinkable" in the GUI designer.


patflynn:
LGTM

PR #643, opened 2016-04-19
Init login flow on connect / deploy

### LINKED ISSUE #642: It would be nice to trigger the auth flow on deployment if the user is not logged in.
Either that or have a link to an action that starts the auth flow where 'You must be looged in to perform this action' is displayed.

etanshaul:
fixes #642 

Automatically inits the login flow (open up browser etc.) when the user attempts to connect or deploy when not logged in.


patflynn:
LGTM.

Issue #606, opened 2016-04-06
'Add User' followed with 'Deny' breaks Account Plugin UI


meltsufin:
Steps taken:

1. Login to Google Cloud using the toolbar icon.
1. In the browser, 'Allow'
1. Click the account toolbar icon again and select "Add Account"
1. In the browser, 'Deny'

Observe that the toolbar icon has changed to logged out state and is now gray. When you click it, you still see your logged in user from before, but now the "Sign Out" button is disabled. The plugin is effectively tricked into thinking that the user has logged out.

chanseokoh:
#647 fixed this issue but was not auto-closing.

PR #647, opened 2016-04-20
Restore previous active user on aborted login

### LINKED ISSUE #644: IntellijGoogleLoginService.logIn() line 165 removes the active user prematurely.
the user can still cancel the login and if they do there will be no active user even though you're still logged in.

etanshaul:
Fixes #644, #606 

Note: the common auth API service that performs the auth flow (opens the browser window etc.) will only do so if the user is not logged in, requiring us to log out prior to initializing the auth flow for the case of multiple accounts.

Therefore, for adding a second account when already logged in, the least invasive update is to keep a reference to the current active user and restore it if the login fails or is aborted.

### DIFF
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/login/IntellijGoogleLoginService.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/login/IntellijGoogleLoginService.java
@@ -169,6 +169,7 @@ public class IntellijGoogleLoginService implements GoogleLoginService {
       @Nullable final IGoogleLoginCompletedCallback callback) {
     UsageTrackerProvider
         .getInstance().trackEvent(LoginTracking.CATEGORY, LoginTracking.LOGIN, "login.start", null);
+    final CredentialedUser lastActiveUser = users.getActiveUser();
     users.removeActiveUser();
     uiFacade.notifyStatusIndicator();
 
@@ -223,8 +224,19 @@ public class IntellijGoogleLoginService implements GoogleLoginService {
           };
           users.addUser(new CredentialedUser(state, localCallback));
         }
-        else if (callback != null) {
-          callback.onLoginCompleted();
+        else {
+          // Login failed (or aborted), so restore the last active user, if any
+          restoreLastActiveUser();
+
+          if (callback != null) {
+            callback.onLoginCompleted();
+          }
+        }
+      }
+
+      private void restoreLastActiveUser() {
+        if (lastActiveUser != null) {
+          setActiveUser(lastActiveUser.getEmail());
         }
       }
     }.queue();


google-account-plugin/src/com/google/cloud/tools/intellij/login/IntellijGoogleLoginService.java
```diff
@@ -169,6 +169,7 @@ public void logIn(@Nullable final String message,
       @Nullable final IGoogleLoginCompletedCallback callback) {
     UsageTrackerProvider
         .getInstance().trackEvent(LoginTracking.CATEGORY, LoginTracking.LOGIN, "login.start", null);
+    final CredentialedUser activeUser = users.getActiveUser();
     users.removeActiveUser();
```
chanseokoh: Just a question: I was wondering what will happen if we just don't call `removeActiveUser()`.

etanshaul: Then attempting to add an account when already logged in will not work: to initiate the auth flow the user has to be logged out, otherwise the flow will be skipped.

etanshaul: Oh, just to point out, `removeActiveUser()` also logs out.

google-account-plugin/src/com/google/cloud/tools/intellij/login/IntellijGoogleLoginService.java
```diff
@@ -169,6 +169,7 @@ public void logIn(@Nullable final String message,
       @Nullable final IGoogleLoginCompletedCallback callback) {
     UsageTrackerProvider
         .getInstance().trackEvent(LoginTracking.CATEGORY, LoginTracking.LOGIN, "login.start", null);
+    final CredentialedUser activeUser = users.getActiveUser();
```
patflynn: maybe rename to lastActiveUser?

google-account-plugin/src/com/google/cloud/tools/intellij/login/IntellijGoogleLoginService.java
```diff
@@ -223,8 +224,19 @@ public void onLoginCompleted() {
           };
           users.addUser(new CredentialedUser(state, localCallback));
         }
-        else if (callback != null) {
-          callback.onLoginCompleted();
+        else {
+          restorePreviousUser();
```
patflynn: could be helpful to add a comment here that this is for the case that the login failed or was cancelled

google-account-plugin/src/com/google/cloud/tools/intellij/login/IntellijGoogleLoginService.java
etanshaul: Yea, that is better.

patflynn:
otherwise you're good to merge.

patflynn:
@etanshaul Etan, can you merge this fix into Android Studio as well?

etanshaul:
Ah, yes. No problem

PR #653, opened 2016-04-25
Decouple version from app.yaml/Dockerfile in flex deploy

### LINKED ISSUE #639: Make deployment version part of main configuration
The user should be able to easily override the default version configuration without having to specify app.yaml and Dockerfile. Version configuration should also be given more prominence by moving it outside of the "Custom Configuration" panel.

deployment_config

etanshaul:
fixes #639 

image

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
@@ -16,11 +16,10 @@
 
 package com.google.cloud.tools.intellij.appengine.cloud;
 
-import com.google.cloud.tools.intellij.appengine.util.AppEngineUtil;
-import com.google.common.base.Preconditions;
 import com.google.cloud.tools.intellij.appengine.cloud.AppEngineDeploymentConfiguration.ConfigType;
-import com.google.cloud.tools.intellij.util.GctTracking;
 import com.google.cloud.tools.intellij.stats.UsageTrackerProvider;
+import com.google.cloud.tools.intellij.util.GctTracking;
+import com.google.common.base.Preconditions;
 
 import com.intellij.openapi.project.Project;
 import com.intellij.remoteServer.runtime.Deployment;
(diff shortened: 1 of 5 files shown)


etanshaul:
@patflynn, @meltsufin

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -137,7 +137,7 @@ public AppEngineDeployAction createAutoDeploymentAction(
         artifactToDeploy,
         defaultAppYaml(),
         defaultDockerfile(artifactType),
-        AppEngineUtil.generateVersion(),
+        version,
```
patflynn: how does the user generate the version now?=

etanshaul: If the user doesn't provide a version ID by typing one into the textfield then one will be generated automatically (see `AppEngineDeployAction` line 113).

patflynn: ah interesting.  I'm not crazy about that duplication.  How about we just make the versionId param to AppEngineDeployAction @NotNull?

etanshaul: The duplication between `createCustomDeploymentAction` and `createAutoDeploymentAction`?

patflynn: No, The fact that we are generating the version here, but then the DeployAction actually calls generate version itself if no version is passed.  I'd rather have that logic confined to the helper.

etanshaul: The only place the version is being generated is in DeployAction. Regardless, you are right its better to confine it to the helper. I'll update it.

etanshaul:
refactored version generation per conversion. @patflynn let me know if that's what you had in mind.

patflynn:
LGTM

PR #654, opened 2016-04-26
app.yaml file chooser filter

### LINKED ISSUE #652: Dockerfile and app.yaml file browsers could filter file types

etanshaul:
fixes #652

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
@@ -37,7 +37,6 @@ import com.intellij.openapi.ui.popup.BalloonBuilder;
 import com.intellij.openapi.ui.popup.JBPopupFactory;
 import com.intellij.openapi.util.Comparing;
 import com.intellij.openapi.util.Condition;
-import com.intellij.openapi.util.SystemInfo;
 import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.openapi.vfs.LocalFileSystem;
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -179,12 +187,26 @@ public boolean value(VirtualFile file) {
...
     appYamlPathField.addBrowseFolderListener(
         GctBundle.message("appengine.appyaml.location.browse.window.title"),
         null,
         project,
-        FileChooserDescriptorFactory.createSingleFileDescriptor());
+        FileChooserDescriptorFactory
+            .createSingleFileDescriptor().withFileFilter(new Condition<VirtualFile>() {
+          @Override
+          public boolean value(VirtualFile file) {
+            return Comparing.equal(file.getName(), "app.yaml", false);
```
patflynn: Messing around I do seem to be able to call `gcloud preview app deploy bla.yaml`..  I'm just wondering if users might do things like app.qa.yaml, app.prod.yaml, and even Dockerfile.qa, and then get any names right (if they need to be) during staging.

etanshaul: Yea I was thinking about that - my original reasoning is that we would guide them on the best practice. But you give a good example here. We may want to table this PR for now.

It's worth noting that regardless of the naming of the file they choose via the IJ plugin, it will be copied over to the target directory with the canonical names (so they will always be passed to gcloud this way). This doesn't affect your point though.

patflynn:
Yeah. Let's not tighten the straight jacket just yet on this one.

etanshaul:
closing this and associated issue #652 for now

Issue #562, opened 2016-03-30
Need to turn off gsutil for gcloud based deployment


patflynn:
export CLOUDSDK_APP_USE_GSUTIL 0 should be set for the deploy gcloud invocation.  Otherwise when gcloud delegates to gsutil for moving files to gcs the gcloud credentials will be used instead of the provided credential override.

etanshaul:
This is an environment variable that needs to be set?

patflynn:
yeah.

etanshaul:
I wonder if it makes more sense to run the gcloud config set command instead of exporting an env var from within java.

patflynn:
my concern with using gcloud config set is that it's a global configuration
that would effect cli based gcloud invocations if we failed to set it back.

I think sticking to the environment variable for now is safer, and
apparently the whole gsutil path will be removed in the near future.

etanshaul:
any idea how to validate that the setting was actually used during deployment?

patflynn:
you could logout of gcloud at the CLI and make sure deployment still works

PR #658, opened 2016-04-26
Replace deprecated stop

### LINKED ISSUE #657: Replace deprecated gcloud stop command
`gcloud preview app modules stop` is now deprecated (https://cloud.google.com/sdk/gcloud/reference/preview/app/modules/stop)

etanshaul:
fixes #657 

An open question I have is if version names must be unique across modules. The new recommended stop command does not take a module name as input so I assume the answer is yes, but I'd like to test this out.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
@@ -86,6 +86,7 @@ public abstract class AppEngineAction implements Runnable {
     commandLine.addParameter("--project=" + appEngineHelper.getProjectId());
     commandLine
         .addParameter("--credential-file-override=" + credentialsPath.getAbsolutePath());
+    commandLine.addParameter("--quiet");
     commandLine.withParentEnvironmentType(ParentEnvironmentType.CONSOLE);
     commandLine.getEnvironment().put("CLOUDSDK_METRICS_ENVIRONMENT", "gcloud-intellij");
     commandLine.getEnvironment().put("CLOUDSDK_APP_USE_GSUTIL", "0");
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStopAction.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStopAction.java
@@ -30,8 +30,6 @@ import com.intellij.remoteServer.runtime.log.LoggingHandler;
 
 import org.jetbrains.annotations.NotNull;
 
-import java.util.Set;
-
 import javax.swing.SwingUtilities;
 
 /**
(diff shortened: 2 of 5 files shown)


etanshaul:
It looks like "modules" were replaced with "services". So the module related code I killed off in this PR will have to be reintroduced.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
```diff
@@ -86,6 +86,7 @@ protected void executeProcess(
     commandLine.addParameter("--project=" + appEngineHelper.getProjectId());
     commandLine
         .addParameter("--credential-file-override=" + credentialsPath.getAbsolutePath());
+    commandLine.addParameter("--quiet");
     commandLine.withParentEnvironmentType(ParentEnvironmentType.CONSOLE);
     commandLine.getEnvironment().put("CLOUDSDK_METRICS_ENVIRONMENT", "gcloud-intellij");
     commandLine.getEnvironment().put("CLOUDSDK_APP_USE_GSUTIL", "0");
```
chanseokoh: A random question: what is the effect/purpose of CLOUSDK_APP_USE_GSUTIL=0? It seems related to service-side metrics.

etanshaul: @chanseokoh - we added it specifically for the purposes of using the credential file override. See #562 for more info.

etanshaul:
Ready for review. After chat with @arouzrokh, the "module" wording is being changed to "services". I will introduce a separate PR to update this across our entire plugin.

patflynn:
I thought that modules stop required a module name and version

etanshaul:
Its using a "service" and a version now. Service is effectively the same thing as a module.

patflynn:
I guess what's confusing is the swapping of subcommand from module to
versions and then the service name being passed as a flag.

If it's just a rename from module to service should it go from:
gcloud preview app modules stop default --version 123
to
gcloud preview app services stop default --version 123
?

In this CL we seem to be switching it to:
gcloud preview app versions stop 123 --service default

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStopAction.java
```diff
@@ -65,11 +65,14 @@ public void run() {
...
         appEngineHelper.getGcloudCommandPath().getAbsolutePath());
 
-    commandLine.addParameters("preview", "app", "modules", "stop");
+    commandLine.addParameters("preview", "app", "versions", "stop");
+    commandLine.addParameters(versionToStop);
+
+    commandLine.addParameter("--service");
     for(String module : modulesToStop) {
       commandLine.addParameter(module);
     }
```
chanseokoh: Is it possible to give multiple modules (services) after `--service`? It doesn't seem so by the help message from `gcloud preview app versions stop --help`. (However, multiple versions are okay.)

etanshaul: Yea good catch. It looks like only a single service can be given to the stop command. However, the deploy command looks like its still set up to return multiple modules (services). So I need to determine if this would ever happen, and if so how do we deal with it.

etanshaul:
@patflynn you are right that the command is changing form too (aside from just swapping module and service):

`gcloud preview app versions stop [versions] --service [service]`

patflynn:
That is weird man.

etanshaul:
The positioning of the --service flag could be moved too (or even omitted). The following are equivalent to what I showed in the previous comment:

`gcloud preview app versions stop --service [service] [versions]`

`gcloud preview app versions stop [service]/[version]` (for stopping just 1 version)

chanseokoh:
Perhaps with the second form we can stop multiple service/version pairs. But since we always deploy one "deployable" (which I believe corresponds to "service"), I think we always get one service/version pair when we deploy an app from our tool.

etanshaul:
Interestingly I tried that resource path format but could not get the command to execute. I consistently get back:

`WARNING: No matching versions found.`

 @arouzrokh - do you know if this feature is still in the works?

patflynn:
it's all crystal clear to me now. LGTM!

etanshaul:
Update logic to accept only a single service output from the deployment (and throwing an exception otherwise).

patflynn:
🚢

Issue #656, opened 2016-04-26
Enable Google code style checking


meltsufin:
See: GoogleCloudPlatform/app-tools-lib-for-java#17

PR #659, opened 2016-04-28
Adding checkstyle to build

etanshaul:
see #656 

Testing build with checkstyle enabled. 
Checkstyle warnings have NOT been addressed yet so this is expected to fail.


etanshaul:
@loosebazooka @meltsufin @patflynn 

Before I spend time fixing all of checkstyles complaints (and there are a lot of them) could you take a quick look to see if this setup looks good to you? I'll go ahead and fix its complaints afterwards in this PR.

etanshaul:
Note the travis build is failing as expected due to checkstyle violations

build.gradle
```diff
@@ -116,6 +116,38 @@ subprojects {
...
+    }
+
+    tasks.withType(Checkstyle).each { checkstyleTask ->
+        checkstyleTask.doLast {
+            reports.all { report ->
+                if(report.name == "xml") {
+                    def outputFile = report.destination
+                    if (outputFile.exists() && outputFile.text.contains("<error ")) {
+                        throw new GradleException("There were checkstyle warnings! For more info check $outputFile")
+                    }
```
chanseokoh: Aha, I see. So, we want to discuss if PMD and FindBugs should also fail locally?

etanshaul: Yea that was the idea.

loosebazooka: I don't know if I would submit a CL that broke travis. Perhaps defer throwing an exception till all style violations are fixed, and then enable fail hards after?

build.gradle
```diff
@@ -116,6 +116,38 @@ subprojects {
...
+    ]
+    dependencies {
+        checkstyleConfig ("com.puppycrawl.tools:checkstyle:${versions.checkstyle}") {
+            transitive = false
+        }
+        checkstyle "com.puppycrawl.tools:checkstyle:${versions.checkstyle}"
+    }
+    checkstyle {
+        config = resources.text.fromArchiveEntry(configurations.checkstyleConfig, 'google_checks.xml')
+        ignoreFailures false
```
loosebazooka: What's ignoreFailures do? Is is different than your exception throwing below?

build.gradle
etanshaul: My plan was to fix the checkstyle issues in this PR before submitting it. I was just looking for some eyes on the config before I spend all the time needed to fix the checkslye warnings.

build.gradle
etanshaul: Yea you're right that it makes sense that ignoreFailures wouldn't be needed with that explicit exception workaround - but I tried removing it or setting it to true and the gradle check task hangs.

etanshaul: Never mind - I had a problem locally with my gradle process. ignoreFailures can be removed.

chanseokoh: I wonder if this can be the other way around: use "ignoreFailures false" and remove throwing an exception explicitly. This "ignoreFailures" worked nicely with PMD and FIndBugs.

etanshaul:
Added comment to explain the workaround for failing the build on checkstyle warnings.

build.gradle
chanseokoh: Aha, just was your last comment.

build.gradle
patflynn: I'm down with this change as is.  One option if you're hesitant on failing locally the build due to PMD,findbugs,checkstyle etc, is to set it up as a pre-commit or pre-push hook.

etanshaul: I discussed with with @meltsufin and @chanseokoh yesterday and I think the best option is to fail the build locally. This way any of those verification problems are discovered early.

patflynn: fine with me.

patflynn: just one note: I'd prefer the style changes and the configuration of the plugin were separate PRs.

etanshaul: Oh ok. In that case, I won't merge this until a separate PR is merged to fix the violations.

Issue #655, opened 2016-04-26
Use gcloud auto generated version


etanshaul:
Now that gcloud's deploy output includes the version ID of the deployment, we can use this instead of generating our own. (see internal #120830848)

chanseokoh:
Closed by #660.

chanseokoh:
Note: the new `gcloud` is now out. If users update `gcloud`, they will see the new JSON format.

PR #660, opened 2016-04-29
Parse version/service and remove generateVersion()

chanseokoh:
See #655.

Note that this should not be merged until the internal #120830848 makes to public.

This PR
- removes generateVersion()
- parses the JSON output from `gcloud app deploy` to get version and service info (this is a bit ugly)

Unlike JavaScript, parsing JSON in Java was a bit of a pain. Because we don't know the exact format/type of the JSON output and it may change in the future, I took an approach of manually going over each JSON element to try best to extract the version/service info. I wonder if there is a nicer way.

An example output:
```json
        {
          "configs": [],
          "versions": [
            {
              "id": "20160429t112518",
              "last_deployed_time": null,
              "project": "springboot-maven-project",
              "service": "default",
              "traffic_split": null,
              "version": null
            }
          ]
        }
```

Probably, I should move `parseDeployOutputToModule()` to the `AppEngineUtil` class and write a test.

Tried basic testing, and undeploying (stopping) works correctly.


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
```diff
@@ -247,15 +241,72 @@ public void run(@NotNull ProgressIndicator indicator) {
...
+    */
 
-    if(deployOutput == null || deployOutput.keySet().size() != 1) {
-      throw new AssertionError("Expected a single module output from flex deployment.");
+    JsonReader reader = new JsonReader(new StringReader(jsonOutput));
+    try {
+      reader.beginObject(); // Top-level is a single object.
+
+      // Look for "versions" in the top-level object.
+      while (reader.hasNext()) {
```
etanshaul: Might be cleaner if instead you create a separate object respresenting the deploy output that you can marshal the json to.

chanseokoh: That would be nice, actually. My only concern is the possibility of future changes to the JSON output. Will that kind of marshallable object resilient/flexible enough to adapt changes in JSON. I hadn't tested it, but it feels like if anything is out of the type slightly, Gson will fail.

Related to that, it is not clear, for example, if `"traffic_split": null` will have a string type or may have an array.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
```diff
@@ -247,15 +241,72 @@ public void run(@NotNull ProgressIndicator indicator) {
...
+              String nested_name = reader.nextName();
+              if ("id".equals(nested_name)) {
+                versionService.version = reader.nextString(); // "id" found. (String)
+              } else if ("service".equals(nested_name)) {
+                versionService.service = reader.nextString(); // "service" found. (String)
+              } else {
+                reader.skipValue(); // Skip a value if not "id".
+              }
+            }
+            reader.endObject();
```
chanseokoh: I feel like this will fail if "versions" contain multiple elements. I will need to fix this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
etanshaul: We don't need the traffic split info right? So that one can be ignored. 

Wouldn't you're current setup also fail if they change the output? E.g. you expect a versions at the top level and then look inside for specific attributes; if this changes then you wouldn't be able to extract the output any more.

chanseokoh: I see. So I can just omit fields that I am interested in the marshallable object, right? That will certainly do.

etanshaul: I'm not sure if you omit them or if you control it via gson annotations in your "domain" object that you create to represent the deploy output.

chanseokoh:
- Did an overhaul to use a JSON-matching class for deserialization as @etanshaul suggested. Now this is much better. It resulted in a few FindBugs warnings of accessing seemingly uninitialized values of this JSON-matching class. I will think about what I can do with the warnings.

- Also moved the static JSON-parsing function to `AppEngineUtil`.

- Added parsing tests.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
```diff
@@ -16,34 +16,64 @@
...
 public class AppEngineUtil {
 
+  // Just to return two strings.
+  public static class VersionService {
+    public String version = null;
+    public String service = null;
+  }
+
+  // Holds de-serialized JSON output of gcloud app deploy.
+  private static class DeployOutput {
```
etanshaul: Having these objects as static classes in AppEngineUtil doesn't feel fight. I would suggest extracting them out into their own class (you can still keep them a single file if you prefer).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
```diff
@@ -16,34 +16,64 @@
...
+import java.lang.reflect.Type;
+import java.util.List;
 
 /**
  * App Engine utility methods
  */
 public class AppEngineUtil {
 
+  // Just to return two strings.
+  public static class VersionService {
```
etanshaul: I don't think you need this. If you like, you could make a helper getters in your DeploymentOutput class to give you these values.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
```diff
@@ -16,34 +16,64 @@
...
+    public String version = null;
+    public String service = null;
+  }
+
+  // Holds de-serialized JSON output of gcloud app deploy.
+  private static class DeployOutput {
+    private static class versionElement {
```
etanshaul: Capitalization. How about just `Version` instead of VersionElement? Also the nesting here feels a bit odd.

etanshaul: The nesting might be ok actually. I'd just consider if AppEngineUtil is the right place for this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
chanseokoh: Yeah, I want to avoid VersionService too. I will add getters in DeployOutput.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
chanseokoh: Yeah, the parsing function is too-specific. Where could be a good place for this? Going back?

etanshaul: I don't know if it makes sense for it to be static since its related to an instance of the deployment. I think it would be fine to put your updated gson logic back into a private method of the DeployAction class.

chanseokoh:
Come to think of it, a fix to this #655 is tricky to roll out. Unless a user upgrades our plugin and gcloud at the same time, things will not work. We are not in an ideal world, so probably this PR needs to attempt to parse JSON first with the old format and later with the new format.

etanshaul:
We already face this situation with the cloud sdk version. If the deployment fails (for whatever reason) we print out a message suggesting to the user to upgrade to the latest version. See AppEngineDeployAction#processTerminated

chanseokoh:
Or, since everything is BETA, we could just not care about breakages caused by using wrong gcloud.

chanseokoh:
Okay, I will do the same thing to print out an upgrade message.

etanshaul:
You shouldn't have to. Its already handled. Just make sure the deployment fails if the output isn't what you expected.

chanseokoh:
I noticed that this parsing routine is executed only when you click 'Undeploy'. Deployment can go fine but parsing can fail.

etanshaul:
Ah you're right. Its a little tricky then because this parsing occurs outside the context of a command line process.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
```diff
@@ -184,11 +182,9 @@ public void processTerminated(final ProcessEvent event) {
 
   private class DeploymentRuntimeImpl extends DeploymentRuntime {
```
patflynn: could this be broken out and named something like AppEngineDeploymentRuntime?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
```diff
@@ -221,18 +217,19 @@ public void run() {
     }
 
     private void stop(@NotNull UndeploymentTaskCallback callback) {
-      String moduleToStop;
+      DeployOutput deployOutput;
       try {
-        moduleToStop = parseDeployOutputToModule(deploymentOutput);
```
patflynn: what do you think about parsing this in the constructor?

etanshaul: Not that performance is really a concern here but just to explain the original reason I did this: having it in the stop() method lazily does the json parsing (doesn't happen until the user actually tries to undeploy which may never happen in the first place).

patflynn: Yeah that's reasonable.  But if our parsing of the deployment result is
broken, I would prefer that we broke immediately, rather than lazily during
stop.  For one we'd be much more likely to notice the breakage.

etanshaul: Yea. That sounds good to me.

chanseokoh: SGTM. I'll move it to the constructor. The constructor will not throw any exception.

chanseokoh:
Notable changes since last commit:
- Refactored DeploymentRuntimeImpl to be a top-level class.
- Moved parsing to the constructor.
- If the JSON parsing fails, the 'Undeploy' menu item will be disabled (grayed out).
- Added one more test case for parsing.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployActionTest.java
```diff
@@ -82,16 +80,12 @@ public void testCreateApplicationDefaultCredentials() throws Exception {
...
     assertEquals(clientId, jsonMap.get("client_id"));
     assertEquals(clientSecret, jsonMap.get("client_secret"));
     assertEquals(refreshToken, jsonMap.get("refresh_token"));
     assertEquals("authorized_user", jsonMap.get("type"));
-  }
-
-  @After
-  public void deleteCredentialFile() {
     credentialFile.delete();
   }
```
chanseokoh: This change is necessary because `@After` method is called for each `@Test`.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRuntimeImpl.java
```diff
@@ -0,0 +1,181 @@
...
+  // If failed to extract service/version info for the deployed app, undeploy will not be enabled.
+  @Override
+  public boolean isUndeploySupported() {
+    return deployOutput != null;
+  }
+
+  @Override
+  public void undeploy(@NotNull final UndeploymentTaskCallback callback) {
+    if (deployOutput == null) {
+      callback.errorOccurred(GctBundle.message("appengine.stop.modules.version.execution.error"));
```
chanseokoh: This check is probably not necessary since 'Undeploy' will be disabled and users will not be able to get into this function if `deployOutput == null`. (See the above `isUndeploySupported()` function.) I left the check here just in case.

chanseokoh:
Tested this with both on the public gcloud (old format) and the internal gcloud (new format). Undeploy works correctly with the new JSON format.

I still need to address issues reported by FindBugs that is confused by Gson's magical field initialization (i.e., de-serialization). I will have to use FindBugs annotations.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRuntimeImpl.java
```diff
@@ -0,0 +1,184 @@
...
+    this.project = project;
+    this.appEngineHelper = appEngineHelper;
+    this.loggingHandler = loggingHandler;
+
+    try {
+      deployOutput = parseDeployOutputToService(deploymentOutput);
+    } catch (JsonParseException e) {
+      loggingHandler.print(
+          GctBundle.message("appengine.deployment.version.extraction.failure") + "\n");
+      Logger.getInstance(AppEngineDeploymentRuntimeImpl.class).warn(
```
patflynn: Loggers should be statically initialized at the top of the class. 

Also I think we probably want to have this kind of error reported right?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRuntimeImpl.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.List;
+
+import javax.swing.SwingUtilities;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+/**
+ * Takes care of undeploy and stop of App Engine applications.
+ */
+class AppEngineDeploymentRuntimeImpl extends DeploymentRuntime {
```
patflynn: you don't need to call this ..Impl anymore.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRuntimeImpl.java
```diff
@@ -0,0 +1,184 @@
...
+    this.loggingHandler = loggingHandler;
+
+    try {
+      deployOutput = parseDeployOutputToService(deploymentOutput);
+    } catch (JsonParseException e) {
+      loggingHandler.print(
+          GctBundle.message("appengine.deployment.version.extraction.failure") + "\n");
+      Logger.getInstance(AppEngineDeploymentRuntimeImpl.class).warn(
+          "Could not retrieve service/version info of deployed application", e);
+      deployOutput = null;
```
patflynn: if parsing failed, this object seems to be fundamentally useless. I'd rather we didn't construct it at all. What do you think?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRuntimeImpl.java
chanseokoh: I'll move the logger.

If you mean reporting as a crash report, in the long run that would be the right direction. But we will have a lot of such (false) reports when users upgrade our plugin while they don't upgrade gcloud (unless we put a special logic to identify and exclude errors caused by the old gcloud format).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRuntimeImpl.java
```diff
@@ -0,0 +1,184 @@
...
+            null) {
+          @Override
+          public void run(@NotNull ProgressIndicator indicator) {
+            ApplicationManager.getApplication().invokeLater(appEngineStopAction);
+          }
+        });
+  }
+
+  @VisibleForTesting
+  static DeployOutput parseDeployOutputToService(String jsonOutput) throws JsonParseException {
```
patflynn: drop the ..toService.  parseDeployOutput(jsonOutput). It's cleaner.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRuntimeImpl.java
etanshaul: This is a little subtle:

If the deploy succeeds, but the parsing fails, then the only way to indicate a successful deploy is by constructing that class (and passing it to `callback.succeeded`). 

This _shouldn't_ happen, but I can imagine a case where the deploy succeeded, but gcloud returns an unexpected deploy output -> in this case I think we'd want to still indicate a successful deploy.

etanshaul: One improvement would be to move the parsing to the DeployAction. Then if its successful construct the Runtime with undeploy enabled (and disabled if not). 

We still have that useless Runtime but thats the pattern that IJ uses: even if undeploy was not supported at all we still have to construct it and override `isUndeploySupported` because of the callback api.

chanseokoh: I agree on that JSON parsing failure is still a successful deployment, because it is true that the app is actually deployment and running in this case.

As @etanshaul said, even if we move the parsing to DeployAction, we will still have to construct (possibly another) DeploymentRuntime anyway to override `isUndeploySupported` to return false. In this sense, I think it is not a bad idea to do parsing in this constructor. Ideas?

patflynn: I think etan's suggestion to do the parsing in the action and construct an
appropriate runtime based off of the result of that could work.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRuntimeImpl.java
chanseokoh: Interesting! Anymore... means, like, since Java 1.7? Or like, because it is no longer a nested function?

patflynn: no, what I meant was that the ..Impl  in the class name isn't necessary now that you've qualified it as an AppEngineDeploymentRuntime.  You can just call it that.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRuntimeImpl.java
etanshaul: @chanseokoh just noting that there's no need to construct a different Runtime depending on if isUndeploySupported. You can just pass in a boolean to the constructor.

As far as I can see, you can still do the parsing in the deploy action.

chanseokoh: @etanshaul Aha, nice catch.

chanseokoh:
- Parsing is done in `AppEngineDeployAction`. Moved tests as well.
- `DeployOutput` Gson-object methods can now return null.
- Renamed to `AppEngineDeploymentRuntime`.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRuntimeImpl.java
chanseokoh: Done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
```diff
@@ -167,7 +155,20 @@ public void onTextAvailable(ProcessEvent event, Key outputType) {
...
       try {
         if (event.getExitCode() == 0) {
-          callback.succeeded(new DeploymentRuntimeImpl(deploymentOutput.toString(), version));
+
+          // Parse JSON output to retrieve service/version info.
+          DeployOutput deployOutput = null;
+          try {
+            deployOutput = parseDeployOutput(deploymentOutput.toString());
+          } catch (JsonParseException e) {
+            getLoggingHandler().print(
```
etanshaul: you can use consoleLogLn here instead. Also, did we settle on not being notified on this kind of failure (I don't think we get a report on logger.warn).  Otherwise, lgtm.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRuntime.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  @Override
+  public boolean isUndeploySupported() {
+    return service != null && version != null;
+  }
+
+  @Override
+  public void undeploy(@NotNull final UndeploymentTaskCallback callback) {
+    if (!isUndeploySupported()) {
```
etanshaul: oh, you can get rid of this block since it can never be executed.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
patflynn: I think it's OK to error here.  You could conceivably explain the error and suggest their might be an incompatibility with gcloud that could be fixed with a component update.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRuntime.java
```diff
@@ -0,0 +1,112 @@
...
+    this.project = project;
+    this.appEngineHelper = appEngineHelper;
+    this.loggingHandler = loggingHandler;
+    this.service = service;
+    this.version = version;
+  }
+
+  @Override
+  public boolean isUndeploySupported() {
+    return service != null && version != null;
```
patflynn: should this be OR and not AND?

chanseokoh: Both of them are required to undeploy an app, right?

etanshaul: Yes, for our purposes we want both. I think you had it correct. But see my other comment about the possible bug.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
```diff
@@ -167,7 +155,20 @@ public void onTextAvailable(ProcessEvent event, Key outputType) {
...
+                GctBundle.message("appengine.deployment.version.extraction.failure") + "\n");
+            logger.warn("Could not retrieve service/version info of deployed application", e);
+          }
+
+          callback.succeeded(
+              new AppEngineDeploymentRuntime(project, appEngineHelper, getLoggingHandler(),
+                                             deployOutput.getService(), deployOutput.getVersion()));
```
etanshaul: wouldn't this throw an NPE if the parsing fails since deployOutput will be null?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRuntime.java
patflynn: ah I had this misread.  My brain expects to see !(service == null || version == null). This is fine though.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
chanseokoh: Wow.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
chanseokoh: For both `consoleLogLn()` and `logger.error()`, they actually don't need text localization. However, not using `GctBundle.message()` results in embedding a very long string in the code such as below. It will also make a bit difficult to maintain consistent messaging if it's outside the resrouce bundle.

"Failed to parse service/version info from the ''gcloud'' deployment output. \nPlease make sure that you are using the latest version of the Google Cloud SDK.\nRun ''gcloud components update'' to update the SDK. (See: https://cloud.google.com/sdk/gcloud/reference/components/update.)"

Suggestions?

patflynn: I think what you have there is fine.

chanseokoh:
- `consoleLogLn`, and print the message whenever it fails to get version and service.
- `logger.error` instead of `logger.warn`. This allows user bug reporting.
- No unnecessary check at the beginning of `AppEngineDeploymentRuntime.undeploy()` as @etanshaul suggested.

When using the current `gcloud`:
old_gcloud
old_gcloud_events

When using the new `gcloud`:new_gcloud

patflynn:
If all comments are addressed: :ship:

PR #671, opened 2016-05-03
Replace deprecated FindBugs annotation with a recommended one

chanseokoh:
We can now use the new FindBugs annotation `SuppressFBWarnings` in place of the deprecated `SuppressWarnings`, whose name overlaps with the Java-standard `SuppressWarnings`.


patflynn:
lgtm

PR #672, opened 2016-05-03
Addressing checkstyle issues

etanshaul:
See #656 

Still pending fixes from account plugin module.


patflynn:
LGTM

etanshaul:
@chanseokoh lets try and get this in - all warnings are accounted for and its up to date with master. Obviously it can't be reviewed line by line, but most of it was automated. Maybe you can help do a couple sanity checks by running the plugin.

See #659 which will be merged in right after this one. Note that I ended up including the google checkstyle config file in our build (rather than pulling from the jar) because I needed to add a couple of suppressions (which isn't possible, as far as I could tell, without modifying the config).

chanseokoh:
com.google.cloud.tools.intellij.login.IntellijGoogleLoginService:265 - Linebreaking is not correct.

com.google.cloud.tools.intellij.login.CancellableServerReceiver:87 - Single <p/> seems sufficient.
com.google.cloud.tools.intellij.login.CancellableServerReceiver:193 - ditto.

com.google.cloud.tools.intellij.login.ui.GoogleLoginAction:57 - There should be a space between `(JComponent)` and `event`.

com.google.cloud.tools.intellij.appengine.cloud.AppEngineDeploymentConfigurator:88 - Linebreaking is not correct.

com.google.cloud.tools.intellij.appengine.cloud.AppEngineDeploymentRunConfigurationEditor:168 - My logic is that the second `.equal` should be intended.

com.google.cloud.tools.intellij.appengine.util.PsiUtils:51, 68, 83, 96, 111, 121, 134, 144 - `.` after the first comment sentence.

I will look at the rest and get back to you soon.

etanshaul:
Thank you for actually look through these, I know it doesn't get more boring than this. I agree with most of your comments so far - sometimes the auto-formatting isn't ideal (even if it passes checkstyle).

etanshaul:
comments so far addressed

chanseokoh:
Let's just get this in. From what I have checked so far, it seems safe.

etanshaul:
Agreed.

PR #674, opened 2016-05-10
Remove direct use of cloud sdk and use the common app-tools library

### LINKED ISSUE #669: Switch over to use the common lib api for app engine deployment operations
Adding this for discussion. We may want to hold off on this until its farther along.

etanshaul:
fixes #669 

Still pending publishing of latest common app-tools jar (as this PR depends on the latest changes) as well as support for setting environment variables in app-tools.

@chanseokoh you can start looking


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
```diff
@@ -66,47 +63,47 @@ public AppEngineAction(
...
+    this.processRunner = processRunner;
 
     credentialsPath = createApplicationDefaultCredentials();
     if (credentialsPath == null) {
       callback.errorOccurred(
-          GctBundle.message("appengine.deployment.credential.not.found",
+          GctBundle.message("appengine.action.credential.not.found",
               appEngineHelper.getGoogleUsername()));
-      return;
+      return null;
```
chanseokoh: I checked that if we fail here and return null, a `NullPointerException` will be propagated out of the Cloud SDK library (by `Preconditions.checkNotNull()`).

etanshaul: yes thanks. if the Sdk can't be created then I shouldn't be executing the command. I'll update this and look into some test coverage.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
```diff
@@ -138,22 +138,26 @@ protected File createApplicationDefaultCredentials() {
...
-    File tempCredentialFilePath = null;
+    File tempCredentialFilePath;
     try {
       tempCredentialFilePath = FileUtil
           .createTempFile(
               "tmp_google_application_default_credential",
               "json",
               true /* deleteOnExit */);
       Files.write(jsonCredential, tempCredentialFilePath, Charset.forName("UTF-8"));
+      credentialsPath = tempCredentialFilePath;
```
chanseokoh: Not a big deal, but I just wonder why you added this here as `credentialsPath` will be set anyway after returning from this function.

etanshaul: No you're right - it was left over by mistake from some refactoring. Updated; also updated error handling based on your other comment.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
```diff
@@ -120,90 +123,115 @@ public void run() {
...
-                  project, appEngineHelper, getLoggingHandler(),
-                  deployOutput != null ? deployOutput.getService() : null,
-                  deployOutput != null ? deployOutput.getVersion() : null));
-        } else if (cancelled) {
-          callback.errorOccurred(GctBundle.message("appengine.deployment.error.cancelled"));
-        } else {
-          logger.warn("Deployment process exited with an error. Exit Code:" + event.getExitCode());
-          callback.errorOccurred(
-              GctBundle.message("appengine.deployment.error.with.code", event.getExitCode()));
+    processRunner.setExitListener(new ProcessExitListener() {
```
chanseokoh: You created the StopExitListner class for the stop action but not for this. Thoughts?

chanseokoh:
Just made a comment. Other than that, LGTM :+1:

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
etanshaul: Yes. Valid point. For deploy, the stop listener would need access to the raw deploy output and I was running into some weird behavior passing in a reference to it (since its assembled by callbacks) so I left it with the intention of getting back to it.  I'll address this since it helps readability to extract it out.

etanshaul:
Updated based on your last comments. Thanks for the review. 

I'll merge this in once the common lib is updated with last feature we need to support this PR (https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/issues/39#issuecomment-218308024) and when a new jar is published.

etanshaul:
@chanseokoh please take a quick look. Changes since last time:

- No longer pulling from local maven repo
- Updated our actions according to the updated app-tools api
- Ability to terminate a running a process (needed to stop a deployment) got lost in the shuffle from app-tools; I will create an issue here to track this

chanseokoh:
I'm looking at it.

One thing I noticed while I tried deploying an app. Previously, the console was printing the message `Updating service [default]...` while it was waiting for the deployment to complete. It no longer prints that. Instead, the second last line is `DONE` (from the Docker build).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
```diff
@@ -44,72 +43,74 @@
...
+   * Kill any executing process for the action.
+   */
   protected void cancel() {
-    // kill any executing process for the action
-    if (processHandler != null
-        && !processHandler.isProcessTerminating()
-        && !processHandler.isProcessTerminated()
-        && processHandler.getProcess() != null) {
+    if (processRunner != null
+        && processRunner.getProcess() != null) {
```
chanseokoh: This fits in a line.

etanshaul:
This is what I'm seeing on the output (are you seeing something else?):

image

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
etanshaul: updated

chanseokoh:
Ah, sorry. I mean, eventually I see what you see, but when it is taking a long time while waiting, it appears to be done after printing 'DONE'.  After taking a long time, it eventually prints out the rest of the output starting with `Updating service [default]...`. This is at least happens on my laptop. Can you check this again?

chanseokoh:
It is actually not a big deal since the icon of the deployment will show that it is running.

etanshaul:
Oh yea - I did see that too. I don't know yet if it's due to an update in gcloud or if the process output is sent to the outputLineListener line by line only after the next line comes in.

chanseokoh:
We've been though this PR, and it LGTM.

We shouldn't forget to set the ENVIRONMENT_VERSION, so I'll create an issue for that.

etanshaul:
yes, please do.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java
```diff
@@ -16,6 +16,11 @@
 
 package com.google.cloud.tools.intellij.appengine.cloud;
 
+import com.google.cloud.tools.app.api.AppEngineException;
+import com.google.cloud.tools.app.impl.cloudsdk.internal.process.DefaultProcessRunner;
```
patflynn: if these are all 'impl' packages it's confusing why clients need them.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
```diff
@@ -180,34 +210,33 @@ public void processTerminated(final ProcessEvent event) {
...
+  }
+
+  /**
+   * Parse the raw json output of the deployment.
+   *
+   * @return an object modeling the output of a deploy command
+   * @throws JsonParseException if unable to extract the deploy output information needed
+   */
   @VisibleForTesting
   static DeployOutput parseDeployOutput(String jsonOutput) throws JsonParseException {
```
patflynn: it feels like this should be pushed into the common library.

PR #678, opened 2016-05-13
Support custom gcloud deploy flags

### LINKED ISSUE #530: should we have a flags field in mvm deployment run configs?

chanseokoh:
Fixes #530.

This is a crude MVP (minimum viable product), adding a single field.

I observed two problems:

- It's visible only when selecting 'Custom'. This itself isn't a problem, but users who wish to use custom flags are also required to specify `app.yaml` and `Dockerfile` together.
- Say a user provided custom flags and clicked on 'Apply'. The flags are now saved in the configuration. The user changes her mind and selects 'Auto'. She no longer sees her previous flags. (The flags have been saved and not cleaned, so if she selects 'Custom' again, the saved flags will be seen in the field.) However, if she runs this 'Auto' deployment, the flags hidden but now shown will be used.

I particularly see the latter problem confusing. Note that this behavior is consistent with how `app.yaml` and `Dockerfile` are saved.


chanseokoh:
In fact, for the second problem above, a user doesn't have to click 'Apply'.

etanshaul:
We should probably hold off on merging this until #674 is merged in, because that PR transitions us to using the app-tools common lib which doesn't support passing in arbitrary flags (yet, if ever).

loosebazooka:
we should probably do something about arbitrary flags, given that it'd be
difficult to really fully expose the full set of flags from gcloud.

etanshaul:
cool: https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/issues/70

chanseokoh:
Thanks. Let's hold off this until GoogleCloudPlatform/app-tools-lib-for-java#47 is done.

We could at least discuss if it is a good idea to put the field under 'Custom'.

chanseokoh:
Closing this for now. I may reopen this later, but that seems a long way down.

PR #683, opened 2016-05-16
Restore user specified source path when IDE init

### LINKED ISSUE #682: UserSpecifiedPathDeploymentSource is not properly de-serialized when the IDE starts up
I will fix it now.

chanseokoh:
Fixes #682.

The file path value was previously empty. So, after restarting IJ, clicking on the play button for a saved deploy config that uses a custom WAR/JAR was failing.

(However, interestingly, the value was displayed correctly when a config window was opened.)


etanshaul:
lgtm

PR #684, opened 2016-05-16
Fix deployment config validation order

etanshaul:
fixes #679 

Performs validation before saving the settings.

Note that doesn't affect the state of the deployment source when clicking 'ok' / 'apply' with a validation error. This only affects the portion of the deployment config where an extension point is provided, e.g. everything other than the name, server, and deployment.

PR #685, opened 2016-05-16
Validation order

### LINKED ISSUE #679: Deployment settings are lost permanently when changing a deployment source and clicking 'Apply' or 'Ok'
Steps to reproduce:

1. Create a deploy configuration. Provide some valid settings, and close the config window.
2. Open the config window again.
3. Change a deployment source (e.g., from Maven source to file system source).
 - At this point, all settings will be reset. However, if we click 'Cancel', they values will be restored when we open the config window later.
4. Click 'Ok' or 'Apply'. This will trigger validation error. Click 'Cancel' to close the window.
5. Open the window again. Deployment source selection has been changed, and all the settings are lost.

etanshaul:
fixes #679
migrated from  #684

Performs validation before saving the settings.

Note that doesn't affect the state of the deployment source when clicking 'ok' / 'apply' with a validation error. This only affects the portion of the deployment config where an extension point is provided, e.g. everything other than the name, server, and deployment.


chanseokoh:
What a bummer. Looks like there shouldn't exist my merge commit.

chanseokoh:
Squashing the commits in the old way may fix this?

etanshaul:
Yep. That was my next thing to try.

PR #694, opened 2016-05-19
Migrate sending metrics to Cloud-platform-wide Google Analytics

### LINKED ISSUE #693: Additionally send usage data to the unified internal metrics collection system
We currently send our usage data to a GA property exclusive to IntelliJ. Ultimately, we want to migrate this traffic to the centralized internal metrics collection system, where analysts can query various Cloud-related metrics across all products, sites, marketing campaigns, documents, etc.

Until almost all of our plugin users complete upgrade, we need to send usage data to both the current IntelliJ-specific Google Analytics (GA) property and the centralized GA property during the transition time. Eventually, we will turn down the IntelliJ-specific GA property.

chanseokoh:
Fixes #693.

I tried to re-use the existing facilities for sending usage data as much as possible. In the long run, we would probably want to overhaul this.

I will be doing actual ping tests in various ways.

### DIFF
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
@@ -56,7 +56,7 @@ public class GoogleUsageTracker implements UsageTracker {
     this.analyticsId = UsageTrackerManager.getInstance().getAnalyticsProperty();
   }
 
-  private static final List<? extends NameValuePair> analyticsBaseData = ImmutableList
+  private static final List<BasicNameValuePair> analyticsBaseData = ImmutableList
       .of(new BasicNameValuePair("v", "1"),         // Protocol version
           new BasicNameValuePair("t", "pageview"),  // Note the "pageview" type, not "event"
           new BasicNameValuePair("ni", "0"),        // Non-interactive? Report as interactive.
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
@@ -16,16 +16,14 @@
 
 package com.google.cloud.tools.intellij.stats;
 
-import com.google.common.base.Strings;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 
 import com.intellij.ide.util.PropertiesComponent;
-import com.intellij.openapi.application.ApplicationInfo;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.updateSettings.impl.UpdateChecker;
-import com.intellij.util.PlatformUtils;
 
 import org.apache.http.NameValuePair;
 import org.apache.http.StatusLine;
@@ -36,12 +34,10 @@ import org.apache.http.client.utils.HttpClientUtils;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.message.BasicNameValuePair;
-import org.jetbrains.annotations.NonNls;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.List;
 
 /**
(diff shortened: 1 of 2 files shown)


chanseokoh:
@loosebazooka Can you also look at `build.gradle`? I did not really understand why `input.property()` was used previously; it was just passing an independent map to `expand()`.

patflynn:
why not just switch entirely?

chanseokoh:
@patflynn Actually, I would love to stop sending usage information to the old (i.e., current) GA right away. However, if we decided to send pings to the new GA property only, we will see partial traffic coming only from users who updated our plugin. If this is ok, I can really gut out the old code and start fresh. (Perhaps we could aggregate the old GA and the new GA traffic at the dashboard level though. Not tried this yet.)

patflynn:
Yeah let's not keep the old code path. Updated plugins will go to the new
project, and non-updated plugins will go to our old project. That's fine.

chanseokoh:
Question: if on Android Studio, will we send pings only to the Android GA property, or to both Android GA and Cloud GA?

chanseokoh:
To clarify my last question, my understanding is that the Android Studio team has their own GA property (which I am not aware of and have no access to). @loosebazooka, is this correct? And, probably, every metrics event happening from our code on Android Studio goes to that GA property, doesn't it? So, the answer to my last question is probably 'No, and keep sending pings to Android GA only', right?

For IntelliJ, we also have our own GA property currently, but we will abandon it and migrate to use the unified Cloud GA property.

chanseokoh:
@patflynn @etanshaul PTAL.

It turns out that we cannot just get rid of the old code since our code may be synced with Android Studio later. We decided not to change the `trackEvent()` interface at the moment. Considering all this, I think it is best in many aspects to just modify existing code a bit to achieve our goal.

To recap, we will now send pings to the Cloud-platform-wide GA Property. Consequently, users who don't upgrade will continue to send pings to the old (i.e., current) GA Property, which we will eventually deprecate.

etanshaul:
Is there some background info on this you could point me to if you'd like me to review this? I don't have the info on what needs to be collected and where it needs to be sent.

chanseokoh:
I think for efficiency and productivity, it's best for you to review just the syntactic structure and programmatic soundness here. I'll ensure the correctness of the functionality of sending metrics to the Cloud GA. I tested and verified that I can see my pings appearing in the target GA Property.

However, if you really want to go deep in reviewing, I think you can take a look at [this implementation](https://github.com/google/cloud-reporting/blob/master/src/main/java/com/google/cloud/metrics/MetricsUtils.java#L183). I basically referenced that implementation along with [the GA protocol and parameter reference].(https://developers.google.com/analytics/devguides/collection/protocol/v1/reference).

chanseokoh:
And feel free to raise any suspicion or ask any question if are curious about anything, even if it's not related to the review task.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
```diff
@@ -36,64 +34,76 @@
...
+   * @param eventAction On IntelliJ, this is the primary event action.
+   * @param eventLabel On IntelliJ, reported as a metadata key.
+   * @param eventValue On IntelliJ, reported as a metadata value.
+   */
   @Override
   public void trackEvent(@NotNull String eventCategory,
       @NotNull String eventAction,
       @Nullable String eventLabel,
       @Nullable Integer eventValue) {
+    if (eventValue != null) {
```
etanshaul: whats the point of this if statement if you are using preconditions#checkArgument right after?

etanshaul: maybe the better question is what was your intent here? the if statement wrapping the precondition check is confusing.

chanseokoh: I could then do `checkArgument(eventValue == null || eventLabel != null)`. That is, if `eventValue` is defined, it must be that `eventLabel` is defined as well. Will this work?

etanshaul: Oh I misread this, I didn't catch that the precondition is checking a different variable. Maybe instead do it like this for readability (if it has the same logic):

`Preconditions.checkArgument(eventValue != null && eventLabel != null)`

etanshaul: Thats actually even harder to read to me. Just leave it how you had it now that I understand what you meant. Just to confirm: if eventValue is defined but eventLabel is null, do you really want to throw an exception?

chanseokoh: I can decide with some discretion, so I will enforce this rule that eventLabel-eventValue is basically a key-value metadata pair. In this sense, a value without key doesn't make much sense.

etanshaul: agree that it doesn't make sense, but it feels arbitrary that you are doing a precondition on the label (and not the value for example). why not, when you are consuming them, just guard against `eventValue != null && eventLabel != null`. This isn't that big of a deal, but it just made the code confusing (to me at least).

chanseokoh: Sounds good. I will just guard the invalid case with a if-statement.

etanshaul: Other than that, lgtm without having looked into the analytics details.

patflynn:
Just curious if you've tested this out with analytics opted in?

patflynn:
LGTM otherwise

chanseokoh:
Yes, I set a test (sort of) GA tracking ID (with the $ANALYTICS_ID env) on my machine and opted in on my local run/debug sessions. I can see pings appear on the target GA.

PR #696, opened 2016-05-20
Do not use gcloud credential for gsutil

chanseokoh:
Regression introduced by #674. Should be 0. (See #562.)


etanshaul:
nice catch. lgtm.

PR #697, opened 2016-05-23
update IJ target version

patflynn:
we should not be developing against versions of IJ with the recently disclosed security vulnerability.


chanseokoh:
LGTM

PR #699, opened 2016-05-24
Set up new usage tracking facilities

chanseokoh:
Adds one new class for new opt-in usage reporting. The new class utilizes [the cloud-reporting library](https://github.com/google/cloud-reporting).

I put the class under `common-lib` because it didn't feel right to place it under `google-account-plugin`.

I suspect that there is an issue regarding class loading and class version mismatch. The above library uses many other libraries internally, and those libraries probably overlap with what we currently have. I observed ClassCastExceptions in other places (e.g., while casting `org.apache.http.message.BasicNameValuePair` to `org.apache.http.NameValuePair`). This must to be resolved. Not sure what's really happening.


common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/UsageTracker.java
```diff
@@ -0,0 +1,91 @@
...
+import com.intellij.ide.util.PropertiesComponent;
+import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.updateSettings.impl.UpdateChecker;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Map;
+
+public class UsageTracker {
```
patflynn: every public class requires javadoc.

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/UsageTracker.java
```diff
@@ -0,0 +1,91 @@
...
+
+  private static final Logger LOGGER = Logger.getInstance(UsageTracker.class);
```
patflynn: our old code doesn't respect this, but Logger is not immutable and therefore shouldn't be all_caps.

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/UsageTracker.java
```diff
@@ -0,0 +1,91 @@
...
+
+  private static final String USAGE_TRACKER_KEY = "GOOGLE_CLOUD_TOOLS_USAGE_TRACKER_OPT_IN";
+  private static final String ANALYTICS_TRACKING_ID;
+
+  // UUID for this IntelliJ installation. Will be used to uniquely identify a user.
+  private static final String clientUuid =
+      UpdateChecker.getInstallationUID(PropertiesComponent.getInstance());
+
+  static {
```
patflynn: you generally don't want to be reading from disk or doing things by side effect in static code as it makes it impossible to properly unit test.

patflynn:
@loosebazooka should take a look at this

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/UsageTracker.java
patflynn: usage tracking should be a service as in its current implementation. It operates by side-effect and needs to be mocked out for unit testing.

patflynn:
I think the reason tracking is in the account plugin is because the account plugin is reponsible for binding the tracking service to be used by other plugins like Cloud Tools and Firebase in Android Studio.

I'll let @loosebazooka comment.

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/UsageTracker.java
chanseokoh: Alright, I actually kind of expected this.

So, I considered this class fundamentally stateless (everything is `static`), but you're suggesting that it's more suitable to to make a single instance and register it as a service, right? That makes sense. I'll go ahead and change it.

patflynn: the class performs IO, so think of it as inheriting the state of the remote
data it operates in. To test a class that calls into these static methods
would mean unavoidably doing IO which is bad for a unit test.

I'll send you a link off this PR with some guidelines.

chanseokoh:
Seems like my new class will anyway end up being registered as a service under the google-account-plugin, because FWIW, common-lib isn't a standalone plugin. It feels awkward that this has nothing to do with Google accounts or login, but I guess there's no good alternative.

patflynn:
yes I agree it's awkward.  It's going to get more awkward if other google
plugins want to do tracking as well.  Something we'll have to put more
thought into at some point.

chanseokoh:
I did some revamp. Now `IdeaUsageTracker` gets every dependency via injection and is registered as an application-level service in `google-account-plugin`.

I still need to add a couple of tests. Also, I believe I will need to define ExtensionEndpoints or something like that to expose this to the `google-cloud-tools-plugin` too? Will work on that.

build.gradle
```diff
@@ -84,7 +84,7 @@ subprojects {
         }
         testCompile 'org.hamcrest:hamcrest-all:1.3'
         compile 'joda-time:joda-time:2.9.2'
-        compile 'com.google.code.findbugs:annotations:2.0.3'
+        compile 'com.google.code.findbugs:annotations:3.0.1'
```
patflynn: you should put this into a separate PR.

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
```diff
@@ -0,0 +1,119 @@
...
+ * Actual Tracking ID of a target GA Property (ANALYTICS_TRACKING_ID) will be read from a config
+ * property "usage.analytics.tracking.id". (Note that we don't hardcode the Tracking ID in our
+ * codebase. See "google-account-plugin/build.gradle" for how we set the property when we release
+ * our plugin.)
+ */
+public class IdeaUsageTracker {
+
+  private static final Logger logger = Logger.getInstance(IdeaUsageTracker.class);
+
+  // UUID for this IntelliJ installation. Will be used to uniquely identify a user.
```
patflynn: the 'Will be used to uniquely identify a user' is misleading. I'd just leave it out.

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
```diff
@@ -0,0 +1,119 @@
...
+   * @param optedIn A predicate (effectively an 0-argument lambda function) returning if a user has
+   *     opted in for client-side usage reporting.
+   */
+  public void init(@Nullable String ideaTypePostfix,
+      @Nullable String analyticsTrackingId, @Nullable Predicate<Object> optedIn) {
+    // Do the basic checking to see if "google-account-plugin/build.gradle" has expanded the
+    // Tracking ID property. Analytics Tracking ID has the form of "UA-xxxxxxxx-y"
+    // (https://support.google.com/analytics/answer/1032385?hl=en).
+    if (analyticsTrackingId != null && analyticsTrackingId.startsWith("UA-")) {
+      Preconditions.checkNotNull(optedIn);
```
etanshaul: this will throw an NPE if null, yet you have the parameter marked @Nullable

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
```diff
@@ -0,0 +1,119 @@
...
+      UpdateChecker.getInstallationUID(PropertiesComponent.getInstance());
+
+  private String ideaType;
+  private String analyticsTrackingId;
+  private Predicate<Object> optedIn;
+
+  /**
+   * Initialize the tracker.
+   *
+   * @param ideaTypePostfix String indicating IntelliJ IDEA type (e.g., Community Edition or
```
patflynn: is it just CE or UE? or also Android Studio and other IJ SKUs?  There's an enum of these values somewhere.

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
```diff
@@ -0,0 +1,119 @@
...
+   *      Ultimate Edition) appended to the event type string reported to GA.
+   * @param analyticsTrackingId In our IntelliJ codebase, the Tracking ID is supposed to come from
+   *     "google-account-plugin/resources/config.properties". The property name is
+   *     "usage.analytics.tracking.id".
+   * @param optedIn A predicate (effectively an 0-argument lambda function) returning if a user has
```
patflynn: perhaps i'll understand this later, but it seems very odd to pass this as a Predicate<Object>. An explanation in the javadoc would help.

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
```diff
@@ -0,0 +1,119 @@
...
+  private String ideaType;
+  private String analyticsTrackingId;
+  private Predicate<Object> optedIn;
+
+  /**
+   * Initialize the tracker.
+   *
+   * @param ideaTypePostfix String indicating IntelliJ IDEA type (e.g., Community Edition or
+   *      Ultimate Edition) appended to the event type string reported to GA.
+   * @param analyticsTrackingId In our IntelliJ codebase, the Tracking ID is supposed to come from
```
patflynn: the config.properties is an implementation detail that's independent of the constructor. It looks like it's already explained in the class javadoc anyway.

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
chanseokoh: Okay, it seems to be `com.intellij.util.PlatformUtils`. I think I can entirely get rid of `ideaTypePostfix` by using this util class.

google-account-plugin/src/com/google/cloud/tools/intellij/AccountPluginInitializationComponent.java
```diff
@@ -53,9 +56,26 @@ public void initComponent() {
...
+
+    Predicate<Object> userOptedIn = new Predicate<Object>() {
+      @Override
+      public boolean apply(@javax.annotation.Nullable Object object) {
+        return UsageTrackerManager.getInstance().hasUserOptedIn();
+      }
+    };
+
+    IdeaUsageTracker usageTracker = ServiceManager.getService(IdeaUsageTracker.class);
+    usageTracker.init(platformPrefix, trackingId, userOptedIn);
```
etanshaul: Why not just pass in a boolean `UsageTrackerManager.getInstance().hasUserOptedIn()` here?

chanseokoh: We create and init a single instance of this class. The previous `GoogleUsageTracker` does the same thing: checking the opt-in status every time it tries to send a ping. I think it makes sense, because a user can opt in at an arbitrary point when using the IDE.

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
chanseokoh: Feedback on this is welcome. If this were Java 8, I would have used a simple lambda function returning `bool` (the opt-in status). I was looking for a similar thing in Java 7 and just found a `Predicate` class. I could use something more pleasant than `Predicate`.

google-account-plugin/src/com/google/cloud/tools/intellij/AccountPluginInitializationComponent.java
etanshaul: Oh so basically you're using the Predicate as a way of dynamically getting the opt in status. Why not just call `UsageTrackerManager.getInstance().hasUserOptedIn()` each time its needed? This seems like a convoluted use of Predicate unless I still don't understand?

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
chanseokoh: The intention is basically like throwing an `IllegalArgumentException` in this case (i.e., `optedIn` should not be null if the Tracking ID is set up). Perhaps I should use `checkArgument` instead of `Preconditions` as in [here](https://engdoc.corp.google.com/eng/doc/devguide/java/practices/preconditions.md?cl=head). I will fix this.

etanshaul: or just make the paramater `@NotNull`

google-account-plugin/src/com/google/cloud/tools/intellij/AccountPluginInitializationComponent.java
chanseokoh: That will do, actually. This revamp was started to make this class suitable for testing. So, using `Predicate` is drived by testing here, since `hasUserOptedIn()` requires accessing an actual IJ datastore (or something like that). The catch is to put this `hasUserOptedIn()` via Dependency Injection so that tests can conveniently make it return true or false. But perhaps, this DI may be too stretched, and there may be a way to mock the datastore in the background in testing???

chanseokoh: FYI, `hasUserOptedIn()` is old code from `UsageTrackerManager` which will be gone eventually. `hasUserOptedIn()` will eventually become `PropertiesComponent.getInstance().getBoolean("GOOGLE_CLOUD_TOOLS_USAGE_TRACKER_OPT_IN", false)`.

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
chanseokoh: Just wanted to allowed callers to pass null if they don't want to set up usage tracking for their convenience. We will see this after deciding whether we will continue to pass a `Predicate` (or alike).

google-account-plugin/src/com/google/cloud/tools/intellij/AccountPluginInitializationComponent.java
etanshaul: I understand the testing concern. At a minimum, you should make it a `Predicate<UsageTrackerManager>` instead to make it clearer. If I can think of a better way I'll let you know.

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
chanseokoh: Yeah, it has nothing to do with this class. I just wanted to leave a doc, so that whoever reads this code later can quickly figure out how our plugin actually retrieves a Tracking ID. (It took me quite a while to figure this out previously with the existing code.)

google-account-plugin/src/com/google/cloud/tools/intellij/AccountPluginInitializationComponent.java
chanseokoh: Perhaps, instead of `Predicate`, passing `PropertiesComponent` can work. (However, I actually want to use a Java 8 lambda function for this, since it will look simple and do the job while looking cool(?).)

etanshaul: Up to you. It feels like a strightly strange use case of Predicate to me. Passing in UsageTrackerManger or PropertiesComponent directly (so you can mock them) should do it too.

chanseokoh: Agreed. Unless this is Java 8, this is confusing. I will make this pass `PropertiesComponents`. A small downside is that it couples more tightly with IntelliJ, but this class is anyway bound to IntelliJ.

chanseokoh:
I simplified this quite a bit and fixed some problems.

chanseokoh:
About `PropertiesComponent`: Turns out that it is an interface with a static `getInstance()` method, so it can be easily mocked in testing (I believe).

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
```diff
@@ -0,0 +1,116 @@
...
+    // Do the basic checking to see if "google-account-plugin/build.gradle" has expanded the
+    // Tracking ID property. Analytics Tracking ID has the form of "UA-xxxxxxxx-y"
+    // (https://support.google.com/analytics/answer/1032385?hl=en).
+    if (analyticsTrackingId != null && analyticsTrackingId.startsWith("UA-")) {
+      this.analyticsTrackingId = analyticsTrackingId;
+    }
+  }
+
+  private boolean hasUserOptedIn() {
+    return PropertiesComponent.getInstance().getBoolean(USAGE_TRACKER_KEY, false);
```
patflynn: Weren't you using the FlagReader for this before. Why change?

patflynn: if you want to make this unit testable you'll need  to emulate the UsageTrackerManager and provide a way to make the PropertiesComponent injectable.

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  /**
+   * Send an opt-in usage event to Google Analytics. Do not send any PII.
+   */
+  public void trackEvent(@NotNull String eventName, @Nullable Map<String, String> metadata) {
+    if (analyticsTrackingId != null && hasUserOptedIn()) {
+      try {
+        new MetricsSender(analyticsTrackingId).send(buildEvent(eventName, metadata));
```
patflynn: Do we have to create a new sender for every track event?

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
```diff
@@ -0,0 +1,116 @@
...
+
+  /**
+   * Initialize the tracker.
+   *
+   * @param analyticsTrackingId Tracking ID of a Property in Google Analytics to which this class
+   *     will send usage pings. (In our IntelliJ codebase, the Tracking ID is supposed to come
+   *     from "google-account-plugin/resources/config.properties". The property name is
+   *     "usage.analytics.tracking.id".)
+   */
+  public void init(@Nullable String analyticsTrackingId) {
```
patflynn: if no analytics tracking ID is present why are we even creating this tracker?

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  private boolean hasUserOptedIn() {
+    return PropertiesComponent.getInstance().getBoolean(USAGE_TRACKER_KEY, false);
+  }
+
+  /**
+   * Send an opt-in usage event to Google Analytics. Do not send any PII.
+   */
+  public void trackEvent(@NotNull String eventName, @Nullable Map<String, String> metadata) {
```
patflynn: @chanseokoh 
@loosebazooka 

I believe Appu previously had an an interface for this kind of stuff to support the Android Studio vs IntelliJ implementations. Is this not needed anymore?

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private static final Logger logger = Logger.getInstance(IdeaUsageTracker.class);
+
+  private static final String USAGE_TRACKER_KEY = "GOOGLE_CLOUD_TOOLS_USAGE_TRACKER_OPT_IN";
+
+  // UUID for this IntelliJ installation.
+  private static final String CLIENT_UUID =
+      UpdateChecker.getInstallationUID(PropertiesComponent.getInstance());
+
+  private static final String EVENT_TYPE = "gcloud-intellij-" + PlatformUtils.getPlatformPrefix();
```
patflynn: can we delegate to PluginInfoService for this?

google-account-plugin/build.gradle
```diff
@@ -53,11 +53,18 @@ configurations {
 }
 
 processResources {
+    // TODO: we will soon remove ANALYTICS_ID all together.
```
patflynn: @loosebazooka to review this.

google-account-plugin/resources/META-INF/plugin.xml
```diff
@@ -57,6 +57,8 @@
                         serviceImplementation="com.google.cloud.tools.intellij.DefaultAccountPluginConfigurationService"/>
     <applicationService serviceInterface="com.google.cloud.tools.intellij.login.GoogleLoginService"
                         serviceImplementation="com.google.cloud.tools.intellij.login.IntellijGoogleLoginService"/>
+    <applicationService serviceInterface="com.google.cloud.tools.intellij.analytics.IdeaUsageTracker"
```
patflynn: I'm confused that we're creating a new usage tracking service while leaving the old one.

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
```diff
@@ -0,0 +1,116 @@
...
+  public void trackEvent(@NotNull String eventName) {
+    trackEvent(eventName, null);
+  }
+
+  public void trackEvent(
+      @NotNull String eventName, @NotNull String metadataKey, @NotNull String metadataValue) {
+    trackEvent(eventName, ImmutableMap.<String, String>of(metadataKey, metadataValue));
+  }
+
+  @VisibleForTesting
```
patflynn: Is there a test for this class?

google-account-plugin/resources/META-INF/plugin.xml
chanseokoh: The plan is to do this in multiple steps to lessen the review burden. The first step is to introduce and set up new facilities. The second is to change exist code to use it. The last is to remove old code, which is pretty large and complex.

patflynn: You'll definitely need to do an Appu review.  A lot of the current
architecture is to support Android Studio tracking so we need to make sure
the new changes support those requirements.

common-lib/src/main/java/com/google/cloud/tools/intellij/analytics/IdeaUsageTracker.java
chanseokoh: This is to avoid callers checking if tracking is enabled every time they tried to use it. They can then make a single call to `trackEvent` without worrying about anything. It's actually the same mechanism as before.

google-account-plugin/resources/META-INF/plugin.xml
chanseokoh: Hmm... I thought Android would go a separate way from here, since it will use its own GA property. We can't use the cloud-reporting lib for Android Studio. There should be different facilities for AS.

What could be the best way?

patflynn: You should talk to Appu about this.  The main point is that Android Studio
also wants to track our events and we don't want to duplicate tracking
calls for both properties.

chanseokoh: Alright, as we talked offline, if our code has to support both Android GA and Concord GA, I think it's better to use the existing code, which is more generic. The cloud-reporting library this PR is trying to use is Concord-specific. We will go back to PR #694.

PR #700, opened 2016-05-25
AE standard deployment support

### LINKED ISSUE #665: Our App Engine deployment run config should allow the selection of App Engine standard modules as deployment sources
IntelliJ Ultimate has built in App Engine support, which handles build and staging of modules with the App Engine facet.  Our deployment UI should allow selecting an IJ App Engine module as a deployment source.

etanshaul:
fixes #665 

Base set of functionality for supporting deployment of projects to App Engine standard. Currently works only in IJ UE when the legacy app engine facet has been applied to the project.

Deployment window / sources for an App Engine standard project:

image

Still working on cleanup/refactoring.


google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -21,17 +21,18 @@ plugin.compatibility.error.message=You are running potentially incompatible vers
 plugin.compatibility.error.name.and.version={0} ({1})
 plugin.compatibility.error.update.link=Please {0}click here{1} to ensure that you have updated all of your plugins.
 
-appengine.flex.tools.menu.item.label=Deploy to App Engine flexible environment...
+appengine.name=App Engine (BETA)
+appengine.tools.menu.item.label=Deploy to App Engine...
```
patflynn: doesn't ultimate already have this menu item from their provided AE integration?

etanshaul: Here's the menu items showing both:

image

Theirs opens the old plugin.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java
```diff
@@ -121,45 +156,35 @@ private File stage() {
...
+          AppEngineFlexDeploymentArtifactType.typeForPath(deploymentArtifactPath))
           : CloudSdkUtil.getFileFromFilePath(deploymentConfiguration.getDockerFilePath());
 
       copyFile(stagingDirectory, "app.yaml", appYamlPath);
       copyFile(stagingDirectory, "Dockerfile", dockerFilePath);
     } catch (IOException ex) {
       logger.warn(ex);
       callback.errorOccurred(GctBundle.message("appengine.deployment.error.during.staging"));
-      return null;
     }
```
chanseokoh: If an error occurred, previously we were not executing `deploy()` after this.

etanshaul: Oops. That got lost when I broke apart staging. Thanks.

patflynn:
when I tried this out I got:
Deploying 'appengine-wiz:war exploded'...
Created temporary staging directory: /tmp/gae-mvm0.tmp
Error: Could not find or load main class com.google.appengine.tools.admin.AppCfg

etanshaul:
Is that app you tested available publicly on github so I can test that out? I do not get that error with the app I am testing so I'm wondering if it has to do with the flags I am setting during stage.

patflynn:
I'm using the IntelliJ new project wizard and picking an App Engine project.

etanshaul:
I'll have to look deeper into your configuration. When I deploy a project from the wizard to AE standard  completes successfully: 

image

patflynn:
Let's look at my config tomorrow morning when I'm back in the office.

chanseokoh:
How about doing `gcloud components install/update app-engine-java`?

etanshaul:
note: redesign in progress. general functionality will remain the same (std deploy currently works properly except for a certain edge cases discussed offline).

etanshaul:
@patflynn, @chanseokoh ptal.

I went ahead and reorganized the logic surrounding the various App Engine operations. 

The primary goal is to improve readability / testability / error handling by removing the inheritance and improving separation between the various operations.

The new design uses `Runner` classes with are java `Runnables` to control the various actions. They take in "operation" classes which execute the individual steps. For example:

`AppEngineFlexibleDeployRunner` has:
- `AppEngineFlexibleStage`
- `AppEngineDeploy`

and similarly

`AppEngineStandardDeployRunner`  has:
- `AppEngineStandardStage`
- `AppEngineDeploy`

The individual operation classes each throw specific exceptions related to the operation they perform, and are caught and handled by the runner classes.

TODO: more test coverage, and re-implement the deployment cancel operation.

patflynn:
Thanks for doing this.  The PR itself is pretty big. What do you think about booking some time with Chanseok and I to talk us through it?  A high bandwidth discussion up front about the high level design will probably save a lot of back and forth.

etanshaul:
Totally agree. The fact that this also has the new standard deploy functionality makes independent review even tougher (although looking just at my last commit can help).

chanseokoh:
Sounds good.

etanshaul:
@patflynn, @chanseokoh 

I pushed a couple more commits since our review. I addressed the error handling things we discussed (see details on the commit message), and I re-implemented the ability to abort running deploys.

Because this PR is already pretty big, my hope it to get this merged in and to then open smaller issues to continue enhancing this as needed. Therefore I only implemented the couple things I thought were essential prior to merging (unless you find anything I forgot).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploy.java
```diff
@@ -42,141 +39,91 @@
...
-
-  /**
-   * Perform a deployment from the given staging directory.
-   */
-  private void deploy(@NotNull File stagingDirectory) {
-    final StringBuilder rawDeployOutput = new StringBuilder();
-    CloudSdk sdk;
-
-    ProcessOutputLineListener stdErrListener = new ProcessOutputLineListener() {
+    ProcessOutputLineListener outputListener = new ProcessOutputLineListener() {
```
patflynn: I wouldn't really understand what the difference between outputListener and deployOutputListener is if I didn't know a bit about the library already.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploy.java
```diff
@@ -42,141 +39,91 @@
...
+        loggingHandler,
+        startListener,
+        outputListener,
+        deployOutputListener,
+        deployExitListener);
     CloudSdkAppEngineDeployment deployment = new CloudSdkAppEngineDeployment(sdk);
     deployment.deploy(configuration);
   }
 
+  AppEngineHelper getHelper() {
```
patflynn: why are these all package protected?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploy.java
```diff
@@ -198,37 +145,29 @@ public void exit(int exitCode) {
 
           if (deployOutput == null
               || deployOutput.getService() == null || deployOutput.getVersion() == null) {
-            consoleLogLn(
-                GctBundle.message("appengine.deployment.version.extract.failure") + "\n");
+            loggingHandler.print(
```
patflynn: I've got a terrible memory, but didn't we say messages to logs should not be localized, only messages in the UI elements?

chanseokoh: Part of the reason was to reuse a long error message in several places (a compromise we took), but I guess this is no longer the case here. In this case `appengine.action.error.update.message` is particularly unnecessary. So, I think we can log a plain string here.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploy.java
etanshaul: Sure - I'll come up with better naming.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploy.java
etanshaul: Because they are only needed by the Runner classes. I didn't find it necessary to further expose them.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploy.java
etanshaul: @patflynn yes we did say that. However `loggingHandler` is what is used to print messages to IJ UI console output window. Perhaps renaming the variable would make that less confusing.

patflynn: I guess I'm thinking that perhaps the console output should also remain unlocalized, since the rest of the console output won't be.

etanshaul: oh right. My thinking is that the console window sort of blurs the line between UI and logging output. It _can_ be the output raw logs pipped to it. But its also the main / only form of certain information messages that the user get within the IDE when performing certain actions. But you are right that the rest of the console output (coming from the common lib) will not be localized.

patflynn: I agree it's blurry.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
```diff
@@ -115,6 +155,24 @@ public AppEngineDeploymentConfiguration createDefaultConfiguration(
...
+            new File(server.getConfiguration().getCloudSdkHomePath()),
+            getEnvironment())
+    );
+  }
+
+  /**
+   * Lazily initialize the environment because this class is instantiated on IDE boot time
+   * and the project may not have yet been fully constructed.
+   */
+  public Environment getEnvironment() {
```
patflynn: Is standard flex going to be a project level setting? Isn't it determine at the point where the user selects the deployment source?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineFlexibleDeployRunner.java
```diff
@@ -0,0 +1,95 @@
...
+  @Override
+  public void run() {
+    File stagingDirectory;
+
+    try {
+      stagingDirectory = deploy.getHelper().createStagingDirectory(deploy.getLoggingHandler());
+    } catch (IOException ioe) {
+      deploy.getCallback().errorOccurred(
+          GctBundle.message("appengine.deployment.error.creating.staging.directory"));
+      logger.error(ioe);
```
patflynn: isn't an IOException the kind of exception that we don't want to report?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineFlexibleStage.java
```diff
@@ -0,0 +1,93 @@
...
+   * App Engine flexible environment.
+   */
+  public void stage(@NotNull File stagingDirectory) {
+    try {
+      File stagedArtifactPath =
+          copyFile(
+              stagingDirectory,
+              "target" + AppEngineFlexDeploymentArtifactType.typeForPath(deploymentArtifactPath),
+              deploymentArtifactPath);
+      stagedArtifactPath.setReadable(true /* readable */, false /* ownerOnly */);
```
patflynn: we should have a comment somewhere explaining why we're doing this. i.e. the Docker bug that makes flex break if you don't.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineHelper.java
```diff
@@ -16,71 +16,103 @@
...
+
+  /**
    * The path to the gcloud command on the local file system.
    */
   File getGcloudCommandPath();
 
   /**
+   * The App Engine environment.
+   */
+  Environment getEnvironment();
```
patflynn: as mentioned earlier I'm wondering whether this is the right scope for 'Environment'.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineHelper.java
```diff
@@ -16,71 +16,103 @@
...
+   * @throws IOException if the staging fails
    */
-  AppEngineStopAction createStopAction(
+  File createStagingDirectory(LoggingHandler loggingHandler) throws IOException;
+
+  /**
+   * Creates an {@link CloudSdk} object that is used in execution of various App Engine actions.
+   *
+   * @param loggingHandler logging messages will be output to this
+   * @param startListener the "callback" listener used for fetching the running process.
```
patflynn: remove the '.' from the end of the line.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineHelper.java
```diff
@@ -16,71 +16,103 @@
...
-   * @param undeploymentTaskCallback a callback for handling completions of the operation
-   * @return the action that will perform the stop operation
+   * @return the file representing the staging directory
+   * @throws IOException if the staging fails
    */
-  AppEngineStopAction createStopAction(
+  File createStagingDirectory(LoggingHandler loggingHandler) throws IOException;
+
+  /**
+   * Creates an {@link CloudSdk} object that is used in execution of various App Engine actions.
```
patflynn: Creates a ..

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStopRunner.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  @Override
+  public void cancel() {
+    if (process != null) {
+      process.destroy();
+    }
+  }
+
+  private void setProcess(Process process) {
```
patflynn: There does seem to be overlap between the runners. I wonder if you wouldn't just want a generic command executor.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -16,40 +16,94 @@
...
   private static final String DEFAULT_WAR_DOCKERFILE_PATH
       = "/generation/src/appengine/mvm/war.dockerfile";
 
+  private final Project project;
   private final File gcloudCommandPath;
+  private final Environment environment;
+
+  private File credentialsPath;
+
+  public enum Environment {
```
patflynn: Environment, if it should continue to exist, shouldn't be scoped to the Helper.

patflynn: public types should always have javadoc.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
etanshaul: Currently, it determined based on the project itself: if the project has the AE standard facet from the old plugin, then its standard. Otherwise, its flex. This came from the document that we went over earlier, but is obviously open for change.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -16,40 +16,94 @@
...
+
+    @Override
+    public String toString() {
+      return label; // TODO message bundle
+    }
+  }
 
-  public CloudSdkAppEngineHelper(@NotNull File gcloudCommandPath) {
+  /**
+   * Initialize the helper.
```
patflynn: Definitely leave out method and constructor javadocs if they don't feel like they add any information.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineFlexibleDeployRunner.java
etanshaul: I would have thought this is exactly the kind of thing we want to be notified about: creating the staging directory failed due to some exceptional condition (which should not happen under normal conditions).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -79,61 +139,148 @@ public File defaultDockerfile(DeploymentArtifactType deploymentArtifactType) {
...
+      return new AppEngineFlexibleDeployRunner(deploy, flexibleStage);
+    }
   }
 
-  @NotNull
   @Override
-  public AppEngineStopAction createStopAction(
+  public File createStagingDirectory(LoggingHandler loggingHandler) throws IOException {
+    File stagingDirectory = FileUtil.createTempDirectory(
+        "gae-mvm" /* prefix */,
```
patflynn: maybe we should change this prefix to something like gae-flex-staging-project-service-version

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
etanshaul: Thats an unfortunate consequence of checkstyle which requires this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -79,61 +139,148 @@ public File defaultDockerfile(DeploymentArtifactType deploymentArtifactType) {
...
+
+    return new CloudSdk.Builder()
+        .sdkPath(getGcloudCommandPath())
+        .async(true)
+        .addStdErrLineListener(stdErrListener)
+        .addStdOutLineListener(stdOutListener)
+        .exitListener(exitListener)
+        .startListener(startListener)
+        .appCommandCredentialFile(credentialsPath)
+        .appCommandMetricsEnvironment("gcloud-intellij")
```
patflynn: I'd expect the gcloud-intellij value should come from pluginInfoService as well.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineFlexibleDeployRunner.java
patflynn: But IO in particular is typically a transient failure. Like out of space,
or the user doesn't have permission to write to disk or something like
that.  We can leave it and see if we get too much noise.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
patflynn: Facets are at the module level though aren't they?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineFlexibleDeployRunner.java
etanshaul: I see your point. For this particular one I think we should probably change it to warn.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStopRunner.java
etanshaul: Yea as I refactored and added more functionality, some overlap did creep in. I'll look into a pattern to improve this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -79,61 +139,148 @@ public File defaultDockerfile(DeploymentArtifactType deploymentArtifactType) {
...
-      ConfigType deploymentType, DeploymentArtifactType artifactType) {
+      AppEngineDeploymentConfiguration deploymentConfiguration,
+      File artifactToDeploy) {
 
-    StringBuilder labelBuilder = new StringBuilder("deploy.flex");
-    switch (deploymentType) {
-      case AUTO:
+    StringBuilder labelBuilder = new StringBuilder("deploy");
+
+    if (environment == Environment.APP_ENGINE_STANDARD) {
```
patflynn: I do respect the way Chanseok made his code fail if a new Environment value gets added. Although I do admit that possibility seems remote.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
```diff
@@ -16,13 +16,72 @@
...
+        }
+      }
+    }
+
+    return false;
+  }
+
+  public static boolean isAppEngineStandardArtifactType(Artifact artifact) {
+    String artifactId = artifact.getArtifactType().getId();
+    return "exploded-war".equalsIgnoreCase(artifactId);
```
patflynn: I think there can be exploded-war artifacts that aren't app engine apps. but maybe that's ok?

etanshaul: Sure there can be. But being exploded-war is not a sufficient condition for being considered and ae standard app. It also has to have the facet. I think the naming of the method is not ideal.

etanshaul: Actually maybe the naming is ok. What I am implying is thats its "the type of artifact needed to be considered an AE _ app", not that if it has that type then it must be an AE app.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/AppEngineFlexibleDeployRunnerTest.java
```diff
@@ -0,0 +1,118 @@
...
+    when(helper.createStagingDirectory(any(LoggingHandler.class))).thenReturn(new File("myFile.jar"));
+    when(deploy.getHelper()).thenReturn(helper);
+    when(deploy.getCallback()).thenReturn(callback);
+    when(deploy.getDeploymentConfiguration()).thenReturn(deploymentConfiguration);
+
+    deployRunner = new AppEngineFlexibleDeployRunner(deploy, stage);
+  }
+
+  @Test
+  public void testCreateStagingDirectory_Error() throws IOException {
```
patflynn: I think the testBla_case pattern is supposed to be camel case after the underscore too. so this should be 'testCreateStagingDirectory_error()'

etanshaul: I quickly dug into where this convention can from and most sources point to this: http://osherove.com/blog/2005/4/3/naming-standards-for-unit-tests.html

Its super old, but it does support the casing that I had so I'll leave it for now.

etanshaul:
@chanseokoh - one of the updates I made was to get the metrics environment name (sent to the common lib for analytics) from the plugin info service class rather than hard-coding it. I pulled it from the userAgent string "gcloud-intellij-cloud-tools-plugin/0.9.5-beta-SNAPSHOT (IntelliJ IDEA Community Edition/IC-143.2370)". Formerly it was hardcoded to "gcloud-intellij". 

Can you confirm if this is ok for analytics?

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/AppEngineFlexibleDeployRunnerTest.java
elharo: https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names

etanshaul: Thanks for that. I'll update it.

chanseokoh:
@etanshaul `getUserAgent()` doesn't work here because it contains `/`, space, `(`, etc. I think we should define a new method in the `PluginInfoService` that returns just the first token. We should return a tool name only.

etanshaul:
@patflynn - I have addressed most of your comments that I think needed to be fixed in this PR. The only change still to be made is determining the correct place for the `Environment` concept.

The biggest update is https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/700/commits/442975063d27f98a46ad70995368f92404041a42 where I removed duplication between runner by creating a new concept of `AppEngineTask'`s.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineFlexibleStage.java
etanshaul: Is this bug documented somewhere? I didn't write this particular part originally so I'm trying to get some context.

etanshaul:
Regarding the environment name for metrics:

I'm almost tempted to restore the hardcoding of "gcloud-intellij". I don't know if this belongs in `PluginInfoService` which is scoped to a specific plugin (e.g. account, or cloud-tools). What we really are after is the overall name of the project as defined externally. Thoughts?

chanseokoh:
@etanshaul Yeah, I also got the same trouble because of the reason you mentioned. I'll also use "gcloud-intellij" in another place (for client-side reporting), so in this sense, defining a common constant somewhere may still be better. But where exactly? Perhaps something like `com.google.cloud.tools.intellij.util.IntelliJPlatform` in `common-lib`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
etanshaul: Sorry I missed this - yes facets are at the module level. I'm thinking this Environment concept should eventually live on the AE facet once we merge in the plugins. Currently we don't have an explicitly defined facet so I'm thinking we can leave this as is, with a todo to refactor this once merged. Are you ok with this approach?

etanshaul:
@chanseokoh - I changed my mind about the placement of the platform name. I think `PluginInfoService` is the right place as is the is the base abstract class of all plugins. See https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/700/commits/58a6617c9b8cf9f356cd9b8b099b30961c53de50

chanseokoh:
@etanshaul Thanks for putting it in `PluginInfoService`.

etanshaul:
@patflynn @chanseokoh please take a look https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/700/commits/2cb7a0166040b383bf1138484b771ee9c451838d

Unfortunately this was more involved than I thought so this is another non-trivial addition to the PR.

This include changes to how the App Engine deployment environment is determined and where the environment information lives. 

With these changes, environment info now lives at the `DeploymentSource` level. App Engine deployable sources (`MavenBuildDeploymentSource`, `UserSpecifiedPathDeploymentSource`, and `AppEngineArtifactDeploymentSource`) now implement `AppEngineDeployable` which provides the targeted environment (currently flexible or standard). 

Also, since target environment is at the deployment source level, a deployment configuration for a given project can now (with adherence to some rules) mix and match deployments to flex and standard (e.g. a multi-module project that has an AE standard module and another module that can be deployed to flex). See `AppEngineUtil` for some documentation on these rules.

patflynn:
Thanks for doing this so quickly.  I'll check this out today.

common-lib/src/main/java/com/google/cloud/tools/intellij/BasePluginInfoService.java
```diff
@@ -41,6 +41,7 @@
   private final String userAgent;
   private final IdeaPluginDescriptor plugin;
   private final FlagReader flagReader;
+  private static final String PLUGIN_NAME_EXTERNAL = "gcloud-intellij";
```
patflynn: this change belongs in a separate PR.

@chanseokoh I'm actually concerned that the new scheme loses the IDE platform the plugin is being used from.  When we interact with APIs directly we specify which IDE the plugin is in (Android Studio vs IntelliJ).  Can we just add this information to the user-agent?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
```diff
@@ -112,9 +69,20 @@ public AppEngineDeploymentConfiguration createDefaultConfiguration(
...
+          String.format("Deployment source with name %s is not deployable to App Engine.",
+              source.getPresentableName()));
+      return null;
+    }
+
     return new AppEngineDeploymentRunConfigurationEditor(
         project,
         source,
-        new CloudSdkAppEngineHelper(new File(server.getConfiguration().getCloudSdkHomePath())));
+        ((AppEngineDeployable) source).getEnvironment(),
```
patflynn: if source has the environment and is already a parameter, why does the client need to redundantly pass it here?

etanshaul: Good point. I'll fix that.

common-lib/src/main/java/com/google/cloud/tools/intellij/BasePluginInfoService.java
chanseokoh: @etanshaul I had a discussion with @patflynn. We will do some overhaul about this thing, so we don't need to think about this much in this PR. Let's leave this field here for now.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
etanshaul: fixed.

build.gradle
```diff
@@ -118,6 +118,7 @@ subprojects {
             // JDepend manages to include a different version of itself in its own transitive closure.
             force "jdepend:jdepend:${project.jdepend.toolVersion}"
         }
+        resolutionStrategy.cacheChangingModulesFor 0, 'seconds'
```
patflynn: this should go in a separate PR. :)

etanshaul: I want to put EVERYTHING in this PR :) just kidding, I'll move it you're right.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurator.java
```diff
@@ -112,9 +69,19 @@ public AppEngineDeploymentConfiguration createDefaultConfiguration(
   public SettingsEditor<AppEngineDeploymentConfiguration> createEditor(
       @NotNull DeploymentSource source,
       @NotNull RemoteServer<AppEngineServerConfiguration> server) {
+    if (!(source instanceof AppEngineDeployable)) {
```
patflynn: just curious if this can happen?

etanshaul: No, it can't happen currently based on the logic that collects the deployment sources. I was just being defensive.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -114,14 +118,20 @@ public AppEngineDeploymentRunConfigurationEditor(
...
 
-    appEngineCostWarningLabel.setText(
-        GctBundle.message("appengine.flex.deployment.cost.warning",
-            COST_WARNING_OPEN_TAG,
-            COST_WARNING_HREF_OPEN_TAG,
-            COST_WARNING_HREF_CLOSE_TAG,
-            COST_WARNING_CLOSE_TAG));
-    appEngineCostWarningLabel.addHyperlinkListener(new BrowserOpeningHyperLinkListener());
-    appEngineCostWarningLabel.setBackground(editorPanel.getBackground());
+    AppEngineEnvironment environment = ((AppEngineDeployable) deploymentSource).getEnvironment();
```
patflynn: should this just be the type accepted by the constructor?

etanshaul: This class currently also uses features of the DeploymentSource type

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineEnvironment.java
```diff
@@ -0,0 +1,38 @@
...
+  APP_ENGINE_FLEX("appengine.environment.name.flexible");
+
+  private String label;
+
+  AppEngineEnvironment(String label) {
+    this.label = label;
+  }
+
+  @Override
+  public String toString() {
```
patflynn: I'd call this method getLocalizedLabel or something like that.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineFlexDeploymentArtifactType.java
```diff
@@ -23,16 +23,17 @@
 import java.util.Locale;
 
 /**
- * Represents the supported Java artifact types that we can deploy to App Engine (Managed VMs).
+ * Represents the supported Java artifact types that we can deploy to App Engine flexible.
```
patflynn: to deploy to App Engine flexible environment.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineFlexibleDeployTask.java
```diff
@@ -0,0 +1,78 @@
...
+import com.google.cloud.tools.appengine.cloudsdk.process.ProcessStartListener;
+import com.google.cloud.tools.intellij.util.GctBundle;
+
+import com.intellij.openapi.diagnostic.Logger;
+
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * Runnable that executes task responsible for deploying an application to the App Engine
```
patflynn: executes 'a' task

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineFlexibleStage.java
patflynn: Is the app-tools-lib not taking care of staging for us at this point?

etanshaul: It can (and should) but I was saving this refactoring for a separate PR.

build.gradle
chanseokoh: I think this PR can't be merged without this caching update because this PR already synced up with the latest app-tools-lib that has gone re-organization.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
```diff
@@ -16,13 +16,206 @@
...
+   * appengine-web.xml etc.) then no option is provided to deploy any of its artifacts to the
+   * flexible environment.
+   *
+   * @return a list of {@link AppEngineArtifactDeploymentSource}'s
+   */
+  public static List<AppEngineArtifactDeploymentSource> createArtifactDeploymentSources(
+      @NotNull Project project) {
+    List<AppEngineArtifactDeploymentSource> sources = Lists.newArrayList();
+
+    for (Module module : ModuleManager.getInstance(project).getModules()) {
```
patflynn: can't there be artifacts that aren't related to a module?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
```diff
@@ -16,13 +16,206 @@
...
+  }
+
+  /**
+   * Creates a list of module deployment sources available for deployment to App Engine. Currently,
+   * all module based sources target the App Engine flexible environment:
+   *
+   * <p>Maven based deployment sources are included if there are no App Engine standard artifacts
+   * associated with the same module.
+   *
+   * <p>User browesable jar/war deployment sources are added only if there are no App Engine
```
patflynn: s/browesable/browsable/

build.gradle
etanshaul: @chanseokoh The problem is that Travis won't let us merge it. 

We have an annoying chicken and egg problem here though: if I move this to a separate PR, that PR will fail (because it would not have been sync'd up with the common-lib) - if I sync it, it will basically be throw away since all those classes have been refactored by this PR. If you all are ok with it this time, I'll leave the common-lib syncing and cache update all in this PR. Next time I can be more careful with the timing.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
patflynn: oh I see. Should AppEngineDeployable extend DeploymentSource?

build.gradle
patflynn: OK with me.  Pretty soon we'll need to start having app-tools-lib version
their releases.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
etanshaul: The idea is that app engine deployable sources of type `DeploymentSource` implement `AppEngineDeployable`. So its actually the other way around.

build.gradle
patflynn: In fact we'll need a versioned library for our release.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
patflynn: Yeah but is there a case where an AppEngineDeployable will not also be a
DeploymentSource?  I'm just wondering if you can avoid this cast by having
the constructor take AppEngineDeployable as its source.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
etanshaul: I don't think so. All projects (I will confirm) have modules. And artifacts are the output of one or more modules.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineEnvironment.java
etanshaul: Really? With toString its the string representation of an AppEngineEnvironment which seemed pretty clear to me. It also has the benefit of not having to call a method for things like logging. I don't have strong feelings about this though.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
patflynn: @etanshaul   I thinkm this is the last question I have before we merge this and iterate.

etanshaul: Sounds good. I will validate this to make sure what I have makes sense and then merge if things look ok to you.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineEnvironment.java
patflynn: For me toString() returns pretty printed object state, but isn't part of
the stuff displayed in a UI unless the type is explicitly some kind of UI
element that would have text.

When I saw it called earlier, I wondered why we weren't localizing, but if
I had read environment.localizedLabel(), I wouldn't have wondered that.

Another case might be you are logging errors including environment
information.  Now when you're looking at feedback you see that the crash
was caused in an environment, but the environment is in Swahili so you're
not quite sure which one it is.q

etanshaul: Good point. Updated.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
etanshaul: As we discussed, it doesn't make sense for there to be an isolated artifact without a module. By definition artifacts are a specification of the output of a project. Without at least one module, there is no source code so I can't really imagine what an artifact would be useful for in this case.

patflynn:
:ship: when ready.

etanshaul:
test for cla bot

PR #698, opened 2016-05-24
Enable cancelling deployment

chanseokoh:
Fixes #687.

This breaks build until the new app-tools common lib is released.

PR #701, opened 2016-05-25
Remove unnecessary inner class

chanseokoh:
One less class.

Obviously, I misunderstood Etan's suggestion (https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/698#discussion_r64420758) in my previous merged PR, so I'm fixing it.


etanshaul:
Thanks lgtm. This is what I had in mind originally. Trivial suggestion if you agree with it: instead of setting the member variably directly, create a private setter and do `setProcess(process)` within the start listener (and you can get rid of the forced renaming of the process variable).

PR #708, opened 2016-06-02
upgrade gradle-intellij-plugin version and use newer gradle plugin sy…

etanshaul:
…ntax


etanshaul:
ignore this for now, running into some build problems

build.gradle
```diff
@@ -18,13 +18,11 @@ buildscript {
...
         maven { url "https://plugins.gradle.org/m2/" }
     }
-    dependencies {
-        classpath 'gradle.plugin.org.jetbrains:gradle-intellij-plugin:0.0.43'
-    }
 }
 
 plugins {
     id 'net.researchgate.release' version '2.3.4'
+    id 'org.jetbrains.intellij' version '0.1.5'
```
patflynn: I think, but I don't remember why, that this method of plugin import broke stuff.  Have you tested this change out?

etanshaul: Its the new way to do it since gradle 2.1. I only recently upgraded the gradle wrapper on the project.

Its the version of the plugin that's breaking stuff. Not sure why yet.

etanshaul: I think it has something to do with muli-plugin projects like ours. I actually dug up on old issue you were contributing to: https://github.com/JetBrains/gradle-intellij-plugin/issues/17 . My suspicion is that the new version of the plugin is not compatible with the workaround you came up with.

patflynn: This has been a valid way to import when we first set this up. But it did
cause breakages, I think related to evaluation order of stuff or something.
(sorry my memory is not so good) @loosebazooka might remember better.

etanshaul: Could be. The new "plugins" syntax is new to gradle > 2.1: https://docs.gradle.org/2.1/release-notes#new-and-noteworthy

Either way its the version of the gradle-intellij-plugin thats causing the problem here (I tested it with the new import method but without upgrading the version and everything works fine).

loosebazooka: There was something weird going on, maybe it had something to do with subprojects? I really have no idea... it was so long ago, check the logs?!

etanshaul:
An update on this so it does just sit here: I tried unsuccessfully to upgrade to the latest version of the gradle-intellij plugin but kept hitting a wall. It has to do with the fact that we have subprojects. Previously, the plugin would sandbox the Account plugin first then the Cloud Tools plugin afterwards, correctly creating two folders. 

After upgrading the plugin, the sandbox logic doesn't work the same way: after staging one of the plugins in /build/idea-sandbox/plugins the next staging wipes out the contents of the folder (rather then appending to it like before) resulting in only a single plugin being sandboxed. I tried a couple things including giving the subprojects different root sandbox directories and then manually (in build.gradle) copying things over, but this didn't work. 

This isn't blocking anything but I just wanted to update. Any ideas welcome.

etanshaul:
lgtm! maybe add a note on this pr with a link to their fix for this.

etanshaul:
Oh. Also, we might not want to change the way we are defining the dependency in the PR in case it has other side-effects (e.g. leave it in the dependencies section like before).

aslo:
Ok, sounds good, I'll make that change. 

It also looks like we're failing CI builds since the newer plugin version [requires java 1.8](https://github.com/JetBrains/gradle-intellij-plugin/blob/master/build.gradle#L10). I'll update the travis config as well to reflect this.

etanshaul:
Oh it requires 8? I think thats a problem since we are targeting IJ 15+ which doesn't yet support java 8.

patflynn:
I guess we can still run our builds on java 8 but target 7.

etanshaul:
@patflynn but since we use Travis to generate the archives we upload to the JB repository, won't we have a binary compatibility issue if we build with java 8?

etanshaul:
obsolete. closing

Issue #702, opened 2016-05-26
Undeploy doesn't work


chanseokoh:
Just tried undeploying an app right after successful deployment. It completed immediately, but the app was in fact not stopped.

chanseokoh:
Strange...

I got this problem when I testing things at home on my external and personal project. As I said, undeploy seemed successful although it actually wasn't.

So, I just tried this at office. No problem when I deploy/undeploy to an internal project. When I tried on to my external project, deploy works fine but undeploy fails with permission denied:

> ERROR: (gcloud.preview.app.versions.stop) Error Response: [403] Operation not allowed
> 
> Details: [
>   [
>     {
>       "@type": "type.googleapis.com/google.rpc.ResourceInfo",
>       "description": "The \"appengine.services.list\" permission is required.",
>       "resourceType": "gae.api"
>     },
>     {
>       "@type": "type.googleapis.com/google.rpc.DebugInfo",
>       "detail": "[ORIGINAL ERROR] generic::permission_denied: Operation not allowed [google.rpc.error_details_ext] { code: 7 message: \"Operation not allowed\" details { type_url: \"type.googleapis.com/google.rpc.ResourceInfo\" value: \"\\n\\007gae.api\\\"5The \\\"appengine.services.list\\\" permission is required.\" } }"
>     }
>   ]
> ]
> 
> Failed to undeploy 'Filesystem JAR or WAR file - target-new.war': Stop application failed with exit code: 1

I will do some more tests, since I just noticed some strange bug of gcloud itself which doesn't make sense at all.

etanshaul:
What was the bug that you noticed? Is it related to this?

chanseokoh:
Seems somewhat related... I think I found another bug here by accident. It may be the case that the undeploy action isn't using the login credential from our plugin but the credential from the gcloud. The internal bug 28977605 can explain the permission denied error.

etanshaul:
If that were the case, couldn't we reproduce this with an internal account too? I.e. make sure you're logged out of gcloud and attempt to undeploy using the IDE.

chanseokoh:
Indeed. I tried logging out on the console. Didn't do anything on the IDE. Undeploy gives me the same permission error. Then I tried logging into my personal account on the console. I set the default project to an arbitrary string on the console. It gave me the same error. I set the project to a correct one. Didn't do anything on the IDE. Then the undeploy worked successfully. It is certain that undeploy is using the gcloud's credential.

[...]

PR #714, opened 2016-06-08
Set cloud project name on the app engine stop config

### LINKED ISSUE #713: App Engine stop configuration needs the cloud project name
Without this, it can only pick up the default project name from the CLI. We want to explicitly send the cloud project name to the common-lib / gcloud.

etanshaul:
fixes #713


chanseokoh:
LGTM assuming that you did a test.

This PR also closes #702.

PR #715, opened 2016-06-08
Update copyright header template

### LINKED ISSUE #712: incorrect copyright header
We should be using the Google copyright header for new files in this project and not the AOSP copyright header.

s/
Copyright (C) 2016 The Android Open Source Project
/
Copyright 2016 Google Inc. All Rights Reserved.
/

chanseokoh:
In relation to #712.

- `.idea/copyright/Android_Copyright.xml` renamed to `.idea/copyright/Google_Copyright.xml`
- `.idea/copyright/profiles_settings.xml`

### DIFF
--- a/.idea/copyright/Android_Copyright.xml
+++ b/.idea/copyright/Android_Copyright.xml
@@ -1,6 +1,0 @@
-<component name="CopyrightManager">
-  <copyright>
-    <option name="myName" value="Android Copyright" />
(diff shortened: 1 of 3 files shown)


etanshaul:
lgtm. but you may want to confirm that we are ok changing copyrights on files that previously had the old version (rather than just for new files going forward). I don't see why not, but IANAL

chanseokoh:
Right, I also thought that's the usual way of updating copyrights. I expect to update this PR accordingly. @patflynn will say something.

patflynn:
for now let's just update the template so new files have the right one.

chanseokoh:
You mean updating only `Google_copyright.xml` and `profiles_settings.xml` and leaving the current copyright headers intact?

chanseokoh:
Updating only the template and the IDE setting. We can think later about what to do with the existing headers.

etanshaul:
lgtm

patflynn:
:shipit:

PR #716, opened 2016-06-08
Fix label message clipping

chanseokoh:
The label was clipped like below, so I fixed it.

deploy_config2


patflynn:
it looks like the clipping is a width issue, why does changing the height fix it?  Also I think we probably don't want fixed width as we don't know the true string length when localized.

chanseokoh:
I think the height is for the entire editor. I tried several times changing only a flag for the field, but whenever I do that, the GUI Designer also changes the height. What's weird, but I decided to let the Designer do what it wants to do, because this will nevertheless happen when the next time the form is changed.

patflynn:
:8ball: lgtm

PR #718, opened 2016-06-09
Use the common library for flexible environment staging

### LINKED ISSUE #717: App Engine flexible environment staging should use the common app-tools library

etanshaul:
fixes #717 

@patflynn @chanseokoh 

A couple things to note:
- I renamed and moved the jar/war versions of the Dockerfile templates so that we wouldn't have to do any programmatic renaming prior to sending them to the common lib
- I am still manually staging the artifact file. The reason is that this file could have any name based on the project configuration. Our Dockerfile templates currently hardcode a pointer to `target.jar` / `target.war`. The simplest path here was to just move the artifact to the staging directory and rename it. The other more complicated option (but minimizes manual staging on our part) is to make the Dockerfile a dynamic template (like freemarker or something) where we can pass in the name of the artifact rather then using `target` for everything. Thoughts on this?

### DIFF
--- a/google-cloud-tools-plugin/resources/generation/src/appengine/mvm/app.yaml
+++ b/google-cloud-tools-plugin/resources/generation/src/appengine/mvm/app.yaml
@@ -1,4 +1,4 @@
-runtime: custom
+runtime: java
 vm: true
 env_variables:
   'DBG_ENABLE': 'true'
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
@@ -129,7 +129,7 @@ public class AppEngineDeploymentConfiguration extends
     this.version = version;
   }
 
-  public boolean isAuto() {
-    return getConfigType() == ConfigType.AUTO;
+  public boolean isCustom() {
+    return getConfigType() == ConfigType.CUSTOM;
   }
 }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineFlexibleStage.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineFlexibleStage.java
@@ -16,16 +16,16 @@
 
 package com.google.cloud.tools.intellij.appengine.cloud;
 
+import com.google.cloud.tools.appengine.api.deploy.DefaultStageFlexibleConfiguration;
+import com.google.cloud.tools.appengine.cloudsdk.CloudSdkAppEngineFlexibleStaging;
 import com.google.cloud.tools.intellij.appengine.util.CloudSdkUtil;
 
 import com.intellij.openapi.diagnostic.Logger;
-import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.remoteServer.runtime.log.LoggingHandler;
 
 import org.jetbrains.annotations.NotNull;
 
 import java.io.File;
-import java.io.IOException;
 
 /**
  * Stages an application in preparation for deployment to the App Engine flexible environment.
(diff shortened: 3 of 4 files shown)


etanshaul:
Discussing this with @meltsufin we realized this needs a bit of work:

Doing the manual staging of the artifact ourselves and then relying on the common lib for the rest of the staging results in a double copying of the artifact (the common lib assumes nothing is pre-staged and will re-copy everything over).

If we rely on gcloud to generate the Dockerfile (and not use templates) then we don't have to do any renaming of the artifact. If the user chooses to override the Dockerfile, then we may be able to assume that they correctly name the artifact pointer in the file. This way we can rely on the common-lib for all of the staging.

etanshaul:
Actually. If we use `runtime: custom` in our app.yaml then relying on gcloud to generate the Dockerfile is not an option.

etanshaul:
Ok I ran another test:

If we change `runtime: custom` to `runtime: java` in our app.yaml, then we can rely on gcloud entirely to perform the staging. It will generate a Dockerfile for and all we need to do is point it to the artifact and app.yaml. @patflynn is there a specific reason you can think of why we'd want to use custom instead of java for the runtime?

patflynn:
The only reason is that the field didn't exist when we built this feature.
For 'auto' configuration this wounds like the right approach.

etanshaul:
Your saying using `runtime: java` sounds like the right approach for auto configs? If so, I agree and this would simplify a few things: we could get rid of the Dockerfile templates entirely, we can rely 100% on the common library for flexible staging.

patflynn:
yeah that's what I'm saying, but don't we need to templates for custom?

etanshaul:
That's true. If we want to continue providing the ability to generate a Dockerfile then yes. But in the custom case, I think its safe to assume that its up the user to customize the generated Dockerfile to correctly point to the artifact (i.e. change target.jar/war to whatever is necessary). So we'll keep the templates, but they'll only be used in the 'custom' scenario.

patflynn:
I'd also be curious to see what the generated docker file looks like
compared to our templates.

etanshaul:
Here it is. I tested that earlier:

FROM gcr.io/google_appengine/openjdk8
ADD gs-spring-boot-0.1.0.jar /app/
ENTRYPOINT ["java", "-jar", "/app/gs-spring-boot-0.1.0.jar"]

patflynn:
@chanseokoh
@erezhaba

I'm surprised that they're not using the docker-entrypoint bash like we
do.  I thought that was required for cloud debugger to work.

etanshaul:
I've made the following updates:

- Offload all staging activity to flexible
- For 'auto' flexible deployments do not provide a Dockerfile (pending confirmation that we are ok using the gcloud generated one)
- Dockerfile templates for jar/war restored to their original form - these will be used by 'custom' flexible deployments

patflynn:
Let's wait until we can confirm with chanseok and erez that the Dockerfile
generated from runtime: java is actually working for the CDB.  If not, I'd
like to fix that before we start using it.

chanseokoh:
The [base Dockerfile](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/blob/async-support/openjdk8/src/main/docker/Dockerfile#L59) already contains the ENTRYPOINT, so it's going to be okay. The question is whether the currently published image was built from this source. I will try to do some test.

patflynn:
What happens when an image redefines the ENTRYPOINT of its parent?

patflynn:
@gregw should https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/tree/async-support/openjdk8 have an example of a a derived image that starts a jar? i.e. the openjdk8 on app engine flex use case.

chanseokoh:
@patflynn Ah, I missed that @etanshaul's Dockerfile redefines ENTRYPOINT. @etanshaul How did you generate the Dockerfile above exactly?

etanshaul:
The generated Dockerfile comes from a template in our plugin. So we can define it however we want. Not sure if that answers your question.

patflynn:
I think the issue is with the docker image generated when using runtime:
java since it redefines the entrypoint and doesn't run
docker-entrypoint.bash

etanshaul:
As a sort of side note: I realized that for custom deployments, app.yaml needs to have runtime: custom in order for a Dockerfile to be supplied. So we'll need an alternate app.yaml template assuming we go with the current proposed strategy.

chanseokoh:
@etanshaul The [Dockerfile](https://github.com/GoogleCloudPlatform/gcloud-intellij/blob/master/google-cloud-tools-plugin/resources/generation/src/appengine/mvm/jar.dockerfile) template in our plugin is different from what you posted above. It however matches what `gcloud preview app gen-config` generates. The `app-tools-lib` seems to run the `gen-config` command according to the lib source code, so I suspect you somehow got it from the lib.

And you're right that we have to set `runtime: cumstom` if we want to supply a custom `Dockerfile`.

In any case, I'm still doing a test to verify something.

etanshaul:
Yes, what I pasted above is the Dockerfile generated by gcloud. That was the source of this discussion about redefining entrypoint.  I think the confusion here is when we use the word 'generated' which is overloaded to mean "generated by gcloud" or "generated by the IJ plugin UI when doing a custom flex deploy".

chanseokoh:
Sorry, the answer below could have given first and eliminated some unnecessary communications.

@patflynn You're right that we need the following Dockerfile (which is the template we supply in our plugin) for CDBG to work:

```
FROM gcr.io/google_appengine/openjdk8
ADD target.jar /app/
ENTRYPOINT ["/docker-entrypoint.bash"]
CMD ["java","-jar","/app/target.jar"]
```

The following Dockerfile (that @etanshaul pasted above and what `gcloud gen-config` generates currently) doesn't set up CDBG:

```
FROM gcr.io/google_appengine/openjdk8
ADD target.jar /app/
ENTRYPOINT ["java", "-jar", "/app/target.jar"]
```

`gen-config` should be fixed to generate the first Dockerfile.

On our IntelliJ-side, this means that we need to always set `runtime: custom` and supply our custom Dockerfile even for the AUTO deploy mode (if we want to provide a workaround for this).

Side note: the good news is that CDBG currently works for JAR deployment (if DBG_ENABLE is set in `app.yaml`), whereas CDBG is broken for WAR deployment.

patflynn:
So we need to update the app.yaml template to include DBG_ENABLE?

chanseokoh:
We added DBG_ENABLE some time ago, so CDBG currently works for JAR-only deployment from our plugin.

chanseokoh:
Just a note about our JAR-deploy template:

```
FROM gcr.io/google_appengine/openjdk8
ADD target.jar /app/
ENTRYPOINT ["/docker-entrypoint.bash"]
CMD ["java","-jar","/app/target.jar"]
```

At the current state, I verified that the ENTRYPOINT above is redundant. (It's in the base Dockerfile [as mentioned previously](https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/718#issuecomment-225016698).)

patflynn:
@etanshaul 
I guess this work can continue now.

etanshaul:
this is extremely outdated. closing.

PR #724, opened 2016-06-10
Fix missing user specified artifact when the project has a standard f…

### LINKED ISSUE #721: unable to deploy jar springboot module when it's in a standard based project
Even though I have a spring-boot jar based module in my project, when I select 'deployment' in our cloud deployment run config, the only presented option is the exploded war artifact from my App Engine standard module.

Steps to reproduce:

1) Create a new app engine standard project using the IJ wizard
2) Deploy using our cloud intregation. [works well]
3) Add a new gradle baseed module using spring initializr
4) add a sample controller like so:
```
package com.example;

import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
import org.springframework.stereotype.Controller;
import org.springframework.web.bind.annotation.RequestMapping;
import org.springframework.web.bind.annotation.ResponseBody;

/**
 * Created by paflynn on 6/10/16.
 */
@Controller
@EnableAutoConfiguration
public class ExampleController {

    @RequestMapping("/")
    @ResponseBody
    String home() {
        return "yo mama is weak";
    }

    public static void main(String[] args) {
        SpringApplication.run(ExampleController.class, args);
    }
}
```

5) verify that it runs by running the gradle bootRun task. (make sure your gradle is configured to use jdk8
6) create a new App Engine cloud deployment run config
7) observe that the only available deployment source is the standard exploded war artifact.

patflynn:
…acet somewhere

Closes #721

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
@@ -24,10 +24,12 @@ import com.google.common.collect.Lists;
 
 import com.intellij.facet.Facet;
 import com.intellij.facet.FacetManager;
+import com.intellij.openapi.module.JavaModuleType;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.module.ModuleManager;
 import com.intellij.openapi.module.ModulePointer;
 import com.intellij.openapi.module.ModulePointerManager;
+import com.intellij.openapi.module.ModuleType;
 import com.intellij.openapi.project.Project;
 import com.intellij.packaging.artifacts.Artifact;
 import com.intellij.packaging.artifacts.ArtifactManager;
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
```diff
@@ -106,15 +106,22 @@ private AppEngineUtil() {
...
+          }
+        }
       }
     }
 
-    Artifact[] artifacts = ArtifactManager.getInstance(project).getArtifacts();
-    if (!containsAppEngineStandardArtifacts(project, Arrays.asList(artifacts))) {
+    // I kind of expected the logic for adding the user specific deployment source to be in the
+    // artifact based deployment sources.
+    if (hasNonAppEngineStandardJavaModules) {
```
etanshaul: My original thinking here was since the "user specified" source kind of lives outside of the module/project hierarchy (the user can select _anything_) that we'd only show it if the project has no AE standard artifacts. But, on the other hand, this doesn't match the logic I was using for maven based deployment sources.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
```diff
@@ -106,15 +106,22 @@ private AppEngineUtil() {
...
+            moduleDeploymentSources.add(createMavenBuildDeploymentSource(project, module));
+          }
+        }
       }
     }
 
-    Artifact[] artifacts = ArtifactManager.getInstance(project).getArtifacts();
-    if (!containsAppEngineStandardArtifacts(project, Arrays.asList(artifacts))) {
+    // I kind of expected the logic for adding the user specific deployment source to be in the
+    // artifact based deployment sources.
```
etanshaul: With regard to the comment: I _think_ it could be an artifact deployment source (currently `UserSpecifiedDeploymentSource` is a module deployment source which is why its here) . But thinking about this more, it might make sense to even create a new type of DeploymentSource (neither module, nor artifact). We can look at this in a separate PR to see if that makes sense.

etanshaul:
There is a bug with the logic for enabling user browseable deployment sources:

This is a project I have that is entirely AE standard - all modules have appengine-web.xml etc. I am debugging this.

image

etanshaul:
I found the source of the problem in the above screenshot and it points to a slightly more widespread issue (mainly with the logic prior to this PR):

The AE standard project I was testing with has an EAR module that contains no artifacts (all the artifacts are produced by the other war modules in the project). Therefore `containsAppEngineStandardArtifacts` was returning a (primitive) false for the ear module resulting in the logic thinking that it's a valid flexible module (its not) - hence the incorrect Filesystem deployment source.

The logic for determining what is actually a flexible module needs to be looked at.

patflynn:
Yeah I think we should just stop trying to guess and always let the user add a user specified artifact.  That might allow us to remove a bunch of code wouldn't it?

patflynn:
What do you think about merging this and then fixing that?

etanshaul:
Yea you're right - the suggestion of always having a user specified artifact does simplify things. Sure - go ahead and merge (since this does fix this other issue) and we can continue iterating on other PRs.

Issue #709, opened 2016-06-03
Git push should trigger the same checking as travis.


patflynn:
When a user pushes, we should run the same tests as travis to notify the user if the CI is likely to fail.

etanshaul:
Do you mean something like a client-side [git hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) that runs gradle check (or whatever)? That could be cool.

patflynn:
yeah!

chanseokoh:
I could do something like below locally, if the purpose is to explicitly trigger a check locally and show a notification locally for the sake of convenience:

```bash
#!/bin/sh
git push @*
if [ $? == 0 ] && [ in a git repo directory ] && [ -e gradlew ]; then
  gradlew clean check # Perhaps we could do this in a background
  if [ $? != 0 ]; then
    notify-send -t 120000 "Check failed" "Hey, your check failed."
  fi
fi
```

patflynn:
To which hook? pre-commit or pre-push?

chanseokoh:
I wasn't specifically referring to the git hook. Not sure what it is. The above script is just a random stuff I could do locally for myself, to throw some random idea. Not really recommending it.

patflynn:
I'd prefer a solution checked into .git/hooks so it runs for everybody by
default.

PR #725, opened 2016-06-10
Install pre push hook

patflynn:
related to #709 

Just an example.  Not sure I want to force this.  Having to wait minutes to push ended up being annoying.

### DIFF
--- a/pre-push
+++ b/pre-push
@@ -1,0 +1,18 @@
+#!/bin/sh
+# From gist at https://gist.github.com/chadmaughan/5889802
+# Copy into your .git/hooks and chmod +x
+
+# stash any unstaged changes
+git stash -q --keep-index
+
+# run the tests with the gradle wrapper
+./gradlew check
+
+# store the last exit code in a variable
+RESULT=$?
+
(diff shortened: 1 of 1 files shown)


etanshaul:
Agree that it might be annoying to wait. Just taking a step back - if the check fails up in travis then we do get notified by email. That's sort of one of the reasons travis is there for - to be a line of defense for our master branch. I wonder if we really need this extra step?

patflynn:
The idea is that running locally is a fair bit faster then waiting to find
out you forget to optimize imports by travis.

This change doesn't force anybody to use it.

chanseokoh:
One problem I see is not restoring the stash when one cancels the check in the middle (Ctrl+C).

etanshaul:
We should be able to catch the sigint using a [trap](http://linuxcommand.org/wss0160.php).

patflynn:
Without objection I'm just going to merge this so people can use it for reference (if they want to).  Feel free to add a sigint trap to make it more robust.

etanshaul:
cool with me. lgtm

PR #728, opened 2016-06-13
Flexible compat environment determination

### LINKED ISSUE #720: App Engine standard apps with vm=true identified as being deployed to 'standard environment'
when a standard app has <vm>true</vm> the target environment in the deployment run configuration should be idenfied as 'flexible environment'

standard-should-be-flex

etanshaul:
fixes #720 

@patflynn, @chanseokoh 

Currently, the only use of determining this is for labeling in the UI.


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineEnvironment.java
```diff
@@ -23,15 +23,25 @@
...
+  APP_ENGINE_FLEX_COMPAT("appengine.environment.name.flexible");
 
   private String label;
 
   AppEngineEnvironment(String label) {
     this.label = label;
   }
 
+  public boolean isStandard() {
+    return this == APP_ENGINE_STANDARD || this == APP_ENGINE_FLEX_COMPAT;
```
chanseokoh: Isn't #720 saying flex-compat should be identified as flex env?

etanshaul: I'm glad you called this method out:

Yes, its identified as flex, but its structured like a standard project (has an appengine-web.xml etc.). I'm thinking I might just create a new helper `isFlexCompat()` and just let the callers handle it the way they like - this will be identical to the current logic, but hopefully clearer.

etanshaul:
Refactored the helper methods to not clump standard and flex-compat environments together under `isStandard()`.

patflynn:
There are two kinds of environment: Standard and Flexible.

Compat is not an 'environment', it's a specific runtime configuration that
runs in the flexible environment. I don't think it makes sense for our
Environment model to be the place to capture the information about the
runtime.

etanshaul:
That is a fair point. I'll think about how to refactor this.

etanshaul:
Updated:
- removed the 'compat' concept from the Environment enum, instead flex-compat projects are marked as targeting the flexible environment. It's now up to any client that needs to distinguish between flex and flex-compat to make the determination. For example, to determine if the deployment path should be a 'standard' deployment path it now looks like this:

```
if (targetEnvironment.isStandard()
        || (targetEnvironment.isFlexible() && AppEngineUtil.isFlexCompat(project, source))) { ...
```

With these changes, the UI labeling issue is resolved implicitly.

I haven't moved the logic that loads the appengine-web.xml out of the static helper as I'm still looking into if this makes sense.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/dom/AppEngineStandardDomElement.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.intellij.util.xml.DomElement;
+import com.intellij.util.xml.GenericDomValue;
+
+/**
+ * DOM representation of the App Engine standard environment appengine-web.xml file.
+ */
+public interface AppEngineStandardDomElement extends DomElement {
+
+  GenericDomValue<String> getVm();
```
patflynn: I think this can return Boolean. and then can be isVm().   We actually have the beginning of this in the Android Studio plugin uder com.google.gct.studio.appengine.dom

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -137,9 +138,10 @@ public CancellableRunnable createDeployRunner(
         wrapCallbackForUsageTracking(
             callback, deploymentConfiguration, source.getFile(), targetEnvironment));
 
-    if (targetEnvironment == AppEngineEnvironment.APP_ENGINE_STANDARD) {
+    if (targetEnvironment.isStandard()
+        || (targetEnvironment.isFlexible() && AppEngineUtil.isFlexCompat(project, source))) {
```
patflynn: we have a real terminology problem around identifying 'standard framework' vs 'standard environment'.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/dom/AppEngineStandardDomElement.java
etanshaul: Yea that would be cleaner. I'll try it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
patflynn: it seems like environment is actually not the thing that decides which runner to use. I wonder if we couldn't change teh logic to be like

if (isStandardFrameworkApp(project, source) return newStandardFrameworkAppDeployer(..); else return create newNonStandardFrameworkAppDeployer(..);

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/dom/AppEngineStandardDomElement.java
loosebazooka: Boolean can be null, no?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
etanshaul: yea we really do. Now that you mention it - if AppEngineEnvironment wasn't thought of as the target cloud environment, but instead we thought of it as the 'framework type', then possibly my code prior to these last changes would make more sense:

3 framework types, 'standard', 'flexible', 'flex-compat' (maybe even 'standard-compat' would make more sense).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
```diff
@@ -113,13 +124,50 @@ private AppEngineUtil() {
...
+      artifact = ((AppEngineArtifactDeploymentSource) source).getArtifact();
+      if (artifact != null) {
+        return isFlexCompat(project, artifact);
+      }
+    }
+
+    return false;
+  }
+
+  private static boolean isFlexCompat(@NotNull Project project,
```
patflynn: I'd rather we not determine this in a static context. Could this be a method on the AppEngineArtifactDeploymentSource?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
patflynn: for frameworks, I'd say there's just 'standard' and 'non-standard'.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
etanshaul: I'd need to look into this. Currently in order to even create that kind of deployment source we need to first determine if the artifact is flex-compat. So doing that would require other refactoring.

patflynn:
Are you still working on some points or should I take a last look?

etanshaul:
I think it makes sense to merge in the current state (once you are good with it) and then improve in next PRs.

etanshaul:
There is one last issue where the IJ DOM manager sometimes returns cached elements (and doesn't pick up changes in appengine-web.xml) that I'm looking into here too.

patflynn:
OK :ship:

etanshaul:
I haven't merged this in yet because I'm getting passed this last hurdle. It doesn't have anything to do with the xml file itself being cached like I said. 

Its due to the fact that the bundled IJ plugin registers its own file description extension for appengine-web.xml leading two descriptions. We then get nondeterministic behavior when the description is loaded from the cache - when our plugin's description is loaded things are fine, but when the legacy one is loaded then it fails to read the xml file. Its not as simple as just unregistering their extension unfortunately. I'm still looking into it.

etanshaul:
@patflynn please take a look at the latest commit: https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/728/commits/6ad825d85d6c313cf8f1d6622e1d3fafcaff831c

Essentially I'm now relying on the file descriptor and dom element model that comes with the bundled plugin instead of defining our own for now. The negative consequence is having to do a little but of ugly xml traversing (with all those nested null checks).

patflynn:
Can this code ever execute in CE?

etanshaul:
Good question. No - this only runs if there is the bundled app engine facet as well.

patflynn:
LGTM

PR #732, opened 2016-06-14
add gradle distribution with sources

patflynn:
apparently this gives us better autocompletion and stuff.  Also gets rid of the annoying IJ notification to do this.


etanshaul:
that's cool. lgtm.

PR #734, opened 2016-06-15
Save and load artifact deployment source types

### LINKED ISSUE #729: Artifact deployment source not deserialized on reboot
Steps:
1) Create a deployment run config with an Artifact type as the deployment source and save it.
2) Reboot the IDE
3) Attempt to run the previously created configuration
=> Stack trace (java.lang.RuntimeException: Invalid deployment source selected for deployment)

etanshaul:
fixes #729 

@patflynn, @chanseokoh 

This error occurs when deployment configurations containing an AE artifact deployment sources are created. Once these configurations are created, if the user closes the project and then attempts to reload the configuration, this fix will correctly recreate the artifact from the serialized state that was saved by the IDE. Without this fix, the `AppEngineArtifactDeploymentSource` will not be recreated - only its platform supplied superclass - which will miss out on AE specific details such as environment, and cause downstream errors during deployment.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSourceType.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSourceType.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright (C) 2016 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.intellij.appengine.cloud;
+
+import com.google.common.base.Optional;
+import com.google.common.base.Predicate;
+import com.google.common.collect.Iterables;
+
+import com.intellij.openapi.project.Project;
+import com.intellij.packaging.artifacts.Artifact;
+import com.intellij.packaging.artifacts.ArtifactManager;
+import com.intellij.packaging.artifacts.ArtifactPointerManager;
+import com.intellij.remoteServer.configuration.deployment.ArtifactDeploymentSource;
+import com.intellij.remoteServer.configuration.deployment.DeploymentSourceType;
+import com.intellij.remoteServer.impl.configuration.deployment.DeployToServerRunConfiguration;
+
+import org.jdom.Element;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Arrays;
+
+/**
+ * A {@link DeploymentSourceType} loading and saving of a {@link AppEngineArtifactDeploymentSource}.
+ */
+public class AppEngineArtifactDeploymentSourceType
+    extends DeploymentSourceType<ArtifactDeploymentSource> {
+
+  private static final String SOURCE_TYPE_ID = "appengine-artifact-source";
+  private static final String NAME_ATTRIBUTE = "name";
+
+  public AppEngineArtifactDeploymentSourceType() {
+    super(SOURCE_TYPE_ID);
+  }
+
+  @NotNull
+  @Override
+  public AppEngineArtifactDeploymentSource load(@NotNull Element tag, @NotNull Project project) {
+    final String name = tag.getAttributeValue(NAME_ATTRIBUTE);
+
+    Artifact[] artifacts = ArtifactManager.getInstance(project).getArtifacts();
+    Optional<Artifact> artifact = Iterables.tryFind(Arrays.asList(artifacts),
+        new Predicate<Artifact>() {
+          @Override
+          public boolean apply(Artifact artifact) {
+            return artifact.getName().equals(name);
+          }
+        }
+    );
+
+    Element settings = tag.getChild(DeployToServerRunConfiguration.SETTINGS_ELEMENT);
+
+    if (artifact.isPresent() && settings != null) {
+      String environment = settings.getAttributeValue(
+          AppEngineDeploymentConfiguration.ENVIRONMENT_ATTRIBUTE);
+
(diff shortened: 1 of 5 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSourceType.java
```diff
@@ -0,0 +1,85 @@
...
+          }
+        }
+    );
+
+    Element settings = tag.getChild(DeployToServerRunConfiguration.SETTINGS_ELEMENT);
+
+    if (artifact.isPresent() && settings != null) {
+      String environment = settings.getAttributeValue(
+          AppEngineDeploymentConfiguration.ENVIRONMENT_ATTRIBUTE);
+
```
patflynn: don't we resolve environment based off of the artifact in the normal flow?  Why are we storing it for the deserialization flow?

patflynn:
The link to the error is helpful, but it would also be nice to describe what was the underlying cause and how this fixes it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSourceType.java
etanshaul: good point. I think that we can use the normal flow here too. I'll refactor it.

etanshaul:
Which link do you mean? Do you mean add a description in the PR? If so, no problem.

patflynn:
Yeah sorry. I meant that the link to the issue was helpful, but it doesn't
really say what the underlying problem was, so the PR should provide a bit
of context.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSourceType.java
```diff
@@ -0,0 +1,85 @@
...
+      return new AppEngineArtifactDeploymentSource(
+          AppEngineEnvironment.valueOf(environment),
+          ArtifactPointerManager.getInstance(project).createPointer(artifact.get()));
+    } else {
+      return new AppEngineArtifactDeploymentSource(
+          ArtifactPointerManager.getInstance(project)
+              .createPointer(tag.getAttributeValue(NAME_ATTRIBUTE)));
+    }
```
chanseokoh: Can we use just `name` in place of `tag.getAttributeValue(NAME_ATTRIBUTE)`? Possibly with an elaborated variable name.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSourceType.java
```diff
@@ -0,0 +1,85 @@
...
+        new Predicate<Artifact>() {
+          @Override
+          public boolean apply(Artifact artifact) {
+            return artifact.getName().equals(name);
+          }
+        }
+    );
+
+    Element settings = tag.getChild(DeployToServerRunConfiguration.SETTINGS_ELEMENT);
+
```
chanseokoh: Seems like we can skip the above `tryFind` search if we check `settings != null` first (unless this kind of refactoring makes code more complex).

etanshaul: Maybe I didn't catch your point, but the tryFind is to filter our our target Artifact from the list (it returns the first that matches the predicate).

chanseokoh: Sorry, let me put this way. I mean in the current logic, if `settings == null`, it will always take the `else` branch (note this branch doesn't use `artifact`) regardless of the outcome of `artifact`. So, if we need to do tryFind only when `settings != null`.

etanshaul: I see. thanks.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSourceType.java
etanshaul: yes. I'll update this too.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSourceType.java
etanshaul: I think this actually might not be the way to do it. In order to get the environment via the "normal" flow (for AE standard artifacts) it has to perform a check for the existence of the app engine facet. 

The problem is that this deserialization logic happens before the facet is loaded.

etanshaul:
@chanseokoh refactored based on your comments

patflynn:
@etanshaul That's weird.  Run configs are not dumb aware so the project state should be available.

etanshaul:
Yea. I'm looking at it more carefully to confirm. Also I added a description of the problem to the first PR comment.

etanshaul:
@patflynn So if even though its not dumb aware it has to do with the order of the loading. At the time this configuration is deserialized, the project's modules have even been loaded yet.

patflynn:
that's odd. The artifacts have loaded but the modules haven't?

etanshaul:
Agree its odd. But its exactly what I'm seeing. I guess its just as valid to do it in that order (establish the link from the modules after the artifacts have been loaded).

patflynn:
Thank you for the updated description1

patflynn:
@etanshaul is there a risk that the environment should be different at deserialization from what it was at serialization?

etanshaul:
I was about to say that there isn't (because what gets serialized is the results of the full blown environment resolution logic). 

But I realized that a Deployment Configuration can contain multiple artifacts - each with potentially different environments - and I am only persisting a single environment. I need to investigate this aspect. Thanks for asking that question.

etanshaul:
False alarm on my last comment:

It works the same way as any of the other fields, like version for example (which can be different for every source). I think it stores the configuration for each source.

To answer your original question again: I don't see any risk here.

PR #737, opened 2016-06-16
Sync up with app-tools - remove setting of gsUtil

etanshaul:
@patflynn


chanseokoh:
I saw the changes in the app-tools before. LGTM. I just wonder why the other currently open PR builds successfully despite that `appCommandGsUtil()` doesn't exist anymore. (Perhaps it was built before app-tools pushed a snapshot recently?)

etanshaul:
Yep. It was build prior to their latest deploy.

PR #744, opened 2016-06-20
Restructure metrics event actions

chanseokoh:
It makes more sense to me to alter the current event structure.

Basically, I'm enforcing my rule that the second parameter to `trackEvent()` are event actions.


patflynn:
Is the purpose of this to make 'eventLabel' purely for qualifying an action with some extra dimension?

chanseokoh:
Yes, that's my intention. 'eventAction' is a concrete action that we want to track. 'eventLabel' and 'eventValue' is some extra metadata info (key-value pair).

chanseokoh:
Of course, we can reconsider the design from the ground.

patflynn:
No I think this is reasonable.

LGTM

PR #745, opened 2016-06-21
Don't show custom deployment options when flex compat

### LINKED ISSUE #740: Don't allow custom deployments for flex compat
In the deployment configuration dialog, if the project is flex compat, we should not be allowing for Dockerfile (yet) selection and app.yaml.

etanshaul:
fixes #740


patflynn:
:ship:

PR #749, opened 2016-06-21
Create pre-build task for app engine artifact sources

### LINKED ISSUE #748: App Engine Standard staging fails due to invalid source directory
Steps:

1) Create a basic app engine standard app via the project wizard
2) Select the exploded war as the deployment source
2) Attempt to deploy to app engine

Investigating to see if we need to sync with changes from app-tools.

screen shot 2016-06-21 at 2 35 01 pm

etanshaul:
fixes #748 

Since `AppEngineArtifactDeploymentSourceType` extends `DeploymentSourceType` which does not provide its own implementations of the build task creation methods, we need to add them.

@patflynn

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSourceType.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSourceType.java
@@ -20,10 +20,12 @@ import com.google.common.base.Optional;
 import com.google.common.base.Predicate;
 import com.google.common.collect.Iterables;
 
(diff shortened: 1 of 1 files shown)


chanseokoh:
I see the added implementation is precisely from `ArtifactDeploymentSourceType`, which also extends `DeploymentSourceType<ArtifactDeploymentSource>`, so LGTM.

I guess the reason we can't extend `ArtifactDeploymentSourceType` is because of our own `SOURCE_TYPE_ID`?

etanshaul:
Yes your guess is exactly right. When these types are deserialized, we specifically need our type to be created. And unless we can provide a custom `SOURCE_TYPE_ID` this is not possible. I don't consider the duplication to be a big deal because its very simple almost pass-through code.

chanseokoh:
LGTM.

PR #750, opened 2016-06-22
Track deploy and stop task execution events

### LINKED ISSUE #746: We should track deployment-type operations (stop, deploy ..) for analytics

chanseokoh:
Fixes #746 to track both deploy and stop events.

I don't see a great value of the previous deploy event tracking (this was what I implemented about 3 months ago) , which is triggered only after successful deployment.

Now we track the events when at the start of each task execution. Does this sound good?

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardDeployTask.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardDeployTask.java
@@ -58,7 +58,7 @@ public class AppEngineStandardDeployTask implements AppEngineTask {
   public void execute(ProcessStartListener startListener) {
     UsageTrackerProvider.getInstance()
         .trackEvent(GctTracking.CATEGORY, GctTracking.APP_ENGINE_DEPLOY,
-            isFlexCompat ? "flex.compat" : "standard", null);
+            isFlexCompat ? "flex-compat" : "standard", null);
 
     File stagingDirectory;
 
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardDeployTask.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardDeployTask.java
@@ -18,7 +18,9 @@ package com.google.cloud.tools.intellij.appengine.cloud;
 
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessExitListener;
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessStartListener;
+import com.google.cloud.tools.intellij.stats.UsageTrackerProvider;
 import com.google.cloud.tools.intellij.util.GctBundle;
+import com.google.cloud.tools.intellij.util.GctTracking;
 import com.google.common.annotations.VisibleForTesting;
 
 import com.intellij.openapi.diagnostic.Logger;
@@ -37,16 +39,27 @@ public class AppEngineStandardDeployTask implements AppEngineTask {
 
   private AppEngineDeploy deploy;
   private AppEngineStandardStage stageStandard;
+  private boolean isFlexCompat;
 
+  /**
+   * @param isFlexCompat does not change any behavior of actual deployment. Provided solely
+   *     for the purpose of Analytics usage reporting.
+   */
   public AppEngineStandardDeployTask(
       @NotNull AppEngineDeploy deploy,
-      @NotNull AppEngineStandardStage stageStandard) {
+      @NotNull AppEngineStandardStage stageStandard,
+      boolean isFlexCompat) {
     this.deploy = deploy;
     this.stageStandard = stageStandard;
+    this.isFlexCompat = isFlexCompat;
   }
 
   @Override
   public void execute(ProcessStartListener startListener) {
+    UsageTrackerProvider.getInstance()
+        .trackEvent(GctTracking.CATEGORY, GctTracking.APP_ENGINE_DEPLOY,
+            isFlexCompat ? "flex.compat" : "standard", null);
+
     File stagingDirectory;
 
     try {
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStopTask.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStopTask.java
@@ -17,7 +17,9 @@
 package com.google.cloud.tools.intellij.appengine.cloud;
 
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessStartListener;
+import com.google.cloud.tools.intellij.stats.UsageTrackerProvider;
 import com.google.cloud.tools.intellij.util.GctBundle;
+import com.google.cloud.tools.intellij.util.GctTracking;
 
 import com.intellij.openapi.diagnostic.Logger;
 
@@ -42,6 +44,9 @@ public class AppEngineStopTask implements AppEngineTask {
 
   @Override
   public void execute(ProcessStartListener startListener) {
+    UsageTrackerProvider.getInstance()
+        .trackEvent(GctTracking.CATEGORY, GctTracking.APP_ENGINE_STOP, null, null);
+
     try {
       stop.getHelper().stageCredentials(stop.getDeploymentConfiguration().getGoogleUsername());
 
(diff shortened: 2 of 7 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardDeployTask.java
```diff
@@ -47,6 +49,9 @@ public AppEngineStandardDeployTask(
 
   @Override
   public void execute(ProcessStartListener startListener) {
+    UsageTrackerProvider.getInstance()
```
etanshaul: I wonder if there's value in tracking compat vs non-compat deployments. This would have to be added here since compat deployments follow the 'standard' deployment flow.

etanshaul:
lgtm besides my question about compat tracking. I think the code is cleaner now with this refactoring too.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStopTask.java
```diff
@@ -42,6 +44,9 @@ public AppEngineStopTask(AppEngineStop stop, String module, String version) {
 
   @Override
   public void execute(ProcessStartListener startListener) {
+    UsageTrackerProvider.getInstance()
```
patflynn: is environment available at this point to add as a dimension?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -265,41 +261,6 @@ public void deleteCredentials() {
...
-      } else {
-        labelBuilder.append(".custom");
-      }
-    }
-    final String eventLabel = labelBuilder.toString();
-
-    return new DeploymentOperationCallback() {
-      @Override
-      public Deployment succeeded(@NotNull DeploymentRuntime deploymentRuntime) {
-        UsageTrackerProvider.getInstance()
```
patflynn: don't we still want to track success/failure/cancellation?

chanseokoh: In that case, let's revive this wrapper again.

etanshaul: I was thinking about that too. I think cancellation definitely sounds like something good for this (since its user initiated). But I'm not totally sure myself if success vs failure makes sense? Hard failures will be caught and reported to us anyway.

patflynn: I think it will be nice to be able to easily compare completion rate of
deployments in our dashboards without having to join them against our
reported exceptions (which have to be manually triggered by the user).

etanshaul: Yea fair enough. I forgot about the manual triggering component too.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStopTask.java
chanseokoh: Not available here, but If we really want, we could change the AppEngineStopTask constructor and pass through some info up to here. (I didn't do it ATM.)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardDeployTask.java
chanseokoh: Done.

Determining if this is compat wasn't readily possible (due to required context missing), so I ended up trickling down that info from callers after changing function signatures.

One alternative is to ping events at the time of creating a Deploy Runner instead of doing it inside `execute()`. This also sounds a fine solution.

chanseokoh:
For deploy cancellation, I am only reporting the action without any context ATM. To provide if it is flex, standard, etc, I think I need to change the current design structure somehow.

chanseokoh:
Let's take a look at the code.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStopTask.java
etanshaul: Well I think we know the environment implicitly since its only flexible environments which currently support stopping. Its true that some code is needed to know if its compat or not.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardDeployTask.java
```diff
@@ -37,16 +39,27 @@
...
     this.deploy = deploy;
     this.stageStandard = stageStandard;
+    this.isFlexCompat = isFlexCompat;
   }
 
   @Override
   public void execute(ProcessStartListener startListener) {
+    UsageTrackerProvider.getInstance()
+        .trackEvent(GctTracking.CATEGORY, GctTracking.APP_ENGINE_DEPLOY,
+            isFlexCompat ? "flex.compat" : "standard", null);
```
etanshaul: Might be ok, but I noticed you are overloading the `flex.[_]` label.

In `AppEngineFlexibleDeployTask` its used for auto vs custom. Here it indicates compat. I guess as long as we can run whatever filters we want within analytics then this is fine.

chanseokoh: Right. I changed it to "flex-compat".

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -137,11 +137,16 @@ public CancellableRunnable createDeployRunner(
         targetEnvironment,
         wrapCallbackForUsageTracking(callback, deploymentConfiguration, targetEnvironment));
 
-    if (targetEnvironment.isStandard()
-        || (targetEnvironment.isFlexible() && AppEngineUtil.isFlexCompat(project, source))) {
-      return createStandardRunner(loggingHandler, source.getFile(), deploy);
+    if (targetEnvironment.isStandard()) {
```
etanshaul: I'd need to see it to confirm, but I think all this code will be easier to read if you restore it to the way it was, then extract out a local `isFlexCompat` variable to use as needed. This gets rid of the nested conditionals and the comments you added to the unnamed parameters.

chanseokoh: What do you think of below?

```java
    boolean isFlexCompat =
        targetEnvironment.isFlexible() && AppEngineUtil.isFlexCompat(project, source);
    if (targetEnvironment.isStandard() || isFlexCompat) {
      return createStandardRunner(loggingHandler, source.getFile(), deploy, isFlexCompat);
    } else if (targetEnvironment.isFlexible()) {
      return createFlexRunner(loggingHandler, source.getFile(), deploymentConfiguration, deploy);
    } else {
      throw new AssertionError("Invalid App Engine target environment: " + targetEnvironment);
    }
```

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardDeployTask.java
etanshaul: Hmm "flex-compat" isn't really an environment (we had an extended discussion about this in another PR). I almost prefer it the way you had it despite the overloading. But maybe @patflynn can give his opinion on this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
etanshaul: yea thats better for sure.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStopTask.java
patflynn: nevermind. let's leave it out for now.

patflynn:
works for me.

LGTM when you've changed that conditional.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardDeployTask.java
chanseokoh: Let's interpret that as "a standard app deployed to the flex env that is compatible with the standard". I think it's fine since this is just a label in Analytics and not a strong type in code.

etanshaul: fine with me. lgtm

PR #752, opened 2016-06-22
Experimenting with new tracking API

patflynn:
Sending this out for thoughts.  I'm finding myself constantly wondering
what the null parameters to the trackEvent method represent.  Using this
kind of fluent interface makes the calling point more legible at the
cost of some interface craziness.

@etanshaul 
@chanseokoh


chanseokoh:
So, this sounds like adding new methods to `UsageTracker` doesn't break Android Studio? They are plugging in their own implementation of this interface currently.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTracker.java
```diff
@@ -33,4 +34,26 @@ void trackEvent(@NotNull String eventCategory,
...
+
+  interface PartialTrackingEventAction extends SendsEvent {
+
+    PartialTrackingEventLabel withAction(String value);
+
+    interface PartialTrackingEventLabel extends SendsEvent {
+
+      PartialTrackingEventValue andLabel(String value);
+
+      interface PartialTrackingEventValue extends SendsEvent {
```
etanshaul: any reason why this is nested?

chanseokoh:
Ah, I think to add new methods, it cannot be an interface anymore knowing that AS is using it.

patflynn:
We would modify AS at the time that we import these changes into that
code-base. For now they are on an older version that this change would not
affect.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTracker.java
patflynn: Yeah. each nested interface will never be constructed outside of the
context of its parent.  i.e. the parent must be specified before you can
specify the child.

etanshaul: Yea that makes sense. I guess it will always be the case that this order should be enforced.

chanseokoh: This is a strange-looking pattern I'm not aware of. What is this called?

In any case, we should make eventAction mandatory. Also, what I think ideal is that eventLabel/eventValue is a metadata pair and a user can provided many:

.addMetadata(key1, value1)
.addMetadata(key2, value2)

Currently, I settled down adding only one pair with the current `UsageTracker` interface. In that regard, changing the name `eventLabel` to `metadataKey` may be good.

etanshaul:
I'm good with this. :ship:

chanseokoh:
Just to confirm: if we change `UsageTracker`, we should change whatever implementation the Android Studio team wrote for that interface later if we ever sync with AS. We don't own that part of the code, according to what @appu said.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
```diff
@@ -100,6 +102,46 @@ public void trackEvent(@NotNull String eventCategory,
...
+  }
+
+  private class TrackingEventBuilder implements
+      PartialTrackingEventAction, PartialTrackingEventLabel, PartialTrackingEventValue {
+    private String category;
+    private String action;
+    private String label;
+    private Integer value;
+
+    public TrackingEventBuilder(String category) {
```
etanshaul: Oh, can the constructor be private?

etanshaul: Never mind, you aren't using it completely like a traditional builder.

chanseokoh:
However, I guess just adding a new method in the interface wouldn't be a problem, since we can always provide an no-op implementation in AS then.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
patflynn: you're right I think that it should be private.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTracker.java
etanshaul: the Sequential Builder Pattern? I just made that up :)

patflynn: interesting. so category and action are required for every event and label and value are not. is that right? 

I want to be more specific rather than more general with my labeling.  The intellij plugin developer shouldn't have to go look at the analytics pipeline to understand what value should be passed to what key.

patflynn: It's trying to be fluent, and maybe the scoping of the interfaces is
gratuitous. I wouldn't be super opposed to breaking them up as siblings. I
definitely would have done that if I thought there were use-cases where the
client might actually store one of these partial types as a left-hand-side
value.

chanseokoh: We actually have a lot of flexibility with what we can pass/provide. The current parameters enforced by `UsageTracker` don't map 1:1 well conceptually anyway.

The backend basically expects the notions of 'event type' and 'event name'. (It's not even certain if these are strictly technical requirements or something we can by-pass.)

After that, there is a notion of 'metadata' in the backend that we can supply in the form of multiple key-value pairs. This is absolutely optional.

There also some uninteresting fields that we can set/pass, but we don't need to be concerned about them.

However, we are not limited just by 'event type', 'event name', and 'metadata'. We can pass whatever (short) information we want to include, by the means of forming a custom (virtual) 'page url' in the way we want: for example, '/virtual/some-string/another-string/and.so.on/'. (This will need some extra handling when we want to extract certain info from the backend, but not a big deal since it will be simple string operations.)

So, the interface design is largely conceptual and up to our discretion.

chanseokoh: The reason I said 'eventAction' is mandatory in the current interface is that, without it, the event says nothing because the only info being pass is the category which is just "com.google.cloud.tools".

google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTracker.java
```diff
@@ -33,4 +34,81 @@ void trackEvent(@NotNull String eventCategory,
...
+
+        /**
+         * Sets the optional scalar value to be associated with this tracking event.
+         *
+         * @param value an optional scalar value that will be recorded as a metric against this
+         *              tracking event
+         * @return a fluent interface for sending the tracking event ping
+         */
+        SendsEvent setValue(Integer value);
```
chanseokoh: The old interface accepted an Integer, but it doesn't have to.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTracker.java
```diff
@@ -33,4 +34,81 @@ void trackEvent(@NotNull String eventCategory,
...
+   *    .andLabel("flex")
+   *    .setValue(1)
+   *    .send();
+   * }
+   *
+   * @param category this is used to identify the client being used. It will be typically the same
+   *                 for all invocations from the same plugin.
+   * @return a fluent interface that allows the user to set the 'action' being tracked by the event
+   */
+  //TODO: Check all invocations fo track event and consider defaulting 'category' to a global value.
```
chanseokoh: Currently, it is always "com.google.cloud.tools" for the both plugins. So, as discussed, removing the parameter seems a good idea.

PR #754, opened 2016-06-22
Follow the convention /virtual/eventType/eventName

chanseokoh:
The backend team just told me that the we should following the below convention when constructing a virtual page url:

`/virtual/<eventType>/<eventName>`

That is, the custom dimension cd19 has to follow /virtual. The eventType value will always be "gcloud-intellij".

The above convention is not ideal, so I asked the backend team if there can be some flexibility. In the mean time, I think we make the URL follow the convention for now.

### DIFF
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
@@ -81,17 +81,15 @@ public class GoogleUsageTracker implements UsageTracker {
         List<BasicNameValuePair> postData = Lists.newArrayList(analyticsBaseData);
         postData.add(new BasicNameValuePair("tid", analyticsId));
 
-        postData.add(new BasicNameValuePair("cd19", eventCategory));  // Event type
+        String externalPluginName =
+            ServiceManager.getService(IdeaAccountPluginInfoService.class).getExternalPluginName();
+        postData.add(new BasicNameValuePair("cd19", externalPluginName));  // Event type
         postData.add(new BasicNameValuePair("cd20", eventAction));  // Event name
         postData.add(new BasicNameValuePair("cd16", "0"));  // Internal user? No.
         postData.add(new BasicNameValuePair("cd17", "0"));  // User signed in? We will ignore this.
 
-        IdeaAccountPluginInfoService pluginInfoService =
-            ServiceManager.getService(IdeaAccountPluginInfoService.class);
-
         // Virtual page information
-        String virtualPageUrl =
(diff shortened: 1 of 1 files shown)


google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
```diff
@@ -81,17 +81,15 @@ public void trackEvent(@NotNull String eventCategory,
         List<BasicNameValuePair> postData = Lists.newArrayList(analyticsBaseData);
         postData.add(new BasicNameValuePair("tid", analyticsId));
 
-        postData.add(new BasicNameValuePair("cd19", eventCategory));  // Event type
```
patflynn: so we're not using category anymore ey?

chanseokoh: I see some options to chose, since the requirement is just that the url should be /virtual/eventType/eventName.

1. Have eventType some kind of high-level category like
/virtual/appengine/deploy-success (dot or dash doesn't matter.)
/virtual/cloud-debugger/create-breakpoint

In this case, there is no room to specify that it is from our plugin. That's very inconvenient. The backend team says we could pass our plugin name as metadata, but that still sounds inconvenient.

2. Set com.google.cloud.tools to eventType (we won't use `.getExternalPluginName()`.)
/virtual/com.google.cloud.tools/appengine.deploy

3. Ignore category completely
/virtual/gcloud-intellij/appengine.deploy

patflynn: It looks to me that the 'plugin name' and 'category' are basically passing
the same information.  In fact plugin name is better because
com.google.cloud.tools is too broad and en compasses all our plugins.

I think your strategy is good, but if we're ignoring 'category' shouldn't
we remove it as a parameter to the event call?

chanseokoh: I like removing it. We can do that!

chanseokoh: I mean we can do that with your new design, but not with this interface to be backward-compatible.

patflynn: Don't worry about Android Studio.  I checked their code-base, they do not
use the UsageTracker  interface outside of our plugin.

chanseokoh:
I'll merge this after getting a LGTM.

patflynn:
LGTM

PR #755, opened 2016-06-23
Updates our app-tools-lib-for-java to the new name and released version.

patflynn:
@chanseokoh 
@meltsufin 
@etanshaul


chanseokoh:
LGTM.

Will we retain the sonatype repo setting for a snapshot version, probably for the purpose of development?

patflynn:
that's a good question. I think we'll leave it for now.  I need to check out what level of snapshot checking our release plugin does.

PR #757, opened 2016-06-23
Service lookup should use interface

patflynn:
The ServiceManager basically maps interface types to implementations
based off of the plugin.xml.  Make sure to retrieve the implementations
using the interface type and not the implementation type.

I'm also moving the initialization to the constructor since the client
type should never change at runtime.

### DIFF
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.tools.intellij.stats;
 
+import com.google.cloud.tools.intellij.AccountPluginInfoService;
 import com.google.cloud.tools.intellij.IdeaAccountPluginInfoService;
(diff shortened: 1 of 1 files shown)


patflynn:
@etanshaul 
@chanseokoh

etanshaul:
lgtm

chanseokoh:
This seems to solve the NPE problem on my desktop.

etanshaul:
the login related NPE? thats weird. maybe test again without this change and see if you can even reproduce the npe.

chanseokoh:
Reproducible. It happens without this change. I think what @patflynn said is right; we should use an interface type when calling .getService().

etanshaul:
Ok that's great that it fixes that. Yea definitely should be using an interface..but I just didn't immediately see how your login related npe is fixed by this.

patflynn:
I suspected, but I'm still confused why it worked for us.

patflynn:
@etanshaul what do you think of cutting a new release with this fix?

etanshaul:
yes, we should. I'll do taht

PR #760, opened 2016-06-24
Update tracking api clients

patflynn:
I've also removed the category from the tracking API since it's global for a given plugin.

### DIFF
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
@@ -44,7 +44,7 @@ import java.util.List;
 /**
  * Google Usage Tracker that reports to Cloud Tools Analytics backend.
  */
-public class GoogleUsageTracker implements UsageTracker {
+public class GoogleUsageTracker implements UsageTracker, SendsEvents {
 
   private static final Logger logger = Logger.getInstance(GoogleUsageTracker.class);
 
@@ -70,11 +70,7 @@ public class GoogleUsageTracker implements UsageTracker {
           new BasicNameValuePair("cid",             // UUID for this IntelliJ client
               UpdateChecker.getInstallationUID(PropertiesComponent.getInstance())));
 
-  /**
-   * Send a (virtual) "pageview" ping to the Cloud-platform-wide Google Analytics Property.
-   */
-  @Override
-  public void trackEvent(@NotNull String eventCategory,
+  public void sendEvent(@NotNull String eventCategory,
       @NotNull String eventAction,
       @Nullable String eventLabel,
       @Nullable Integer eventValue) {
@@ -87,7 +83,7 @@ public class GoogleUsageTracker implements UsageTracker {
 
         List<BasicNameValuePair> postData = Lists.newArrayList(analyticsBaseData);
         postData.add(new BasicNameValuePair("tid", analyticsId));
-        postData.add(new BasicNameValuePair("cd19", externalPluginName));  // Event type
+        postData.add(new BasicNameValuePair("cd19", eventCategory));  // Event type
         postData.add(new BasicNameValuePair("cd20", eventAction));  // Event name
         postData.add(new BasicNameValuePair("cd16", "0"));  // Internal user? No.
         postData.add(new BasicNameValuePair("cd17", "0"));  // User signed in? We will ignore this.
@@ -108,8 +104,8 @@ public class GoogleUsageTracker implements UsageTracker {
   }
 
   @Override
-  public PartialTrackingEventAction trackEvent(String category) {
-    return new TrackingEventBuilder(this, category);
+  public PartialTrackingEventLabel trackEvent(String action) {
+    return new TrackingEventBuilder(this, externalPluginName, action);
   }
 
   private static void sendPing(@NotNull final List<? extends NameValuePair> postData) {
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
@@ -71,7 +71,7 @@ public class GoogleUsageTracker implements UsageTracker, SendsEvents {
               UpdateChecker.getInstallationUID(PropertiesComponent.getInstance())));
 
   /**
-   * Sends an event to Analytics.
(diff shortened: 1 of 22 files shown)


patflynn:
@etanshaul

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
```diff
@@ -87,7 +86,7 @@ public void trackEvent(@NotNull String eventCategory,
 
         List<BasicNameValuePair> postData = Lists.newArrayList(analyticsBaseData);
         postData.add(new BasicNameValuePair("tid", analyticsId));
-        postData.add(new BasicNameValuePair("cd19", externalPluginName));  // Event type
+        postData.add(new BasicNameValuePair("cd19", eventCategory));  // Event type
```
chanseokoh: Why are we passing `externalPluginName` as `eventCategory` in `trackEvent()`? This parameter is always `externalPluginName` here.

chanseokoh: Ah, because of the no-op implementation?

In any case, we then need to make sure that `virtualPageUrl` below is

`"/virtual/" + eventCategory + "/" + eventAction`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/git/GcpCheckoutProvider.java
```diff
@@ -69,7 +69,8 @@ public String getVcsName() {
   @Override
   public void doCheckout(@NotNull final Project project, @Nullable final Listener listener) {
     UsageTrackerProvider.getInstance()
-        .trackEvent(GctTracking.CATEGORY, GctTracking.VCS_CHECKOUT, null, null);
+        .trackEvent(GctTracking.VCS_CHECKOUT)
+        .ping();
```
chanseokoh: This can fit into one line.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/git/UploadSourceAction.java
```diff
@@ -114,7 +114,8 @@ public void update(AnActionEvent event) {
   @Override
   public void actionPerformed(final AnActionEvent event) {
     UsageTrackerProvider.getInstance()
-        .trackEvent(GctTracking.CATEGORY, GctTracking.VCS_UPLOAD, null, null);
+        .trackEvent(GctTracking.VCS_UPLOAD)
+        .ping();
```
chanseokoh: Ditto.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTracker.java
```diff
@@ -16,99 +16,64 @@
...
-   * @param category this is used to identify the client being used. It will be typically the same
-   *                 for all invocations from the same plugin.
-   * @return a fluent interface that allows the user to set the 'action' being tracked by the event
+   * @param action is typically a specific operation the user has performed in the plugin, and is
+   *               often prefixed with a domain such as 'appengine.' or 'clouddebugger.'
+   * @return a fluent interface for setting the remaining parameters of a tracking ping
    */
-  //TODO: Check all invocations fo track event and consider defaulting 'category' to a global value.
-  PartialTrackingEventAction trackEvent(String category);
+  PartialTrackingEventLabel trackEvent(String action);
```
etanshaul: Do we have naming issue with `PartialTrackingEventLabel`? My concern is that given this: 

```
    UsageTrackerProvider.getInstance().trackEvent(GctTracking.APP_ENGINE_STOP).ping();
```
Notice that track event returns a "partial" label yet its totally legitimate to send a ping from this. Note this is only a naming concern. Maybe its OK because it is also a `PingsAnalytics`..

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/util/GctTracking.java
```diff
@@ -31,6 +29,8 @@
   public static final String APP_ENGINE_STOP = "appengine.stop";
 
   public static final String PROJECT_SELECTION = "project.selection";
```
chanseokoh: PROJECT_SELECTION is not used anymore. Remove it?

chanseokoh:
FYI, it is and has been that  the category value (or `externalPluginName` currently) is same for the both account and cloud-tools plugins. Going forward, the value we will use is "gcloud-intellij" (previously "com.google.cloud.tools"). I think using the same name containing "intellij" makes sense, because if the account plugin used its own plugin name, we would not be able to distinguish pings from the account plugin in AS and the one in IntelliJ (unless we use different names).

google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTracker.java
patflynn: No you're right. The 'partials' don't make sense anymore.  Updating.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/git/GcpCheckoutProvider.java
patflynn: will fix.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/git/UploadSourceAction.java
patflynn: will fix

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
patflynn: good catch

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/util/GctTracking.java
patflynn: will fix.

patflynn:
@chanseokoh I think you're right that it's ok for the account plugin to report the same client.  I will want the android studio install to differentiate itself when we get there. Will file a bug.

chanseokoh:
I tested it with the latest commit. Working fine. :ship:

patflynn:
Thanks chanseok.

Issue #447, opened 2016-02-17
Remove hardcoding of IJ versions from plugin.xml


etanshaul:
Open for tracking purposes - pending enhancement to the gradle-intellij-plugin

See the hardcoded fix we used to close #438 to control IJ versions.

Once the referenced issue in the JetBrains gradle-intellij-plugin is addressed, we should removed the hardcoded versions and revert back to utilizing the plugin to automate the generation of plugin.xml.

etanshaul:
closed as "won't fix". this gives us more control

PR #762, opened 2016-06-27
IJ build version

etanshaul:
I noticed we were still targeting 15.0.3 in our plugin configuration (this determines which version of the IDE the plugin supports from the plugin repository). 

This updates it to 15.0.6.

@patflynn

### DIFF
--- a/google-account-plugin/resources/META-INF/plugin.xml
+++ b/google-account-plugin/resources/META-INF/plugin.xml
@@ -20,7 +20,7 @@
   <!-- "version" set by gradle-intellij-plugin -->
   <!-- "idea-version since-build" set by gradle-intellij-plugin -->
   <!-- Workaround for gradle-intellij-plugin limitation: https://github.com/JetBrains/gradle-intellij-plugin/issues/66 -->
-  <!-- Setting version to 15.0.6 and up. https://confluence.jetbrains.com/display/IDEADEV/IntelliJ+IDEA+15+143.1821.5+Release+Notes -->
+  <!-- Setting version to 15.0.6 and up. https://confluence.jetbrains.com/display/IDEADEV/IDEA+15+EAP -->
   <idea-version since-build="143.2370.31" />
 
   <description>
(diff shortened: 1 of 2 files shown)


google-account-plugin/resources/META-INF/plugin.xml
```diff
@@ -20,8 +20,8 @@
   <!-- "version" set by gradle-intellij-plugin -->
   <!-- "idea-version since-build" set by gradle-intellij-plugin -->
   <!-- Workaround for gradle-intellij-plugin limitation: https://github.com/JetBrains/gradle-intellij-plugin/issues/66 -->
-  <!-- Setting version to 15.0.3 and up. https://confluence.jetbrains.com/display/IDEADEV/IntelliJ+IDEA+15+143.1821.5+Release+Notes -->
-  <idea-version since-build="143.1821.5" />
+  <!-- Setting version to 15.0.6 and up. https://confluence.jetbrains.com/display/IDEADEV/IntelliJ+IDEA+15+143.1821.5+Release+Notes -->
```
chanseokoh: Let's change the numbers in the URL too.

chanseokoh: Hmm. The following page doesn't seem to exist though.

https://confluence.jetbrains.com/display/IDEADEV/IntelliJ+IDEA+15+143.2370.31+Release+Notes

etanshaul: Oh right thanks.

etanshaul: Yea. I'm trying to find a good URL to link to. There aren't really good mappings between the full version numbers and the public ones (I determine them mainly by launching the desired version of IJ, e.g. 15.0.6 and then looking at the about page to get the full version number).

patflynn: doesn't this get set by the gradle-intellij-plugin at build time?

etanshaul: @patflynn - no, see https://github.com/GoogleCloudPlatform/gcloud-intellij/issues/447

The basic idea is that the plugin doesn't currently provide a way to configure the "since" and "until" portions of the version selection to achieve what we want so we are manually entering the versions for now. Once the plugin is updated (https://github.com/JetBrains/gradle-intellij-plugin/issues/66#issuecomment-214879784).

etanshaul:
@chanseokoh, I updated the url to a page that references the mapping between 15.0.6 and the full version number.

chanseokoh:
LGTM

patflynn:
lgtm

Issue #731, opened 2016-06-14
The static App Engine utility methods that perform IO should be factored out as collaborators that can be mocked


patflynn:
Two reasons to do this:
1) Testing:  If we want to be able to unit test effectively we need to be able to swap out the stuff that depends on the deep IJ class hierarchy and infratructure.  Also as a general principle static methods should be pure functions.  They must not depend on hidden state or have side-effects.
2) Modeling:  I think it may actually improve readability if we move some of the static code into some of the model objects since the static methods are often describing a property of those objects.

patflynn:
I'm going to tinker around with this this morning.

etanshaul:
@patflynn are you working on this? If not, I can start with it.

patflynn:
Yeah go ahead. Planning has interrupted this.

What do you think about refactoring most of the 'Helper' features as an IJ
service with minimal context?  Also some of the utility methods look like
they belong on the objects they query.

I'd like the work to address this architectural issue to take a decent stab
at improving the test coverage of the infrastructure we're providing for
App Engine support.

And finally in small PRs when possible. (I know that's hard)

etanshaul:
Yes. I agree with your points here. I'll tackle this.

etanshaul:
closed as fixed

PR #764, opened 2016-06-28
Refactoring from static helpers to injected service

etanshaul:
In progress.

First push to address #731 

- Moves some of the static utility methods into an IJ service

Unit tests pending.

(I'll address the remainder of #731 in separate PR's).


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineProjectHelper.java
```diff
@@ -0,0 +1,129 @@
...
+   */
+  @Nullable
+  public abstract AppEngineEnvironment getAppEngineArtifactEnvironment(@NotNull Project project,
+      @NotNull Artifact artifact);
+
+  /**
+   * Returns the project's appengine-web.xml or null if it does not exist.
+   */
+  @Nullable
+  public abstract XmlFile loadAppEngineStandardWebXml(@NotNull Project project,
```
chanseokoh: For now, this method is used only by the Impl class. Does it need to be exposed (public)?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineProjectHelper.java
```diff
@@ -0,0 +1,129 @@
...
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Collection;
+
+/**
+ * A set of helper methods for inspecting an App Engine project's structure and configuration.
+ */
+public abstract class AppEngineProjectHelper {
```
patflynn: How about renaming this to AppEngineService?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineProjectHelper.java
etanshaul: I will either make it private or expose it in the impl with @VisibleForTesting if it makes sense to unit test it. I'm still working on the testing part.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineProjectHelper.java
etanshaul: Sure. What do you think about `AppEngineProjectService`. I like the "Project" part because these methods all deal with the inspection of an AE project's structure.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineProjectHelper.java
chanseokoh: I find that the methods below are also consumed only by the Impl, although they may still make sense to be public for future usage?

isAppEngineStandardArtifact
hasAppEngineStandardFacet
isAppEngineStandardArtifactType
isAppEngineFlexArtifactType

etanshaul: agree. my previous comment on this applies to all these cases. I'll make sure to address this before finalizing the pr.

etanshaul:
@patflynn, @chanseokoh ready for review

Sorry for the size, but it was kind of unavoidable since everything is tightly coupled. Here's a summary:

- Moved the static utility methods to a new injectable `AppEngineProjectService` to support querying project structure related App Engine things (i.e. things that rely on the project, modules, artifacts..)

- Create an injectable `AppEngineAssetProvider` to externalize loading of things like the `appengine-web.xml`. This is primarily done so it can be mocked in unit tests.

- Created test coverage for the helpers in `AppEngineProjectService` including the use of test `appengine-web.xml` files located in the `testData` directory.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectService.java
```diff
@@ -0,0 +1,123 @@
...
+public abstract class AppEngineProjectService {
+
+  public static final String APP_ENGINE_STANDARD_FACET_NAME = "Google App Engine";
+
+  public static AppEngineProjectService getInstance() {
+    return ServiceManager.getService(AppEngineProjectService.class);
+  }
+
+  /**
+   * Given a deployment source, returns the xml tag corresponding to the project's appengine-web.xml
```
patflynn: shouldn't it be '.. the deployment source's appengine-web.xml'?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectService.java
```diff
@@ -0,0 +1,123 @@
...
+
+  /**
+   * @see AppEngineProjectService#getFlexCompatXmlConfiguration(Project, DeploymentSource).
+   */
+  @Nullable
+  public abstract XmlTag getFlexCompatXmlConfiguration(@NotNull Project project,
+      @Nullable Artifact artifact);
+
+  /**
+   * Determines if a project is set up like an App Engine standard project but is configured in
```
patflynn: ditto here.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectService.java
```diff
@@ -0,0 +1,123 @@
...
+      Collection<Artifact> artifacts);
+
+  /**
+   * @see AppEngineProjectService#containsAppEngineStandardArtifacts(Project, Module).
+   */
+  public abstract boolean isAppEngineStandardArtifact(@NotNull Project project,
+      @NotNull Artifact artifact);
+
+  /**
+   * True if the artifact has the App Engine standard facet.
```
patflynn: {@code true}

patflynn: do artifacts have facets?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectService.java
```diff
@@ -0,0 +1,123 @@
...
+   */
+  public abstract boolean hasAppEngineStandardFacet(@NotNull Project project,
+      @NotNull Artifact artifact);
+
+  /**
+   * True if the artifact type is an exploded-war.
```
patflynn: {@code true}

patflynn: true if it's an exploded-war and has an appengine-web.xml?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectServiceImpl.java
```diff
@@ -0,0 +1,199 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.jetbrains.idea.maven.project.MavenProject;
+import org.jetbrains.idea.maven.project.MavenProjectsManager;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Set;
+
+public class AppEngineProjectServiceImpl extends AppEngineProjectService {
```
patflynn: class javadoc is required.

patflynn: I'm not a fan of the ...Impl naming pattern.  Is there anything else about the class that you could use to distinguish it from possible other implementations of this interface?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectServiceImpl.java
```diff
@@ -0,0 +1,199 @@
...
+
+  private AppEngineAssetProvider assetProvider;
+
+  public AppEngineProjectServiceImpl() {
+    assetProvider = AppEngineAssetProvider.getInstance();
+  }
+
+  @Nullable
+  @Override
+  public XmlTag getFlexCompatXmlConfiguration(@NotNull Project project,
```
patflynn: I've wondered why the client code would want an XmlTag rather than the actual value of some attribute that the appnengine-web.xml describes.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectServiceImpl.java
```diff
@@ -0,0 +1,199 @@
...
+    return null;
+  }
+
+  @Override
+  public boolean isFlexCompat(@NotNull Project project, @NotNull DeploymentSource source) {
+    return isFlexCompat(project, getArtifact(source));
+  }
+
+  @Override
+  public boolean isFlexCompat(@NotNull Project project, @Nullable Artifact artifact) {
```
patflynn: do you really even want to accept null?

google-cloud-tools-plugin/testData/descriptor/appengine-web.xml
```diff
@@ -0,0 +1,21 @@
...
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<appengine-web-app xmlns="http://appengine.google.com/ns/1.0">
+    <application>app</application>
+    <version>1</version>
+</appengine-web-app>
```
patflynn: missing new line.

patflynn:
Thanks for doing this. You move quick.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectService.java
etanshaul: definitely

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectService.java
etanshaul: Indirectly - artifacts are related to modules (in that they are produced by them); and modules have facets. I can update the comment to make it a bit clearer.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectService.java
etanshaul: Not for this method: this method is checking the Artifact**Type** so its just looking for exploded-war.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectServiceImpl.java
etanshaul: Fair question - this was done for convenience so that the UI code could determine if the flex-compat "type" was 'vm' or 'env'. I think I can refactor this like you suggest.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectServiceImpl.java
etanshaul: I'll update this so that the caller that could potentially pass null will handle it instead.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectServiceImpl.java
etanshaul: hmm I can't really think of any other possible implementations at the moment. Do you prefer the **Default**... naming pattern instead? I actually prefer Default, not sure why I went with impl. If I can come up with a more meaningful name though I'll update it.

patflynn: Sure. Default is fine in this case.

etanshaul:
@patflynn please see https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/764/commits/903f798a523a322421cc4e587b1be4074ad2998e for updates based on review

I also refactored some logic in `DefaultAppEngineProjectService` to reduce duplication in checking the contents of the flex-compat xml tag.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectService.java
patflynn: You missed this one.  s/project/Deployment Source

patflynn:
LGTM

PR #776, opened 2016-07-11
Suppress missing javadoc warning

etanshaul:
For some things like constructors that are well named and don't do much, its a little annoying to be forced to have to write a javadoc. I think its better to have no comment than a redundant one.

This suppresses this checkstyle check.

@aslo please take a look


aslo:
LGTM!

PR #777, opened 2016-07-11
Suppress checkstyle javadoc trailing period check

etanshaul:
Removing the checkstyle check that forces you to have a period at the end of the first sentence of a javadoc. Usually this is ok, but for things like the following it just gets in the way:

```
/**
 * @author user
 */

```

@aslo  ptal

### DIFF
--- a/google_checks.xml
+++ b/google_checks.xml
@@ -197,6 +197,7 @@
         <module name="JavadocTagContinuationIndentation"/>
         <module name="SummaryJavadoc">
             <property name="forbiddenSummaryFragments" value="^@return the *|^This method returns |^A [{]@code [a-zA-Z0-9]+[}]( is a )"/>
+            <property name="period" value=""/>
         </module>
         <module name="JavadocParagraph"/>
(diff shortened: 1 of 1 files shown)


aslo:
LGTM!

patflynn:
not sure about this change. I don't think the example is valid Google style. All public class and interfaces require javadoc including a descriptive comment.

etanshaul:
@patflynn - all of the code we integrated from the old plugin has author attribution in the source files like the example I showed. I didn't want to have to update them all with a trailing period just to pass checkstyle. But maybe because its now part of our plugin it makes sense to do so.

elharo:
Wow. This checkstyle check may be actively wrong. It's extremely pedantic, but the official rule is no period unless  there's a following sentence i.e. 

@author user

but

@author user. This person is really cool.

etanshaul:
That makes sense. I think that the rules just aren't granular enough to distinguish between a full sentence that would need a period and an author annotation with just a name as in your first example.

Issue #768, opened 2016-07-06
Warn users if legacy google-app-engine plugin is installed


etanshaul:
And prompt to remove the old JB plugin.

Depends on #767. Needed only until old plugin is debundled from IJ.

etanshaul:
this may be relevant to this too (same kinda of warning thing):

https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/651

PR #780, opened 2016-07-12
Integrate JetBrains google-app-engine-plugin into Google's plugin

### LINKED ISSUE #767: Merge functionality of current plugin offering with the legacy google-app-engine-plugin
This issue to is to track this effort.

etanshaul:
Fixes #767 

Base set of functionality needed to integrate the old plugin into ours to get everything running side by side with passing tests and passing static analysis checks.

More info coming..


etanshaul:
@chanseokoh @joaoandremartins @aslo 

This PR merges the [JetBrains GAE plugin](https://plugins.jetbrains.com/plugin/4254?pr=) into ours. It brings the JetBrains source into our module hierarchy while preserving the git history and attribution.

I made the minimal set of changes possible to get all the functionality working side by side (removing conflicts, making modifications where necessary etc.).  All of the features provided by the old plugin are integrated into the google-cloud-tools-plugin module or a submodule of it.

When a user is running Ultimate Edition, then extra functionality is loaded (provided by the "ultimate" plugin).

There currently will be conflicts if running this project in IJ with the old JB plugin enabled. We will handle this in #768. For testing this out its best to manually disable the JB plugin:

image

There are also several other changes that are needed but can be separated into other PRs - deprecation of all usages of the old appcfg sdk, updating local run to use our common lib, etc.

I know this PR is pretty massive and tough to review so we can discuss in person if needed.

(Note about the branches: I created a new branch call `integrate-jetbrains` to which all PRs related to this effort will be merged for now. I wanted to keep master separate for any releases we may need to make prior to merging this all in).

etanshaul:
As a guide, you can run through the JetBrains tutorial for their plugin and see that it still works with their plugin disabled (since the functionality has been merged in - App Engine standard project wizard, local run/debug, code whitelisting..) - excludes deploy since we have our own separate implementation:

https://www.jetbrains.com/help/idea/2016.1/getting-started-with-google-app-engine.html

chanseokoh:
Thanks for the link to the tutorial. I'll do some testing.

BTW, the build seems failing.

etanshaul:
Yea for some reason some tests are failing on Travis, I'll investigate
tomorrow.

etanshaul:
I figured out the test failure problem: some new tests rely on .iml files which were being ignored by .gitignore so I added an exclusion for test data. It should pass now hopefully.

chanseokoh:
I'm running UE. After disabling the JB AE plugin, I don't see the Java App Engine project wizard in Java. Interestingly, Java EE wizards also disappear.

chanseokoh:
I noticed that New Project Wizard doesn't really work in CE. (UE seems fine.)

- There is no `web` directory and `WEB-INF` is under root. (In UE, it's under `web`.)
- `index.jsp` doesn't exist. (In UE, it's under `web`.)
- A local run configuration is not setup automatically. In fact, CE can't create a new one, so there's no way to run a local dev server.

etanshaul:
Yes you are right. This is currently by design. The features you reference rely on UE functionality.

For the local dev server - I would have thought this would work in CE, but since support for AE standard apps rely on UE features this may not be the case. I'll look into this a little more.

etanshaul:
@chanseokoh - I looked into running the local dev server and this also relies on UE. It integrates into IJ's server functionality which is not available in CE. So everything you saw from your last comment is expected.

The only thing that is strange is it still lets you create a an app engine standard app from the wizard using Community Edition. We may want to disallow this later and instead only allow creation of a flex template (in a future PR).

chanseokoh:
Yeah, disallowing the wizard for standard seems like a good idea for now. (Or, at least adding `index.jsp` would be necessary, but not being able to run the local server is confusing and not consistent with the tutorial.)

chanseokoh:
I have played around a bit with creating a new project and running the local dev app server. They are working fine, so that's cool.

We should figure out what we will do with the licence header?

etanshaul:
Thanks for checking this out.

Do you mean the Jetbrains headers? That's intentional - we plan to keep their attribution (we also have the full git history).

chanseokoh:
Ah, yeah. Then it looks good. It is certainly unreasonable to review all of the JetBrains code, so I think we should move on and merge in the code after @joaoandremartins and @aslo also check that basic things are working, shouldn't we?

etanshaul:
Yes, agree.

aslo:
I'm currently going through the sample tutorial and some other testing, will let you know if I find anything that needs to be addressed.

etanshaul:
That's great. Thanks.

aslo:
I'm experiencing some issues with App Engine deployment. I created issue https://github.com/GoogleCloudPlatform/gcloud-intellij/issues/786 with some more detail.  I'll continue digging into the issue in the meantime.

aslo:
Ok, I'm satisfied - as far as I can tell this looks good!

etanshaul:
reopening pr to try merging differently (need to properly preserve history)

PR #651, opened 2016-04-25
Check for plugin compatibility

etanshaul:
fixes #650 

Adds a notification window if plugin versions are not equal:

image

Todo: finalize and externalize wording.

Note: the check is owned by the cloud tools plugin. Once this is pushed out and users update their plugin, any subsequent plugin updates where the versions are mismatched will show this warning.

PR #788, opened 2016-07-18
Warn users if legacy google-app-engine plugin is installed

### LINKED ISSUE #768: Warn users if legacy google-app-engine plugin is installed

aslo:
fixes #768 

This PR implements features to ease the transition to the merged version of our plugin, containing code that used to live in the  bundled "Google App Engine Integration" plugin.

* Adds a startup check for the existence of the bundled "Google App Engine Integration" plugin (similar to https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/651). With the changes in https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/780, the two cannot be running at the same time, as our plugin has been merged with theirs. The user is given the option to disable the old plugin.
* On project initialization, checks if 1) the old plugin has been removed, and 2) if the old plugin's facets are present. If so, it attempts to migrate them to the new plugin's facets 

screenshot from 2016-07-18 16 07 16

TODO:
- [x] detect when the old plugin is installed, and alert the user
- [x] make sure no exceptions are thrown when loading both plugins
- [x] when the user clicks our link, we open the plugin settings panel and highlight the old plugin


aslo:
@etanshaul Do you have any feedback on the wording here?  It would also be nice if there were some notice about this change from Jetbrains that we could link to here, as well. Does something like that exist?

etanshaul:
cool! does the ide still throw an exception if both plugins are enabled (i.e. prior to removing the old plugin using the new dialog)?

etanshaul:
There is no notice anywhere. I think we'll be ok for now with the wording we settle on here.

google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -144,6 +144,7 @@ Code inspections for App Engine Java code.</description>
                      level="ERROR" implementationClass="com.google.cloud.tools.intellij.appengine.validation.RestSignatureInspection"/>
 
     <postStartupActivity implementation="com.google.cloud.tools.intellij.PluginCompatibilityCheck"/>
+    <postStartupActivity implementation="com.google.cloud.tools.intellij.BundledAppEnginePluginExistenceCheck"/>
```
etanshaul: How about `ConflictingAppEnginePluginCheck`?

aslo:
Yep, it will throw an exception. The user would actually see something like this: 

screenshot from 2016-07-18 16 19 59

At first I was thinking that this would be acceptable since the exception doesn't prevent us from warning the user, and if anything, it provides clear evidence that action needs to be taken. Now I'm having second thoughts - what do you think?

google-cloud-tools-plugin/resources/META-INF/plugin.xml
aslo: :+1:

etanshaul:
For wording, here's a first take at tweaking it - I only made minor changes:

```
You are running an old plugin (Google App Engine Integration) that conflicts with the installed Google Cloud Tools for Java plugin.  

Please click here to deactivate the old plugin.
```

I just changed a couple words and shortened it (also notice the multi-line). Feel free to iterate on this too.

etanshaul:
I wonder if there's a way to direct link to deactivate the old plugin (rather than just opening the plugin page). Something like opening up the plugin page, disabling the old plugin automatically and having that row highlighted would be cool, but I don't know if its possible.

aslo:
I was wondering that as well - I'll look into that option a bit more.

etanshaul:
Looks good! Just need to fix the checkstyle problems (and maybe see if we can have better deep linking into the plugin page).

etanshaul:
Just curious - was the update to the FacetType id not needed? I see you reverted it.

aslo:
I originally changed the facet name to avoid an exception that IDEA throws at startup when it tries to load duplicate facets.  I reverted because I realized that changing the facet name will break any existing projects that reference it. 

Now I'm trying to look into other strategies for avoiding that exception.

etanshaul:
Good call.

Just off the top of my head - what if, on plugin initialization, we check for the conflicting facet type (coming from the old plugin) and then remove or rename them? This would depend on the plugin initialization hook (e.g. CloudToolsPluginInitializationComponent) being called before our facet is loaded and after the old facet is loaded.

etanshaul:
I just noticed that the IDE error dialog (when a plugin throws an exception for instance) has a direct link to disable a plugin:

image

Clicking on  "disable plugin" opens the following:

image

aslo:
@etanshaul The only remaining issue here is that duplicate facet exception that's thrown when intellij attempts to load two `google-app-engine` facets (one from this plugin, and one from the old bundled plugin). 

The exception is annoying, but as far as I can tell there really aren't any other negative consequences to the user. The exception will be immediately followed by our warning that the old plugin should be disabled. I'd be ok with that behavior if you are. 

If not, I'd love any suggestions for how to prevent the duplicate facet from being loaded and/or suppress
s the exception.

etanshaul:
remind me - why can't we rename our facet? Is it because of a concern for users who were previously using the old plugin with a facet configured?

aslo:
Indeed. As we discussed, I'm going to see if we can 1) give our facet a new name, 2) detect that a legacy project is using the old facet, and 3) reconfigure that project to point to our new facet.

aslo:
Closing because #780 has been merged into master. Will reopen new PR with master as base.

PR #795, opened 2016-07-27
Updates for IDEA 2016.2

### LINKED ISSUE #772: Investigate potential breaking API changes in IJ EAP 2016.2
When I target release candidate 2016.2 EAP:

ideaVersion = IU-162.1120.6 ([see here](https://confluence.jetbrains.com/display/IDEADEV/IDEA+2016.2+EAP) for the build #'s)

I get compilation errors. For example: `BreakpointConfigurationPanel:247`

It looks like for that instance `WatchRootNode#getAllChildren` was removed in this EAP.

aslo:
fixes #772 

This PR contains a few tweaks that are needed in order for the plugin to work on IDEA 2016.2 (aka 162.* builds). These changes are all backwards compatible, and the plugin built with these changes, targeting 2016.2, should still work for prior releases.


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: java
 jdk:
-  - openjdk7
+  - oraclejdk8
```
etanshaul: we may have discussed this already, but do we really want to build using Java 8 given that we are targeting Java 7 for IJ < 2016?

gradle.properties
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-ideaVersion = IC-15.0.6
+ideaVersion = IU-162.1447.7
```
etanshaul: did you mean to commit this?

aslo: Indeed I did.  Even though the changes in this PR are backward-compatible, bug fixes and behavior changes to the 2016.2 IDEA platform meant that I had to make some changes to our tests, so that they pass in new versions (but necessarily fail when building against older versions).  

I committed this change because this code will not pass a CI build targeting the older versions of the IDEA platform, even though it should run, once built, without issue for those versions. (The change from jdk 7->8 in travis.yml follows from this - the newer IDEA platform needs java 8).

etanshaul: Makes sense. So locally we'll have to update the target IJ to older versions when testing. Also even though travis builds with java 8 we should keep the jdk target locally to 7.

etanshaul: And when testing with IJ < 2016.2 locally, the tests will fail. I guess there's no way around this without maintaining multiple versions of our plugin (which seems like a more painful thing to do for now).

etanshaul:
@patflynn - what do you think about these changes? I'm in favor of doing this so that we can avoid having to maintain multiple versions of our plugin (one for IJ < 2016.2 and another for < 2016.2).

gradle.properties
etanshaul: @aslo - is the specific build number (162.14...) needed or can we achieve the same thing with 2016.2 here? If so that would make it clearer.

etanshaul: Also, you note that you changed it from IC to IU.

aslo: Good call, `2016.2` would definitely be clearer. I'll also update back to `IC`

etanshaul:
maybe @patflynn wants to take a quick look too

Otherwise LGTM.

patflynn:
I guess we should talk about this. As long as we want to support 15 I would expect us to be building and testing against it. I don't think we can change the idea version to 2016 until we drop support for 15 or we choose to manage two separate branches (if that's necessary).

aslo:
@patflynn Since the code is, for the most part, compatible with both versions, we could start running our travis builds against multiple targets. It would look something like this: 

```
language: java
matrix: 
- openjdk7
  env: IDEA_VERSION=15.0.6
- oraclejdk8
  env: IDEA_VERSION=2016.2
script: ./gradlew check -PideaVersion=$IDEA_VERSION
# etc ...
```

patflynn:
define 'for the most part' :)

aslo:
The plugin's src code is backwards compatible, but the tests currently are not, meaning that the same tests will not pass on the older releases. 

The only "flaky" test is the [EndpointImplicitUsageProviderTest](https://github.com/GoogleCloudPlatform/gcloud-intellij/blob/master/google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/validation/EndpointImplicitUsageProviderTest.java), which fails on builds that target `15.0.*` IDEA releases, because there is a bug in the IDEA platform code from those releases, that was apparently fixed in 2016.2. 

I'm currently looking into ways to modify the tests such that they can pass on both versions, without loosening up the test's assertions too much.

etanshaul:
Its weird that we had a unit test that was asserting a bug in the platform.

etanshaul:
I like the updates. As we discussed, the only open thing for me is figuring out if we want to use that build matrix in travis (and if so which versions should we specify since right now it seems a little arbitrary).

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/validation/EndpointImplicitUsageProviderTest.java
```diff
@@ -15,32 +15,69 @@
...
+    if (psiClass.getName().contains(ASSERT_IMPLICIT_USAGE_TRUE)) {
+      assertTrue(endpointUsageProvider.isImplicitUsage(psiClass));
+    } else if (psiClass.getName().contains(ASSERT_IMPLICIT_USAGE_FALSE)) {
+      assertFalse(endpointUsageProvider.isImplicitUsage(psiClass));
+    } else {
+      fail("Class name must contain an assertion key. (Either '" +
+          ASSERT_IMPLICIT_USAGE_TRUE+"' or '"+ASSERT_IMPLICIT_USAGE_FALSE+"')");
+    }
+
+    for (PsiMethod m : implicitUsageMethods) {
```
etanshaul: I don't think checkstyle will let you have single letter var names

aslo: :+1:, will fix

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/validation/EndpointImplicitUsageProviderTest.java
```diff
@@ -15,29 +15,69 @@
...
+import com.intellij.psi.PsiMethod;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+
 /**
  * Tests for {@link EndpointImplicitUsageProvider}.
- * See http://www.jetbrains.org/intellij/sdk/docs/basics/testing_plugins/testing_highlighting.html
  */
 public class EndpointImplicitUsageProviderTest extends EndpointTestBase {
```
aslo: I updated this test such that it will now pass on both versions of intellij. Note that it no longer tests the highlighting of java classes - instead it directly tests methods on the `EndpointImplicitUsageProvider` by constructing `PsiElements` from the same test data as before.

aslo:
@patflynn Mind taking one more look at this? I believe all comments have been addressed. Note that now we're running a build matrix on travis with both IJ versions.

patflynn:
So the plan is that we will release two different binaries for our next release?  How does this change effect the automated release process?

aslo:
Good point - the intent is to only release one binary. As currently configured, travis will run the `deploy` phase twice (once for each build in the build matrix). I'm going to update the travis config to only deploy when building against the `15.0.6` intellij sdk.

patflynn:
lgtm

PR #797, opened 2016-07-27
Initialize the login widget from the credential datastore at startup

### LINKED ISSUE #794: Closing and restarting IntelliJ causes Google login state to reset
Steps to reproduce:

1) Install the latest version of gcloud-intellij (0.9.7-beta)
2) Login using the Google login widget
2) close IntelliJ
3) restart Intellij

result:
The login widget shows that I'm logged out

expected result:
The login widget shows my logged in state from before closing IntelliJ.

This used to work, but I'm not sure when it stopped working. It should be tested in Android Studio as well.

patflynn:
Fixes #794.

Really the service should initialize itself.  I haven't implemented it
that way because of some pretty gnarly circular dependencies in the
initialization.

Basically the service calls itself through delegates during
initialization. In addition multiple instances of the datastore are
actually instantiated during intitialization.

### DIFF
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/AccountPluginInitializationComponent.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/AccountPluginInitializationComponent.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.tools.intellij;
 
+import com.google.cloud.tools.intellij.login.Services;
 import com.google.cloud.tools.intellij.login.util.TrackerMessageBundle;
 import com.google.cloud.tools.intellij.stats.UsageTrackerManager;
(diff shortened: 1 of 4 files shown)


patflynn:
woops. forgot to run tests.

etanshaul:
You mind explaining why this was happening prior to the refactoring of login as a service? I'm not seeing the connection.

patflynn:
Prior to the refactor to a service GoogleLogin was managed as a singleton
with a getInstance method that performed construction and initialization as
separate steps.  If you try to initialize the state of the component during
construction it'll fail since there are code paths that depend on the
component existing already at initialization.

etanshaul:
Oh I see. So the initialization was lost during the refactoring. Thanks.

LGTM

patflynn:
yep. I should have mentioned that!

PR #798, opened 2016-07-28
Merge JetBrains plugin integration into master

etanshaul:

I want to go ahead and merge the integrate-jetbrains branch into master. @patflynn correctly pointed out that if we need to patch production we can always branch off of a release tag.

Note this branch has already been reviewed, but feel free to look again.

For any open pull requests you can either close it and reopen a new one with master as the target referencing the closed pr ([I really wish](https://github.com/isaacs/github/issues/18) GitHub had the ability to change the target branch of a PR), or just merge into integrate-jetbrains then sync up master afterwards.

Let me know if you see any issues with this.


etanshaul:
Also, prior to merging this I'm going to have to temporarily enable merge commits in our github settings so that we get all the history without squashing.

patflynn:
makes sense.

aslo:
Sounds good to me!

joaoandremartins:
SGTM

google-cloud-tools-plugin/build.gradle
```diff
@@ -31,12 +31,25 @@ intellij {
 
 project.afterEvaluate {
     prepareSandbox.dependsOn ':google-account-plugin:prepareSandbox'
+
```
patflynn: @loosebazooka can you look over these Gradle changes?

google-cloud-tools-plugin/build.gradle
```diff
@@ -31,12 +31,25 @@ intellij {
...
+    //todo: currently we need this to include 'ultimate' module to the plugin zip; it would be better to support such inclusion in 'intellij' plugin
+    def task = project.tasks.findByName("prepareSandbox")
```
loosebazooka: preferably task is better named.

google-cloud-tools-plugin/build.gradle
```diff
@@ -31,12 +31,25 @@ intellij {
...
+    task.dependsOn project('ultimate').jar
+    task.doLast {
+        copy {
+            from files(project('ultimate').jar)
```
loosebazooka: does this work? I usually end up doing project.jar.getArchivePath

etanshaul: yea it works. but I can change it if you think another way is cleaner.

google-cloud-tools-plugin/jps-plugin/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+sourceSets.main.java.srcDirs = ['src']
+sourceSets.main.resources.srcDirs = ['resources']
+sourceSets.test.java.srcDirs = ['testSrc']
+sourceSets.test.resources.srcDirs = ['testData']
+
+jar.archiveName = "google-app-engine-jps-plugin.jar"
+
+// Overriding paths in root build file since this module is nested one level deeper
+pmd {
+    ruleSetFiles = files('../../custom-pmd-ruleset.xml');
```
loosebazooka: I think (maybe?) you can reference project root directory instead of relative paths, but I'm not actually sure if that's the correct approach.

google-cloud-tools-plugin/runtime/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+
+// Overriding paths in root build file since this module is nested one level deeper
+pmd {
+    ruleSetFiles = files('../../custom-pmd-ruleset.xml');
+}
+findbugs {
+    excludeFilter = file('../../findbugs-excludefilter.xml')
+}
+
+jar.archiveName = "appEngine-runtime.jar"
```
loosebazooka: seems like kind of weird naming here camelCase and hyphens?

etanshaul: totally. that was inherited from the merged plugin. I can update this.

google-cloud-tools-plugin/ultimate/build.gradle
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+sourceSets.main.resources.srcDirs = ['resources']
+sourceSets.main.java.srcDirs = ['src']
+sourceSets.test.java.srcDirs = ['testSrc']
+
+intellij {
+    type 'IU'
+
+    //actually only 'JavaEE' and 'GwtStudio' plugins are required, the others are transitive dependencies
```
loosebazooka: I'm not sure what this comment means, that we're putting these plugin dependencies in because we **want** explicitly define them or because we **have** to

google-cloud-tools-plugin/ultimate/build.gradle
```diff
@@ -0,0 +1,53 @@
...
+    plugins 'JavaEE', 'GwtStudio', 'Groovy', 'junit', 'git4idea',
+            'maven', 'gradle', 'properties', 'java-i18n',
+            'DatabaseTools', 'PersistenceSupport', 'JavaScriptLanguage', 'CSS', 'jsp'
+}
+
+// Overriding paths in root build file since this module is nested one level deeper
+pmd {
```
loosebazooka: similar to above question about relative paths.

google-cloud-tools-plugin/jps-plugin/build.gradle
loosebazooka: But it would help if the paths of this plugin ever changed without having to worry about this arbitrary pathing to pmd and findbugs configuration.

google-cloud-tools-plugin/build.gradle
loosebazooka: I don't have a preference, just do what works for you, I was just confused. Didn't know files could resolve a task into an archive like that.

google-cloud-tools-plugin/ultimate/build.gradle
etanshaul: I think (its another thing we inherited) it means that only the JavaEE and GwtStudio (although I'm not sure about this one) are used directly by the plugin. The other ones are transitive dependencies of two that need to be explicitly defined.

etanshaul: And yes, it doesn't work without defining all of them. I'll try and clarify the comment.

loosebazooka: Oh so the transitive ones are required so that Jave EE and GwtStudio work, that makes sense. Is there some way we can remove gwt support @patflynn ?

google-cloud-tools-plugin/jps-plugin/build.gradle
etanshaul: Yea I had meant to ask you about this earlier. This was the best workaround I could come up with, but if I can reference the root directory then I might be able to remove these overrides. I'll try.

loosebazooka: I think you can access rootProject.projectDir ?

etanshaul: thats cool. I'll test it out.

etanshaul:
@loosebazooka thanks for the gradle review. I've pushed the updates.

etanshaul:
@loosebazooka was there anything else here or does it look ok to you? I'll go ahead and merge if it looks good.

google-cloud-tools-plugin/build.gradle
```diff
@@ -48,9 +61,19 @@ dependencies {
...
+    from sourceSets.test.output
+}
+
 configurations {
     // this is already obtained from google-api-java-client-min-repackaged
     compile.exclude group: 'com.google.api-client', module: 'google-api-client'
+    tests
+}
+
+artifacts {
```
loosebazooka: just out of curiosity, are we exporting test jars somewhere?

loosebazooka:
The gradle stuff looks fine to me, I haven't looked at the other non-gradle files though.

google-cloud-tools-plugin/build.gradle
etanshaul: I'm not sure I understand? You can see that task "testJar" right above the code you referenced

etanshaul: Also the 'utlimate' module also uses this tests configuration (see its build.gradle:40)

PR #801, opened 2016-08-01
Show exception message on flexible staging

### LINKED ISSUE #800: Exception on flexible staging
The following occurs do to an invalid path to the deployment artifact during flexible staging. This should not occur during normal execution:

- If the deployable is an artifact deployment source, then the internal platform should correctly pick up the deployable path
- If the deployable is a maven source, then the path is picked up by the maven plugin

At a minimum we should should present the exception message "/a/b/c.jar (The system cannot find ...)" to the user in the deployment console in IJ.

Description:
java.io.FileNotFoundException
Device Model:
unknown

Device Release:
2016.2

App Version:
2016.2

Stack trace:
java.lang.RuntimeException: java.io.FileNotFoundException: 
(The system cannot find the path specified)
	at com.google.cloud.tools.intellij.appengine.cloud.AppEngineFlexibleStage.stage(AppEngineFlexibleStage.java:81)
	at com.google.cloud.tools.intellij.appengine.cloud.AppEngineFlexibleDeployTask.execute(AppEngineFlexibleDeployTask.java:67)
	at com.google.cloud.tools.intellij.appengine.cloud.AppEngineRunner.run(AppEngineRunner.java:39)
	at com.intellij.openapi.application.TransactionGuardImpl$6.run(TransactionGuardImpl.java:291)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.a(LaterInvocator.java:326)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.run(LaterInvocator.java:310)

etanshaul:
fixes #800 

Since we cannot control the structure of the project and what may be causing this exception, printing the exception message to the console will at least help the user debug the issue. This will still throw and IDE exception, but will also show the following on the deployment console:

image


patflynn:
Could we validate the existence of the files specified in the run config at
launch?

akerekes:
If it throws and IDE error, then this will be reportable by the users as
bug, right? Would it work without making it an IDE error?

etanshaul:
@patflynn I updated the code to also check for the existence of the deployment source. If the source does not exist, then the deployment will not be started, the IDE will not throw an exception, and the user will be shown the following:

image

etanshaul:
I separated out the error cases

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -113,12 +113,18 @@ public CancellableRunnable createDeployRunner(
...
+    if (!(source instanceof AppEngineDeployable)) {
+      callback.errorOccurred(GctBundle.message("appengine.deployment.invalid.source.error"));
       throw new RuntimeException("Invalid deployment source selected for deployment");
     }
 
+    if (source.getFile() == null
+        || !source.getFile().exists()) {
+      callback.errorOccurred(GctBundle.message("appengine.deployment.source.not.found.error",
+          source.getFilePath()));
+      throw new RuntimeException("Deployment source selected for deployment not found");
```
patflynn: do we really want to throw an exception in this case?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -113,12 +113,18 @@ public CancellableRunnable createDeployRunner(
       AppEngineDeploymentConfiguration deploymentConfiguration,
       DeploymentOperationCallback callback) {
 
-    if (source.getFile() == null
-        || !(source instanceof AppEngineDeployable)) {
-      callback.errorOccurred(GctBundle.message("appengine.deployment.source.not.found.error"));
+    if (!(source instanceof AppEngineDeployable)) {
+      callback.errorOccurred(GctBundle.message("appengine.deployment.invalid.source.error"));
       throw new RuntimeException("Invalid deployment source selected for deployment");
```
patflynn: I guess this shouldn't happen so this case really involves an error on our part?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
etanshaul: I think it makes sense to not throw an exception here and just return. But for the other case (not an instance of AppEngineDeployable) to keep the exception.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
etanshaul: Agreed.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
etanshaul: On second thought, I think its ok to keep the exception here: if the DeploymentSource was created in the first place then there should be a valid path to a deployable file (or exploded war directory). It is an exceptional case that this will not happen.

What do you think?

patflynn: I'm not sure.  Is it not possible that the deployment source was create when there was a valid file, but that file is gone at the time of deployment?

etanshaul:
@patflynn I updated the logic to not throw an exception when the deployable file cannot be located. I am still throwing one with the deployable is not of the expected type.

etanshaul:
I will write some tests to cover this.

etanshaul:
done

patflynn:
:+1:

patflynn:
woops. sorry. hit close by accident.

PR #802, opened 2016-08-01
Refactor cloud SDK configuration to google plugin setting

### LINKED ISSUE #787: Move cloud SDK setting to a "global" google plugin setting
The motivation for this is that we now will have components that rely on the SDK that are not tied to a Cloud configuration. Most notably the local server run for AE standard apps.

related to: #781

etanshaul:
Fixes #787 

@patflynn 
@joaoandremartins 
@aslo 

More details / screenshots coming.


etanshaul:
This PR moves the Cloud SDK path setting to an IntelliJ applicationService `CloudSdkService`. There is also a new UI view into this setting via the Google settings within the IJ settings:

image
Also note how these google settings were moved to be sub menu items of a "Google" parent.

The user can still configure the Cloud SDK when setting up a cloud, however the logic was updated to store the setting in the new location. If the user had previously configured the sdk via the new google setting, then it will be pre-populated with this value when creating a new cloud.

The UI for Cloud SDK panel was refactored into its own class so that it can be reused by the various UI components that use it: Google Settings, Cloud, New AE standard wizard (still in the works).

aslo:
What about the reverse case - if a user had previously set up a cloud, will this setting be migrated to this new location? Or will they have to reconfigure at the root plugin level?

etanshaul:
They'll have to reconfigure it. But not necessarily at the root plugin level since the Cloud configuration will still have the cloud sdk selector (it just persists it on the root plugin setting and not on the cloud configuration now). 

I'll think about how/if we can make this more user friendly for backwards compatibility.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleSettingsConfigurable.java
```diff
@@ -84,73 +65,34 @@ public String getHelpTopic() {
...
-
-    JPanel usageTrackerGroup = new JPanel(new BorderLayout());
-    usageTrackerGroup
-        .setBorder(new TitledBorder(new EtchedBorder(EtchedBorder.LOWERED), "Usage Tracker"));
-    usageTrackerGroup.add(enableUsageTrackerBox, BorderLayout.NORTH);
-    usageTrackerGroup.add(privacyPolicyText, BorderLayout.SOUTH);
-    return usageTrackerGroup;
   }
-}
+}
```
joaoandremartins: Needs extra line.

joaoandremartins:
LGTM

patflynn:
reviewing now. sorry for the delay

google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTrackerConfigurable.java
```diff
@@ -0,0 +1,96 @@
...
+
+  @NotNull
+  @Override
+  public String getId() {
+    return "google.settings.tracker";
+  }
+
+  @Nullable
+  @Override
+  public Runnable enableSearch(String option) {
```
patflynn: just curious what this does?

google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTrackerConfigurable.java
```diff
@@ -0,0 +1,96 @@
...
+    return null;
+  }
+
+  @Nls
+  @Override
+  public String getDisplayName() {
+    return "Usage Tracking";
```
patflynn: localization?

google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTrackerConfigurableProvider.java
```diff
@@ -41,11 +41,12 @@ public Configurable createConfigurable() {
...
+    // Create a sub-menu item for the cloud SDK and hide the usage tracker if not avaible
     if (PlatformUtils.isIntelliJ() && UsageTrackerManager.getInstance()
         .isUsageTrackingAvailable()) {
       return true;
     } else {
       return false;
     }
   }
-}
+}
```
patflynn: missing new lines.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTrackerPanel.java
```diff
@@ -0,0 +1,104 @@
...
+
+    enableUsageTrackerBox.setSelected(usageTrackerManager.hasUserOptedIn());
+    privacyPolicyText.setText(TrackerMessageBundle.message(
+        "settings.privacy.policy.comment",
+        PRIVACY_POLICY_URL));
+
+
+    // Disable checkbox if usage tracker property has not been configured
+    if (!usageTrackerManager.isUsageTrackingAvailable()) {
+      enableUsageTrackerBox.setEnabled(false);
```
patflynn: I know this is only for the case where people are building the plugin themselves, but I'm wondering if we should have clearer messaging if usage tracking is disabled.  Either by removing the setting completely, or by having clear text that indicates why it can't be turned off/on.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
```diff
@@ -175,7 +112,13 @@ public boolean canCheckConnection() {
     return false;
   }
 
-  public String getCloudSdkDirectory() {
-    return cloudSdkDirectoryField.getText();
+  @VisibleForTesting
+  public CloudSdkPanel getCloudSdkPanel() {
```
patflynn: package protected?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
```diff
@@ -91,7 +92,7 @@ public AppEngineServerConfiguration createDefaultConfiguration() {
   @Override
   public RemoteServerConfigurable createServerConfigurable(
       @NotNull AppEngineServerConfiguration configuration) {
-    return new AppEngineCloudConfigurable(configuration, null);
```
patflynn: I'm just wondering if the server configuration model is something we want to bypass or whether we should proxy the cloud sdk service through the configuration model.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineServerConfiguration.java
```diff
@@ -17,24 +17,12 @@
...
 
 /**
  * Model for the IntelliJ application scoped 'Cloud' configurations.  This is a base configuration
  * used by App Engine deployment runtime configurations. It's primarily the bits that can be re-used
- * across deployments, such as auth and project.
+ * across deployments.
  */
 public class AppEngineServerConfiguration extends
     ServerConfigurationBase<AppEngineServerConfiguration> {
-
```
patflynn: this class doesn't appear to do much anymotre.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkConfigurable.java
```diff
@@ -0,0 +1,93 @@
...
+  @Nullable
+  @Override
+  public Runnable enableSearch(String option) {
+    return null;
+  }
+
+  @Nls
+  @Override
+  public String getDisplayName() {
+    return "Cloud SDK";
```
patflynn: localization.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkConfigurable.java
```diff
@@ -0,0 +1,93 @@
...
+  @NotNull
+  @Override
+  public String getId() {
+    return "google.settings.sdk";
+  }
+
+  @Nullable
+  @Override
+  public Runnable enableSearch(String option) {
+    return null;
```
patflynn: how does returning null effect searchability?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkConfigurable.java
```diff
@@ -0,0 +1,93 @@
...
+
+  @Nls
+  @Override
+  public String getDisplayName() {
+    return "Cloud SDK";
+  }
+
+  @Nullable
+  @Override
+  public String getHelpTopic() {
```
patflynn: note to self: investigate online help.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkConfigurable.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Override
+  public void reset() {
+    cloudSdkPanel.reset();
+  }
+
+  @Override
+  public void disposeUIResources() {
+
```
patflynn: is this important?  In the other component you set the panel to null?

google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTrackerPanel.java
etanshaul: If usage tracking is not available, then that sub-menu item (under 'Google') will not be available at all. See UsageTracekrConfigurableProvider#canCreateConfigurable()

patflynn: ok cool.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
etanshaul: Interesting question. If we proxied it, it would take care of @aslo point about backwards compatibility (with users that previously had a server configuration). The only downside I see here is duplication.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineServerConfiguration.java
etanshaul: Yes, if I don't go down the proxy route, then I'll remove it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -0,0 +1,127 @@
...
+public class CloudSdkPanel {
+
+  private TextFieldWithBrowseButton cloudSdkDirectoryField;
+  private JLabel warningMessage;
+  private JPanel cloudSdkPanel;
+
+  public CloudSdkPanel(@NotNull CloudSdkService cloudSdkService) {
+    warningMessage.setVisible(false);
+
+    final String cloudSdkDirectoryPath
```
patflynn: shouldn't we only be detecting the path if sdkService.cloudSdkHomePath has not been set?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkConfigurable.java
etanshaul: yes, thanks. I missed this one

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -0,0 +1,127 @@
...
+  public JLabel getWarningMessage() {
+    return warningMessage;
+  }
+
+  @VisibleForTesting
+  public TextFieldWithBrowseButton getCloudSdkDirectoryField() {
+    return cloudSdkDirectoryField;
+  }
+
+  private DocumentAdapter getSdkDirectoryFieldListener() {
```
patflynn: why return this as a method rather than a static constant?

google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTrackerConfigurable.java
etanshaul: According to the docs:

> @return an action to perform when this configurable is opened when a search filter query is entered by the user in setting dialog. This action, for example, can select something in a tree or a list embedded in this setting page that matches the query. 

We don't need this. I'll look into possibly using a non-searchable implementation of `Configurable` so we don't have to return null

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
etanshaul: no reason. I can refactor this.

etanshaul:
There is an issue with this implementation that I noticed. Moving the SDK configuration to an IntelliJ service does not take care of persistence. This service needs to backed by some serializable configuration.  

Proxying the server configuration will not be sufficient since there are some actions that need the SDK independently of a server config (e.g. local run).

I'm looking at this now.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanelTest.java
```diff
@@ -0,0 +1,113 @@
...
+  private JLabel warningMessage;
+  private TextFieldWithBrowseButton cloudSdkDirectoryField;
+
+  private static final String CLOUD_SDK_EXECUTABLE_PATH = new File("/a/b/c/gcloud-sdk/bin/gcloud").getAbsolutePath();
+  private static final String CLOUD_SDK_DIR_PATH = new File("/a/b/c/gcloud-sdk").getAbsolutePath();
+
+  private static final String MISSING_SDK_DIR_WARNING = "Please select a Cloud SDK home directory.";
+
+  @Override
+  public void setUp() throws Exception {
```
patflynn: I'm guessing BasePluginTestCase does not play nicely with UI components?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
etanshaul: this is for convenience: if its not set in the SdkService, but is on the users path, then we can populate it here for them.

patflynn: I Guess what I'm saying is shouldn't we be detecting the path only if service.getCloudSdkHomePath == null? And shouldn't we be initializing the cloudSdkDirectoryField to service.getCloudSdkHomePath if it isn't null?

etanshaul: Oh. Yea for the cases where the SdkService is configured, it is redundantly checking the path. I'll refactor this.

Regarding the second point - the initialization happens via the reset() method of that same class which gets called when the panel is loaded after construction. I think I can move the path logic to here too.

patflynn: aha. I did not realize that reset was called on initialization.

etanshaul:
@patflynn I've addressed the remaining issues in the last 3 commits. Will merge as discussed after the build completes.

patflynn:
:ship:

etanshaul:
closed via #808

PR #804, opened 2016-08-02
Remove unused exception

chanseokoh:


patflynn:
@loosebazooka 

I'm wondering how this used to be used.  Otherwise LGTM

loosebazooka:
I have no idea what this is.

loosebazooka:
It possible it was added for compatibility when we moved out of android studio.

PR #810, opened 2016-08-04
Only build pushes to master

etanshaul:
Just testing the behavior first

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: java
+branches:
+  only:
+  - master
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


patflynn:
would this stop CI on our individual feature branches? Do we want that?

etanshaul:
The hope is that all PRs (for any branch) will still have CI. But only _pushes_ to master will be built. This avoid the duplicate building for PRs (once for the PR and again for the push).

I'm assuming here that the feature branches you are referring to will be part of a PR intended to eventually be merged to master.

Issue #782, opened 2016-07-13
Update local run/debug to use the common library


etanshaul:

joaoandremartins:
@etanshaul as discussed, going to take over this. Feel free to assign other existing similar issues on the run/debug with commlib to me!

joaoandremartins:
Currently, running an app locally is very tightly coupled with running a Java process (because in any other way, e.g., an API call, the debugger can't work).

So, for this work, my suggestion is to first change all SDK location mechanisms to be performed using the library and, on a second stage, change the Java process run into an API call (if possible).

etanshaul:
The CloudSdk builder has a callback (startListener) which gives the running process. I wonder if this can be used? Check out AppEngineRunner#run() to see how I'm using it for deploy.

etanshaul:
Also regarding the changing of the SDK locations - I'm also working on #787 and #781 which are tightly coupled to each other, and to this. 

So we may want to coordinate a plan - my approach currently is to keep all the old SDK (appcfg) code as is, deprecate it, and move the new gcloud SDK code alongside it. After that is done, and after all consumers use the gcloud sdk, we can remove the deprecated appcfg code.

joaoandremartins:
The current implementation with `AppEngineServerConfigurationType` relies on `JavaCommandLineStartupPolicy` to run code on start and shutdown, and to run a Java process with arguments given by us.

The whole IDEA framework seems to be geared towards Run/Debug Configurations only running processes, whereas we only want to make a library call.

http://www.jetbrains.org/intellij/sdk/docs/basics/run_configurations/run_configuration_execution.html

Because the com.intellij.javaee framework appears to be limited in functionality and severely lacks documentation, I'm going to give it a shot at creating a new Run/Debug Configuration for devappserver, with a ProgramRunner that only calls the library.

patflynn:
@joaoandremartins should this be linked to your last PR?

PR #811, opened 2016-08-04
Use common library to run standard apps locally

joaoandremartins:
@etanshaul @aslo PTAL

This is just a proof of concept, it still needs considerable work.

Just interested in your high level opinion on the direction this is headed.


google-cloud-tools-plugin/ultimate/src/com/intellij/appengine/server/run/CloudSdkStartupPolicy.java
```diff
@@ -0,0 +1,134 @@
...
+              throws ExecutionException {
+            AppEngineServerModel runConfiguration =
+                (AppEngineServerModel) commonModel.getServerModel();
+
+            // This is the place we have access to the debug jvm flag provided by IJ in the
+            // Startup/Shutdown tab. We need to add it here.
+            String jvmDebugFlag = map.get("").trim();
+            runConfiguration.addJvmFlag(jvmDebugFlag);
+
+            AppEngineRunTask runTask = new AppEngineRunTask(runConfiguration, null, null, null);
```
etanshaul: will this eventually take in the listeners?

joaoandremartins: Nope. I actually just added a new AppEngineRunTask constructor that only takes in a runConfiguration and am using it here.

etanshaul:
@joaoandremartins yes this approach looks straightforward and makes sense to me.

I still wonder to what degree the server configuration UI can be customized, beyond what is injected via `AppEngineRunConfigurationEditor`. For example, I remember noticing differences in the UI that was shown for tomcat and the UI that was shown for the old App Engine local run. Maybe these differences were all in the custom, non platform, part of the UI..

joaoandremartins:
@etanshaul @patflynn @aslo PTAL

This PR:
- Refactors AppEngineRunner into AppEngineExecutor,
- Changes Files by Paths in several locations,
- Removes AppEngineSdk on the ultimate surface (and a bit of the main surface, which is fully dealt with in #822),
- Moves from App Engine SDK to Cloud SDK (through common lib) to run local projects, reusing the Application Servers framework.

Some work left:
- Reuse CloudSdkPanel on the Configure button when https://youtrack.jetbrains.com/oauth?state=%2Fissue%2FIDEA-110316 is fixed,
- Add "Build artifact" task on newly created configurations (by default, a project is created with a Run/Debug configuration which contains the Build task)

Also, the config UI looks like this:

screenshot from 2016-08-18 11 04 57
screenshot from 2016-08-18 11 05 10
screenshot from 2016-08-18 11 05 26
screenshot from 2016-08-18 11 05 37
screenshot from 2016-08-18 11 05 48

I'm inclined to remove the Python and Custom VM Runtime panels.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
```diff
@@ -118,8 +118,8 @@ public void connect(@NotNull ConnectionCallback<AppEngineDeploymentConfiguration
 
       if (!Services.getLoginService().isLoggedIn()) {
         callback.errorOccurred(GctBundle.message("appengine.deployment.error.not.logged.in"));
-      } else if (CloudSdkUtil.isCloudSdkExecutable(
-          CloudSdkUtil.toExecutablePath(CloudSdkService.getInstance().getCloudSdkHomePath()))) {
+      } else if (CloudSdkUtil.isCloudSdkExecutable(CloudSdkUtil.toExecutablePath(
+          CloudSdkService.getInstance().getCloudSdkHomePath().toString()))) {
```
etanshaul: what is getCloudSdkHomePath is null?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineRunTask.java
```diff
@@ -0,0 +1,73 @@
...
+import com.google.cloud.tools.appengine.cloudsdk.process.ProcessOutputLineListener;
+import com.google.cloud.tools.appengine.cloudsdk.process.ProcessStartListener;
+import com.google.cloud.tools.intellij.appengine.sdk.CloudSdkService;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Represents an App Engine Standard run task. (i.e., devappserver)
+ */
+public class AppEngineRunTask implements AppEngineTask {
```
etanshaul: maybe `AppEngineLocalRunTask`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -111,15 +111,15 @@ JLabel getWarningMessage() {
 
   public boolean isModified() {
     return !Comparing.strEqual(getCloudSdkDirectory(),
-        CloudSdkService.getInstance().getCloudSdkHomePath());
+        CloudSdkService.getInstance().getCloudSdkHomePath().toString());
```
etanshaul: null check for getCloudSdkHomePath?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -111,15 +111,15 @@ JLabel getWarningMessage() {
...
   }
 
   public void apply() throws ConfigurationException {
     CloudSdkService.getInstance().setCloudSdkHomePath(getCloudSdkDirectory());
   }
 
   public void reset() {
-    setCloudSdkDirectoryText(CloudSdkService.getInstance().getCloudSdkHomePath());
+    setCloudSdkDirectoryText(CloudSdkService.getInstance().getCloudSdkHomePath().toString());
```
etanshaul: same comment about null

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
```diff
@@ -32,8 +35,10 @@ public DefaultCloudSdkService() {
   }
 
   @Override
-  public String getCloudSdkHomePath() {
-    return propertiesComponent.getValue(CLOUD_SDK_PROPERTY_KEY);
+  public Path getCloudSdkHomePath() {
```
etanshaul: @Nullable

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineRunConfigurationEditor.java
```diff
@@ -84,25 +113,78 @@ private void onArtifactChanged() {
...
+    adminPort.setText(serverModel.getAdminPort() != null
+        ? String.valueOf(serverModel.getAdminPort()) : "");
+    authDomain.setText(serverModel.getAuthDomain());
+    storagePath.setText(serverModel.getStoragePath());
+    logLevel.setSelectedItem(serverModel.getLogLevel());
+    maxModuleInstances.setText(serverModel.getMaxModuleInstances() != null
+        ? String.valueOf(serverModel.getMaxModuleInstances()) : "");
+    useMtimeFileWatcher.setSelected(serverModel.getUseMtimeFileWatcher());
+    threadsafeOverride.setText(serverModel.getThreadsafeOverride());
+    jvmFlags.setDialogCaption("Server Parameters");
```
etanshaul: localize ui messages in this class

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
```diff
@@ -0,0 +1,142 @@
...
+          public String getDisplayString() {
+            return "App Engine Plugins Core library";
+          }
+
+          @Override
+          public OSProcessHandler createProcessHandler(String string, Map<String, String> map)
+              throws ExecutionException {
+            AppEngineServerModel runConfiguration;
+            try {
+              // Getting the clone so the debug flags aren't added to the persisted settings.
```
etanshaul: Can you explain a bit more? Why don't you want these settings persisted?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
```diff
@@ -0,0 +1,142 @@
...
+              runConfiguration = (AppEngineServerModel) commonModel.getServerModel().clone();
+            } catch (CloneNotSupportedException ee) {
+              throw new ExecutionException(ee);
+            }
+
+            // This is the place we have access to the debug jvm flag provided by IJ in the
+            // Startup/Shutdown tab. We need to add it here.
+            String jvmDebugFlag = map.get("");
```
etanshaul: the key is really an empty string?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerHelper.java
```diff
@@ -32,14 +33,11 @@
   public ApplicationServerInfo getApplicationServerInfo(
       ApplicationServerPersistentData persistentData)
       throws CantFindApplicationServerJarsException {
-    final AppEngineSdk sdk = ((AppEngineServerData) persistentData).getSdk();
-    String version = sdk.getVersion();
-    return new ApplicationServerInfo(sdk.getLibraries(),
-        "AppEngine Dev" + (version != null ? " " + version : ""));
+    return new ApplicationServerInfo(new File[]{}, "Cloud SDK");
```
etanshaul: Assuming this is the place where the local run config name (visible in the UI) is set, I'm more in favor of calling it something like App Engine Local Server or something like that. Users don't really need to care too much about the Cloud SDK (aside from initially giving the path). Plus Cloud SDK isn't clear to me that its about running a local dev server.

etanshaul:
@joaoandremartins regarding this error you were seeing when doing a local run with `<vm>true</true>`:

image

I spoke to @meltsufin about this and it happens with Java 8. They implemented a fix in the maven plugin to add a [JVM flag ](https://www.google.com/url?q=https%3A%2F%2Fgithub.com%2FGoogleCloudPlatform%2Fapp-maven-plugin%2Fblob%2Fmaster%2Fsrc%2Fmain%2Fjava%2Fcom%2Fgoogle%2Fcloud%2Ftools%2Fmaven%2FRunMojo.java%23L213&sa=D&sntz=1&usg=AFQjCNEd68bVI8NVpocDuwGyQT9kLXbuug)to get around it.

We can do something similar in the meantime, but I will file something against the common-lib to move this logic there.

etanshaul:
Is piping the output of the the dev appserver to the IJ console window on the roadmap here? Currently it looks like it just shows "running" and "connected".

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineRunTask.java
```diff
@@ -0,0 +1,73 @@
...
+  private RunConfiguration runConfig;
+  private ProcessOutputLineListener stdOutListener;
+  private ProcessOutputLineListener stdErrListener;
+  private ProcessExitListener exitListener;
+
+  public AppEngineRunTask(@NotNull RunConfiguration runConfig) {
+    this.runConfig = runConfig;
+  }
+
+  public AppEngineRunTask(@NotNull RunConfiguration runConfig,
```
etanshaul: This constructor is never used.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineRunTask.java
```diff
@@ -0,0 +1,73 @@
...
+      ProcessOutputLineListener stdErrListener,
+      ProcessExitListener exitListener) {
+    this(runConfig);
+    this.stdOutListener = stdOutListener;
+    this.stdErrListener = stdErrListener;
+    this.exitListener = exitListener;
+  }
+
+  @Override
+  public void execute(ProcessStartListener startListener) {
```
etanshaul: Not for this PR, but something to think about - do we want analytics pings for local dev server runs? @chanseokoh

google-cloud-tools-plugin/ultimate/build.gradle
```diff
@@ -36,6 +36,9 @@ dependencies {
     compile files(project(':google-cloud-tools-plugin').sourceSets.main.output)
     compile files(project(':google-cloud-tools-plugin:jps-plugin').sourceSets.main.output)
     compile files("${System.properties['java.home']}/../lib/tools.jar")
+    compile ('com.google.cloud.tools:appengine-plugins-core:0.1.4') {
```
etanshaul: Since this is also defined in the google-cloud-tools-plugin module maybe we should extract out the version to a variable so we don't have to update it in two places when upgrading the lib.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineRunTask.java
patflynn: yes.

chanseokoh: Never hurts sending whatever is interesting (less PII), and actually it looks like a very good idea.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
joaoandremartins: Added null check.

patflynn: If it's possible for the cloudSdkHomePath to be invalid why not expose that as a method on the service? e.g. cloudSdkService.hasValidSdk() or something like that.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineRunTask.java
patflynn: +1

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
joaoandremartins: The settings we get here are already persisted. We don't want to persist them including the jvm debug flag that we're adding.
If we used the same reference, each Debug invocation would add the same debug flag over and over again. (e.g., -agentlib:jdwp=transport=dt_socket,address=127.0.0.1:37180,suspend=y,server=n).

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
joaoandremartins: Indeed. And nothing can change it AFAIU.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
joaoandremartins: The issue isn't that the path isn't valid, but rather it isn't set. That's the only way propertiesComponent.getValue(CLOUD_SDK_PROPERTY_KEY) can be set to null. Paths.get() actually doesn't perform any validation.
But we're still not doing any SDK path validation here, and the IDE is exploding when we set an invalid path. This would be solved by reusing CloudSdkPanel, which does the error checking.

@etanshaul should we move SDK checking to the CloudSdkService level, so that it doesn't let you create an SDK with an incorrect path?

etanshaul: @patflynn previously agreed that we don't want to prevent the user from creating an invalid SDK. There is an exception: the Cloud UI forces the user to have a valid path. So bottom line is that we can't always assume its set or valid.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineExecutor.java
```diff
@@ -23,14 +23,14 @@
 import com.intellij.openapi.vcs.impl.CancellableRunnable;
 
 /**
- * Runner of {@link AppEngineTask}'s.
+ * Executor of {@link AppEngineTask}'s.
  */
-public class AppEngineRunner implements CancellableRunnable {
+public class AppEngineExecutor implements CancellableRunnable {
```
patflynn: @etanshaul 

I'm noticing that this is a 'generic' executor but calls APP_ENGINE_DEPLOY_CANCEL ping on cancel.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
etanshaul: Also, CloudSdkPanel does not force the user to have a non-empty or valid SDK. Its up to users of the panel (other panels that include it) decide if they want to override this behavior.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
```diff
@@ -27,7 +29,7 @@ public static CloudSdkService getInstance() {
     return ServiceManager.getService(CloudSdkService.class);
   }
 
-  public abstract String getCloudSdkHomePath();
+  public abstract Path getCloudSdkHomePath();
 
   public abstract void setCloudSdkHomePath(String path);
```
patflynn: if we return a Path, shouldn't we accept a Path?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineExecutor.java
etanshaul: Good catch. Thats a bug.

etanshaul: https://github.com/GoogleCloudPlatform/gcloud-intellij/issues/831

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
joaoandremartins: Ah, yes. The big red warning led me to think I couldn't click yes, but I was just able to. In that case, I think I'll need to replicate the SDK check on my editor. Or maybe factor SDK checking. i.e., what Patrick suggested earlier.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
joaoandremartins: I thought about this, but we're really only using the String part of the Path to set the value in propertiesComponent.
In practice, we'd be converting to a Path to call the method, and then the method would do the opposite conversion to String.

That being said, which would you prefer? String or Path?

google-cloud-tools-plugin/ultimate/build.gradle
joaoandremartins: Done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudType.java
joaoandremartins: The functionality I'm looking for is actually already in CloudSdkUtil.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -110,16 +111,22 @@ JLabel getWarningMessage() {
...
 
   public boolean isModified() {
-    return !Comparing.strEqual(getCloudSdkDirectory(),
-        CloudSdkService.getInstance().getCloudSdkHomePath());
+    if (getCloudSdkDirectory() != null) {
+      return Paths.get(getCloudSdkDirectory())
+          .compareTo(CloudSdkService.getInstance().getCloudSdkHomePath()) != 0;
+    }
+
+    return CloudSdkService.getInstance().getCloudSdkHomePath() != null;
```
joaoandremartins: @etanshaul can you validate this is the desired behaviour?

etanshaul: Moving my comment to the right place:

What about the case where one is null, and the other is empty? Just trying to make sure this does what we expect: e.g. if the field is empty string, but the SDK from the service is null, will this return true?

joaoandremartins: After trying this with SDK as null and got an NPE.

Changing so that comparison is made if both are not null, otherwise returns true if any of them isn't null and false if both are null.

Let me know if that's what we want.

joaoandremartins:
Ah yes, I didn't tell you but I was in talks with the Java team and they told me everything would be fixed when they were able to push 1.9.42 to gcloud (they were having some issues), so I didn't worry any more about it. There is a flag you can pass (-Dappengine.user.timezone=UTC) to make it work, but since this will eventually be fixed, I'm not going to do anything there.

The issue with output piping is a little bit weirder. I must have done something that caused output not to be redirected anymore.

joaoandremartins:
I can't find anything I did that could have caused this and the OSProcessHandler interface changed in the meantime, so it could be that IJ changed something that caused the process to not attach to a console anymore.

I'll keep looking but, if this is on the IJ side, I'll have to open an issue with them.

meltsufin:
This is interesting. So, we won't need to add '-Dappengine.user.timezone=UTC' for Java 8, when App Engine SDK 1.9.42 is pushed to gcloud.

joaoandremartins:
Ah, output piping is made possible by a fix I made locally for the common library that needs to be pushed first. It stopped working when I changed the lib version from 0.1.5-SNAPSHOT to 0.1.4.

What do you think of pushing this as is, pushing the fix to the common lib, then changing the commlib version here to 0.1.5?

joaoandremartins:
The fix is actually already deployed to the lib (https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/184). I'm guessing we now only need to push a new library version, update the version there to 0.1.6-SNAPSHOT and set 0.1.5 here.

@meltsufin what do you think?

etanshaul:
Why don't we release the new version of the lib with the fix, and create a quick PR to update the version here first. That way we can verify it in this PR.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
etanshaul: Ok. I'm sure I understand this (why would they be persisted over and over again..seems like a symptom of something that should be refactored). But if there's no other way I'll take your word for it.

joaoandremartins:
@etanshaul @patflynn @aslo I think this is ready. PTAL

patflynn:
Got an error trying to add an Application Server to the run configuration:
[ 258370]  ERROR - l.treeView.AbstractTreeUpdater - com.intellij.openapi.roots.ui.configuration.libraryEditor.LibraryTableTreeBuilder 
--------------Creation trace: 
java.lang.Throwable
	at com.intellij.openapi.roots.impl.libraries.LibraryImpl.<init>(LibraryImpl.java:130)
	at com.intellij.openapi.roots.impl.libraries.LibraryImpl.<init>(LibraryImpl.java:102)
	at com.intellij.openapi.roots.impl.libraries.LibraryTableBase$LibraryModel.createLibrary(LibraryTableBase.java:292)
	at com.intellij.openapi.roots.impl.libraries.LibraryTableBase$LibraryModel.createLibrary(LibraryTableBase.java:286)
	at com.intellij.j2ee.appServerIntegrations.impl.ApplicationServersManagerImpl$ApplicationServersManagerModifiableModelImpl.createNewApplicationServer(ApplicationServersManagerImpl.java:354)
	at com.intellij.j2ee.appServerIntegrations.impl.ApplicationServersManagerImpl$ApplicationServersManagerModifiableModelImpl.createNewApplicationServer(ApplicationServersManagerImpl.java:333)
	at com.intellij.j2ee.appServerIntegrations.impl.ApplicationServersManagerImpl.createServer(ApplicationServersManagerImpl.java:106)
	at com.google.cloud.tools.intellij.appengine.facet.impl.AppEngineUltimateWebIntegration.getOrCreateAppServer(AppEngineUltimateWebIntegration.java:158)
	at com.google.cloud.tools.intellij.appengine.facet.impl.AppEngineUltimateWebIntegration.setupDevServer(AppEngineUltimateWebIntegration.java:143)
	at com.google.cloud.tools.intellij.appengine.facet.AppEngineFacetEditor.onFacetInitialized(AppEngineFacetEditor.java:194)
	at com.intellij.facet.impl.ui.FacetEditorImpl.onFacetAdded(FacetEditorImpl.java:144)
	at com.intellij.facet.impl.ProjectFacetsConfigurator.commitFacets(ProjectFacetsConfigurator.java:234)
	at com.intellij.openapi.roots.ui.configuration.ModulesConfigurator$6.run(ModulesConfigurator.java:321)
	at com.intellij.openapi.application.impl.ApplicationImpl.runWriteAction(ApplicationImpl.java:1010)
	at com.intellij.openapi.roots.ui.configuration.ModulesConfigurator.apply(ModulesConfigurator.java:290)
	at com.intellij.openapi.roots.ui.configuration.projectRoot.ModuleStructureConfigurable.apply(ModuleStructureConfigurable.java:391)
	at com.intellij.openapi.roots.ui.configuration.ProjectStructureConfigurable.apply(ProjectStructureConfigurable.java:342)
	at com.intellij.openapi.options.ex.SingleConfigurableEditor$1.run(SingleConfigurableEditor.java:181)
	at com.intellij.openapi.project.DumbPermissionServiceImpl.allowStartingDumbModeInside(DumbPermissionServiceImpl.java:35)
	at com.intellij.openapi.project.DumbService.allowStartingDumbModeInside(DumbService.java:281)
	at com.intellij.openapi.options.ex.SingleConfigurableEditor.doOKAction(SingleConfigurableEditor.java:176)
	at com.intellij.openapi.ui.DialogWrapper$OkAction.doAction(DialogWrapper.java:1833)
	at com.intellij.openapi.ui.DialogWrapper$DialogWrapperAction.actionPerformed(DialogWrapper.java:1799)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2348)
	at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402)
	at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259)
	at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:252)
	at java.awt.Component.processMouseEvent(Component.java:6535)
	at javax.swing.JComponent.processMouseEvent(JComponent.java:3324)
	at java.awt.Component.processEvent(Component.java:6300)
	at java.awt.Container.processEvent(Container.java:2236)
	at java.awt.Component.dispatchEventImpl(Component.java:4891)
	at java.awt.Container.dispatchEventImpl(Container.java:2294)
	at java.awt.Component.dispatchEvent(Component.java:4713)
	at java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4888)
	at java.awt.LightweightDispatcher.processMouseEvent(Container.java:4525)
	at java.awt.LightweightDispatcher.dispatchEvent(Container.java:4466)
	at java.awt.Container.dispatchEventImpl(Container.java:2280)
	at java.awt.Window.dispatchEventImpl(Window.java:2750)
	at java.awt.Component.dispatchEvent(Component.java:4713)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:758)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:76)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:86)
	at java.awt.EventQueue$4.run(EventQueue.java:731)
	at java.awt.EventQueue$4.run(EventQueue.java:729)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:76)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:728)
	at com.intellij.ide.IdeEventQueue.a(IdeEventQueue.java:866)
	at com.intellij.ide.IdeEventQueue._dispatchEvent(IdeEventQueue.java:650)
	at com.intellij.ide.IdeEventQueue.dispatchEvent(IdeEventQueue.java:381)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:201)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:109)
	at java.awt.WaitDispatchSupport$2.run(WaitDispatchSupport.java:184)
	at java.awt.WaitDispatchSupport$4.run(WaitDispatchSupport.java:229)
	at java.awt.WaitDispatchSupport$4.run(WaitDispatchSupport.java:227)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.awt.WaitDispatchSupport.enter(WaitDispatchSupport.java:227)
	at java.awt.Dialog.show(Dialog.java:1084)
	at com.intellij.openapi.ui.impl.DialogWrapperPeerImpl$MyDialog.show(DialogWrapperPeerImpl.java:792)
	at com.intellij.openapi.ui.impl.DialogWrapperPeerImpl.show(DialogWrapperPeerImpl.java:465)
	at com.intellij.openapi.ui.DialogWrapper.invokeShow(DialogWrapper.java:1638)
	at com.intellij.openapi.ui.DialogWrapper.show(DialogWrapper.java:1587)
	at com.intellij.ide.actions.ShowStructureSettingsAction.showDialog(ShowStructureSettingsAction.java:49)
	at com.intellij.ide.actions.ShowStructureSettingsAction.actionPerformed(ShowStructureSettingsAction.java:38)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher$3.performAction(IdeKeyEventDispatcher.java:595)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.processAction(IdeKeyEventDispatcher.java:646)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.b(IdeKeyEventDispatcher.java:485)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.dispatchKeyEvent(IdeKeyEventDispatcher.java:215)
	at com.intellij.ide.IdeEventQueue._dispatchEvent(IdeEventQueue.java:622)
	at com.intellij.ide.IdeEventQueue.dispatchEvent(IdeEventQueue.java:381)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:201)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:105)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:93)
	at java.awt.EventDispatchThread.run(EventDispatchThread.java:82)

joaoandremartins:
I can't reproduce. Can you enumerate which steps you took?

joaoandremartins:
Now that #822 is merged, let me know if you have any other comments, otherwise I plan to merge this until the end of today. @etanshaul @aslo @patflynn

patflynn:
That stacktrace happened to me on Windows. Perhaps test on that?

Hold on on merging. I'm going to need more time to review.

etanshaul:
I'm looking at the latest changes now.

etanshaul:
Where do we stand on sending the output of the dev server to the IJ console window given that we are now updated to the latest version? I didn't notice and use of the output listeners that would need to be given to the common lib.

joaoandremartins:
The IJ Application Servers way of logging stuff to the console is to just give it the Process object that later gets attached to the console, which is why I created AppEngineStandardRunTask. It should be logging fine now.

joaoandremartins:
Actually, the best might be for me to finish the merge. I'll tell you here when I've pushed the merged version.

etanshaul:
Yea doing the merge before final review makes sense. Let me know if you run into any issues from my last PR.

joaoandremartins:
OK, I tested a bit and this looks good for review. @etanshaul @aslo @patflynn 

I also still need to test on Windows. Let me know if you have any ideas on how to do that.

etanshaul:
@joaoandremartins - does debug work for you? I'm getting timeouts. Run seems to be working great.

etanshaul:
I think I found a bug related to attempting to run the dev server with a missing SDK:

Steps:
1) Configure a local run and give it the path to the sdk. Save the run config
2) Open Settings. Go to Google -> SDK. Clear out the path to the SDK and save it with an empty SDK.
3) Run the config you created in step 1.

Outcome:
image

Expected outcome:
Some kind of message saying the SDK is missing. Here is what happens for deploy under the same scenario after running it with no SDK:

image

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
```diff
@@ -124,107 +138,524 @@ public int getDefaultPort() {
...
+  @Override
+  public String getPythonStartupArgs() {
+    return settings.getPythonStartupArgs();
+  }
+
+  public void setPythonStartupArgs(String pythonStartupArgs) {
+    settings.setPythonStartupArgs(pythonStartupArgs);
+  }
+
+  private static final String SEPARATOR = "!@#";
```
etanshaul: what is "!@#" ?

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurableTest.java
```diff
@@ -35,14 +35,16 @@
...
+import java.nio.file.Paths;
 
 public class AppEngineCloudConfigurableTest extends PlatformTestCase {
   private AppEngineCloudConfigurable appEngineCloudConfigurable;
   private SystemEnvironmentProvider environmentProvider;
   private CloudSdkService cloudSdkService;
   private TextFieldWithBrowseButton cloudSdkDirectoryField;
 
-  private static final String CLOUD_SDK_DIR_PATH = new File("/a/b/c/gcloud-sdk").getAbsolutePath();
+  private static final Path CLOUD_SDK_DIR_PATH = Paths.get("/a/b/c/gcloud-sdk");
```
etanshaul: I think Paths.get is not very smart about doing platform independent stuff. Instead I think its better to break it up yourself:

Paths.get("my", "path", "...")

Its just a unit test, so its not a huge deal unless we start running them on other platforms.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineRunConfigurationEditor.java
```diff
@@ -84,25 +113,80 @@ private void onArtifactChanged() {
...
+    gcsBucketName.setText(serverModel.getDefaultGcsBucketName());
   }
 
   protected void applyEditorTo(CommonModel commonModel) throws ConfigurationException {
     final AppEngineServerModel serverModel = (AppEngineServerModel) commonModel.getServerModel();
+    serverModel.setPort(validateInteger(port.getText(), "port"));
+    serverModel.setArtifact(getSelectedArtifact());
+    serverModel.setAdvancedSettings(enableAdvanced.isSelected());
+
+    if (enableAdvanced.isSelected()) {
```
etanshaul: What if the user enables advanced settings, modifies the config, and then hides the advanced settings. Would we not want those changes saved in this case?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
```diff
@@ -124,107 +138,524 @@ public int getDefaultPort() {
...
+  public String getDefaultGcsBucketName() {
+    return settings.getDefaultGcsBucketName();
+  }
+
+  public void setDefaultGcsBucketName(String defaultGcsBucketName) {
+    settings.setDefaultGcsBucketName(defaultGcsBucketName);
+  }
+
+  /**
+   * This class is used to serialize run/debug config settings. It only supports basic types (e.g.,
```
etanshaul: Just curious what you mean by "basic" types?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerHelper.java
```diff
@@ -0,0 +1,49 @@
...
+import com.intellij.javaee.appServerIntegrations.ApplicationServerPersistentDataEditor;
+import com.intellij.javaee.appServerIntegrations.CantFindApplicationServerJarsException;
+import com.intellij.javaee.oss.server.JavaeePersistentData;
+
+import java.io.File;
+
+/**
+ * @author nik
+ */
+public class AppEngineServerHelper implements ApplicationServerHelper {
```
etanshaul: I deleted this class in the SDK cleanup PR because its no longer doing anything. I don't think we should bring it back.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerIntegration.java
```diff
@@ -48,4 +55,9 @@ public ApplicationServerPersistentDataEditor createNewServerEditor() {
     // project wizard because there is a special 'Google App Engine' option
     return null;
   }
+
+  @Override
+  public ApplicationServerHelper getApplicationServerHelper() {
```
etanshaul: See the version of this class in master - this way you can get rid of `AppEngineServerHelper` completely.

etanshaul:
I don't know exactly how I did this, but I managed to get the local run into an unstoppable state. Clicking the stop button doesn't do anything. And trying to re-run, or debug says port 8080 is already in use.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerIntegration.java
joaoandremartins: Yeah I saw that, but thought I'd use the helper to name the application servers, e.g., App Engine Local Server. Otherwise they show up as Generic

etanshaul: Ah I didn't realize that. Sounds good.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
joaoandremartins: That's a string I'm hoping won't be in a jvm flag, that I'm using as a separator to serialize a list into a string. Let me know if you've got better alternatives.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurableTest.java
joaoandremartins: Ah yes, I didn't really look into the separators, but they are platform-specific so will change it.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineRunConfigurationEditor.java
joaoandremartins: Changed it around so that the configuration is always persisted with all settings, but if the advanced settings flag isn't checked, the RunConfiguration cloned object is trimmed down to its basic settings at startup time.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
joaoandremartins: I had a hard time choosing this word because I'm not too sure what can and can't be serialized. I had issues serializing lists and I read somewhere that custom objects couldn't be serialized by XmlSerializer. Didn't want to use primitive type because String isn't primitive.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerHelper.java
joaoandremartins: It's naming the Application Servers "App Engine Local Server".

joaoandremartins:
Hum yes, I think I've run into this flakiness a few times as well. But I think this happens inside the IDE and we don't have any control over it..

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
etanshaul: oh - maybe call it DELIMITER instead? For some reason it didn't click what when I first looked at it. I don't have a better suggestion for this.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
etanshaul: maybe objects that implement Serializable? Could be that the list issue was because the items in the list did not.

patflynn:
random question: the old dev app server had a flag to control whether the datastore was cleared at restart.  Is that not available through the new flags? I think that's probably one of the most likely things a user will want to control.

patflynn:
Also: Any of you guys seeing IJ crash after being open for a while on this branch?

I'm consistently seeing it quit with a non zero exit code if I leave it open and go do stuff.

etanshaul:
Have you noticed this ever on any other branch (I haven't)? I haven't had this branch open for a long period of time, but I'll test it.

patflynn:
FYI I've been able to reproduce the earlier stacktrace on Windows.

joaoandremartins:
1) I can add a TODO to add the Datastore flag (and possibly all the others). There will have to be another library version too.
2) IntelliJ crashing or just a run/configuration crash? I always have a runIdea window open on the side and I haven't seen IJ crashing yet. But I have seen the run/debug session interrupted on the IDE side and still going outside, like Etan described. I don't know what's causing this and am thinking of looking into it after this PR.
3) @patflynn let me know when you can show me your reproduction of the issue.

patflynn:
I just reproduced the same crash on master so it's probably not your PR.
Something for us to look into but not blocking for your merge.

There's a lot of work to do to create a polished and easy to use interface
for configuring the server,  but most if it we can do by iterating in later
PRs. The one thing I'd like to see fixed before we merge is the checkbox
being used to expand the advanced settings.  I don't think that's the right
use of a checkbox. Perhaps for now you can just always show the panel (also
fix the left margin spacing). That keeps it simple and we can think about
how to improve things later. If you really want to hide the settings you
should use a button. Typically it's a '+' when collapsed and '-' when
expanded. Probably worth looking around the IJ UI to look for examples of
that pattern.

joaoandremartins:
I'm not sure if you're testing my code, since the part that adds Application Servers is mostly new. Can you ensure you're not running the old plugin?

How do you suggest improving the UI, other than changing the Advanced Settings functionality? Why shouldn't you use a checkbox for it?
My original idea was to do exactly that -- use a +/- tab. Kinda like the Before Launch Tasks menu. But this functionality isn't provided by the Swing editor, which is why I fell back to using the checkbox. It is still pretty intuitive to me, so I'd like to understand your point better.

I don't think we should be always showing the panel.

Given the size of this PR, I suggest instead adding TODOs and submitting as is. What do you think?

joaoandremartins:
As for the spacing issue, AFAIK it's caused by the scroll bar added on the right when you toggle the advanced settings. I looked into it, but thought this is the desired behaviour.

I might be able to do something about the artifact and port boxes dislocating to the right, if that's what you were referring to.

patflynn:
A checkbox is a stateful element of a form. When you check a box you expect
that to represent state in your configuration, but in fact it's just a UI
control here.  It's just the wrong widget for the action you want to
perform.

patflynn:
that's interesting and strange. I can't imagine it would be desired. It
looks very raw with the flags left aligned flush against the border,
especially with the enclosing form having the natural margins you would
expect.

joaoandremartins:
@patflynn give it another go. Hopefully you like it better now!

patflynn:
thanks.  I'm just going to do a last quick pass through of the code right now.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -110,16 +111,22 @@ JLabel getWarningMessage() {
   }
 
   public boolean isModified() {
-    return !Comparing.strEqual(getCloudSdkDirectory(),
-        CloudSdkService.getInstance().getSdkHomePath());
+    CloudSdkService sdkService = CloudSdkService.getInstance();
+
+    return Paths.get(getCloudSdkDirectory() != null ? getCloudSdkDirectory() : "")
```
patflynn: why isn't this just using Paths.equals?

etanshaul:
sorry if we talking about this already, but where do we stand on this potential bug: https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/811#issuecomment-241505706

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtilLegacy.java
```diff
@@ -97,6 +95,7 @@ public void customize(JList list, Artifact value, int index, boolean selected,
   @Nullable
   public static AppEngineFacet findAppEngineFacet(@NotNull Project project,
       @NotNull Artifact artifact) {
+    // TODO(joaomartins): Find out why the GAE facet isn't being added to Gradle projects.
```
patflynn: Is this a bug?  Could you include a reference to the github issue here?

etanshaul:
lgtm pending my last comment

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineRunConfigurationEditor.java
```diff
@@ -84,25 +107,75 @@ private void onArtifactChanged() {
...
     final Artifact artifact = serverModel.getArtifact();
     myArtifactComboBox.setSelectedItem(artifact);
     if (artifact == null && myArtifactComboBox.getItemCount() == 1) {
       myArtifactComboBox.setSelectedIndex(0);
+      BuildArtifactsBeforeRunTaskProvider.setBuildArtifactBeforeRun(
+          commonModel.getProject(), commonModel, (Artifact) myArtifactComboBox.getSelectedItem());
     }
-    myServerParametersEditor.setDialogCaption("Server Parameters");
-    myServerParametersEditor.setText(serverModel.getServerParameters());
+    port.setText(serverModel.getPort() != null
```
patflynn: consider extracting a method for all these int to string conversions.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
joaoandremartins: Ah, changed.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
```diff
@@ -47,24 +49,33 @@
...
+
+  private AppEngineModelSettings settings = new AppEngineModelSettings();
 
   @Override
   public J2EEServerInstance createServerInstance() throws ExecutionException {
-    return new AppEngineServerInstance(myCommonModel);
+    return new AppEngineServerInstance(commonModel);
   }
 
+  // TODO(joaomartins): Consider adding something here so we get a nice log pane
```
patflynn: could you open an issue for this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtilLegacy.java
joaoandremartins: Done.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineRunConfigurationEditor.java
joaoandremartins: Done.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
```diff
@@ -47,24 +49,33 @@
...
+    RunConfiguration, Cloneable {
 
-  private ArtifactPointer myArtifactPointer;
-  private int myPort = 8080;
-  private String myServerParameters = "";
-  private CommonModel myCommonModel;
+  private ArtifactPointer artifactPointer;
+  private CommonModel commonModel;
+
+  private AppEngineModelSettings settings = new AppEngineModelSettings();
```
patflynn: just wondering why this Settings object needs to exist. Couldn't you just put all the fields directly on the AppEngineServerModel?

joaoandremartins:
@etanshaul that was a bug and I resolved it by checking if the SDK is empty and returning early, then throwing an ExecutionException from CloudSdkStartupPolicy, which triggers a balloon notification.

However, I'm just realising that I didn't cover the case where the SDK is invalid, so I'll need to add that.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
```diff
@@ -124,107 +138,506 @@ public int getDefaultPort() {
...
+    Path appYaml = Paths.get(artifactPointer.getArtifact().getOutputPath());
+    appYamls.add(appYaml.toFile());
+    return appYamls;
   }
 
-  public void setServerParameters(String serverParameters) {
-    myServerParameters = serverParameters;
+  @Override
+  public String getHost() {
+    return null;
```
patflynn: should this be returning settings.getHost()?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
joaoandremartins: Artifact and CommonModel can't be persisted, so we'd need to go to some lengths to make that work. We could maybe get away with artifact by just keeping a String pointer, but the same can't be done with CommonModel.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
joaoandremartins: I was thinking of working on these immediately after I push this PR. (i.e., these TODOs would effectively be my issue list). But I can create an issue for this one, sure.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
```diff
@@ -124,107 +138,506 @@ public int getDefaultPort() {
...
+
+  public void setPythonStartupArgs(String pythonStartupArgs) {
+    settings.setPythonStartupArgs(pythonStartupArgs);
+  }
+
+  private static final String DELIMITER = "!@#";
+
+  @Override
+  public List<String> getJvmFlags() {
+    return Arrays.asList(settings.getJvmFlags() != null
```
patflynn: I think guava Splitter would give simpler code.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
```diff
@@ -124,107 +138,506 @@ public int getDefaultPort() {
...
+        ? settings.getJvmFlags().split(DELIMITER) : new String[]{});
```
patflynn: also random though: Do the other run configs split their flags? Since the user enters them as a free form string I wonder if it's not safest to just store exactly what they have inputted.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
joaoandremartins: oh, I meant to get around this but got distracted. Good catch!

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerEditor.java
```diff
@@ -36,22 +41,45 @@
...
 
   private JPanel myMainPanel;
+  // TODO(joaomartins): Replace with CloudSdkPanel when
+  // https://youtrack.jetbrains.com/issue/IDEA-110316 gets fixed.
   private TextFieldWithBrowseButton mySdkHomeField;
+  private JLabel warningMessage;
 
   public AppEngineServerEditor() {
     mySdkHomeField
         .addBrowseFolderListener("Google App Engine SDK", "Specify Google App Engine Java SDK home",
```
patflynn: we need to update these labels to Cloud SDK from Google App Engine SDK

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerHelper.java
patflynn: @etanshaul 
@joaoandremartins 

did you guys figure out what to do here?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerHelper.java
```diff
@@ -0,0 +1,49 @@
...
+
+  public ApplicationServerInfo getApplicationServerInfo(
+      ApplicationServerPersistentData persistentData)
+      throws CantFindApplicationServerJarsException {
+    return new ApplicationServerInfo(new File[]{},
+        GctBundle.getString("appengine.run.server.name"));
+  }
+
+  public ApplicationServerPersistentData createPersistentDataEmptyInstance() {
+    return new JavaeePersistentData();
```
patflynn: just curious what this does?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.Map;
+
+/**
+ * Runs a Google App Engine Standard app locally with devappserver, through the tools lib.
+ */
+public class CloudSdkStartupPolicy implements ExecutableObjectStartupPolicy {
+
+  // The startup process handler is kept so the process can be explicitly terminated, since we're
+  // not delegating that to the framework.
+  OSProcessHandler startupProcessHandler;
```
patflynn: should this be private or @VisibleForTesting?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerHelper.java
etanshaul: I think @joaoandremartins confirmed that it is needed if we want to provide a default name for the local server run config.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
```diff
@@ -0,0 +1,155 @@
...
+      @Override
+      public ExecutableObject getDefaultScript(final CommonModel commonModel) {
+        return new ExecutableObject() {
+          @Override
+          public String getDisplayString() {
+            return GctBundle.getString("appengine.run.startupscript.name");
+          }
+
+          @Override
+          public OSProcessHandler createProcessHandler(String string, Map<String, String> map)
```
patflynn: can  you rename the parameters to this method to be more descriptive?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
```diff
@@ -0,0 +1,155 @@
...
+            if (jvmDebugFlag != null) {
+              runConfiguration.addAllJvmFlags(Arrays.asList(jvmDebugFlag.trim().split(" ")));
+            }
+
+            AppEngineStandardRunTask runTask =
+                new AppEngineStandardRunTask(runConfiguration);
+            AppEngineExecutor executor = new AppEngineExecutor(runTask);
+            executor.run();
+
+            Process devappserverProcess = executor.getProcess();
```
patflynn: devAppServer?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
```diff
@@ -0,0 +1,155 @@
...
+            if (devappserverProcess == null) {
+              throw new ExecutionException(
+                  GctBundle.message("appengine.cloudsdk.location.missing.message"));
```
patflynn: do we know that null process means the path is wrong?  is it nullable?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
```diff
@@ -0,0 +1,155 @@
...
+          public OSProcessHandler createProcessHandler(String string, Map<String, String> map)
+              throws ExecutionException {
+            startupProcessHandler.destroyProcess();
+
+            ProcessBuilder dummyProcess = new ProcessBuilder("true");
+            try {
+              return new OSProcessHandler(dummyProcess.start(),
+                  GctBundle.getString("appengine.run.shutdownscript"));
+            } catch (IOException ioe) {
+              return null;
```
patflynn: definitely don't catch and ignore.  We want to see errors.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
joaoandremartins: I wanted to protect myself against the fact that space might be a valid character in a jvm flag but, actually, I don't think it is and I think that's how the flags would be separated. There is even a bug where if the user adds more than one flag separated by spaces, we only count it as one because of how setJvmFlags works.

So, I'm just going to use space as a separator.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerHelper.java
joaoandremartins: Yes, we need this just to name the application servers.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
joaoandremartins: Switched to Splitter.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerEditor.java
joaoandremartins: Done.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerHelper.java
joaoandremartins: This just persists an empty configuration for this application server. In the IJ plugin, this was used to persists an application server of a specific App Engine SDK version. Since we don't support versioning, we have no use for this, but can't return null due to NPE.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
joaoandremartins: Set to private.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
joaoandremartins: Done. string -> workingDirectory, map -> envVariables.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
joaoandremartins: Got rid of this variable since we're now checking SDK validity at the beginning of this method, not AppEngineStandardRunTask.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
joaoandremartins: This was deleted.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
patflynn: I'm heading out. Thanks for taking all the feedback. Feel free to merge
when you've covered my last batch of comments.

Good work.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
joaoandremartins: Done.

joaoandremartins:
Fixes #782

Issue #781, opened 2016-07-13
Remove appcfg sdk usages from integrated plugin


etanshaul:
One example - new GAE standard project wizard:

![image](https://cloud.githubusercontent.com/assets/1735744/16809157/36210b48-48ed-11e6-84b2-23c7a90518ac.png)

PR #815, opened 2016-08-10
Use common lib for AE class whitelisting

etanshaul:
see #781 - This is another piece of removing the old SDK. I am continuing to break that issue into smaller reviewable pieces.

The common library now includes a class, [AppEngineJreWhiteList](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/bb90335836af95f1938be5d49fdcdaabb7a86650/src/main/java/com/google/cloud/tools/appengine/api/whitelist/AppEngineJreWhitelist.java), which checks if a fully qualified classname is whitelisted for AE. 

Previously, the whitelist class was reflectively loaded from the appcfg jar which required some parsing and caching to make the check efficient. The previous logic was removed with a simple direct call to the common lib `AppEngineJreWhiteList#contains()`.

In future PRs related to #781, `AppEngineSdk` (which was brought in from the old plugin) will be removed completely and merged into our new `CloudSdkService`.


etanshaul:
@patflynn / @joaoandremartins / @aslo

joaoandremartins:
Are there already unit tests for the new whitelist mechanism? If not, could you add some?

etanshaul:
The new whitelist mechanism is really just calling `AppEngineJreWhiteList#contains()` (part of the common-lib which itself is already unit tested). Unless you had something else in mind?

joaoandremartins:
Yep, if it's already unit tested, then LGTM.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacet.java
```diff
@@ -68,35 +59,4 @@ public static AppEngineFacet getAppEngineFacetByModule(@Nullable Module module)
...
-    }
-    final DomElement root = element.getRootElement();
-    if (!ReflectionUtil.isAssignable(domClass, root.getClass())) {
-      return null;
-    }
-    return (T) root;
-  }
-
-
-  public boolean shouldRunEnhancerFor(@NotNull VirtualFile file) {
```
patflynn: are these just not used?

patflynn:
:+1: This looks a whole let simpler. I guess we've tested it in action?

I'm wondering what the performance characteristics of the new common lib mechanism are.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacet.java
etanshaul: `shouldRunEnhancerFor` was not used in the old plugin either (I will double check that it was called outside by the framework - its not overriding anything so I don't think it is).

`getRootElement` is downstream from the old tools menu action which is now replaced by our new menu item (and we handle this DOM stuff already).

etanshaul:
I think it will be faster than the old implementation:

The common lib just exposes a HashSet. So contains operations are constant time. Previously, they were parsing the file manually, caching it and then doing the lookup - we no longer have to load any files up manually.

patflynn:
how does the common lib produce the hashset? Is it a static map?

etanshaul:
yes, a static Set

patflynn:
:ship:

PR #817, opened 2016-08-10
Rename packages from old plugin

etanshaul:
Some of the packages brought in from the old JB plugin still had the old name. This PR renames them to conform with our package naming.


etanshaul:
@patflynn I updated the commit without messing with the ©

patflynn:
LGTM

PR #818, opened 2016-08-11
Set user-agent to "Automated" for Analytics event pings

chanseokoh:
Previously, it was `"Apache-HttpClient/4.4.1 (Java/1.7.0_101)"`. Setting it to `"Automated"` prevents GA from filtering out our pings.

`"Automated"`actually comes from [here](https://github.com/google/cloud-reporting/blob/master/src/main/java/com/google/cloud/metrics/MetricsUtils.java#L78).


etanshaul:
lgtm

PR #822, opened 2016-08-16
Remove old sdk from plugin

### LINKED ISSUE #781: Remove appcfg sdk usages from integrated plugin

etanshaul:
fixes #781 

@patflynn, @joaoandremartins 

Removes the remaining references to the old SDK from the project.

- Remove appcfg from the facet editor
- Refactor all the helpers related to app engine standard (appengine-web.xml schema etc) into our CloudSdkService

Open issues:
- Determine how to support datanucleus (jdo/jpa) in the new plugin given that the libs are no longer bundled in the SDK. This support is broken by this PR, but will be fixed in a future one.


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
```diff
@@ -40,4 +79,168 @@ public String getCloudSdkHomePath() {
...
+  }
+
+  @Override
+  public boolean isClassInWhiteList(@NotNull String className) {
+    return AppEngineJreWhitelist.contains(className);
+  }
+
+  @NotNull
+  @Override
+  public List<String> getUserLibraryPaths() {
```
joaoandremartins: This method should probably use the common library to get the SDK libraries. i.e., using getJarPath()

etanshaul:
For reference, here are the libraries that were previously configured in the bundled JB plugin:

image

And here are the libraries configured when choosing datanucleus / jdo in the former bundled plugin:

image

In the current state of our plugin, here is what is configured:
image
jsr107 is missing

And in our plugin for persistence:
image
The libs are not resolved until we determine how to handle this in a future PR.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
etanshaul: it does pull it from gcloud. but since its not on the classpath I don't think we can use getJarPath?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
```diff
@@ -16,21 +16,60 @@
...
  * Default implementation of {@link CloudSdkService} backed by {@link PropertiesComponent} for
  * serialization.
  */
 public class DefaultCloudSdkService extends CloudSdkService {
 
+  private static final Logger logger = Logger.getInstance(DefaultCloudSdkService.class);
+
   private PropertiesComponent propertiesComponent;
   private static final String CLOUD_SDK_PROPERTY_KEY = "GCT_CLOUD_SDK_HOME_PATH";
+  private static final String JAVA_TOOLS_BASE_PATH
```
joaoandremartins: Anything that assumes internal knowledge of the SDK should probably be better left to the library.
If you don't want to deal with removing this in this PR, should we add a TODO here or so?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
joaoandremartins: Since the libraries are under the Cloud SDK folder, in this case, we could add "appengine-endpoints" and "jsr107" to the keys returned by getJarPath(). We wouldn't need to assume that those jars are located at JAVA_TOOLS_BASE_PATH + "lib/opt/user", for example.

etanshaul: Checkout my screenshots below (2nd to last one). Some of the lib's are actually directories. I wonder if we can use a similar mechanism to what you are suggesting.

etanshaul:
@patflynn / @joaoandremartins - any guesses as to why [jsr107](https://jcp.org/en/jsr/detail?id=107) would have been added as a lib by the old plugin? This is not bundled in gcloud so I'm wondering if we want or need to bring this in.

joaoandremartins:
Not sure. @ludoch might know, since I recall this being added to Eclipse as well.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/descriptor/AppEngineWebSchemaProvider.java
```diff
@@ -73,10 +74,14 @@ public XmlFile getSchema(@NotNull @NonNls String url, @Nullable Module module,
       if (facet != null) {
         final File file;
         if (isApplicationXmlFile(baseFile)) {
-          file = facet.getSdk().getApplicationSchemeFile();
+          file = CloudSdkService.getInstance().getApplicationSchemeFile();
         } else {
-          file = facet.getSdk().getWebSchemeFile();
+          file = CloudSdkService.getInstance().getWebSchemeFile();
```
patflynn: what's the difference between these two files? (web vs application scheme)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFrameworkDetector.java
```diff
@@ -56,10 +52,6 @@ public FrameworkType getFrameworkType() {
 
   @Override
   public void setupFacet(@NotNull AppEngineFacet facet, ModifiableRootModel model) {
```
patflynn: does this have to be implemented  if it's a no-op?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
```diff
@@ -150,9 +146,10 @@ private void addSupport(final Module module,
     }
 
     final Project project = module.getProject();
-    webIntegration.addDevServerToModuleDependencies(rootModel, sdk);
+    webIntegration.addDevServerToModuleDependencies(rootModel);
 
-    final Library apiJar = addProjectLibrary(module, "AppEngine API", sdk.getUserLibraryPaths(),
+    final Library apiJar = addProjectLibrary(module, "AppEngine API",
```
patflynn: So does this happen once on creating the facet?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineCommunityWebIntegration.java
```diff
@@ -82,18 +81,15 @@ public void setupJpaSupport(@NotNull Module module, @NotNull VirtualFile persist
   }
 
   @Override
-  public void setupRunConfiguration(@NotNull AppEngineSdk sdk,
-      @Nullable Artifact artifact,
-      @NotNull Project project) {
+  public void setupRunConfiguration(@Nullable Artifact artifact, @NotNull Project project) {
```
patflynn: why is this class called ..Community..?  (Lots of questions from ignorance here. sorry)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
```diff
@@ -27,7 +35,44 @@ public static CloudSdkService getInstance() {
     return ServiceManager.getService(CloudSdkService.class);
   }
 
+  @Nullable
   public abstract String getCloudSdkHomePath();
 
   public abstract void setCloudSdkHomePath(String path);
+
```
patflynn: a lot of this kind of stuff does kind of feel like it could be passed straight through to the common lib. Wrapping it as an IJ service isn't terrible though, but then the name of the service seems a bit off.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineCommunityWebIntegration.java
etanshaul: So the way it works is that plugin.xml in our main cloud-tools module defines a WebIntegration inteface. The same config file then injects a Community implementation for that module.

Similarly, the ultimate module injects an overriding Ultimate implementation. This module is only loaded when running UE.

The CE version is basically just a noop implementation.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFrameworkDetector.java
etanshaul: it doesn't. I'll remove it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
etanshaul: Yes, but I believe the trigger is adding the framework support (which in turn creates the facet).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/descriptor/AppEngineWebSchemaProvider.java
etanshaul: I was wondering that too. Maybe appengine-application.xml is an old concept? This is what the template looks like:

```
<appengine-application xmlns="http://appengine.google.com/ns/1.0">
  <application>app</application>
</appengine-application>
```

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
etanshaul: By passing through to the common lib do you mean having the common lib expose this functionality for us? That is not a bad idea as I could see this being useful to other clients like ecplise.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/descriptor/AppEngineWebSchemaProvider.java
patflynn: ah I think I remember what this is. It's for multi-module/service java apps that are bundled as EARs.  Only appcfg supports it so it probably can be safely removed. People might be a little surprised when they find that deployment doesn't chain all their separate services..

@ludoch

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
etanshaul: And if I interpreted you correctly, how about we leave it in the service and open an enhancement against the common lib to expose this functionality for us.

I'll think about a better name for the service since its doing a bunch of appengine standard'y stuff now.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
etanshaul: Yea. My suggestion would be to leave this here, open an issue against the lib, and then update this once its supported. Sound good to you?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/descriptor/AppEngineWebSchemaProvider.java
etanshaul: Yea, we currently do not support ear deployments. The individual modules need to be deployed separately so I agree this can be removed if that's what it was for.

etanshaul:
@patflynn, @joaoandremartins ptal

In my last commit I cleaned up the file path related logic for the sdk and the AE standard libs. I'm going to look into test coverage, but you can review in the meantime.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
```diff
@@ -27,7 +35,44 @@ public static CloudSdkService getInstance() {
...
+  @Nullable
+  public abstract File getToolsApiJarFile();
```
joaoandremartins: I'm assuming these are still Files to fit more easily in the IJ framework?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
joaoandremartins: In that case, we could add extra methods to the library to get those directories too. I think the SDK clients shouldn't assume any knowledge of the internal SDK file system arrangement.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
```diff
@@ -40,4 +81,214 @@ public String getCloudSdkHomePath() {
...
+  }
+
+  @Override
+  public boolean isValid() {
+    return !StringUtil.isEmpty(getCloudSdkHomePath())
+        && getToolsApiJarFile() != null
+        && getToolsApiJarFile().exists();
+  }
+
+  private static String findLatestVersion(File dir) {
```
joaoandremartins: Do we still need this if we're not supporting SDK versioning?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
```diff
@@ -16,21 +16,62 @@
...
  * Default implementation of {@link CloudSdkService} backed by {@link PropertiesComponent} for
  * serialization.
  */
 public class DefaultCloudSdkService extends CloudSdkService {
 
+  private static final Logger logger = Logger.getInstance(DefaultCloudSdkService.class);
+
   private PropertiesComponent propertiesComponent;
   private static final String CLOUD_SDK_PROPERTY_KEY = "GCT_CLOUD_SDK_HOME_PATH";
+  private static final Path JAVA_TOOLS_RELATIVE_PATH
```
joaoandremartins: Add TODO to get rid this by adding to the library.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
etanshaul: Yea, thats right.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
etanshaul: Actually, the consumers of this are using it to get the latest version of the endpoints and jsr107 libs.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
etanshaul: Will do.

patflynn:
I'm looking at this now. Sorry for the delay.

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/appengine/jps/model/impl/JpsAppEngineModuleExtensionImpl.java
```diff
@@ -74,7 +77,7 @@ public void applyChanges(@NotNull JpsAppEngineModuleExtensionImpl modified) {
   @Override
   public String getToolsApiJarPath() {
     return FileUtil.toSystemDependentName(
-        JpsArtifactPathUtil.appendToPath(getSdkHomePath(), LIB_APPENGINE_TOOLS_API_JAR));
+        JpsArtifactPathUtil.appendToPath(getSdkHomePath(), LIB_APPENGINE_TOOLS_API_JAR.toString()));
```
patflynn: https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/186

patflynn: maybe a todo to update once the issue in link is closed?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/descriptor/AppEngineWebSchemaProvider.java
patflynn: let's remove the ApplicationSchemeFile handling.

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/appengine/jps/model/impl/JpsAppEngineModuleExtensionImpl.java
etanshaul: Yea definitely. This is inline with @joaoandremartins's suggestions to mark todos to offload things to the common lib where it makes sense.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/descriptor/AppEngineWebSchemaProvider.java
etanshaul: Done - see https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/822/commits/fae3ae03423fb53ba4de4862dc85bf063e37a7e2

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
```diff
@@ -187,7 +184,8 @@ private void addSupport(final Module module,
         LOG.error(ioe);
       }
       final Library library = addProjectLibrary(module, "AppEngine ORM",
-          Collections.singletonList(sdk.getOrmLibDirectoryPath()), sdk.getOrmLibSources());
+          Collections.singletonList(CloudSdkService.getInstance().getOrmLibDirectoryPath()),
```
patflynn: not a biggie, but unless I was worried that CloudSdkService could change mid thread, I'd be tempted to just extract a field or variable and reuse it across invocations.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
patflynn: that works.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
```diff
@@ -40,4 +83,206 @@ public String getCloudSdkHomePath() {
   public void setCloudSdkHomePath(String cloudSdkHomePath) {
     propertiesComponent.setValue(CLOUD_SDK_PROPERTY_KEY, cloudSdkHomePath);
   }
+
+  @Nullable
+  private Path getJavaToolsBasePath() {
```
patflynn: not tempted to just blow up in these null cases?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
```diff
@@ -40,4 +83,206 @@ public String getCloudSdkHomePath() {
...
+    return getLibUserDirectoryPath() != null
+        ? Paths.get(getLibUserDirectoryPath(), "orm").toString()
+        : null;
+  }
+
+  @NotNull
+  @Override
+  // TODO this path is incorrect. We need to determine an alternate strategy for loading these libs
+  // maybe from maven central
+  public VirtualFile[] getOrmLibSources() {
```
patflynn: do we need any of these since we're not doing jpa/jdo support?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
etanshaul: my reasoning was that I didn't want the IDE to explode if gcloud ever changes the way it packages these libs. But maybe having it explode would be a way for us to quickly be informed. What do you think?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
etanshaul: no. but I'd like to remove all of this in a different PR since this one is already a little big.

patflynn: Yeah if this is erroring I'd want to know about it

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
patflynn: if this returns a list of paths should it just return List<Path> or List<File>?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
```diff
@@ -40,4 +83,206 @@ public String getCloudSdkHomePath() {
...
+  @NotNull
+  @Override
+  public List<String> getUserLibraryPaths() {
+    List<String> result = new ArrayList<>();
+    String libUserDirectory = getLibUserDirectoryPath();
+    if (libUserDirectory != null && new File(libUserDirectory).exists()) {
+      result.add(libUserDirectory);
+    }
+
+    if (getJavaToolsBasePath() == null) {
```
patflynn: doesn't libUserDirectory depend on getJavaToolsBasePath not being null?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
etanshaul: it does return a list? did you mean a list of Paths?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
```diff
@@ -40,4 +83,206 @@ public String getCloudSdkHomePath() {
...
+  private static String findLatestVersion(File dir) {
+    String[] names = dir.list();
+    if (names != null && names.length > 0) {
+      String max = Collections.max(Arrays.asList(names));
+      return FileUtil.toSystemIndependentName(new File(dir, max).getAbsolutePath());
+    }
+    return null;
+  }
+
+  private Map<String, Set<String>> loadBlackList() throws IOException {
```
patflynn: do you know if anybody has file an issue in appengine-plugins-core to provide an API for this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
```diff
@@ -40,4 +83,206 @@ public String getCloudSdkHomePath() {
...
+  @Nullable
+  private String getLibUserDirectoryPath() {
+    return getJavaToolsBasePath() != null
+        ? getJavaToolsBasePath().resolve(Paths.get("lib", "user")).toString()
+        : null;
+  }
+
+  private static File[] getJarsFromDirectory(File libFolder) {
+    List<File> jars = new ArrayList<>();
+    final File[] files = libFolder.listFiles();
```
patflynn: I think this does I/O.  Can we make this method non-static?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
etanshaul: If javaToolsBasePath is null it will return an empty list which was just to satisfy the caller that will NPE otherwise. But I suppose this should be refactored.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerHelper.java
```diff
@@ -32,14 +33,14 @@
...
-    String version = sdk.getVersion();
-    return new ApplicationServerInfo(sdk.getLibraries(),
+    final CloudSdkService sdkService = CloudSdkService.getInstance();
+    String version = sdkService.getVersion();
+    return new ApplicationServerInfo(sdkService.getLibraries(),
         "AppEngine Dev" + (version != null ? " " + version : ""));
   }
 
   public ApplicationServerPersistentData createPersistentDataEmptyInstance() {
-    return new AppEngineServerData("");
```
patflynn: just curious about this change. What is it even doing? how does the dta object get used?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/AppEngineServerStartupPolicy.java
```diff
@@ -38,17 +37,14 @@
 public class AppEngineServerStartupPolicy implements JavaCommandLineStartupPolicy {
 
   public JavaParameters createCommandLine(CommonModel commonModel) throws ExecutionException {
-    final AppEngineServerData data = (AppEngineServerData) commonModel.getApplicationServer()
-        .getPersistentData();
-    final AppEngineSdk sdk = data.getSdk();
-    if (StringUtil.isEmpty(sdk.getSdkHomePath())) {
+    final CloudSdkService sdkService = CloudSdkService.getInstance();
+    if (StringUtil.isEmpty(sdkService.getCloudSdkHomePath())) {
       throw new ExecutionException("Path to App Engine SDK isn't specified");
```
patflynn: reword error message from App Engine SDK to Cloud SDK

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/AppEngineServerStartupPolicy.java
```diff
@@ -38,17 +37,14 @@
...
     }
-    final File toolsApiJarFile = sdk.getToolsApiJarFile();
-    if (!toolsApiJarFile.exists()) {
+    final File toolsApiJarFile = sdkService.getToolsApiJarFile();
+    if (toolsApiJarFile == null || !toolsApiJarFile.exists()) {
       throw new ExecutionException(
-          "'" + sdk.getSdkHomePath() + "' isn't valid App Engine SDK installation: '"
-              + toolsApiJarFile.getAbsolutePath() + "' not found");
+          "'" + sdkService.getCloudSdkHomePath() + "' isn't valid App Engine SDK installation: '");
```
patflynn: same here

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/AppEngineServerStartupPolicy.java
etanshaul: this class gets removed by https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/811 so I didn't bother

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
patflynn: oh weird.  I guess my `<>` got removed from my comments by the markup formatting. my suggestion was `List<Path>` or `List<File>`

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerHelper.java
etanshaul: Good point. The config subclass (AppEngineServerData) was removed. This whole class can actually be deleted too as its not doing anything.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
etanshaul: no I don't think so. I can file one though.

etanshaul: do you know where this method blacklist even comes from? It was pulled from old plugin, but I don't see this documented anywhere in AE (whereas the class whitelist is documented).

patflynn: wait wasn't this replaced by the whitelist?

etanshaul: No. My original explanation of all this was incorrect:

There are two separate concepts, a _class_ whitelist (which the common lib is providing us and was previously pulled from the old sdk jar), and a _method_ blacklist which blacklists some methods on classes in the whilelist (for example `System` is whitelisted, but `System#exit` is blacklisted).

Its this blacklist that I don't know where JB got it from. It's a plain text file in the resources folder.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
patflynn: let's leave the nullable returns as is then. I don't think logging.error makes sense since as you say, these methods can be called when the user doesn't have an SDK set.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
patflynn: what I meant is maybe this should be the first condition.  i.e.  shortcircuit if getJavaToolsBasPath is null, then add the libUserDirectory, since you know it exists by that point.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/AppEngineServerStartupPolicy.java
patflynn: ok

etanshaul:
@patflynn / @joaoandremartins I think I've address all of your comments and cleaned things up a bit in my last several commits. ptal.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/AppEngineCloudSdkService.java
```diff
@@ -0,0 +1,290 @@
...
+  private static final Logger logger = Logger.getInstance(AppEngineCloudSdkService.class);
+
+  private PropertiesComponent propertiesComponent;
+  private static final String CLOUD_SDK_PROPERTY_KEY = "GCT_CLOUD_SDK_HOME_PATH";
+  private static final Path JAVA_TOOLS_RELATIVE_PATH
+      = Paths.get("platform", "google_appengine", "google", "appengine", "tools", "java");
+  private Map<String, Set<String>> myMethodsBlackList;
+
+  public AppEngineCloudSdkService() {
+    this.propertiesComponent = PropertiesComponent.getInstance();
```
joaoandremartins: You could initialize propertiesComponent at declaration time, since you're not assigning anywhere else.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/AppEngineCloudSdkService.java
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  @Nullable
+  @Override
+  public String getSdkHomePath() {
```
joaoandremartins: FYI I'm changing this in my PR to return a Path.

joaoandremartins:
Aside from the propertiesComponent nit I pointed out, LGTM.

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/appengine/jps/model/impl/JpsAppEngineModuleExtensionImpl.java
patflynn: still don't see a TODO: here

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/AppEngineCloudSdkService.java
```diff
@@ -0,0 +1,290 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.jar.Attributes;
+
+/**
+ * Default implementation of {@link AppEngineSdkService} backed by {@link PropertiesComponent} for
+ * serialization.
+ *
+ */
+// TODO Offload various path logic for retrieving AE libs to the common library once implemented
```
patflynn: for TODO comments I'm not sure what our github strategy but I'd be fine with "TODO({github-user-name}): ..." which would match our styleguide.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/AppEngineCloudSdkService.java
```diff
@@ -0,0 +1,290 @@
...
+public class AppEngineCloudSdkService extends AppEngineSdkService {
```
patflynn: This renaming has surprised me.  The concept of an 'App Engine SDK' seems to be going away. But let's not churn on it. This is fine and we can think about the right way to model cloud sdk vs App Engine tools at a later date.

patflynn:
Feel free to merge once you address my last few comments.

etanshaul:
I addressed the last couple of comments:

- I renamed the SDK service back to what it was (I agree the 'app engine sdk' wording isn't the best)
- Added a missing todo and added username to them
- Initialize member var at declaration

I'll merge this in once the build is green.

PR #823, opened 2016-08-16
reset idea version to 15.0.6

etanshaul:
A previous PR accidentally changed this version. We still want this set to our earliest supported build.


patflynn:
:+1:

PR #825, opened 2016-08-16
Change user-agent for Analytics HTTP requests

chanseokoh:
Turns out that we have a great degree of flexibility in choosing a user-agent string.

### DIFF
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
@@ -52,6 +52,7 @@ public class GoogleUsageTracker implements UsageTracker, SendsEvents {
 
   private final String analyticsId;
   private String externalPluginName;
+  private String userAgent;
 
   /**
    * Constructs a usage tracker configured with analytics and plugin name configured from its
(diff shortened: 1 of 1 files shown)


etanshaul:
Do you mean it can be set to anything and we still get the same effect?

chanseokoh:
No, the previous string `"Apache-HttpClient/4.4.1 (Java/1.7.0_101)"` was problematic. I checked that this one in the PR works. Eclipse will use its own string as well.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
```diff
@@ -111,10 +114,10 @@ public FluentTrackingEventWithLabel trackEvent(String action) {
     return new TrackingEventBuilder(this, externalPluginName, action);
   }
 
-  private static void sendPing(@NotNull final List<? extends NameValuePair> postData) {
+  private void sendPing(@NotNull final List<? extends NameValuePair> postData) {
```
etanshaul: you made this non-static but I don't see any change to a caller of this?

chanseokoh: It's to use `userAgent`. Another way could be making `userAgent` static (and possibly other fields). Which one do you think is better?

etanshaul:
Oh I see. it looks like sendPing is only called from a non-static context so what you did looks fine.

etanshaul:
lgtm

PR #826, opened 2016-08-17
Removing analytics oauth scopes

patflynn:
They are not used by our plugins in intellij.


etanshaul:
lgtm. but just curious - is there a way to verify this?

patflynn:
We don't actually call any analytics APIs from intellij. It was for Android
Studio and they no longer need them there either.

etanshaul:
oh ok. lgtm then.

PR #832, opened 2016-08-18
Only send undeploy analytics ping for actual deployment cancellations.

### LINKED ISSUE #831: App Engine undeploy analytics ping gets executed for all task cancellations
It should only apply to deployment cancellations.

etanshaul:
fixes #831 

@patflynn what do you think of this approach?

To be determined: do we want a separate type of ping for flex vs standard deploy cancellations. Also, now that I'm thinking about it, do we really care about tracking cancellations (note this is completely different than undeployments).

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineTask.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineTask.java
@@ -21,12 +21,18 @@ import com.google.cloud.tools.appengine.cloudsdk.process.ProcessStartListener;
 /**
  * Encapsulates a task to be run on Google App Engine.
  */
-public interface AppEngineTask {
+public abstract class AppEngineTask {
 
   /**
    * Executes an App Engine task.
    * @param startListener a callback for retrieving the running process
    */
-  void execute(ProcessStartListener startListener);
+  abstract void execute(ProcessStartListener startListener);
+
+  /**
+   * Gets invoked with the task gets cancelled by {@link AppEngineRunner}
+   */
+  void onCancel() {
+  }
 
 }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineTask.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineTask.java
@@ -30,7 +30,7 @@ public abstract class AppEngineTask {
   abstract void execute(ProcessStartListener startListener);
 
   /**
-   * Gets invoked with the task gets cancelled by {@link AppEngineRunner}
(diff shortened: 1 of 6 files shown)


etanshaul:
+ @chanseokoh

chanseokoh:
> do we want a separate type of ping for flex vs standard deploy cancellations

For this kind of thing, I think it's more appropriate to pass this information as metadata (`withLabel()` and `setValue()`).

etanshaul:
@chanseokoh - thanks I'll take that suggestion.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineTask.java
```diff
@@ -21,12 +21,20 @@
 /**
  * Encapsulates a task to be run on Google App Engine.
  */
-public interface AppEngineTask {
+public abstract class AppEngineTask {
```
patflynn: why change this to an abstract base class?

patflynn:
approach looks good to me.

etanshaul:
Updated the ping to include the environment meta-data using @chanseokoh's suggested approach

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineTask.java
etanshaul: so that I can have an empty implementation of the onCancel method - I didn't want to force implementors to override this method if they didn't need it. 

It's a matter of taste I guess, but I prefer this to having a bunch of implementors with empty implementations.

chanseokoh:
Pinging works fine.

selection_001

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineTask.java
patflynn: I'm OK with this for this instance. 

Generally I'm not a big fan of using abstract classes as interfaces because taken to the extreme, inheritors lose the ability to implement multiple interfaces (AKA abstract parent classes). I also tend to like really explicit interfaces that implementors have to think about in terms of the full contract.  One way to combine things is to have an interface + an abstract base class with some convenience implementations.

etanshaul: Yea for sure. The other approach here would be to keep it an interface so that the onCancel method is very explicitly published, and then live with empty implementations of it.

But if we don't feel strongly in this case, I'll leave it as is.

patflynn:
let me know when you're done with commits. Only pending item I see is the ping change to add a label

etanshaul:
that change is done. I added the environment label based on what kind of deployment it is

patflynn:
ship it.

PR #834, opened 2016-08-22
Workaround for NoSuchMethodError exception

### LINKED ISSUE #813: NoSuchMethodError exception in 2016.2.1
Exception class name: java.lang.NoSuchMethodError: com.intellij.util.containers.ContainerUtil.intersection(Ljava/util/Collection;Ljava/util/Collection;)Ljava/util/Collection;
Exception message: com.intellij.util.containers.ContainerUtil.intersection(Ljava/util/Collection;Ljava/util/Collection;)Ljava/util/Collection;
Stack trace:
java.lang.NoSuchMethodError: com.intellij.util.containers.ContainerUtil.intersection(Ljava/util/Collection;Ljava/util/Collection;)Ljava/util/Collection;
	at com.google.cloud.tools.intellij.git.UploadSourceAction.performFirstCommitIfRequired(UploadSourceAction.java:470)
	at com.google.cloud.tools.intellij.git.UploadSourceAction.access$500(UploadSourceAction.java:93)
	at com.google.cloud.tools.intellij.git.UploadSourceAction$1.run(UploadSourceAction.java:211)
	at com.intellij.openapi.progress.impl.CoreProgressManager$TaskRunnable.run(CoreProgressManager.java:635)
	at com.intellij.openapi.progress.impl.CoreProgressManager$3.run(CoreProgressManager.java:170)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:494)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:443)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.executeProcessUnderProgress(ProgressManagerImpl.java:54)
	at com.intellij.openapi.progress.impl.CoreProgressManager.runProcess(CoreProgressManager.java:155)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl$1.run(ProgressManagerImpl.java:128)
	at com.intellij.openapi.application.impl.ApplicationImpl$2.run(ApplicationImpl.java:308)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)

aslo:
Fixes #813 

Implement container intersection method so we don't need to depend on the SDK util method `ContainerUtil.intersection`, which is apparently not available at runtime when running with the 2016.2 release of IDEA.


etanshaul:
Did you check to see if this method is available in Google collections / Apache commons or something like that?

patflynn:
It's kind of weird to me to have an intersection function on Collections
objects. Seems like the behavior around handling duplicates isn't intuitive.

aslo:
@patflynn I agree, this is nonintuitive behavior, will revise.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/git/UploadSourceAction.java
```diff
@@ -468,7 +468,8 @@ public void run() {
         return false;
       }
 
-      Collection<VirtualFile> files2add = Collections.intersection(untrackedFiles, files2commit);
+      Collection<VirtualFile> files2add =
```
patflynn: do we know that the behavior if this other utility method is the same as the one it replaces? 

I'm also a bit concerned we're using a library provided through the IDEA platform classpath that they may not have intended to be used by clients.  

will chat with you about this offline.

etanshaul:
lgtm assuming you confirmed that the intersection / difference operations behave the same way as before

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/git/UploadSourceAction.java
```diff
@@ -468,9 +469,12 @@ public void run() {
         return false;
       }
 
-      Collection<VirtualFile> files2add =
-          CollectionUtils.intersection(untrackedFiles, files2commit);
-      Collection<VirtualFile> files2rm = ContainerUtil.subtract(trackedFiles, files2commit);
+      Set<VirtualFile> files2CommitAsSet = new HashSet<>(files2commit);
+      Set<VirtualFile> untrackedFilesAsSet = new HashSet<>(untrackedFiles);
```
patflynn: this looks a lot better and is more explicit. 

When I look at this performFirstCommitIfRequired method I'm a bit surprised that we have to define this kind of thing. It feels very standard git workflowy. But I don't want to get into changing that.  My main/only concern is that the behavior remains the same for now. It would be cool to have some tests to reflect that.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/git/UploadSourceAction.java
```diff
@@ -467,8 +469,12 @@ public void run() {
         return false;
       }
 
-      Collection<VirtualFile> files2add = ContainerUtil.intersection(untrackedFiles, files2commit);
-      Collection<VirtualFile> files2rm = ContainerUtil.subtract(trackedFiles, files2commit);
+      Set<VirtualFile> files2CommitAsSet = new HashSet<>(files2commit);
+      Set<VirtualFile> untrackedFilesAsSet = new HashSet<>(untrackedFiles);
+      Set<VirtualFile> trackedFilesAsSet = new HashSet<>(trackedFiles);
+
+      Collection<VirtualFile> files2add = Sets.intersection(untrackedFilesAsSet, files2CommitAsSet);
```
etanshaul: the use of '2' here kind of bugs me, but I guess thats not what this PR is about

aslo:
@patflynn So, technically this new `intersection` method does not behave in exactly the same way as it did before, which I actually think is an improvement.  The Jetbrains `intersection` method actually handled duplicates asymmetrically, which is a bug. 

Since we're dealing with Sets of files anyway, I think it's appropriate to dedupe these collections.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/git/UploadSourceAction.java
aslo: Good point. I was just copying from what was there but since I'm in the neighborhood I can update.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/git/UploadSourceAction.java
```diff
@@ -461,19 +463,23 @@ public void run() {
       }, indicator.getModalityState());
       final GcpUntrackedFilesDialog dialog = dialogRef.get();
 
-      final Collection<VirtualFile> files2commit = dialog.getSelectedFiles();
-      if (!dialog.isOK() || files2commit.isEmpty()) {
+      final Collection<VirtualFile> filesToCommit = dialog.getSelectedFiles();
```
etanshaul: ahh so much easier on the eyes now. Thanks.

patflynn:
merge when ready.

PR #838, opened 2016-08-24
change runtime error message for missing creds on deploy

patflynn:
'Application default credentials ' is a specific credential type for GCP. We aren't creating ADCs for deploy in this scenario so I reworded the error.

Generally users should not see this error.


patflynn:
@etanshaul

PR #845, opened 2016-08-25
Remove no longer applicable flexible environment messaging from Cloud…

### LINKED ISSUE #773: Remove app engine flexible price warning from cloud settings
On the Cloud configuration page (Settings -> Clouds) it is no longer appropriate to have the app engine flexible price warning here because we now also support deployment to app engine standard. We also have a price warning in the Deployment UI so we should be covered.

image

etanshaul:
… panel

fixes #773


patflynn:
LGTM

patflynn:
I wonder if we could send them to the global App Engine docs?

etanshaul:
Which docs are these?

etanshaul:
Oh I misread your comment. Yea that would be nice. I'll replace the flex warning with this.

etanshaul:
I took your suggestion - it now points to the AE docs instead.

etanshaul:
image

patflynn:
LGTM

PR #850, opened 2016-08-29
Adds usage metrics for local run/debug and starting a new project.

### LINKED ISSUE #842: add analytics pings to newly integrated JB code

joaoandremartins:
Fixes #842

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
@@ -24,6 +24,7 @@ import com.google.cloud.tools.intellij.appengine.sdk.CloudSdkService;
 import com.google.cloud.tools.intellij.stats.UsageTrackerProvider;
 import com.google.cloud.tools.intellij.util.GctTracking;
 import com.google.common.base.Strings;
+
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
@@ -36,8 +36,6 @@ public class AppEngineStandardRunTask extends AppEngineTask {
   private String runnerId;
 
   /**
-   * {@link AppEngineStandardRunTask} constructor.
-   *
    * @param runConfig local run configuration to be sent to the common library
    * @param runnerId typically "Run" or "Debug", to indicate type of local run
    */
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.tools.intellij.appengine.cloud;
 
+import com.google.api.client.repackaged.com.google.common.base.Strings;
 import com.google.cloud.tools.appengine.api.devserver.RunConfiguration;
 import com.google.cloud.tools.appengine.cloudsdk.CloudSdk;
 import com.google.cloud.tools.appengine.cloudsdk.CloudSdkAppEngineDevServer;
@@ -58,7 +59,7 @@ public class AppEngineStandardRunTask extends AppEngineTask {
 
     UsageTrackerProvider.getInstance()
         .trackEvent(GctTracking.APP_ENGINE_RUN)
-        .withLabel(runnerId != null ? runnerId : "")
+        .withLabel(Strings.nullToEmpty(runnerId))
         .ping();
 
   }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
@@ -37,6 +37,8 @@ public class AppEngineStandardRunTask extends AppEngineTask {
   private String runnerId;
 
   /**
+   * {@link AppEngineStandardRunTask} constructor.
+   *
    * @param runConfig local run configuration to be sent to the common library
    * @param runnerId typically "Run" or "Debug", to indicate type of local run
    */
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
@@ -21,8 +21,11 @@ import com.google.cloud.tools.appengine.cloudsdk.CloudSdk;
 import com.google.cloud.tools.appengine.cloudsdk.CloudSdkAppEngineDevServer;
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessStartListener;
 import com.google.cloud.tools.intellij.appengine.sdk.CloudSdkService;
+import com.google.cloud.tools.intellij.stats.UsageTrackerProvider;
+import com.google.cloud.tools.intellij.util.GctTracking;
 
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * Represents an App Engine Standard run task. (i.e., devappserver)
@@ -30,9 +33,17 @@ import org.jetbrains.annotations.NotNull;
 public class AppEngineStandardRunTask extends AppEngineTask {
 
   private RunConfiguration runConfig;
+  private String runnerId;
 
-  public AppEngineStandardRunTask(@NotNull RunConfiguration runConfig) {
+  /**
+   * {@link AppEngineStandardRunTask} constructor.
+   *
+   * @param runConfig local run configuration to be sent to the common library
+   * @param runnerId typically "Run" or "Debug", to indicate type of local run
+   */
+  public AppEngineStandardRunTask(@NotNull RunConfiguration runConfig, @Nullable String runnerId) {
     this.runConfig = runConfig;
+    this.runnerId = runnerId;
   }
 
   @Override
@@ -46,5 +57,11 @@ public class AppEngineStandardRunTask extends AppEngineTask {
 
     CloudSdkAppEngineDevServer devServer = new CloudSdkAppEngineDevServer(sdkBuilder.build());
     devServer.run(runConfig);
+
+    UsageTrackerProvider.getInstance()
+        .trackEvent(GctTracking.APP_ENGINE_RUN)
+        .withLabel(runnerId != null ? runnerId : "")
+        .ping();
+
   }
 }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
@@ -16,7 +16,6 @@
 
 package com.google.cloud.tools.intellij.appengine.cloud;
 
-import com.google.api.client.repackaged.com.google.common.base.Strings;
 import com.google.cloud.tools.appengine.api.devserver.RunConfiguration;
 import com.google.cloud.tools.appengine.cloudsdk.CloudSdk;
 import com.google.cloud.tools.appengine.cloudsdk.CloudSdkAppEngineDevServer;
@@ -24,7 +23,7 @@ import com.google.cloud.tools.appengine.cloudsdk.process.ProcessStartListener;
 import com.google.cloud.tools.intellij.appengine.sdk.CloudSdkService;
 import com.google.cloud.tools.intellij.stats.UsageTrackerProvider;
 import com.google.cloud.tools.intellij.util.GctTracking;
-
+import com.google.common.base.Strings;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacet.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacet.java
@@ -40,10 +40,6 @@ public class AppEngineFacet extends Facet<AppEngineFacetConfiguration> {
       @NotNull String name,
       @NotNull AppEngineFacetConfiguration configuration) {
     super(facetType, module, name, configuration, null);
-
-    UsageTrackerProvider.getInstance()
-        .trackEvent(GctTracking.APP_ENGINE_SUPPORT_ADDED)
-        .ping();
   }
 
   public static FacetType<AppEngineFacet, AppEngineFacetConfiguration> getFacetType() {
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacet.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacet.java
@@ -40,8 +40,6 @@ public class AppEngineFacet extends Facet<AppEngineFacetConfiguration> {
       @NotNull String name,
       @NotNull AppEngineFacetConfiguration configuration) {
     super(facetType, module, name, configuration, null);
-
-    UsageTrackerProvider.getInstance().trackEvent(GctTracking.APP_ENGINE_CREATE_FACET).ping();
   }
 
   public static FacetType<AppEngineFacet, AppEngineFacetConfiguration> getFacetType() {
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacet.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacet.java
@@ -16,6 +16,8 @@
 
 package com.google.cloud.tools.intellij.appengine.facet;
 
+import com.google.cloud.tools.intellij.stats.UsageTrackerProvider;
+import com.google.cloud.tools.intellij.util.GctTracking;
 import com.intellij.facet.Facet;
 import com.intellij.facet.FacetManager;
 import com.intellij.facet.FacetType;
@@ -38,6 +40,10 @@ public class AppEngineFacet extends Facet<AppEngineFacetConfiguration> {
       @NotNull String name,
       @NotNull AppEngineFacetConfiguration configuration) {
     super(facetType, module, name, configuration, null);
+
+    UsageTrackerProvider.getInstance()
+        .trackEvent(GctTracking.APP_ENGINE_SUPPORT_ADDED)
+        .ping();
   }
 
   public static FacetType<AppEngineFacet, AppEngineFacetConfiguration> getFacetType() {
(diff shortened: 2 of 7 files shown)


google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
```diff
@@ -92,7 +92,7 @@ public OSProcessHandler createProcessHandler(
             }
 
             AppEngineStandardRunTask runTask =
-                new AppEngineStandardRunTask(runConfiguration);
+                new AppEngineStandardRunTask(runConfiguration, programRunner.getRunnerId());
```
etanshaul: Can you explain how this runner id gets populated with "run" or "debug" like you mention in a comment in this PR? It's not clear to from reading the docs in ProgramRunner.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
```diff
@@ -300,6 +302,10 @@ public void addSupport(@NotNull Module module,
       AppEngineSupportProvider.this
           .addSupport(module, rootModel, myFrameworkSupportModel,
               PersistenceApiComboboxUtil.getSelectedApi(myPersistenceApiComboBox));
+
+      UsageTrackerProvider.getInstance()
+          .trackEvent(GctTracking.APP_ENGINE_SUPPORT_ADDED)
```
etanshaul: I believe this will track both new AE projects and AE project imports. What about adding a metric to track usages of the AE standard project wizard? I know there's some overlap there, but it could be nice to have that metric on its own.

joaoandremartins:
@etanshaul @patflynn @aslo 

This captures metrics for local run/debug and new project with GAE facet.
It doesn't capture adding the GAE facet to a project because I didn't locate anywhere in our code we have control when a facet is added. Is this metric that important that warrants extra time spent on it?

I also didn't test run this change because I don't know how to. I'll wait for @chanseokoh to come back to point me to where I would see my pings.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
```diff
@@ -21,18 +21,29 @@
...
  * Represents an App Engine Standard run task. (i.e., devappserver)
  */
 public class AppEngineStandardRunTask extends AppEngineTask {
 
   private RunConfiguration runConfig;
+  private String runnerId;
 
-  public AppEngineStandardRunTask(@NotNull RunConfiguration runConfig) {
+  /**
+   * {@link AppEngineStandardRunTask} constructor.
```
etanshaul: really nitpicking, but seems like a redundant comment

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
joaoandremartins: IJ populates it. It has two ProgramRunners to run Runs and Debugs, and getRunnerId() was the only place I could find that allows us to distringuish between runs and debugs. I used the debugger to check these values.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
joaoandremartins: This actually only seems to track new project wizard. This code path isn't activating on project imports, looking for the gold...

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
```diff
@@ -46,5 +57,11 @@ public void execute(ProcessStartListener startListener) {
 
     CloudSdkAppEngineDevServer devServer = new CloudSdkAppEngineDevServer(sdkBuilder.build());
     devServer.run(runConfig);
+
+    UsageTrackerProvider.getInstance()
+        .trackEvent(GctTracking.APP_ENGINE_RUN)
+        .withLabel(runnerId != null ? runnerId : "")
```
aslo: Minor thing, but you can replace this with a string helper from [Guava](http://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Strings.html#nullToEmpty(java.lang.String)): `Strings.nullToEmpty`. I believe apache commons provides something similar as well.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
etanshaul: It _should_ track imports if you import an ae standard project and then add framework support by clicking on the popup. 

But regardless, for new project's we should probably be explicit about tracking this action rather than tracking it's side effect.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
etanshaul: ah ok cool

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
joaoandremartins: Added label.

joaoandremartins:
@etanshaul @aslo PTAL.

Regarding project imports, the typical way to add GAE support is to click on the Configure popup link. We only have control at AppEngineFacetType.createFacet(), AppEngineFacetType.createDefaultConfiguration() and AppEngineFacet(), which are also used by the other ways of adding facets to modules and give us no way of distinguishing what spawned that facet creation, e.g., explicitly adding a facet, clicking the configure button, new project wizard, etc.

I would suggest adding tracking at the AppEngineFacet() level, which tracks every possible way of adding facets, but a face that is constructed isn't necessarily added to a module, so this kind of tracking would possibly not provide very reliable results...

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
etanshaul: What I meant about being explicit about tracking new projects wizard usages is to add the ping directly on the method that the wizard uses (not on the side effect of adding ae support). It could also use a top event instead of a label.

Didn't you mention that this method gets invoked for imports? If so, then the current logic would incorrectly count imports as new projects, no?

joaoandremartins: I never mentioned this method is invoked for imports -- on the contrary, it's not.

This is the method that the wizard uses.

I'll change to a dedicated top level event here.

etanshaul: Oh ok. I misread your last comment. Sounds good with regards to tracking new projects.

Regarding facets - I see no harm in adding tracking to all additions of the ae facet - even if this means some useless additions (not added to the module, like you mentioned).

joaoandremartins: OK, I'm adding tracking for both GAE facets added through Project Settings -> Facets and every created GAE facet (even if it isn't added).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacetEditor.java
```diff
@@ -172,6 +174,10 @@ public String getHelpTopic() {
   @Override
   public void onFacetInitialized(@NotNull Facet facet) {
     AppEngineWebIntegration.getInstance().setupDevServer();
+
+    UsageTrackerProvider.getInstance()
+        .trackEvent(GctTracking.APP_ENGINE_ADD_FACET)
```
etanshaul: whats the difference between facet addition and facet creation? Is there duplication / overlap?

LGTM otherwise.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacet.java
```diff
@@ -38,6 +40,8 @@ public AppEngineFacet(@NotNull FacetType facetType,
       @NotNull String name,
       @NotNull AppEngineFacetConfiguration configuration) {
     super(facetType, module, name, configuration, null);
+
+    UsageTrackerProvider.getInstance().trackEvent(GctTracking.APP_ENGINE_CREATE_FACET).ping();
```
patflynn: is the facet added to a module in code that we manage?  I'm thinking AE facets added to modules is closer to the signal we want.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
```diff
@@ -300,6 +302,10 @@ public void addSupport(@NotNull Module module,
       AppEngineSupportProvider.this
           .addSupport(module, rootModel, myFrameworkSupportModel,
               PersistenceApiComboboxUtil.getSelectedApi(myPersistenceApiComboBox));
+
+      UsageTrackerProvider.getInstance()
```
patflynn: Is this called when a facet is added or only on new project creation?

joaoandremartins: Only at new project creation time, here.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacetEditor.java
joaoandremartins: An example I just found is when you start IJ and AppEngineFacet() is called, even though you aren't adding any facet (the IDE is constructing the project and constructing an AppEngineFacet). You'd be counting each time the browser is started as a new facet added.

I'm afraid there are other cases where AppEngineFacet() is called but a facet isn't being added by a user that I'm not seeing.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
patflynn: I just tried out creating a new plain Java module, and then adding GAE as new library support, and it seemed to run this code.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacet.java
joaoandremartins: There are two ways of adding a facet to a project, after it's created: Project Settings, Facets and add there or the Configure popup.

We control the Facet UI (AppEngineFacetEditor), so we're using onFacetInitialized to track facets added through Project Settings.

The only place we're able to detect facets added through the Configure popup is in the constructor. However, a constructor ping would also capture all other ways of a facet being created, including invalid ones (e.g., when the IDE is started), which is why I'm tracking this here.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacetEditor.java
patflynn: There are a few classes like the
com.intellij.facet.autodetecting.FacetDetector API that look like they
could be interesting.

There's also com.intellij.facet.ProjectWideFacetListener and
com.intellij.facet.FacetManager.Listener

no idea if they are what we're looking for but might be worth checking out.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
```diff
@@ -300,6 +302,11 @@ public void addSupport(@NotNull Module module,
       AppEngineSupportProvider.this
           .addSupport(module, rootModel, myFrameworkSupportModel,
               PersistenceApiComboboxUtil.getSelectedApi(myPersistenceApiComboBox));
+
+      // Called on New Project or New Module dialogs.
```
patflynn: need to change this comment.

patflynn:
is there a hook we can use to track new project wizard invocations that use app engine?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
joaoandremartins: I added it when you showed me this code was also triggered on New Module. I don't think it's activated in other scenarios, anything I'm missing?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacetEditor.java
joaoandremartins: Well spotted! FacetDetector is deprecated, but we're actually already extending FacetBasedFrameworkDetector, which is the new recommended way. I didn't realise, but we can use the FacetBasedFrameworkDetector.setupFacet() method to track facets added through the Configure popup.

As for ProjectWideFacetListener, I can't figure out how we can use it. It's being used by IJ on a UI component, but I can't really decide on the best place to add it. In any case, since we already have all the metrics we need, I suggest we don't use this, unless you have any suggestions.

joaoandremartins:
The one hook I was able to find is AppEngineSupportProvider.addSupport(), which is where I added the current new project/module tracking.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
patflynn: It's called on new app engine project/module and also when adding framework support to an existing module.

joaoandremartins: By adding framework support, do you mean going to Project Settings and adding the GAE facet there? Or clicking the Framework discovered Configure link?

Doing this doesn't trigger addSupport() for me. Let me know if there's another way I'm overlooking.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFrameworkDetector.java
```diff
@@ -40,6 +43,14 @@ public AppEngineFrameworkDetector() {
   }
 
   @Override
+  public void setupFacet(@NotNull AppEngineFacet facet, ModifiableRootModel model) {
+    UsageTrackerProvider.getInstance()
+        .trackEvent(GctTracking.APP_ENGINE_ADD_FACET)
```
etanshaul: shouldn't this be ADD_SUPPORT, not ADD_FACET?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
```diff
@@ -307,6 +309,11 @@ public void addSupport(@NotNull Module module,
...
+      UsageTrackerProvider.getInstance()
+          .trackEvent(GctTracking.APP_ENGINE_ADD_SUPPORT)
```
etanshaul: Shouldn't this be APP_ENGINE_STANDARD_NEW_PROJECT_WIZARD (or something more concise)? I was under the impression that this is called on new projects, but I may be confusing things now.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
patflynn: go to a module root in the left nav, right click on it, and select 'Add
Framework Support..'

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
joaoandremartins: This is called on new projects, new modules and Add Framework Support, which is why I opted for a more generic name.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFrameworkDetector.java
joaoandremartins: This is when you add the facet to a module through the Configure popup button.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
joaoandremartins: Changed.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
```diff
@@ -310,7 +310,7 @@ public void addSupport(@NotNull Module module,
           .addSupport(module, rootModel, myFrameworkSupportModel,
               PersistenceApiComboboxUtil.getSelectedApi(myPersistenceApiComboBox));
 
-      // Called on New Project or New Module dialogs.
+      // Called on New Project, New Module or Add Framework Support dialogs.
```
patflynn: how about: 'Called when creating a new App Engine module from the 'new project' or 'new module' wizards or upon adding App Engine 'Framework Support' to an existing module'

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
```diff
@@ -21,18 +21,28 @@
...
 /**
  * Represents an App Engine Standard run task. (i.e., devappserver)
  */
 public class AppEngineStandardRunTask extends AppEngineTask {
 
   private RunConfiguration runConfig;
+  private String runnerId;
 
-  public AppEngineStandardRunTask(@NotNull RunConfiguration runConfig) {
+  /**
```
patflynn: You'll probalby get a lint error here.  Google Java style is that you can leave out a javadoc comment completely, or ommit the @param/@return structure comments if they're redundant, but if you have a Javadoc for a method or class it must include a description.

example:

`void obviousMethod(){}  //this is ok (no javadoc above)`

```
/**
 * Does stuff.
**/
void nonObviousMethod(int obviousParam) {};  // OK
```
```
/**
  * Does stuff
  *
  * @param nonObviousParam changes something in the method invocation
**/
void nonObviousMethod(int nonObviousParam){}; //OK
```

```
/**
  * @param nonObviousParam changes something in the method invocation
**/
void obviousMethod(int nonObviousParam){}; //NOT OK. If you have a Javadoc you must have a summary comment.
```

joaoandremartins: I followed Etan's suggestion to remove the unnecessary descriptive comment, but in that case I'll re-add it, because I want to specify what runnerId is.

joaoandremartins:
@etanshaul @patflynn I'd like to submit today if possible. I think all comments were address, but let me know if there's anything I've missed.

patflynn:
:ship:

PR #854, opened 2016-08-31
Fix incorrect Analytics event

chanseokoh:
There was a mistake when refactoring strings in commit 248a3a2c361df80cc7159740e57d6d9d41654a60.

### DIFF
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/stats/LoginTracking.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/stats/LoginTracking.java
@@ -27,4 +27,5 @@ public class LoginTracking {
   public static final String LOGIN_START = "user.login.start";
   public static final String LOGIN_CANCELLED = "user.login.cancelled";
   public static final String LOGIN_COMPLETE = "user.login.complete";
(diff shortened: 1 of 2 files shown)


google-account-plugin/src/com/google/cloud/tools/intellij/stats/LoginTracking.java
```diff
@@ -27,4 +27,5 @@ private LoginTracking() {
   public static final String LOGIN_START = "user.login.start";
   public static final String LOGIN_CANCELLED = "user.login.cancelled";
   public static final String LOGIN_COMPLETE = "user.login.complete";
```
etanshaul: just to confirm - is LOGIN_COMPLETE still being called from the correct place?

chanseokoh: Yes, in the same file, [here](https://github.com/GoogleCloudPlatform/gcloud-intellij/blob/master/google-account-plugin/src/com/google/cloud/tools/intellij/login/IntellijGoogleLoginService.java#L481).

etanshaul:
lgtm

PR #857, opened 2016-08-31
Replaces File for Path in several locations. Removes CloudSdkUtil.

### LINKED ISSUE #856: Use Paths instead of Files
We're making heavy usage of Files instead of the Java 7 recommended Paths.

joaoandremartins:
fixes #856


joaoandremartins:
@etanshaul @patflynn @aslo 

- Swaps File for Path in many locations,
- Removes CloudSdkUtil and two other unused files.

patflynn:
@joaoandremartins use the assignees field to indicate reviewers. It supports multiple reviewers now.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
```diff
@@ -86,16 +89,18 @@ public boolean isModified() {
...
-          GctBundle.message("appengine.cloudsdk.location.invalid.message"));
-    } else {
+    try {
+      new CloudSdk.Builder()
+          .sdkPath(Paths.get(cloudSdkPanel.getCloudSdkDirectory()))
+          .build()
+          .validate();
       CloudSdkService.getInstance().setSdkHomePath(cloudSdkPanel.getCloudSdkDirectory());
+    } catch (AppEngineException aee) {
+      throw new RuntimeConfigurationError(aee.getMessage());
```
patflynn: I guess this shows up as a validation message in the UI?  We should be using a localizable string from our resource bundle for that.   Can we really know what the error condition is with a generic AppEngineException?

joaoandremartins: I thought about this too. We'd need to change AppEngineException to contain some sort of status code to present one of our messages. I can go ahead and implement this change in the common library.

patflynn: I'd prefer a more specific exception vs a status code.

joaoandremartins:
@patflynn so it appears we're moving to a validate-per-SDK strategy in https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/209 and there are plans for a validate-per-action later on, which completely changes the way validate works as we discussed it last week.

In that case, since I think this was the only pending item, I suggest PTAL.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -267,12 +266,8 @@ private File doStageCredentials(String googleUsername) {
     );
     String jsonCredential = new Gson().toJson(credentialMap);
     try {
-      credentialsPath = FileUtil
-          .createTempFile(
-              "tmp_google_application_default_credential",
-              "json",
-              true /* deleteOnExit */);
-      Files.write(jsonCredential, credentialsPath, Charset.forName("UTF-8"));
+      credentialsPath = Files.createTempFile("tmp_google_application_default_credential", "json");
```
patflynn: why did we remove the deleteOnExit param?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -88,15 +80,22 @@ private void checkSdk() {
...
+
       cloudSdkDirectoryField.getTextField().setForeground(JBColor.black);
       warningMessage.setVisible(false);
-    } else {
+    } catch (AppEngineException aee) {
       cloudSdkDirectoryField.getTextField().setForeground(JBColor.red);
       warningMessage.setVisible(true);
-      warningMessage.setText(
-          GctBundle.message("appengine.cloudsdk.location.invalid.message"));
+      if (aee instanceof BadSdkLocationException) {
```
patflynn: did these exception types end up getting merged?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/settings/ImportSettings.java
```diff
@@ -1,185 +0,0 @@
...
-import java.util.List;
-import java.util.Set;
-import java.util.zip.ZipEntry;
-import java.util.zip.ZipException;
-import java.util.zip.ZipFile;
-
-/**
- * Provides functionality to update IDEA setting from a jar containing new settings.
- */
-public class ImportSettings {
```
patflynn: what's going on here?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
joaoandremartins: I think the default for temp files is to delete on exit. (i.e., a behaviour similar to File.deleteOnExit())
I just tested a deployment and the credentials file is deleted after the deployment thread is done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/settings/ImportSettings.java
joaoandremartins: I noticed this code is totally unused and decided to delete it as part of this cleanup.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
patflynn: FileUtil:337 appears to pass 'false' as the default for deleteOnExit

joaoandremartins:
@patflynn @etanshaul PTAL

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineFlexibleStage.java
```diff
@@ -57,38 +55,31 @@ public AppEngineFlexibleStage(
...
-  public void stage(@NotNull File stagingDirectory) {
+  public void stage(@NotNull Path stagingDirectory) {
     try {
-      File stagedArtifactPath =
-          copyFile(
-              stagingDirectory,
-              "target" + AppEngineFlexDeploymentArtifactType.typeForPath(deploymentArtifactPath),
-              deploymentArtifactPath);
-      stagedArtifactPath.setReadable(true /* readable */, false /* ownerOnly */);
+      Path stagedArtifactPath = stagingDirectory.resolve(
```
patflynn: you'll need to test flex deployment with this change.

patflynn:
LGTM assuming @etanshaul is good with it and you've tested flex deployment and verified the app comes up after deployed.

etanshaul:
reviewing the latest now

joaoandremartins:
Yep, flex deployment tested and seems to be working fine.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerEditor.java
```diff
@@ -59,31 +63,42 @@ protected void textChanged(DocumentEvent event) {
...
-      warningMessage.setVisible(true);
-      warningMessage.setForeground(JBColor.RED);
-      warningMessage.setText(GctBundle.getString("appengine.cloudsdk.location.missing.message"));
-    } else if (!CloudSdkUtil.containsCloudSdkExecutable(mySdkHomeField.getText())) {
+    CloudSdk sdk = new CloudSdk.Builder()
+        .sdkPath(Paths.get(sdkHomeField.getText()))
+        .build();
+
+    try {
+      sdk.validateAppEngineJavaComponents();
```
etanshaul: I don't think this is needed with the latest common-lib version. See `AppEngineHelper#createSdk` - the validation happens internally.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
```diff
@@ -86,16 +89,19 @@ public boolean isModified() {
...
    */
   @Override
   public void apply() throws ConfigurationException {
-    if (StringUtil.isEmpty(cloudSdkPanel.getCloudSdkDirectory())
-        || !CloudSdkUtil.containsCloudSdkExecutable(cloudSdkPanel.getCloudSdkDirectory())) {
+    try {
+      new CloudSdk.Builder()
+          .sdkPath(Paths.get(cloudSdkPanel.getCloudSdkDirectory()))
+          .build()
+          .validateCloudSdk();
```
etanshaul: similar to my other comment - doesn't this happen internally now?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerEditor.java
joaoandremartins: Yes, but we're interested in failing early. In other words, letting the customer know they chose a valid or invalid SDK.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
```diff
@@ -86,16 +89,19 @@ public boolean isModified() {
...
   /**
-   * User's shouldn't be able to save a cloud configuration without a valid Cloud SDK configured.
+   * Users shouldn't be able to save a cloud configuration without a valid Cloud SDK configured.
    */
   @Override
   public void apply() throws ConfigurationException {
-    if (StringUtil.isEmpty(cloudSdkPanel.getCloudSdkDirectory())
-        || !CloudSdkUtil.containsCloudSdkExecutable(cloudSdkPanel.getCloudSdkDirectory())) {
+    try {
+      new CloudSdk.Builder()
```
etanshaul: I'm not sure about instantiating a new CloudSdk just for the purposes of validating it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
etanshaul: sorry I was wrong here - it happens when the user attempts to execute gcloud or appcfg, not on creating a new SDK like I said. Still it feels wrong to create a new one just for validation purposes

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
joaoandremartins: Yes, I discussed this with Patrick two weeks ago. If we're moving to a strategy of validating the Cloud SDK per action/method, then we'll probably not validate here anymore. This needs some more thought, but I don't think fixing this should be in the scope of this PR.

etanshaul: I think we still need early validation though - when creating a new cloud we want to warn the user if the sdk is invalid - this happens before any action is invoked.

joaoandremartins: Exactly, which is why I think this needs some more thought.

For the moment, instantiating a CloudSdk object isn't that expensive or critical, so I think we can open a new issue and deal with it later.

etanshaul: fair enough. I'm ok with it - especially since you were able to get rid of that entire utility class that was checking the gcloud path.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -282,9 +282,11 @@ private File doStageCredentials(String googleUsername) {
 
   @Override
   public void deleteCredentials() {
-    if (credentialsPath != null && credentialsPath.exists()) {
-      if (!credentialsPath.delete()) {
-        logger.warn("failed to delete credential file expected at " + credentialsPath.getPath());
+    if (credentialsPath != null && Files.exists(credentialsPath)) {
+      try {
+        Files.delete(credentialsPath);
+      } catch (IOException ioe) {
```
etanshaul: won't this throw other types of exceptions too? You are covered on `NoSuchFileException` because of of the `exists` check, but what about all the others?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -46,20 +46,10 @@
...
   private JPanel cloudSdkPanel;
 
-  public CloudSdkPanel(@NotNull CloudSdkService cloudSdkService) {
+  public CloudSdkPanel() {
     warningMessage.setVisible(false);
     warningMessage.setIcon(RunConfigurations.ConfigurationWarning);
 
-    if (cloudSdkService.getSdkHomePath() == null) {
-      final String cloudSdkDirectoryPath
-          = CloudSdkUtil.findCloudSdkDirectoryPath(SystemEnvironmentProvider.getInstance());
```
etanshaul: where is the path detection now that you got rid of this?

joaoandremartins: It happens inside the tools lib. See DefaultCloudSdkService.getSdkHomePath().

In this case, reset() is called, which checks if the CLOUD_SDK_PROPERTY_KEY is null. If it is, it uses CloudSdk.getSdkPath(), which uses the default PathResolver to find the gcloud binary.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
joaoandremartins: The only checked exception thrown by delete is IOException. What other exceptions did you have in mind?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
etanshaul: oh cool

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
etanshaul: I guess you're right. `SecurityException` is unchecked. The rest are subclasses of `IOException`.

etanshaul:
:ship:

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -35,6 +35,7 @@ appengine.promote.label=Promote the deployed version to receive all traffic
 appengine.promote.info.label={0}To manually promote a version use the {1}Google Cloud Console{2}{3}
 appengine.stop.previous.version.label=Stop previous version
 appengine.environment.label=Environment\:
+appengine.cloudsdk.devappserver.missing=Selected Cloud SDK does not contain the app-engine-java component.
```
etanshaul: Could this use the same messages I added in my recent pr: 

https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/884

etanshaul: I called it appengine.deployment.error.missing.java.components, but we can rename it to something more general.

etanshaul:
I added one more comment, but after that, lgtm.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
joaoandremartins: I thought of giving instructions on how to add the GAE SDK in my message, but thought that it would look a bit ugly in the warning text.
I'm thinking telling the user that the component is missing is a hint big enough to will let them find out they need to "gcloud components install". What do you think?

etanshaul: Oh, it comes up in the UI?

In that case maybe we can split up the message I created in my other PR into 2 parts, and you consume only the first part of it. What do you think?

joaoandremartins: appengine.cloudsdk.java.components.missing=Selected Cloud SDK does not contain the app-engine-java component.
appengine.cloudsdk.java.components.howtoinstall=To install, run 'gcloud components install app-engine-java'.

WDYT?

etanshaul: :+1:

etanshaul: don't forget the line break when concatenating them for the deployment use case

etanshaul:
@patflynn @joaoandremartins what is your opinion on including this PR in the next beta release? Since its ready I see no reason not to wait for the merge.

PR #860, opened 2016-09-02
Fixes to App Engine product naming

devlance:
This brings the plugin in line with the Google usage of the naming so there isn't a discrepancy.


patflynn:
LGTM.

it doesn't appear likely that your changes caused the errors CI is seeing. Will file an issue.

etanshaul:
This PR actually causes legitimate unit test failures in `AppEngineDeploymentRunConfigurationEditorTest`.

The memory leak errors are a side effect we can investigate separately.

patflynn:
aha. Yeah I see it now.  @devlance there's a label message check in the test.

Thanks Etan.

devlance:
Updated the unit tests.

patflynn:
Thanks Brett!

PR #868, opened 2016-09-06
Expose promote and stop-previous-version deployment flags

### LINKED ISSUE #530: should we have a flags field in mvm deployment run configs?

etanshaul:
fixes #530 

Still need to decide on default values (currently I'm using true for both promote and set-previous-version):

image


etanshaul:
A note on the behavior:

For flex deployments (where versions can be stopped), if there is a currently served version, and a new version is deployed with: 
  `promote=false`
  `stopPreviousVersion=true`

the `stopPreviousVersion` flag is ignored and the previous version is not stopped. There is no error or warning.

patflynn:
It sounds like we can improve the UX of this in our UI.

1) stopPreviousVersion should not be visible/settable for standard
deployments.
2) stopPreviousVersion should only be settable if promote is set to true
for Flex deployments.

etanshaul:
Yes. Good points. I will incorporate this.

etanshaul:
@patflynn I incorporated your ideas:

App Engine standard app - stopPreviousVersion is hidden:
image

AE flex app with promote selected - stopPreviousVersion is visible and enabled:
image

AE flex app with promote unselected - stopPreviousVersion is visible but disabled:
image

todo for next commit: don't send stopPrevoiusVersion field to deployment for AE standard apps or when promote is not selected

patflynn:
shouldn't 'stop previous version' be disabled and unchecked when 'promote'
is unchecked?

Also is there any way to make the 'stop previous version' a member of a
subgroup of the promote widget?  i.e. visibly owned by 'promote'?

etanshaul:
I originally didn't want to modify the user's selection based on what promote is (i.e. unchecking stopPreviousVersion) because the field is not applicable (as opposed to false). But I'm open to this.

Regarding making stopPreviousVersion a subgroup - good idea, I'll do that.

patflynn:
ah I see what you're saying. So by default it's unchecked, but if someone
manages to check it, and then unchecks 'promote' then the
'stopPreviousVersion' remains checked.

That makes sense from one standpoint: i.e. remembering the setting of
stopPreviousVersion when promote is checked,  however it could confuse
users to thinking that when promote is not checked that
'stopPreviousVersion' is enabled and will be done at the end of deploy. I
think that's problematic.

etanshaul:
Actually, by default, both promote _and_ stopPreviousVersion are checked (as it currently stands).

I agree with your point in the 2nd paragraph. However one thing to remember is that when promote is not checked, then stopPreviousVersion is ignored and is not sent to the deployment at all (will be implemented soon). 

But yea, I think its still better for it to appear like stopPreviousVersion is false in this case to not confuse the user - although personally when I see disabled fields I don't tend to consider what their values are (to me its implied they are ignored).

patflynn:
I think with the field as a sibling of 'promote' that's definitely not how
I would see it, but perhaps as a child then it'd be ok.

etanshaul:
I'll make it a child (visually) and also uncheck stopPreviousVersion when promote is unchecked - I don't think this has any downsides other than forgetting the user's previous promote selection (and even this concern is questionable).

etanshaul:
@patflynn made some updates:

- Moved the checkboxes to the left of the label to look like a traditional checkbox field
- Made stopPreviousVersion a visual child of promote
- When promote is unchecked, then stopPreviousVersion is automatically unchecked
- Only sending stopPreviousVersion to the deployment if promote is selected and the environment is AE flex

image

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -481,4 +497,18 @@ public void itemStateChanged(ItemEvent itemEvent) {
       resetOverridableFields(overrideCheckbox, field);
     }
   }
+
+  private class PromoteListener implements ItemListener {
```
patflynn: it'd be nice to test the view specific logic around this.

etanshaul: yes, I'll add tests

etanshaul: done - tests added

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditorTest.java
```diff
@@ -108,6 +110,46 @@ public void testUiAppEngineFlexEnvironment() {
...
+    assertTrue(stopPreviousVersionCheckbox.isVisible());
+    assertTrue(stopPreviousVersionCheckbox.isEnabled());
+
+    // Disable the promote checkbox and test that stopPreviousVersion behaves correctly
+    promoteCheckbox.setSelected(false);
+
+    assertFalse(stopPreviousVersionCheckbox.isSelected());
+    assertFalse(stopPreviousVersionCheckbox.isEnabled());
+
+    Disposer.dispose(editor);
```
patflynn: should this go in the teardown?

etanshaul: editor is defined within the method so that it can be instantiated differently per test

patflynn: ah. I should have noticed that.

LGTM

PR #870, opened 2016-09-06
Rename com.google.cloud.tools.intellij.git package to com.google.cloud.tools.intellij.vcs

patflynn:
'git' is not the higher order concept in the source repo integration.


google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -55,8 +55,8 @@ Code inspections for App Engine Java code.</description>
 
     <implicitUsageProvider implementation="com.google.cloud.tools.intellij.appengine.validation.EndpointImplicitUsageProvider"/>
 
-    <checkoutProvider implementation="com.google.cloud.tools.intellij.git.GcpCheckoutProvider"/>
-    <vcsPopupProvider implementation="com.google.cloud.tools.intellij.git.GcpQuickListContentProvider"/>
+    <checkoutProvider implementation="com.google.cloud.tools.intellij.vcs.GcpCheckoutProvider"/>
```
etanshaul: Looking at GcpCheckoutProvider, I see a dependency on the `git4idea` package (e.g. `git4idea.GitVcs`). Is this name changed justified still?

patflynn:
@etanshaul yeah.  This package is for Google Cloud Source Repository related integration.  CSR currently is git-only, but that's not intrinsic to the service.  I would really like to call the package 'sourcerepository' but that seems too long. Another option is 'source', but I was concerned that it could be misconstrued.  I'm open to either option though.

etanshaul:
Ok cool thanks for explaining. I'm fine with vcs. LGTM.

PR #871, opened 2016-09-07
Elysium is an old name.  Rename package to reflect new naming.

patflynn:
The package for now represents the set of GCP specific domain resources,
so I'm renaming to resources.

The API for managing them is now called Resource Manager.


etanshaul:
:ship:

Issue #861, opened 2016-09-02
Seeing transient memory leaks during tests on Travis


patflynn:
com.google.cloud.tools.intellij.appengine.cloud.AppEngineDeploymentRunConfigurationEditorTest > testUiAppEngineFlexEnvironment FAILED
    junit.framework.ComparisonFailure at AppEngineDeploymentRunConfigurationEditorTest.java:106
com.google.cloud.tools.intellij.appengine.cloud.AppEngineDeploymentRunConfigurationEditorTest > testUiAppEngineStandardEnvironment FAILED
    junit.framework.ComparisonFailure at AppEngineDeploymentRunConfigurationEditorTest.java:94
com.google.cloud.tools.intellij.debugger.CloudExecutionStackTest > testGetTopFrame_nullFramesReturnsNull SKIPPED
com.google.cloud.tools.intellij.debugger.CloudExecutionStackTest > testGetTopFrame_nonNull SKIPPED
ERROR: Memory leak detected: com.google.cloud.tools.intellij.appengine.cloud.AppEngineDeploymentRunConfigurationEditor@4ed59f6e of class class com.google.cloud.tools.intellij.appengine.cloud.AppEngineDeploymentRunConfigurationEditor
See the cause for the corresponding Disposer.register() stacktrace:
java.lang.RuntimeException: Memory leak detected: com.google.cloud.tools.intellij.appengine.cloud.AppEngineDeploymentRunConfigurationEditor@4ed59f6e of class class com.google.cloud.tools.intellij.appengine.cloud.AppEngineDeploymentRunConfigurationEditor
See the cause for the corresponding Disposer.register() stacktrace:
	at com.intellij.openapi.util.objectTree.ObjectTree.assertIsEmpty(ObjectTree.java:209)
	at com.intellij.openapi.util.Disposer.assertIsEmpty(Disposer.java:124)
	at com.intellij.openapi.util.Disposer.assertIsEmpty(Disposer.java:120)
	at com.intellij.openapi.application.impl.ApplicationImpl.disposeSelf(ApplicationImpl.java:308)
	at com.intellij.openapi.application.impl.ApplicationImpl.access$400(ApplicationImpl.java:91)
	at com.intellij.openapi.application.impl.ApplicationImpl$5$1.run(ApplicationImpl.java:268)
	at com.intellij.openapi.util.ShutDownTracker$2.run(ShutDownTracker.java:172)
[...]

PR #874, opened 2016-09-07
subclass teardown behavior should occur prior to calling the superclass

patflynn:
I suspect this will fix the memory issue exception

possibly fixes: #861


etanshaul:
This doesn't make this pr invalid, but I did try this and the memory leak error did not go away from what I can remember. To reproduce, keep this change, and reintroduce the unit test failure from that other PR (by changing the localized variable test).

etanshaul:
lgtm because I think this is correct regardless of my last poitn

patflynn:
ah I see.  This exception happens regardless of whether the test passes or fails.

etanshaul:
I only see the leak on test failure. I'm running the test like this:

`./gradlew -Dtest.single=AppEngineDeploymentRunConfigurationEditorTest :google-cloud-tools-plugin:test`

patflynn:
I'm running from the IJ console and see the output on both success and
failure. Perhaps the gradle test task produces more output on failure.

PR #881, opened 2016-09-09
Add more stacktrace output to console in travis build

patflynn:


etanshaul:
Does this not add an extra check execution to the build? LGTM otherwise

etanshaul:
Oh, is the check implied before this change?

patflynn:
yeah. My understanding is that if you don't include the "script" element
it'll detect gradle and automatically call ./gradlew check.  This change
makes it explicit and adds the ''--stacktrace" flag.

etanshaul:
ok. 🚢

patflynn:
It looks like you might be right that this is related to having multiple concurrent builds:
https://discuss.gradle.org/t/gradle-check-fails-but-tests-finish-with-success/10214/5

etanshaul:
Good find. I'd hate to reduce the concurrency given how slow things are as is. But if we have no other choice we can try it.

etanshaul:
And why this tends to happen on the 'pr' build and not the 'push' build is totally perplexing to me.

patflynn:
Maybe one gets pre-empted by the other..

Issue #878, opened 2016-09-08
Investigate transient build failures


etanshaul:
We are seeing occasional build failures that correct themselves after restarting the job.

They usually occur for only one of the builds in the build matrix (one of the targeted versions of IJ). This makes me suspect that there is some issue running these builds in parallel.

patflynn:
yeah this is the error I'm seeing:
```

* What went wrong:
Execution failed for task ':google-cloud-tools-plugin:ultimate:test'.
> Process 'Gradle Test Executor 9' finished with non-zero exit value 137
```

It seems like we're not outputting enough info to see the full cause info.

PR #882, opened 2016-09-12
Temporarily disable parallel environment builds

patflynn:
We can't seem to get green builds with this configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 language: java
-matrix:
-  include:
-  - jdk: openjdk7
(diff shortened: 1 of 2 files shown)


patflynn:
FYI 

@aslo 
@joaoandremartins 

related to #878

patflynn:
I'll rerun this a few times to make sure this is fixing things.  I'm not 100% sure this is the problem.

patflynn:
I guess we know that is not the problem.

So something about the travis java 7 image with IJ 15 is causing the failure maybe.

PR #888, opened 2016-09-13
Only destroy process and track cancellation if process is alive

### LINKED ISSUE #887: Deployment cancellation metric not being tracked correctly
The deployment cancellation client-side tracking is under certain circumstances being invoked too many times, inflating the cancellation metric.

This happens because we keep track of all initiated deployments, and upon cancellation we iterate over them and cancel them all. The problem is that successful deployments are still maintained in the list of deployments causing a subsequent cancellation to get invoked too many times.

See `AppEngineRuntimeInstance#deploy` and `AppEngineRuntimeInstance#disconnect`.

etanshaul:
fixes #887 

@patflynn 
@chanseokoh (since this affects metrics)
@meltsufin (since I think you helped with the original code)

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineExecutor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineExecutor.java
@@ -45,8 +45,14 @@ public class AppEngineExecutor implements CancellableRunnable {
   @Override
   public void cancel() {
     if (process != null) {
-      process.destroy();
-      task.onCancel();
+      // Only destroy the process and signal cancellation if the process hasn't exited
+      // TODO(eshaul) replace this with Process#isAlive when switching to Java 8
+      try {
+        process.exitValue();
+      } catch (IllegalThreadStateException itse) {
+        process.destroy();
+        task.onCancel();
+      }
     }
   }
 
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineExecutor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineExecutor.java
@@ -44,16 +44,21 @@ public class AppEngineExecutor implements CancellableRunnable {
 
   @Override
   public void cancel() {
-    if (process != null) {
-      // Only destroy the process and signal cancellation if the process hasn't exited
-      // TODO(eshaul) replace this with Process#isAlive when switching to Java 8
-      try {
-        // Throws IllegalThreadStateException when the process has not yet terminated
-        process.exitValue();
-      } catch (IllegalThreadStateException itse) {
-        process.destroy();
-        task.onCancel();
-      }
+    // Only destroy the process and signal cancellation if the process hasn't exited
+    if (process != null && isAlive(process)) {
+      process.destroy();
+      task.onCancel();
+    }
+  }
+
+  // TODO(eshaul) replace this with Process#isAlive when switching to Java 8
+  private boolean isAlive(Process process) {
+    try {
+      // Throws IllegalThreadStateException when the process has not yet terminated
+      process.exitValue();
+      return false;
+    } catch (IllegalThreadStateException itse) {
+      return true;
     }
   }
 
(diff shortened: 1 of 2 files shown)


meltsufin:
> @meltsufin (since I think you helped with the original code)
Helped cause the issue that is :)

etanshaul:
@meltsufin I put it much more gently :)  I'm not actually sure when the issue surfaced though since there was a lot of refactoring here.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineExecutor.java
```diff
@@ -45,8 +45,14 @@ public void onStart(Process process) {
   @Override
   public void cancel() {
     if (process != null) {
-      process.destroy();
-      task.onCancel();
+      // Only destroy the process and signal cancellation if the process hasn't exited
+      // TODO(eshaul) replace this with Process#isAlive when switching to Java 8
+      try {
+        process.exitValue();
```
meltsufin: need a comment on this line saying that process.exitValue() throws IllegalThreadStateException when the process has not exited

meltsufin:
LGTM, other than documentation comment

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineExecutor.java
etanshaul: yea, good point. I'll add it

chanseokoh:
LGTM about metrics.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineExecutor.java
```diff
@@ -45,8 +45,15 @@ public void onStart(Process process) {
   @Override
   public void cancel() {
     if (process != null) {
-      process.destroy();
-      task.onCancel();
+      // Only destroy the process and signal cancellation if the process hasn't exited
```
patflynn: is this to deal with when the user cancels after the task is complete?

etanshaul: its to ensure that cancellations are only tracked for non-terminated processes

etanshaul: not after the task is complete - but to make sure that a cancellation of an in progress task is not counted multiple times

etanshaul:
@paflynn are are ok with this change? I can merge it in if so and then begin the release process. I'd like this fix to go in for the release.

chanseokoh:
@etanshaul Don't forget to include https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/854 when you make a release.

etanshaul:
This next release will have everything in master so yes it will be included. Thanks.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineExecutor.java
patflynn: this method cancel() is called by the UI I guess?  Why would it be invoked multiple times?

etanshaul: Check out the info on the issue. We can go over it in more detail in person
too if that's not enough. I'll be in soon.

etanshaul: I missed some details in your question - yes its called by the UI when the user hits the stop button. It would be invoked multiple times because we keep track of all deployments and don't clear out successful ones from the list. The success happens asynchronously so it would be difficult to do this - I think its safer just to attempt to cancel them all (skipping the processes that are already terminated).

etanshaul: For more detail - we keep track of all the deployments for cancellation because the cancellation happens at the "cloud" level - each cloud can have multiple simultaneous running deployments so we need a handle on them all when the user hits the stop button.

PR #909, opened 2016-09-15
Fixes issues with Cloud SDK path

joaoandremartins:
1) Forces user to specify valid SDK path in "Clouds" if an invalid one was persisted in "Other Settings". Before, if the path wasn't modified, user could still specify that path.
2) In 2016.2, ConnectionCallback.errorOccurred() seems to have stopped working. Adding an explicit notification that alerts for the error (Invalid Cloud SDK).


google-cloud-tools-plugin/resources/config.properties
```diff
@@ -15,4 +15,7 @@
 #
 
 # Place cloud tools plugin configuration here.
-feature.appengine.flex=true
\ No newline at end of file
+feature.appengine.flex=true
+
+# Should be used in any Notifications explicitly produced by the plugin.
+notifications.plugin.groupdisplayid=Google App Engine plugin
```
etanshaul: can you explain what is going on here? I'm not sure I understand this or what this file is for?

joaoandremartins: Notifications requires a groupDisplayId String to group notifications for. That String will appear in Settings -> Notifications so you can mute/unmute/etc. notifications of that kind. I'm thinking that all notifications from the plugin should have the same groupDisplayId, and config.properties is the place where this config seemed to make more sense. i.e., there's a "# Place cloud tools plugin configuration here." comment right at the beginning of it.

Feel free to suggest a better place.

etanshaul: Looks like all other usages of Notification just use the appropriate message bundle.

etanshaul:
lgtm

joaoandremartins:
FWIW, Jenkins build fails

hudson.security.AccessDeniedException2: joaomartins@google.com is missing the Overall/Read permission

etanshaul:
Looks like it fails due to a PMD violation

patflynn:
Hey Joao, what's the url you're seeing that AccessDeniedException on?

joaoandremartins:
https://cloud-tools-for-java-testing.appspot.com/job/google-cloud-intellij%20CI%20Build/58/

If I follow Details of google-cloud-intellij CI Build

PR #914, opened 2016-09-16
Fix jenkins test issue

patflynn:
Jenkins builds failed when I moved the workspaces to an SSD. 

Failure was here:
https://cloud-tools-for-java-testing.appspot.com/job/google-cloud-intellij%20CI%20Build%20(IDEA%2015)/75/testReport/junit/com.google.cloud.tools.intellij.appengine.cloud/DefaultAppEngineProjectServiceTest/testGetAppEngineArtifactEnvironment_Standard/

No idea what the reason it would start failing after switch to SSD. Maybe the fact I'm using symbolic links to the workspace is causing problems. Either that or```
 LocalFileSystem.getInstance().findFileByIoFile(
        new File(path));
``` returns a different appengine-web.xml that's not in the content root because.. who knows.


etanshaul:
lgtm assuming this fixes the jenkins build (looks like it does)

PR #915, opened 2016-09-16
Take module deployment sources into account when checking if a project is flex-compat

### LINKED ISSUE #907: Configuration deployment option appears for Compat Maven artifacts
maven_yesconf
explodedwar_noconf

@etanshaul

etanshaul:
Fixes #907 

Still verifying.


etanshaul:
More info:

Now that we can deploy to app engine standard (and flex compat) using maven - since https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/851 - the flex compat determination logic can no longer assume only artifact deployment sources: maven deployment sources _are_ module deployment sources. This PR takes module deployment sources into account too when checking if a source is flex compat.

PR #916, opened 2016-09-19
Configure app engine standard libs and pull them from maven central

### LINKED ISSUE #866: Add configurable libraries to the App Engine Standard facet configuration
Per the document on Java App Engine Dependency Management in IntelliJ and Eclipse.

https://docs.google.com/document/d/1UEqYEWwbgNK8EqPuIUvXTtOfFbV-Oa7cmKxtrU2ARzE/edit#

etanshaul:
fixes #866 

Still a couple things left:
- Add remaining libs - currently only servlet-api
- Add / remove libs via the facet editor

Flow:

User can select libraries in the new project wizard:
image

User is promted to configure the selected lib to be downloaded from maven central:
image

Library is on the classpath and visible in the library settings:
image
image

User can see and configure the libs from the facet editor:
image


etanshaul:
Also, after speak with @akerekes, we think it could be beneficial for maven projects to add the dependency to the POM rather than "manually" to the classpath like this (giving the user the option). This can be a separate enhancement though.

patflynn:
Yeah for Maven/Gradle projects the build files are the source of truth, and
we should never make any native modifications to the classpath. Supporting
those projects can be a separate enhancement like you said, but I think
what you have here for native projects is good.

etanshaul:
Remaining todos:

- How does this functionality interact with maven/gradle projects that may already include these dependencies? In this PR we are not modifying the build files, but we still would need a way to detect if the build systems already include these dependencies and handle it appropriately

- Implement facet functionality - checking / unchecking the libs from the facet editor should add / remove the library

patflynn:
Can we hide this UI from Maven and Gradle based projects for now?

etanshaul:
Yes. That sounds a reasonable solution for this iteration.

etanshaul:
I realized that this solution needs to be modified slightly: rather than persisting and loading the facet lib selections via IntelliJ's property mechanism, they need to be synced with the users actual configured libs.

For example, regardless of the user's library selection within the facet, the user can always go into the Library configuration (within the Module config) and manually remove a library. The facet needs to be aware of and synchronized with this.

patflynn:
Yeah that makes a lot of sense.  I was going to review this this afternoon, is it ready for that or should I hold off?

etanshaul:
Please hold off. There are a few tricky things related to this syncing that I'm trying to figure out. I'll put out a comment here when its ready.

etanshaul:
Syncing between the projects libs and the facet lib UI is implemented.

Remaining todos:
- Get library addition/removal via the facet UI working (not new project wizard which is working already).
- Set the scope on the loaded libraries

etanshaul:

You can start reviewing this. I still have a couple todo's that I'm working on:

- Adding new libs via the facet editor throws an exception. Something is not being disposed properly.
- Need to suppress this functionality for Maven/Gradle projects to prevent native additions to the classpath that could conflict with the build system.

patflynn:
looks like you have some checkstyle issues.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacetEditor.java
```diff
@@ -1,91 +0,0 @@
...
-import org.jetbrains.annotations.Nls;
-import org.jetbrains.annotations.NotNull;
-
-import javax.swing.JComponent;
-import javax.swing.JPanel;
-
-/**
- * @author nik
- */
-public class AppEngineFacetEditor extends FacetEditorTab {
```
patflynn: Do you know why this is showing up as a delete and creation instead of a move from AppEngineFacetEditor to AppengineStandardFacetEditor? Is it a github issue?

etanshaul: I'm not totally sure. I did just do a refactor / rename in the ide.

patflynn:
strange, cause it did the right thing for the form file.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacetConfiguration.java
```diff
@@ -50,17 +58,33 @@ public void readExternal(Element element) throws InvalidDataException {
   public void writeExternal(Element element) throws WriteExternalException {
   }
 
+  public Set<AppEngineStandardMavenLibrary> getLibraries(@NotNull Project project) {
```
joaoandremartins: This method could use some Javadoc. Seems like its converting libraries from the platform representation to ours?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  @Override
+  public String getHelpTopic() {
+    return "Google_App_Engine_Facet";
+  }
+
+  @Override
+  public void onFacetInitialized(@NotNull Facet facet) {
+    AppEngineWebIntegration.getInstance().setupDevServer();
```
joaoandremartins: Why is this required here?

etanshaul: this is not new code (I'm not sure why github didn't catch the file renaming).

I believe this is for when you create a new app engine standard project and you have a local dev server set up by default for you.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacetConfiguration.java
etanshaul: agreed - I need to give the entire pr a look over with regards to docs.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardMavenLibrary.java
```diff
@@ -0,0 +1,117 @@
...
+  public static AppEngineStandardMavenLibrary getLibrary(
+      Predicate<AppEngineStandardMavenLibrary> predicate) {
+    return Iterables.find(
+        Arrays.asList(AppEngineStandardMavenLibrary.values()),
+        predicate,
+        null /*default value*/);
+  }
+
+  /**
+   * Certain maven versions like "LATEST" are displayed differently - e.g. "Latest", so we need to
```
joaoandremartins: I thought that "latest" isn't a valid Maven version, at least since Maven 3?

etanshaul: Yea I wondered that too. Could be that the intellij maven plugin is using 2 internally. I'll check whats going on.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardMavenLibrary.java
```diff
@@ -0,0 +1,117 @@
...
+ * Defines the available App Engine standard maven-sources libraries.
+ */
+public enum AppEngineStandardMavenLibrary {
+  // TODO get display names from bundle (in form too)
+  SERVLET_API("Servlet API", new RepositoryLibraryProperties(
+      "javax.servlet", "servlet-api", "2.5"), DependencyScope.PROVIDED),
+  JSTL("JSP Standard Tag Library (JSTL)", new RepositoryLibraryProperties(
+      "javax.servlet", "jstl", RepositoryUtils.LatestVersionId), DependencyScope.PROVIDED),
+  APP_ENGINE_API("App Engine API", new RepositoryLibraryProperties(
+      "com.google.appengine", "appengine-api-1.0-sdk", RepositoryUtils.LatestVersionId),
```
patflynn: apparently we should be using RepositoryUtils.ReleaseVersionId so that we don't include non GA releases.  I'm a bit concerned as I've read some reports that any non fixed version number is not supported in Maven 3.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * UI panel for configuring App Engine standard environment libraries.
+ */
+public class AppEngineStandardLibraryPanel {
+
+  private JPanel libraryPanel;
+  private JCheckBox servletApi;
+  private JCheckBox jstl;
+  private JCheckBox appEngineApi;
+  private JCheckBox endpoints;
```
joaoandremartins: On top of that, would any kind of metrics here be useful? Like how many users are using an API?

patflynn:
Yeah good idea!

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
```diff
@@ -0,0 +1,197 @@
...
+    this.context = context;
+    libraryListener = new LibraryModificationListener();
+
+    LibraryTablesRegistrar.getInstance()
+        .getLibraryTable(context.getProject()).addListener(libraryListener);
+  }
+
+  @Nls
+  public String getDisplayName() {
+    return "Google App Engine";
```
etanshaul: Ok. I think I will also rename this to Google App Engine Standard since this facet only applies to that env.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
etanshaul: @joaoandremartins metrics would be great here.

@patflynn I can definitely make it data driven. That would make a lot more sense.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
```diff
@@ -0,0 +1,93 @@
...
+  private JCheckBox objectify;
+  private List<JCheckBox> libraryGroup = Arrays.asList(servletApi, jstl, appEngineApi, endpoints,
+      objectify);
+
+  public Set<AppEngineStandardMavenLibrary> getSelectedLibraries() {
```
etanshaul: Yes, I revisit testing this PR once I overcome the couple of barriers I'm facing now with the disposable exceptions.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
```diff
@@ -0,0 +1,93 @@
...
+                    .getLibraryByDisplayName(libraryCheckbox.getText())
+                    : null;
+              }
+            }),
+        Predicates.notNull()
+    ));
+  }
+
+  public void selectLibraries(Set<AppEngineStandardMavenLibrary> libraries) {
+    Collection<String> availableLibraryNames = Collections2.transform(libraries,
```
etanshaul: What this does is select the checkboxes based on what libraries the user has configured. So it needs to be an action - maybe setSelectedLibraries is clearer?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
etanshaul: Oh I remember - the reason I didn't make it data driven at first is so that I can create the gui using the form editor. But I can programatically add them to the form instead.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacetConfiguration.java
```diff
@@ -50,17 +58,33 @@ public void readExternal(Element element) throws InvalidDataException {
...
+    LibraryTable libraryTable = LibraryTablesRegistrar.getInstance().getLibraryTable(project);
+    Library[] libraries =  libraryTable.getLibraries();
+
+    Set<AppEngineStandardMavenLibrary> configuredLibraries = Sets.newHashSet();
+    for (Library configuredLibrary : libraries) {
+      AppEngineStandardMavenLibrary mavenLibrary
+          = AppEngineStandardMavenLibrary.getLibraryByMavenDisplayName(configuredLibrary.getName());
+      if (mavenLibrary != null) {
```
etanshaul: Added a log warning if the library fails to load

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardMavenLibrary.java
etanshaul: I'll update to ReleaseVersionId.

I still need to look at how IJ handles it, but I would assume that if the API allows for non-fixed version numbers that it would handle using the appropriate version of Maven internally (or handles that functionality external to Maven).

etanshaul:
Question now that I've solved the major hurdles with this PR:

Do we want to have any of these libraries selected by default in the new project wizard? I'd expect, at a minimum, the App Engine API to be selected by default. Also, should we rearrange these somehow? Or maybe have a visual distinction between the provided dependency (Servlet API) and the rest? 

Here is how it currently looks (note they are all deselected by default):

image

joaoandremartins:
I would think that Servlet and App Engine APIs would be used by the big majority of app developers in GAE. I would leave the other ones unselected.
I'd also probably place JSTL after App Engine API, since I think it's more likely users will use App Engine APIs rather than JSP?

patflynn:
@ludoch

Actually, now that I think about it. I don't think it even makes sense to
allow the user to not select the servlet and JSP libraries since they're
provided by the runtime environment.

I hear you about the App Engine API.  I'd be tempted to have it checked by
default.

patflynn:
What's weird with the JSP thing is that they're only provided at runtime if
they're used by including a JSP... So it may make sense to expose that to
the user somehow by giving them the option of not selecting it.

etanshaul:
@patflynn what do you mean by exposing it by giving them the option of not selecting it? How is that different from just having it unselected by default like it is now?

etanshaul:
The library UI is now disabled for Maven modules. The last thing is to disable it for gradle projects as well / see if there is a general mechanism for determining if a module is backed by a build system.

Not a lot will change based on this last todo (it will be contained) so please review. It would also be helpful to test out the UI to get a feel for the experience.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  public class LibraryModificationListener implements Listener {
+
+    @Override
+    public void afterLibraryAdded(final Library addedLibrary) {
+      final DependencyScope scope = AppEngineStandardMavenLibrary
+          .getLibraryByMavenDisplayName(addedLibrary.getName()).getScope();
+
+      new WriteAction() {
```
joaoandremartins: Why is this WriteAction necessary?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
```diff
@@ -0,0 +1,121 @@
...
+
+  private boolean enabled = true;
+
+  public AppEngineStandardLibraryPanel(boolean enabled) {
+    this.enabled = enabled;
+  }
+
+  public Set<AppEngineStandardMavenLibrary> getSelectedLibraries() {
+    return Sets.newHashSet(Collections2.filter(
+        Collections2.transform(Arrays.asList(libraryPanel.getComponents()),
```
joaoandremartins: This looks a bit too complicated to just get the selected libraries.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
```diff
@@ -0,0 +1,121 @@
...
+      libraryPanel.setBorder(
+          BorderFactory.createTitledBorder(BorderFactory.createEtchedBorder(), "Libraries"));
+
+      for (AppEngineStandardMavenLibrary library : AppEngineStandardMavenLibrary.values()) {
+        final JCheckBox libraryCheckbox = new JCheckBox(library.getDisplayName());
+        libraryPanel.add(libraryCheckbox);
+
+        // The Servlet API is provided by the AE standard runtime. So we are enabling it by default.
+        if (library == AppEngineStandardMavenLibrary.SERVLET_API) {
+          libraryCheckbox.setSelected(true);
```
joaoandremartins: How did the conversation around the servlet API go? Does it make sense to allow users to disable it?

etanshaul: I was going to not allow disabling, but this made the experience for project imports confusing:

If the user has an existing project, then adds our facet (makes in an AE standard app), then the lib configuration for the servlet would have to be enabled in this case to allow them to add it. It felt cleaner just to leave it editable all the time given this.

That being said, its selected by default in the new project wizard.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
```diff
@@ -172,29 +169,64 @@ private static Artifact findOrCreateWebArtifact(AppEngineFacet appEngineFacet) {
...
+  }
+
+  static Library loadMavenLibrary(final Module module, AppEngineStandardMavenLibrary library) {
+    RepositoryLibraryProperties libraryProperties = library.getLibraryProperties();
+
+    final RepositoryWithVersionAddLibraryAction action = new RepositoryWithVersionAddLibraryAction(
+        module, RepositoryLibraryDescription.findDescription(libraryProperties),
+        libraryProperties.getVersion());
+
+    action.invoke(module.getProject(), null, null);
```
joaoandremartins: Maybe add the argument names of these nulls? e.g. /* argument */

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/RepositoryWithVersionAddLibraryAction.java
```diff
@@ -0,0 +1,83 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.jetbrains.idea.maven.utils.library.RepositoryAddLibraryAction;
+import org.jetbrains.idea.maven.utils.library.RepositoryLibraryDescription;
+import org.jetbrains.idea.maven.utils.library.RepositoryLibrarySupport;
+import org.jetbrains.idea.maven.utils.library.propertiesEditor.RepositoryLibraryPropertiesDialog;
+import org.jetbrains.idea.maven.utils.library.propertiesEditor.RepositoryLibraryPropertiesModel;
+
+/**
+ * Extends {@link RepositoryAddLibraryAction} to allow setting a custom library version in the
```
joaoandremartins: I didn't find this Javadoc very helpful.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
etanshaul: will do

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
etanshaul: modifying and committing the root model needs to be in a write action from what I've seen in the platform code. I'll double check this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
```diff
@@ -110,14 +110,21 @@ public boolean isAppEngineFlexArtifactType(@NotNull Artifact artifact) {
...
     MavenProject mavenProject = projectsManager.findProject(module);
 
-    boolean isMavenProject = projectsManager.isMavenizedModule(module)
-        && mavenProject != null;
+    return mavenProject != null
+        && projectsManager.isMavenizedModule(module);
+  }
+
+  @Override
+  public boolean isJarOrWarMavenBuild(@NotNull Project project, @NotNull Module module) {
```
joaoandremartins: You don't need the project argument here anymore, right?

joaoandremartins:
Are you considering adding more unit tests?

etanshaul:
yes. working on it. right after I add the gradle detection

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
etanshaul: no. I'll remove it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/RepositoryWithVersionAddLibraryAction.java
etanshaul: in what way? you didn't understand it or it was redundant?

patflynn:
Is there any way we can restrict it to only the 2.5 version of the servlet api?

etanshaul:
@patflynn it is restricted to 2.5. See AppEngineStandardMavenLIbrary.SERVLET_API

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
etanshaul: I prefer a functional style for simple mapping/transform operations. But it is a bit cumbersome pre-Java 8.

patflynn:
Have you tested local run configs with this change?  Not sure if it's related but I'm noticing that the local server configuration icon has a red 'x' on it, but looking at the configuration doesn't highlight any errors.  Running the config results in:
`Error running App Engine Local Server: Error running Google App Engine Dev Server. Exception occurred : java.lang.NullPointerException Exception message : null`   

screenshot from 2016-10-04 15 05 18

patflynn:
I should probably test this on master but I'm also seeing two 'app engine local server' dependencies in the module settings.

patflynn:
screenshot from 2016-10-04 15 07 40

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
etanshaul: good point. Ill check

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/RepositoryWithVersionAddLibraryAction.java
joaoandremartins: I'm not familiar with RepositoryAddLibraryAction, but I was hoping I wouldn't have to refer to it to understand what this class does. What does "the action" consist on? And what is done with that custom library version?

patflynn:
I also tried removing the appengine-api sdk dep using the facet editor and now my module shows up as broken with a missing dep.
screenshot from 2016-10-04 15 10 44

etanshaul:
@patflynn 
regarding the app-engine-api removal: were you testing this with an already existing project that had this? This dependency is obsolete and now removed. Instead it comes from maven via the managed dependencies.

patflynn:
No this was a newly created project using the wizard and the facet library
adding feature from this branch.

patflynn:
It is preselected to 2.5, but I can change it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/RepositoryWithVersionAddLibraryAction.java
etanshaul: I removed the method javadoc and updated the class doc trying to make it clearer. Let me know if it helps.

etanshaul:
@patflynn, @joaoandremartins, @aslo

I think I've addressed the comments from review as well as the following remaining items:

- bug where removing a lib via the new UI doesn't remove it from the module
- exclude the new UI from Gradle projects (in addition to Maven)

Only thing left is to look into test coverage. ptal

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -102,6 +102,12 @@ appengine.run.startupscript=Running dev_appserver.py.
 appengine.run.startupscript.name=App Engine Plugins Core library
 appengine.run.shutdownscript=Terminating dev_appserver.py.
 appengine.run.jvmflags.title=Server parameters
+appengine.library.servlet.api.name=Servlet API
+appengine.library.jstl.api.name=JSP Standard Tag Library (JSTL)
+appengine.library.app.engine.api.name=App Engine API
+appengine.library.endpoints.api.name=App Engine Endpoints
+appengine.library.objectify.api.name=Objectify
+appengine.facet.name=Google App Engine
```
patflynn: I think we should probably call it 'App Engine Standard Environment'

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
```diff
@@ -0,0 +1,197 @@
...
+  public void disposeUIResources() {
+    if (libraryListener != null) {
+      LibraryTablesRegistrar.getInstance()
+          .getLibraryTable(context.getProject()).removeListener(libraryListener);
+    }
+  }
+
+  @Override
+  public String getHelpTopic() {
+    return "Google_App_Engine_Facet";
```
patflynn: sure

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardMavenLibrary.java
```diff
@@ -0,0 +1,121 @@
...
+  public RepositoryLibraryProperties getLibraryProperties() {
+    return libraryProperties;
+  }
+
+  public DependencyScope getScope() {
+    return scope;
+  }
+
+  @Nullable
+  public static AppEngineStandardMavenLibrary getLibraryByDisplayName(final String name) {
```
patflynn: can we do the lookup by group+artifact ID instead?

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: ah yes. I intended to do this but forgot. Will update

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacetConfiguration.java
etanshaul: I added it in another location: AppEngineSupportProvider#addMavenLibraries - I think I mistook your comment for that class.

In this case, a warning log is not correct: if the mavenLibrary is null, this just means that the project configured library does not match one of our AE managed libraries and we skip it. This is a normal condition and a warning is not necessary.

patflynn: got it

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardMavenLibrary.java
etanshaul: when necessary, we do do the lookup by group+artifact+version as you suggest via the `AppEngineStandardMavenLibrary#getLibraryByMavenDisplayName`

The `getLibraryByDisplayName` method is used exclusively when its safe to just use the name: currently this is only in the library panel class where we are matching the name of the checkbox to the name of one of our managed libs.

patflynn:
one thing. I may have missed it, but I didn't see the added tracking for which libraries were 'added'

etanshaul:
Do you mean metrics tracking? I can look into this. Although I need to give
some thought to how how we want to track repeated additions: users adds,
removes, adds again etc.

patflynn:
Keep it simple for now.  Just knowing what the total selects a specific
library in the wizard or facet importer is good enough.

etanshaul:
@patflynn see https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/916/commits/1eb9180a55c602665d75bf97af8922b1676929d7 for metrics tracking of library additions.

Note the label will be the name of AppEngineStandardMavelLibrary enum, e.g. SERVLET_API.

PR #919, opened 2016-09-21
Adds another date format to parse dates from CDB.

### LINKED ISSUE #917: Breakpoint.getFinalTime() coming with two different formats from the server
e.g.,
2016-09-21T15:39:21Z
2016-09-21T15:57:04.000Z

a ParseException is being logged when the first one is returned.

The change was already acknowledged by the apis team, and the milliseconds bit is going away. In the meantime, we're interchangeably getting dates in two different formats.

joaoandremartins:
Fixes #917 
@etanshaul @patflynn @aslo

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/BreakpointUtil.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/BreakpointUtil.java
@@ -86,7 +86,6 @@ public class BreakpointUtil {
     dateString = dateString.replaceAll("Z$", "-0000");
 
     SimpleDateFormat iso8601Format = new SimpleDateFormat(ISO_8601_FORMAT);
-    iso8601Format.setLenient(true);
 
     try {
       return iso8601Format.parse(dateString);
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/BreakpointUtil.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/BreakpointUtil.java
@@ -19,6 +19,7 @@ package com.google.cloud.tools.intellij.debugger;
 import com.google.api.client.repackaged.com.google.common.base.Strings;
 import com.google.api.services.clouddebugger.v2.model.StatusMessage;
 import com.google.cloud.tools.intellij.util.GctBundle;
+import com.google.common.collect.ImmutableList;
 
 import com.intellij.openapi.diagnostic.Logger;
 
@@ -26,6 +27,7 @@ import org.jetbrains.annotations.Nullable;
 
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.util.Collection;
 import java.util.Date;
 
 /**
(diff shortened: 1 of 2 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/BreakpointUtil.java
```diff
@@ -84,11 +86,17 @@ public static Date parseDateTime(@Nullable String dateString) {
     dateString = dateString.replaceAll("Z$", "-0000");
 
     SimpleDateFormat iso8601Format = new SimpleDateFormat(ISO_8601_FORMAT);
+    iso8601Format.setLenient(true);
 
     try {
       return iso8601Format.parse(dateString);
```
patflynn: If NO_MS is the new format that's being rolled out, should we check that first?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/BreakpointUtil.java
```diff
@@ -84,11 +86,17 @@ public static Date parseDateTime(@Nullable String dateString) {
     dateString = dateString.replaceAll("Z$", "-0000");
 
     SimpleDateFormat iso8601Format = new SimpleDateFormat(ISO_8601_FORMAT);
+    iso8601Format.setLenient(true);
```
patflynn: apparently this is the default.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/BreakpointUtil.java
```diff
@@ -84,11 +86,17 @@ public static Date parseDateTime(@Nullable String dateString) {
...
 
     try {
```
etanshaul: Might be overkill but what do you think about refactoring to something like this:

```
public static final String supportedFormats = Arrays.asList(ISO_.., ISO..);
..
for (String formatString : formats) {
  try {
   ...
   return format.parse(dateString);
  } catch (..) {
   LOG.error("Failed to parse datetime %s using format %s", ..);
  }
}
```

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/BreakpointUtil.java
joaoandremartins: This shouldn't have been here, forgot to delete it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/BreakpointUtil.java
joaoandremartins: I would prefer your way if we had more than 2 formats. Realistically, we'll have two for a short time and then back to one. Which reminds me, I should also add a TODO to check on the no_ms rollout status and eventually clean up.

etanshaul: Agree I guess. I just don't really like the nested try catch but it's not a big deal.

aslo:
:+1:

etanshaul:
Aside form @patflynn's comments :ship: (although I prefer getting rid of the nested try catch's but I'll leave it up to you :)  )

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/BreakpointUtil.java
joaoandremartins: Makes sense.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/BreakpointUtil.java
```diff
@@ -85,16 +91,14 @@ public static Date parseDateTime(@Nullable String dateString) {
...
+    for (String format : FORMATS) {
       try {
-        iso8601Format = new SimpleDateFormat(ISO_8601_FORMAT_NO_MS);
-        return iso8601Format.parse(dateString);
-      } catch (ParseException pe2) {
-        LOG.error("error parsing datetime " + dateString, pe2);
+        dateFormat = new SimpleDateFormat(format);
+        return dateFormat.parse(dateString);
+      } catch (ParseException pe) {
+        LOG.warn("error parsing datetime " + dateString + " with format " + format);
```
joaoandremartins: Added log.error.

patflynn:
why did you ahve to push a commit to force a jenkins build?  Did it have a transient failure?

joaoandremartins:
Yeah Jenkins is coming up with weird errors that don't seem related to this PR.
First, a CloudSdkPanelTest error was coming up, now it's a different one..

PR #932, opened 2016-09-22
Validate app-engine-java is installed at the run/debug window.

### LINKED ISSUE #931: Local server run isn't checking for app-engine-java in gcloud
Following is logged if app-engine-java isn't installed:

4:24:53 PM Error running App Engine Local Server
           Error running Google App Engine Dev Server.
           Exception occurred : com.google.cloud.tools.appengine.cloudsdk.AppEngineJavaComponentsNotInstalledException
           Exception message : Validation Error: Java App Engine components not installed. Fix by running 'gcloud components install app-engine-java' on command-line.

joaoandremartins:
Produce user friendlier messaging if it's not.

Fixes #931

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
@@ -17,13 +17,20 @@
 package com.google.cloud.tools.intellij.appengine.cloud;
 
 import com.google.cloud.tools.appengine.api.devserver.RunConfiguration;
+import com.google.cloud.tools.appengine.cloudsdk.AppEngineJavaComponentsNotInstalledException;
 import com.google.cloud.tools.appengine.cloudsdk.CloudSdk;
 import com.google.cloud.tools.appengine.cloudsdk.CloudSdkAppEngineDevServer;
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessStartListener;
 import com.google.cloud.tools.intellij.appengine.sdk.CloudSdkService;
+import com.google.cloud.tools.intellij.flags.PropertiesFileFlagReader;
 import com.google.cloud.tools.intellij.stats.UsageTrackerProvider;
+import com.google.cloud.tools.intellij.util.GctBundle;
 import com.google.cloud.tools.intellij.util.GctTracking;
 import com.google.common.base.Strings;
+
+import com.intellij.notification.Notification;
+import com.intellij.notification.NotificationType;
+import com.intellij.notification.Notifications;
 
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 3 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
```diff
@@ -51,12 +53,19 @@ public AppEngineStandardRunTask(@NotNull RunConfiguration runConfig, @Nullable S
...
         .async(true)
-        .startListener(startListener);
+        .startListener(startListener)
+        .build();
 
-    CloudSdkAppEngineDevServer devServer = new CloudSdkAppEngineDevServer(sdkBuilder.build());
+    try {
+      sdk.validateAppEngineJavaComponents();
+    } catch (AppEngineJavaComponentsNotInstalledException ex) {
+      return;
```
aslo: What will the user see if we just return here? Will some kind of validation error be displayed to them?

joaoandremartins: In that case, the process returned by startListener to AppEngineExecutor will be null.
We're now checking for null in CloudSdkStartupPolicy. If null, we say app-engine-java is missing and provide an instruction on how to install it.

elharo:
FYI, this may soon be moot.  in eclipse we're close to just autoinstalling it when it isn't bundled

joaoandremartins:
Still a good check to not explode if, for any reason, app-engine-java isn't installed, I think.

joaoandremartins:
Can I get an approval here when you guys have time? @aslo @patflynn @etanshaul

aslo:
LGTM :shipit:

patflynn:
I don't think we need this code. The next release of Cloud SDK is going to auto-install the java component if it detects that it's trying to run a java app engine module.

patflynn:
We will need to update to the latest version of the plugins lib to benefit from that.

joaoandremartins:
I prefer to have this code. It's always good to check for a null process in the startup policy.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardRunTask.java
patflynn: why not catch this exception at the caller?  Returning null as an undocumented contract for this specific validation error doesn't seem like an obvious API.

joaoandremartins: This method actually doesn't return anything. AppEngineExecutor.getProcess() would. I could try and add some Javadoc there?

I feel like the right way to do this with exceptions would be to throw a checked one, which conflicts with the declaration of `execute()`.

PR #933, opened 2016-09-26
Fix composite object display in debugger

### LINKED ISSUE #929: Cloud Debugger: Objects in the debugger view are not inspectable
For some reason I can't see the values in the session any more:

screen shot 2016-09-22 at 12 19 25 pm

joaoandremartins:
Fixes #929

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudStackFrame.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudStackFrame.java
@@ -148,7 +148,7 @@ public class CloudStackFrame extends XStackFrame {
       this.variable = variable.getVarTableIndex() != null ? variableTable
           .get(variable.getVarTableIndex().intValue())
           : variable;
-      members = variable.getMembers();
+      members = this.variable.getMembers();
     }
 
     @Override


joaoandremartins:
Whoa sorry, this isn't right. Let me get the right files in here.

joaoandremartins:
OK, the change is right now.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudStackFrame.java
```diff
@@ -148,7 +148,7 @@ public MyValue(@NotNull Variable variable, @NotNull List<Variable> variableTable
       this.variable = variable.getVarTableIndex() != null ? variableTable
           .get(variable.getVarTableIndex().intValue())
           : variable;
-      members = variable.getMembers();
+      members = this.variable.getMembers();
```
aslo: Your fix makes sense to me. 

Do you mind explaining the couple lines above this though? What exactly is the `variableTable` and how is its representation of the variable different than the `variable` argument to this constructor? What is the purpose of lines 148-150?

joaoandremartins:
FYI, there are also some issues with the variable viewer.

screenshot from 2016-09-26 12 19 02
screenshot from 2016-09-26 12 19 13
screenshot from 2016-09-26 12 19 28
screenshot from 2016-09-26 12 19 38

So clicking the arrow in these "special" objects just makes the arrow disappear immediately. The Developers Console UI is slightly different.

I've seen this issue way long before, so I don't think it's in the scope of this PR to fix. Might file a new issue.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudStackFrame.java
joaoandremartins: My understanding is that the server compresses (or simply serializes) the variables table in different ways and, somewhere along the way, they changed the way they do it.

We're getting `arguments` and `locals` from CDB. Each of these instances contains a `varTableIndex` attribute, which is a pointer to an index of a `variableTable`, kept alongside these `variables` and `locals`.

The existing code seems to suggest that this variable information used to be stored inside the `arguments` and `locals` instances, and now it's stored alongside them. Hope this is clear!

aslo: Yep, makes sense. Thanks for the explanation.

aslo:
LGTM

PR #940, opened 2016-09-28
Fix issue with local debug

### LINKED ISSUE #928: Server returns an error when I start a local dev server in debug mode
steps to reproduce:

1) checkout this sample https://github.com/GoogleCloudPlatform/cloud-debugger-idea-sample  (probably works with other projects)
2) change the server config to open localhost:8080/hellobrowser
3) start a local debug session
4) observe breakpoint is hit
5) observe that the http response if hellobrowser is:

`bad runtime process port ['']

ERROR: transport error 202: recv failed during handshake: Connection reset by peer
ERROR: JDWP Transport dt_socket failed to initialize, TRANSPORT_INIT(510)
JDWP exit error AGENT_ERROR_TRANSPORT_INIT(197): No transports initialized [debugInit.c:750]`

aslo:
fixes #928

See also https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/473 for discussion of the same issue in the eclipse plugin. Note that this will not fix projects with multiple modules/services.

### DIFF
--- a/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineRunConfigurationEditor.java
+++ b/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineRunConfigurationEditor.java
@@ -65,7 +65,6 @@ public class AppEngineRunConfigurationEditor extends SettingsEditor<CommonModel>
   private JTextField apiPort;
   private JTextField host;
   private JComboBox logLevel;
-  private JTextField maxModuleInstances;
   private JCheckBox useMtimeFileWatcher;
   private JTextField threadsafeOverride;
   private JCheckBox allowSkippedFiles;
@@ -117,7 +116,6 @@ public class AppEngineRunConfigurationEditor extends SettingsEditor<CommonModel>
     authDomain.setText(serverModel.getAuthDomain());
     storagePath.setText(serverModel.getStoragePath());
     logLevel.setSelectedItem(serverModel.getLogLevel());
-    maxModuleInstances.setText(intToString(serverModel.getMaxModuleInstances()));
     useMtimeFileWatcher.setSelected(serverModel.getUseMtimeFileWatcher());
     threadsafeOverride.setText(serverModel.getThreadsafeOverride());
     jvmFlags.setDialogCaption(GctBundle.getString("appengine.run.jvmflags.title"));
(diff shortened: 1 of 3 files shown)


etanshaul:
does this break multi-module projects or just not fix the existing problem with them?

aslo:
It does not fix the issue that already exists. This PR only fixes the issue for single-module projects.

etanshaul:
lgtm

patflynn:
I think instances are actually configurable in our local run config aren't they? Should it be disabled for debug for now?

aslo:
@patflynn Do you mean that we should prevent users from overriding the setting? Not sure what you're referring to with 'local run config'.

joaoandremartins:
Yeah, I think what Patrick is suggesting is to disable setMaxModuleInstances so the user doesn't set it to >1 and break the fix.

https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/master/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java#L127

So this will effectively render scale testing locally impossible IIUC?

joaoandremartins:
So, after looking at the code, if we want to go with forcing maxModuleInstance to 1, I'd argue we should probably make getMaxModuleInstances() in the library always return 1 temporarily.

Alternatively, if we want to go down this PR's route, we should probably hide the Max Module Instances setting from the run/config UI. This is because the full-fledged dev_appserver fix will probably take a while to be implemented.

Although I have to wonder, how did this work successfully until now? Did we just start enabling several VMs in dev_appserver recently?

patflynn:
Sounds reasonable. Let's hide the setting for now.

aslo:
Ok, I agree. I'll hide the setting in the "Edit Run/Debug Configurations" dialog. Then in run mode, it will use whatever the default setting is (I think it's 2), and in debug mode it will use 1.

aslo:
> Although I have to wonder, how did this work successfully until now? Did we just start enabling several VMs in dev_appserver recently?

It could have something to do with the change in the jvm args that we pass to the dev_appserver. We are currently passing "server=n,suspend=y", but if we startup the jvm with "server=y,suspend=n", (which I think we did before) the first of 2 jvm processes is able to start properly and the debugger can attach to it. The second jvm process will fail to start, but it will dump an error message to the console which can be easily ignored.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineRunConfigurationEditor.java
```diff
@@ -117,7 +116,6 @@ protected void resetEditorFrom(CommonModel commonModel) {
     authDomain.setText(serverModel.getAuthDomain());
     storagePath.setText(serverModel.getStoragePath());
     logLevel.setSelectedItem(serverModel.getLogLevel());
-    maxModuleInstances.setText(intToString(serverModel.getMaxModuleInstances()));
```
patflynn: maybe include a comment as to why maxModulesInstances is not included?

aslo: 👍

aslo:
@patflynn Added some comments referencing the issue. Let me know if you think this is good to go.

joaoandremartins:
FYI, after rebase, I'm still unable to debug without this issue.

joaoandremartins:
Nvm, this was an issue with my specific local setup.

Issue #945, opened 2016-10-04
Creating a new App Engine project using the IJ wizard for Java EE projects creates a broken dev appserver config


patflynn:
Steps to reproduce:
1) file -> new project
2)  select 'Java Enterprise' in left margin panel
3) select Google App Engine in the right panel and click next.
4) observe that the local server run configuration has a red 'x' on it and fails when you run it.
5) observe that the local server is listed twice as a library in the module dependencies settings.

patflynn:
Found while testing #916.  There's some screenshots on that PR that may be of help.

joaoandremartins:
I think I know what is causing this, and it only affects Java Enterprise projects. Creating a Java project (with the GAE facet enabled) produces a correct run/debug config.

What's happening is that a template run/config is being created at a time we still don't have any artifact context.

https://github.com/GoogleCloudPlatform/google-cloud-intellij/blob/master/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/AppEngineServerConfigurationType.java#L49

Then, `setupFrameworkSupport()`, where we already have the artifact context, calls `setupRunConfiguration()`.
For reasons not completely clear to me, for a Java Enterprise project, the following line returns a list with the one template run/debug config created before, whereas for a normal Java project, that list is empty.

https://github.com/GoogleCloudPlatform/google-cloud-intellij/blob/master/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineUltimateWebIntegration.java#L113

I don't know if the template should be being returned, judging from a few comments on the IJ SDK:
https://github.com/JetBrains/intellij-community/blob/4e2d1a2e48b7951bb4af70e772873f862cd0c4d7/platform/lang-impl/src/com/intellij/execution/impl/RunManagerImpl.java#L255
https://github.com/JetBrains/intellij-community/blob/4e2d1a2e48b7951bb4af70e772873f862cd0c4d7/platform/lang-impl/src/com/intellij/execution/impl/RunManagerImpl.java#L66

From there, if the list is empty (i.e., Java project), we correctly create a new run/debug config with the artifact set. If the list isn't empty (i.e., Java Enterprise), we don't do anything and the incomplete run/debug config stays there without any artifact.

[...]

PR #948, opened 2016-10-05
Do not pre-select artifact on local GAE run/debug config UI.

joaoandremartins:
Helps on #945

### DIFF
--- a/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineRunConfigurationEditor.java
+++ b/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineRunConfigurationEditor.java
@@ -26,7 +26,6 @@ import com.intellij.openapi.options.SettingsEditor;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.Comparing;
 import com.intellij.packaging.artifacts.Artifact;
-import com.intellij.packaging.artifacts.ArtifactManager;
 import com.intellij.packaging.impl.run.BuildArtifactsBeforeRunTaskProvider;
 import com.intellij.ui.PanelWithAnchor;
 import com.intellij.ui.RawCommandLineEditor;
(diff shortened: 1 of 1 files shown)


joaoandremartins:
FYI this fix reenables the message on the UI for how to fix the config.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineRunConfigurationEditor.java
```diff
@@ -114,11 +114,6 @@ protected void resetEditorFrom(CommonModel commonModel) {
     final AppEngineServerModel serverModel = (AppEngineServerModel) commonModel.getServerModel();
     final Artifact artifact = serverModel.getArtifact();
     myArtifactComboBox.setSelectedItem(artifact);
-    if (artifact == null && myArtifactComboBox.getItemCount() == 1) {
```
joaoandremartins: Indeed... I don't think we have enough information at this point to distinguish a Java project from a Java EE one.

joaoandremartins: There is no NPE when you run the config that I know of.

getItemCount() here returns 1 both for Java and Java EE projects, because we're in the UI editor and, therefore, at least one run config must exist.

What returns 1 for Java EE projects is the list size in the following line:

https://github.com/GoogleCloudPlatform/google-cloud-intellij/blob/master/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineUltimateWebIntegration.java#L113

The list isn't empty, so we move on with the template configuration that doesn't have an artifact associated. (If the list were empty, we would associate an artifact right after)

Then, you would open the editor which, by default, automatically selects the artifact if the project only contains one artifact, which makes the error go away. However, when the UI is closed, the artifact was actually never persisted to the config, so the error comes back.

My fix is to not automatically pre-select the artifact, but let the user do it, so they will have done something for it to go away. We could also just persist the artifact when we automatically pre-select it, but then the error would have gone away without the users having done nothing out of the ordinary, which is a slightly worse experience imo..

joaoandremartins:
Github is saying I don't have write access to merge this PR

patflynn:
Try now.q

PR #949, opened 2016-10-05
Swap Cloud Debugger Icons

aslo:
fixes https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/934

Swaps icons such that cloud debugger run configurations uses the same icon as the app engine dev server and deploy configurations.

Before:
screenshot from 2016-10-05 11 10 02

After:
screenshot from 2016-10-05 11 09 14


aslo:
Ah, that makes sense. Will locate the correct icon and update.

PR #956, opened 2016-10-06
GAE framework support no longer adds local devserver library to dependencies list

### LINKED ISSUE #945: Creating a new App Engine project using the IJ wizard for Java EE projects creates a broken dev appserver config

joaoandremartins:
Only application server does that now for Java EE projects.

Fixes #945

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
@@ -148,8 +148,6 @@ public class AppEngineSupportProvider extends FrameworkSupportInModuleProvider {
       }
     }
 
-    webIntegration.addDevServerToModuleDependencies(rootModel);
-
     addMavenLibraries(librariesToAdd, module, rootModel, webArtifact);
   }
 


joaoandremartins:
In this state, `AppEngineUltimateWebIntegration.addDevServerToModuleDependencies()` will be unused code. I'm wondering if I should remove that too?
We could use possibly use it again if IJ were to check if a dependency already exists in a module before adding it again, but I'm not sure if or when that will happen..

aslo:
Just curious, where does the remaining dev server dependency get added?

joaoandremartins:
The responsible for adding duplicate libraries is https://github.com/JetBrains/intellij-community/blob/master/java/idea-ui/src/com/intellij/ide/util/newProjectWizard/AddSupportForFrameworksPanel.java#L449

For Java EE projects, `selectedFrameworks` will have size 3: "Application Server" , "Web Application", and "Google App Engine". Java projects don't have the "Application Server" bit.

We're always adding the library on our side using the Google App Engine part, through `AppEngineSupportProvider.addSupport()`. Adding the library for the Application Server is completely contained in the framework. I think it only uses `AppEngineServerHelper.getApplicationServerInfo()` to grab the dependency name and contained libraries. I couldn't spot anywhere we have any control over this, and couldn't discover any reliable hack on AppEngineSupportProvider to not add the library if it's already added. (addSupport() is called in different order on different runs, so sometimes you'd see two dependencies, sometimes one)

aslo:
Interesting, thanks!

PR #957, opened 2016-10-06
Correctly setup run configuration when a new AppEngine project is created as Java EE project (#945)

chashnikov:
When a new Java EE project is created it may create its own run configuration in AppServerSupportConfigurable, so we need to reuse it.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/maven/AppEngineFacetImporter.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/maven/AppEngineFacetImporter.java
@@ -85,7 +85,8 @@ public class AppEngineFacetImporter extends
       final String artifactName = module.getName() + ":war exploded";
       final Artifact webArtifact = modelsProvider.getModifiableArtifactModel()
           .findArtifact(artifactName);
-      AppEngineWebIntegration.getInstance().setupRunConfiguration(webArtifact, module.getProject());
+      AppEngineWebIntegration.getInstance().setupRunConfiguration(webArtifact, module.getProject(),
+          null);
     }
   }
 }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/maven/AppEngineFacetImporter.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/maven/AppEngineFacetImporter.java
@@ -86,7 +86,7 @@ public class AppEngineFacetImporter extends
       final Artifact webArtifact = modelsProvider.getModifiableArtifactModel()
           .findArtifact(artifactName);
       AppEngineWebIntegration.getInstance().setupRunConfiguration(webArtifact, module.getProject(),
-          null);
+          null /* existingConfiguration */);
     }
   }
 }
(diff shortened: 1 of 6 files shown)


patflynn:
Hi Nikolay!

Thanks for this PR.  You'll need to sign our CLA before we can merge your
commits.

I think we should have some instructions here

chashnikov:
I've just signed CLA.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/maven/AppEngineFacetImporter.java
```diff
@@ -85,7 +85,8 @@ protected void reimportFacet(IdeModifiableModelsProvider modelsProvider,
       final String artifactName = module.getName() + ":war exploded";
       final Artifact webArtifact = modelsProvider.getModifiableArtifactModel()
           .findArtifact(artifactName);
-      AppEngineWebIntegration.getInstance().setupRunConfiguration(webArtifact, module.getProject());
+      AppEngineWebIntegration.getInstance().setupRunConfiguration(webArtifact, module.getProject(),
+          null);
```
joaoandremartins: Nit: can you add /* existingConfiguration */ after null?

joaoandremartins:
On top of this, it seems like #948 will need to be rolled back. I'll also have to add a line to it, so I can do this in another PR.

Thanks a lot for this PR!

joaoandremartins:
@patflynn this LGTM. Do you want to take a glance here before merge?

patflynn:
@joaoandremartins 
@chashnikov 

You guys have a better understanding of what's going on here. LGTM assuming Nikolay agrees with your modifications to the branch.

joaoandremartins:
I just tested that without my modifications, new run configs wouldn't have an artifact selected by default when there is only one artifact to be chosen from.
I also verified that with my addition, the red cross doesn't appear again. That's because the artifact is persisted into the settings at framework setup time.
Besides that, I just applied some style changes.
So, I think we should be good here!

patflynn:
This PR hasn't triggered a CI build for some reason.

joaoandremartins:
I think Jenkins might be stuck here? Haven't heard back from it in an hour.

patflynn:
ah, of course, it's an external pull request.. need to fix the config for that.

patflynn:
@joaoandremartins for now can you push this change to a project branch?

joaoandremartins:
OK

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/maven/AppEngineFacetImporter.java
chashnikov: I think such comments would pollute code. In IDEA 2016.3 we added a feature specifically for such cases: it [automatically shows](https://blog.jetbrains.com/idea/2016/09/intellij-idea-2016-3-eap-faster-git-log-parameter-hints-and-more/) parameter names for literal values. This is how this code looks in my IDE:
Parameter Hint

joaoandremartins: That's great, thanks for mentioning. I'll keep that in mind.

PR #966, opened 2016-10-07
Warn the user if they are using java 8 on AE standard

aslo:
fixes https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/847

Adds a project startup check to warn the user if their project contains any modules that are using an unsupported language level for their project configuration.


google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -164,8 +164,9 @@
                      key="unique.rest.signature.name" groupKey="inspections.group.name" enabledByDefault="true"
                      level="ERROR" implementationClass="com.google.cloud.tools.intellij.appengine.validation.RestSignatureInspection"/>
 
-    <postStartupActivity implementation="com.google.cloud.tools.intellij.PluginCompatibilityCheck"/>
-    <postStartupActivity implementation="com.google.cloud.tools.intellij.ConflictingAppEnginePluginCheck"/>
+    <postStartupActivity implementation="com.google.cloud.tools.intellij.startup.PluginCompatibilityCheck"/>
```
patflynn: I was kind of expecting this stuff hooked in through a [ProjectComponent](https://upsource.jetbrains.com/idea-ce/file/idea-ce-1731d054af4ca27aa827c03929e27eeb0e6a8366/platform/core-api/src/com/intellij/openapi/components/ProjectComponent.java) or even a [ModuleComponent](https://upsource.jetbrains.com/idea-ce/file/idea-ce-1731d054af4ca27aa827c03929e27eeb0e6a8366/platform/core-api/src/com/intellij/openapi/module/ModuleComponent.java) .  I wonder why Jetbrains introdoced this startup activity API.

aslo: Yeah, that makes sense. I'll move this code over to a project component.

aslo: It appears that postStartupActivity may actually be the right choice here. The issue with using project or module components is that this feature requires us to inspect the project structure (currently using `PSIManager` via the `DefaultAppEngineAssetProvider` which we use to read the module's `appengine-web.xml`).  PSI files appear not to be initialized until later in the project initialization lifecycle (even after the `projectOpened` hooks in both `ModuleComponent` and `ProjectComponent`), which results in an exception when we try to access PSI files via either of these components.

I'm going to revert to the prior implementation using postStartupActivity.

etanshaul: One thing to keep in mind (I'm coming in late to this PR) is that you should probably check for the existence of App Engine framework support / AE facet for this check rather than manually reading the appengine-web.xml again (since it has already been read by the time the framework support has been added).

aslo: We do currently check for the presence of the AE facet (see [below](https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/966/files#diff-98e94be8c6d76533d7afdc7ad8291fb1R71)) in order to determine whether the module is  an app engine module. We're need to access the `appengine-web.xml` in order to inform our logic about whether or not this is an appengine standard project (see `AppEngineProjectService.getInstance().getModuleAppEngineEnvironment(appengineWebXml)`). I'm not sure if there's a better way to do that.

aslo:
@patflynn I agree. But there are a number of posts on the [JB forums](https://intellij-support.jetbrains.com/hc/en-us/community/posts/203366830-Initialise-Project-must-read-PSI-Files) that seem to indicate that a postStartupActivity is the appropriate place to access PSI files.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/module/JavaLanguageLevelSupportCheck.java
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  @Override
+  public void moduleAdded() {
+    if (isModuleUsingUnsupportedLanguageLevel(thisModule)) {
+      warnUser(thisModule);
+    }
+  }
+
+  protected boolean isModuleUsingUnsupportedLanguageLevel(Module module) {
```
aslo: No good reason :)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/module/JavaLanguageLevelSupportCheck.java
```diff
@@ -0,0 +1,184 @@
...
+    // if it's not an app engine module, it's fine
+    if (!hasAppEngineFacet(module)) {
```
aslo: The way I was thinking about it, the reason that java 8 would be unsupported in a module is very much predicated on the module being an app engine module.  I will update the method name though, to make that more clear.

aslo:
@patflynn I believe all comments have been addressed, PTAL when you have a chance.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/UnsupportedJavaVersionCheck.java
```diff
@@ -0,0 +1,177 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import javax.swing.event.HyperlinkEvent;
+
+/**
+ * A StartupActivity that warns the user if they are using an unsupported java version.
```
patflynn: I'd mention the fact that it only cares about App Engine Standard

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/UnsupportedJavaVersionCheck.java
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  private List<Module> findAppEngineModulesUsingUnsupportedLanguageLevel(Project project) {
+    Module[] projectModules = ModuleManager.getInstance(project).getModules();
+    List<Module> invalidModules = new ArrayList<>();
+
+    for (Module module : projectModules) {
+      // if it's not an app engine module, skip it
+      if (!hasAppEngineFacet(module)) {
+        continue;
```
etanshaul: code styling thing - I prefer not using continue and wrapping the remaining code in the negation of this condition.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/UnsupportedJavaVersionCheck.java
```diff
@@ -0,0 +1,177 @@
...
+
+  private static final String RUNTIME_TAG = "runtime";
+  private static final String RUNTIME_TAG_JAVA_8 = "java8";
+  private static final String UPDATE_HREF = "#update";
+  private static final LanguageLevel HIGHEST_SUPPORTED_LANGUAGE_LEVEL = LanguageLevel.JDK_1_7;
+
+  @Override
+  public void runActivity(@NotNull Project project) {
+    List<Module> invalidModules = findAppEngineModulesUsingUnsupportedLanguageLevel(project);
+    if (!invalidModules.isEmpty()) {
```
etanshaul: If you only care if its not empty, why gather them all in the first place? Isn't it enough just to find the first?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/UnsupportedJavaVersionCheck.java
```diff
@@ -0,0 +1,177 @@
...
+    Module[] projectModules = ModuleManager.getInstance(project).getModules();
+    List<Module> invalidModules = new ArrayList<>();
+
+    for (Module module : projectModules) {
+      // if it's not an app engine module, skip it
+      if (!hasAppEngineFacet(module)) {
+        continue;
+      }
+
+      @Nullable
```
etanshaul: Looks weird to have a nullable over a statement

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/UnsupportedJavaVersionCheck.java
```diff
@@ -0,0 +1,177 @@
...
+      return false;
+    }
+    String runtime = rootTag.getSubTagText(RUNTIME_TAG);
+    return RUNTIME_TAG_JAVA_8.equals(runtime);
+  }
+
+  private static void setModuleLanguageLevel(Module module, LanguageLevel languageLevel) {
+    final ModifiableRootModel rootModel = ModuleRootManager.getInstance(module)
+        .getModifiableModel();
+    rootModel.getModuleExtension(LanguageLevelModuleExtension.class)
```
etanshaul: I'm not sure about this, but do you also need to perform this within the write action?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/UnsupportedJavaVersionCheck.java
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * A StartupActivity that warns the user if they are using an unsupported java version.
+ */
+public class UnsupportedJavaVersionCheck implements StartupActivity {
+
+  private static final String RUNTIME_TAG = "runtime";
+  private static final String RUNTIME_TAG_JAVA_8 = "java8";
+  private static final String UPDATE_HREF = "#update";
+  private static final LanguageLevel HIGHEST_SUPPORTED_LANGUAGE_LEVEL = LanguageLevel.JDK_1_7;
```
etanshaul: I'd either rename so that it indicates that its for AE standard, or rename the class altogether (something like `AppEngineStandardUnsupportedJavaVersionCheck`).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/UnsupportedJavaVersionCheck.java
aslo: We actually need the full list of invalid modules. See below in the `NotificationListener` implementation, when we iterate through all the invalid modules we collected above and set them to use the correct language level.

etanshaul: Ah, nice. Missed that part.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/UnsupportedJavaVersionCheck.java
aslo: This is actually a valid use of the `@Nullable` anotation (it's declared as applicable to `ElementType.LocalVariable`).  It's a bit uncommon to use it in this way, but I think it's important to explicitly call out that since it's assigned to the result of a method that's annotated as nullable, it too is nullable. Without the annotation, I can too easily envision someone refactoring this method and not thinking to check that variable for a null value.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/UnsupportedJavaVersionCheck.java
aslo: Fine by me :+1:

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/UnsupportedJavaVersionCheck.java
etanshaul: Ok. But I think that if you pass that var to a method with @NotNull param the ide will warn you anyway. Currently you are passing it to a @Nullable parameter which is why I saw that statement level annotation as redundant.

aslo: The IDE may warn you, but it's not guaranteed to, depending on how an IDE is configured (mine doesn't give me a warning). The way I think about it, we can't necessarily rely on future code readers having the same static analysis tools that we do. The main value of this annotation is for documentation purposes.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/UnsupportedJavaVersionCheck.java
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  private boolean isAppEngineStandard(@Nullable XmlFile appengineWebXml) {
+    AppEngineEnvironment environment = AppEngineProjectService.getInstance()
+        .getModuleAppEngineEnvironment(appengineWebXml);
+
+    return environment == AppEngineEnvironment.APP_ENGINE_STANDARD;
+  }
+
+  private boolean declaresJava8Runtime(@Nullable XmlFile appengineWebXml) {
```
aslo: Sounds good. I was kind of on the fence about whether or not this was too much of a special case to belong in the DefaultAppEngineProjectService, but I'm happy to move it there.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/UnsupportedJavaVersionCheck.java
patflynn: Actually thinking about it more I'm ambivalent.  OK with leaving it for now.

patflynn:
@aslo don't block on my to submit this.  Feel free to merge once you get the LGTM from. @etanshaul

etanshaul:
lgtm after checkstyle fixes

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/UnsupportedJavaVersionCheck.java
etanshaul: Oh, I didn't notice that you addressed this comment?

aslo: Oh sorry, missed this one. Will address.

PR #975, opened 2016-10-10
isModified() should not only return true if the user has enabled tracking

### LINKED ISSUE #969: Usage tracking selection doesn't persist
I found I could not enable the usage tracking. The checkbox would not visually persist.

Replicate
- install plugin
- goto cloud preferences usage tracking dialog
- select usage tracking
- select ok, close preferences
- return to preferences, goto usage tracking

Screenshots of version below:

screen shot 2016-10-10 at 7 37 32 am
screen shot 2016-10-10 at 7 39 01 am

patflynn:
I'm a bit perplexed as to how this ever worked. I'm going to have to dig in the history to figure out what happened here as I'm confident this setting used to persist. 

Fixes #969

### DIFF
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTrackerPanel.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTrackerPanel.java
@@ -80,7 +80,7 @@ public class UsageTrackerPanel {
   }
 
   public boolean isModified() {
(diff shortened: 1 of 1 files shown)


etanshaul:
probably broken by me in https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/808/files

patflynn:
Aha!  Now we know at least when this started.  On the plus side that should mean we're under-counting opt-in users.

PR #982, opened 2016-10-10
always use the google-java-format if the plugin is installed.

patflynn:
@aslo 
@etanshaul 

let me know if you guys think this is a bad idea.


etanshaul:
is the commit message correct? I don't see how having the plugin installed affects this.

aslo:
This looks like just setting the google-java-format for developing the plugin. Is that the intent?

patflynn:
Correct. This change just turns on google-java-format plugin if it's installed in the IntelliJ that you're using to edit google-cloud-intellij

joaoandremartins:
I'm guessing we'd also need to first download and install google-java-format to make this work.

SG, though I've been running into issues with the way google-java-format organises imports. It's incompatible with checkstyle, so I usually fall back to the IJ code style for imports, which is almost what checkstyle requires (need to add some new lines here and there)

patflynn:
Yeah this will only apply if you happen to have the plugin install, which on our work IJ instances it is by default. 

I'll take a look at the imports issue, may just fix that by modifying linting.

PR #994, opened 2016-10-11
Fix deployment sources for flex compat projects; fix maven deploy targets for flex compat

### LINKED ISSUE #993: Flex compat apps showing too many deployment sources
These apps should have the same sources as AE standard apps.

Right now they are showing all the sources available to "regular" flex apps as well (note that, for example, non-exploded-wars should not be available to flex compat apps):

image

etanshaul:
fixes #993 

Flex compat projects should show the same sources as AE standard projects.

Also fixes error when deploying a maven deployment source for flex compat projects: since the cloud environment for these projects is flex, it incorrectly used the war target instead of the exploded directory.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
@@ -93,7 +93,8 @@ public class AppEngineUtil {
       Collection<Artifact> artifacts = ArtifactUtil.getArtifactsContainingModuleOutput(module);
       for (Artifact artifact : artifacts) {
         if ((isStandardModule && projectService.isAppEngineStandardArtifactType(artifact))
-            || (environment.isFlexible() && projectService.isAppEngineFlexArtifactType(artifact))) {
+            || (!isFlexCompat && environment.isFlexible()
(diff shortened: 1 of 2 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
```diff
@@ -93,7 +93,8 @@ private AppEngineUtil() {
       Collection<Artifact> artifacts = ArtifactUtil.getArtifactsContainingModuleOutput(module);
       for (Artifact artifact : artifacts) {
         if ((isStandardModule && projectService.isAppEngineStandardArtifactType(artifact))
-            || (environment.isFlexible() && projectService.isAppEngineFlexArtifactType(artifact))) {
+            || (!isFlexCompat && environment.isFlexible()
```
patflynn: it's getting pretty difficult to parse the logic going on in this class.  We may want to consider a rethink/refactor sometime this quarter with some testing to help document the expectations here.

etanshaul:
@joaoandremartins I'd like to get your approval here too as you helped uncover this issue.

joaoandremartins:
I'm approving without testing. I assume you already tested this.

etanshaul:
I did

PR #1007, opened 2016-10-12
Release v16.10.4

patflynn:
I think we're setting some kind of record today.

On the plus side imagine how long this would have taken on Travis.


joaoandremartins:
Still testing, let me know if you want the approval before.

PR #1013, opened 2016-10-13
Silently download App Engine libraries

### LINKED ISSUE #1006: UX: Users change the version of Servlet API when using our wizard to add it to the project
We currently only support 2.5.  We should make that very explicit if we can't disable them changing it entirely.

Ideally there would be no dialog UIs to import the selected libraries.

etanshaul:
fixes #1006


etanshaul:
Thats correct.

Although we may need to reconsider how to handle the servlet api dependency per @patflynn's comment https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1006#issuecomment-253547071

etanshaul:
@patflynn in your opinion, should we merge this in, or wait for a more robust way to handle the servlet-api dependency (depending on if its flex compat)?

PR #1029, opened 2016-10-17
Updates for 163

### LINKED ISSUE #1000: project broken when tested against IJ EAP preview
https://cloud-tools-for-java-testing.appspot.com/job/google-cloud-intellij%20master%20daily%20build%20orb%20(IDEA%20EAP)/28/console

aslo:
fixes #1000

Updates required to get the plugin to build on the latest 163.* EAP builds. Changes are explained inline.

### DIFF
--- a/common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/BasePluginTestCase.java
+++ b/common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/BasePluginTestCase.java
@@ -19,7 +19,6 @@ package com.google.cloud.tools.intellij.testing;
 import com.intellij.openapi.Disposable;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.extensions.DefaultPluginDescriptor;
-import com.intellij.openapi.extensions.ExtensionPoint;
 import com.intellij.openapi.extensions.ExtensionPointName;
 import com.intellij.openapi.extensions.Extensions;
 import com.intellij.openapi.extensions.PluginId;
@@ -75,17 +74,11 @@ public class BasePluginTestCase {
   protected <N, T extends N> ExtensionPointImpl<T> registerExtensionPoint(
       @NotNull ExtensionPointName<N> name,
       @NotNull Class<T> type) {
-    ExtensionPointImpl<T> extensionPoint = new ExtensionPointImpl<T>(
+    extensionsArea.registerExtensionPoint(
         name.getName(),
         type.getName(),
-        ExtensionPoint.Kind.INTERFACE,
-        extensionsArea,
-        null,
-        new Extensions.SimpleLogProvider(),
-        new DefaultPluginDescriptor(PluginId.getId(type.getName()), type.getClassLoader())
-    );
-    extensionsArea.registerExtensionPoint(extensionPoint);
-    return extensionPoint;
+        new DefaultPluginDescriptor(PluginId.getId(type.getName()), type.getClassLoader()));
+    return extensionsArea.getExtensionPoint(name.getName());
   }
 
   @After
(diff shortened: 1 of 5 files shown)


common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/BasePluginTestCase.java
```diff
@@ -75,17 +74,11 @@ public final void setup() {
   protected <N, T extends N> ExtensionPointImpl<T> registerExtensionPoint(
       @NotNull ExtensionPointName<N> name,
       @NotNull Class<T> type) {
-    ExtensionPointImpl<T> extensionPoint = new ExtensionPointImpl<T>(
```
aslo: `ExtensionPointImpl`'s constructor has been made package private, so we can't use it anymore. Moving forward it's a good idea to avoid depending on impl classes.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineAssetProvider.java
```diff
@@ -65,21 +63,35 @@ public XmlFile loadAppEngineStandardWebXml(@NotNull Project project,
           project, "appengine-web.xml", module.getModuleContentScope()));
     }
 
-    Iterables.filter(appEngineWebXmls, Predicates.notNull());
```
etanshaul: Just curious - why is this invalid? It gets rid 6+ lines of boiler plate code.

findbugs-excludefilter.xml
```diff
@@ -34,7 +34,11 @@
       <Bug pattern="DM_STRING_CTOR" />
       <Bug pattern="RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE" />
       <Bug pattern="RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE" />
+      <Bug pattern="NP_NULL_ON_SOME_PATH" />
       <Bug pattern="NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE" />
+      <Bug pattern="NP_NULL_ON_SOME_PATH_MIGHT_BE_INFEASIBLE" />
+      <Bug pattern="NP_NULL_PARAM_DEREF" />
+      <Bug pattern="NP_GUARANTEED_DEREF" />
```
aslo: There were numerous false-positive findbugs warnings when compiling against the 163.* builds. I'm not sure exactly why this happened, but I did verify that these are false positives, not real issues. Seeing that we were already ignoring similar NP_NULL warnings, this seemed like an appropriate solution.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineAssetProvider.java
aslo: Looks like our comments missed each other.  My original comment on this line:

>This was a bug, unrelated to the version update, but was causing a findbugs warning. Iterables.filter does not modify the list, so the filtered list was actually being ignored.

aslo: We could, though this returns an `Iterator`, and we would still have to convert it into a list immediately afterward in order to sort according to the `AppEngineWebXmlOrdering`. This seemed a bit cleaner to me, though I don't feel that strongly about it.

etanshaul:
Ok, I do like the functional style here, but whatever you prefer. Quickly
look at that code its not 100% clear to me that we need to filter nulls
out. But if we are unsure if the API guarantees non-nulls then its safer to
keep it.

aslo:
@etanshaul I hear you. Updated that section to use a more functional style, and saved us a few LOC.

PR #1030, opened 2016-10-17
Auto select the app engine api when objectify is selected

### LINKED ISSUE #1018: Installing Objectify automatically installs the App engine API SDK.
We should auto-check the AE API if they select objectify.

etanshaul:
fixes #1018 

unit tests pending.

This will current auto select the App Engine API when objectify is chosen. It does not however prevent the user from then unselecting the App Engine API afterwards. I'm debating between adding logic to prevent this, and adding a text label somewhere to indicate that App Engine API is required when Objectify is selected.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
@@ -20,20 +20,18 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Predicates;
 import com.google.common.collect.Collections2;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Sets;
 
 import org.jetbrains.annotations.NotNull;
 
-import java.awt.Component;
-import java.awt.GridLayout;
 import java.awt.event.ItemEvent;
 import java.awt.event.ItemListener;
-import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
 import java.util.Set;
 
 import javax.annotation.Nullable;
-import javax.swing.BorderFactory;
 import javax.swing.JCheckBox;
 import javax.swing.JPanel;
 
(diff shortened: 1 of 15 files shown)


etanshaul:
I'm leaning towards not disabling the App Engine API checkbox (so that it can't be unchecked) after selecting the Objectify library. Thoughts on this?

Other than this point, its ready for review.

joaoandremartins:
Did you mean "so that it _can_ be unchecked"?

joaoandremartins:
If Objectify depends on the App Engine API, than I think it would make sense to both disable the App Engine API button if Objectify is enabled AND adding the text explaining why that is happening.

I can't think of a reason why a user would want Objectify enabled and the App Engine API disabled, and get the subsequent compilation errors.

etanshaul:
@patflynn I updated the library panel per our discussion. Now endpoints and objectify are visually nested underneat the App Engine API:

etanshaul:
My last commit refactored a bunch of the original code in this PR so please re-review.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
```diff
@@ -42,51 +43,77 @@
...
   }
 
   public void toggleLibrary(AppEngineStandardMavenLibrary library, boolean select) {
-    for (Component libraryCheckbox : libraryPanel.getComponents()) {
-      if (libraryCheckbox instanceof JCheckBox
-          && ((JCheckBox) libraryCheckbox).getText().equals(library.getDisplayName())) {
-        ((JCheckBox) libraryCheckbox).setSelected(select);
+    for (JCheckBox libraryCheckbox : libraries) {
+      if (libraryCheckbox.getText().equals(library.getDisplayName())) {
+        libraryCheckbox.setSelected(select);
```
joaoandremartins: Ah, I somehow missed the boolean arg..

etanshaul:
@patflynn ptal to see if the updates based on our offline discussion look good to you

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
```diff
@@ -42,51 +43,77 @@
...
+    // Objectify and Endpoints are dependencies of the App Engine API.
+    objectifyCheckBox.setSelected(false);
+    endpointsCheckBox.setSelected(false);
+    appEngineApiCheckBox.addItemListener(new AppEngineApiListener());
+    objectifyCheckBox.addItemListener(new AppEngineApiDependencyListener());
+    endpointsCheckBox.addItemListener(new AppEngineApiDependencyListener());
   }
 
   public Set<AppEngineStandardMavenLibrary> getSelectedLibraries() {
     return Sets.newHashSet(Collections2.filter(
```
patflynn: not blocking, but just looking at all this transformation code makes me think an immutable bimap would have been a simpler way to handle this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
```diff
@@ -100,21 +127,43 @@ public boolean isEnabled() {
...
+
+    return null;
+  }
+
+  @VisibleForTesting
+  List<JCheckBox> getLibraries() {
+    return libraries;
+  }
+
+  private class AppEngineApiListener implements ItemListener {
```
patflynn: AppEngineApiCheckboxListener? is this actually used more than once? Maybe you can inline it?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
```diff
@@ -100,21 +127,43 @@ public boolean isEnabled() {
...
+      JCheckBox checkbox = (JCheckBox) event.getItem();
+
+      if (!checkbox.isSelected()) {
+        objectifyCheckBox.setSelected(false);
+        endpointsCheckBox.setSelected(false);
+      }
+    }
+  }
+
+  private class AppEngineApiDependencyListener implements ItemListener {
```
patflynn: same here, AppEngineApiDependentCheckboxListener

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
etanshaul: this may be good idea. I'll quickly evaluate it

Issue #1032, opened 2016-10-17
remove feature flag checks from fully GA'd features


patflynn:
app engine deployment and cloud debugging are now turned on for all users. We should remove the feature flag checks and move whatever we can into the plugin.xml

PR #1033, opened 2016-10-17
Unfeature flag the cloud debugger

patflynn:
Relates to #1032.


etanshaul:
Are you planning on submitting individual PRs to unfeature each component?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/CloudToolsPluginInitializationComponent.java
```diff
@@ -62,10 +62,6 @@ public void initComponent() {
     CloudToolsPluginInfoService pluginInfoService = ServiceManager
         .getService(CloudToolsPluginInfoService.class);
 
-    if (pluginInfoService.shouldEnable(GctFeature.DEBUGGER)) {
```
etanshaul: Should we also remove `GctFeature.Debugger`?

patflynn:
yes to both comments.

etanshaul:
ok. lgtm then after the unit test fix

PR #1034, opened 2016-10-18
Provide link to download the cloud sdk

### LINKED ISSUE #1019: Include link to download the cloud sdk
In all locations where the user is asked to select the sdk directory, there should also be a link to the download page in case the user doesn't have it installed.

etanshaul:
fixes #1019


etanshaul:
Here are a couple examples of what this looks like. It appends the download link message to all cloud sdk selection widgets (when in warning mode):

image

image

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanelTest.java
```diff
@@ -34,11 +36,18 @@
 
   private CloudSdkService cloudSdkService;
 
-  private JLabel warningMessage;
+  private JTextPane warningMessage;
   private TextFieldWithBrowseButton cloudSdkDirectoryField;
 
-  private static final String INVALID_SDK_DIR_WARNING =
-      "No Cloud SDK was found in this directory.";
+  private static final String INVALID_SDK_DIR_WARNING = "<html>\n"
```
patflynn: I'm not convinced in the value of checking the exact message text here.  isn't it good enough for the test to verify that an error is being displayed when it should be?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -43,12 +45,19 @@
 public class CloudSdkPanel {
 
   private TextFieldWithBrowseButton cloudSdkDirectoryField;
-  private JLabel warningMessage;
+  private JTextPane warningMessage;
   private JPanel cloudSdkPanel;
+  private JLabel warningIcon;
+
+  private static final String CLOUD_SDK_DOWNLOAD_LINK = "https://cloud.google.com/sdk/docs/"
```
patflynn: I tend to like using the URL type for URL so that any typos are caught early.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanelTest.java
etanshaul: I'd usually agree, but in this cause there are multiple warning messages that can be displayed in the same UI field (one for empty field, and one for invalid field) so this test checks that the right one is being shown.

joaoandremartins:
Can you propagate this to wherever else `appengine.cloudsdk.location.missing.message` is used?

etanshaul:
@joaoandremartins I think it is already. I added this code to the CloudSdkPanel class; so now anywhere there is an sdk input field this will be used.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanelTest.java
patflynn: got it. I wonder if we should look into building an error type in the future.

this is good for now.

joaoandremartins:
There are a couple other places:
AppEngineServerModel
CloudSdkStartupPolicy

etanshaul:
Do those classes have a UI for cloud sdk browsing? If so, they should be using the CloudSdkPanel.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
etanshaul: actually, since this is just a static (hardcoded) url that we verify - and not something generated by user input for example - are you ok with leaving it as a string? I don't really see the added benefit.

joaoandremartins:
`AppEngineServerModel.checkConfiguration()` has its own local dev server UI.
CloudSdkStartupPolicy just throws up on the console.

In general, importing forms from different packages is failing due to https://youtrack.jetbrains.com/issue/IDEA-110316 / https://youtrack.jetbrains.com/issue/IDEA-162625.

etanshaul:
@joaoandremartins please check out the updates to `AppEngineServerModel` and `CloudSdkStartupPolicy`. I fixed / refactored the validation as we discussed; I also added the download link to these too.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
```diff
@@ -162,20 +165,28 @@ public void checkConfiguration() throws RuntimeConfigurationException {
           "App Engine facet not found in '" + artifact.getName() + "' artifact");
     }
 
-    if (CloudSdkService.getInstance().getSdkHomePath() == null) {
+    CloudSdkService cloudSdkService = CloudSdkService.getInstance();
+    if (cloudSdkService.getSdkHomePath() == null
+          || StringUtils.isEmpty(cloudSdkService.getSdkHomePath().toString())) {
```
joaoandremartins: If the home path is empty, I think we should just let validateCloudSdk() take care of it, so I'd just check for null.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
```diff
@@ -64,15 +69,27 @@ public String getDisplayString() {
           public OSProcessHandler createProcessHandler(
               String workingDirectory, Map<String, String> envVariables) throws ExecutionException {
             CloudSdkService sdkService = CloudSdkService.getInstance();
+
             if (sdkService.getSdkHomePath() == null
-                || sdkService.getSdkHomePath().toString().isEmpty()) {
+                || StringUtils.isEmpty(sdkService.getSdkHomePath().toString())) {
```
joaoandremartins: Same as above here.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
```diff
@@ -64,15 +69,27 @@ public String getDisplayString() {
...
               throw new ExecutionException(
-                  GctBundle.message("appengine.cloudsdk.location.missing.message"));
+                  CloudSdkPanel.createErrorMessageWithLink(
+                      GctBundle.message("appengine.cloudsdk.location.missing.message")));
             }
-
-            if (sdkService == null) {
+            try {
```
joaoandremartins: new line before?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
joaoandremartins: Also, what's the benefit of using StringUtils over just String.isEmpty()?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
```diff
@@ -99,11 +116,6 @@ public OSProcessHandler createProcessHandler(
             executor.run();
 
             Process devappserverProcess = executor.getProcess();
-            if (devappserverProcess == null) {
```
joaoandremartins: We shouldn't get rid of this block, because it is still possible that dev_appserver blows up for some reason and we don't want it to look nasty on the IDE.

How about we throw an ExecutionException with a generic message like "The Google App Engine local development server terminated with an error."?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
etanshaul: StringUtils#isEmpty is null safe. but I guess since its doing a toString it doesn't really matter

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
etanshaul: I'd prefer for shit to blow up in this case. This is truely an exception condition and it would be nice to be notified (via an error report) rather than just swallowing it with a UI message. Thoughts?

joaoandremartins: The way it blows with a thrown unchecked exception or an ExecutionException is really similar AFAIK. The only difference is that you'll get a nasty visible stack trace, whereas with an ExecutionException you get a user friendly notification, with a red baloon, the report to google cloud dialog, etc.

This was also a requested fix from #932

etanshaul: If that's actually the case that ExecutionException's don't get swallowed by the platform (and actually report and error) we need to fix some other code I think:

The sdk validation logic (empty sdk, invalid sdk) is also using this exception. In these cases we definitely should not be reporting errors.

etanshaul: Yea exactly - this is a condition that we do not anticipate so I think this is where something went genuinely wrong and we would want to know.

etanshaul:
@joaoandremartins I think this pr needs your approval too since you requested changes

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
etanshaul: I'm checking for empty because we have 2 kinds of error messages for the user. the error message shown when validateCloudSdk fails is different.

joaoandremartins: They should be different. If the SDK location is empty, is because the user set it that way. In that case, we should tell them No SDK was found in that directory, rather than an SDK directory wasn't specified.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
joaoandremartins: In that case, I'd argue the code would be a bit cleaner to revert back to String.isEmpty() (and we'd cut the dependency on that apache lib).

But, ideally, what I'd really like is to get rid of this empty check altogether, for the reasons on the above comment.

etanshaul: I can change it to isEmpty since StringUtils isn't adding anything here for us

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
joaoandremartins: An empty SDK is invalid, and that is going to be caught when you call sdk.validateCloudSdk() shortly after.

I'm considering an empty SDK different from a null SDK. For an SDK to be empty, the user has to have made it that way (previously, it's null, hence unspecified). For me, an empty string falls under "invalid string provided by user", so the messaging should be coherent.

Also, there could be a file system where an empty string is a valid location.

joaoandremartins:
I take it that by resolving the comment, you want to go ahead and merge things your way, so go ahead :)

etanshaul:
Did I resolve it? I didn't think I did anything. Either way, I'll merge it
:) Unless @patflynn wants to weight in.

patflynn:
nothing to add.  carry on.

etanshaul:
cool. @joaoandremartins and I can fight it out later after work and the winner gets to merge his way :)

PR #1035, opened 2016-10-18
Codecov enhancements

aslo:
- adding a coverage badge on our README
- adding a codecov.yml config file
- configuring codecov to write commit statuses, rather than the PR commit. This is a bit less noisy, and allows us to see coverage stats at the commit level.


.codecov.yml
```diff
@@ -0,0 +1,16 @@
+comment: off
+coverage:
+  precision: 2
+  range: "70...100"
```
patflynn: so this is the coverage for the specific PR?

.codecov.yml
```diff
@@ -0,0 +1,16 @@
...
+  round: down
+  status:
+    changes: false
+    patch: true
+    project: true
+parsers:
+  gcov:
+    branch_detection:
```
patflynn: what do these params do?

aslo:
@patflynn @etanshaul @joaoandremartins  Seeking opinions on what we want to do here.  

Currently, codecov adds a single comment on each PR, with detailed code coverage stats. The comment is updated with each new push. The downside of this is that this is a bit noisy, as there is no way to disable email notifications for the code coverage comment. (This will happen maximum one time per PR).

An alternative is to have codecov publish to the github status API (like what jenkins does). We can configure it to block merging if we don't meet a certain coverage threshold (or to just always pass, but still report the coverage percentage).  If we did this, we would get less detailed coverage stats in the PR. We would also need to setup a github service acccount credentials to allow codecov to write statuses to our repo. 

What do you guys think? Leave it as-is, or change it to use github statuses?

joaoandremartins:
I actually thought that the code coverage stats referred only to the files in a PR, but I guess I never read them that carefully.. Which is also why, for me, we can probably move to github statuses. Where would this information be published in that case?

patflynn:
I'm OK with this, although I'd refarain from blocking merging for now.

aslo:
Those commit statuses report both the coverage for the overall repository, and for the changes introduced with the PR. By moving to commit statuses it would just be a bit less noisy, IMO. You would be able to see all the same level of detail by following the links to the codecov.io reports on their site.

.codecov.yml
```diff
@@ -0,0 +1,11 @@
+codecov:
+  bot: aslo
+coverage:
+  status:
+    patch:
+      default:
+        target: 0%
```
aslo: Set a target of 0 so that these statuses will always be green, regardless of code coverage changes. Later, we can change these to use `target: auto`, the default setting, to fail builds if the coverage goes down below its current setting. 

Also note that it's an option to have these statuses never block merging, even when they are red. I still think this is valuable as-is for the time being.

aslo:
@patflynn Finished tweaking the settings. PTAL, and note that the way the statuses appear on this PR is how they will appear on all others that have these changes.

.codecov.yml
patflynn: Red and not blocking seems preferable. Can we make it do that now?

aslo: :taco:

aslo: Sorry, meant :+1:

Issue #1043, opened 2016-10-19
Cloud Sdk warning messaging is incorrect in the dev app server run config


etanshaul:
![image](https://cloud.githubusercontent.com/assets/1735744/19525899/811ba036-95f0-11e6-9bae-594a69d3abf4.png)

etanshaul:
closed via https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1046

PR #1046, opened 2016-10-19
Reuse the CloudSdkPanel in the dev app server run config

### LINKED ISSUE #792: link to console cloud repo after successfull source import is wrong
steps to reproduce:
1) Create a new project in IJ
2) Select VCS -> Import into Version control
3) select GCP
4) Select a project without an existing repo
5) observe a successful import with a link to the repo
6) click link
7) observe you are on an IAM cloud console page.

expected result:
the link brings me to the source navigation view for the uploaded project in the cloud console.

### LINKED ISSUE #1041: Cloud SDK download link is broken in dev server run config windows

etanshaul:
Removes code duplication that was recreating the already existing `CloudSdkPanel` because of IJ bug https://youtrack.jetbrains.com/issue/IDEA-162625.

This bug does not surface if we don't include panels across modules from within the gui forms. If we delete the form (like in the PR), and just instantiate the CloudSdkPanel, then there is no problem including it across modules.

Also happens to fix #1041 and #1043 

Still some work pending:
- I may inline `AppEngineServerEditor` since its not doing much anymore
- @joaoandremartins can we now get rid of the sdk validation logic in `AppEngineServerModel` and `AppEngineStartupPolicy` since we now let the CloudSdkPanel do validation?


joaoandremartins:
This is awesome! Thanks for this work, Etan.

I don't think we can get rid of the SDK checks in CloudSdkStartupPolicy or AppEngineServerModel.

The former because I think there are subtle conditions when you can go through the whole run-devappserver flow without ever specifying an application server, or a valid one (through the "Run Anyway" way, but I think other ways too), and we don't want the IDE to explode on those cases.

The latter because of the same reasons as the former, and also because if you haven't created any Application Server when you get in the UI, this validation is the one that causes the "you need to install Cloud SDK" warning message in the UI (by throwing the ExecutionException).

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/integration/AppEngineServerHelper.java
```diff
@@ -44,6 +50,28 @@ public ApplicationServerPersistentData createPersistentDataEmptyInstance() {
   }
 
   public ApplicationServerPersistentDataEditor createConfigurable() {
-    return new AppEngineServerEditor();
+    final CloudSdkPanel cloudSdkPanel = new CloudSdkPanel();
+
+    return new ApplicationServerPersistentDataEditor<ApplicationServerPersistentData>() {
+      @Override
+      protected void resetEditorFrom(ApplicationServerPersistentData s) {
```
joaoandremartins: Maybe rename 's' to a more meaningful name? like `data` or `persistedConfiguration` (which would make the applyEditorTo arg be called something like `toBePersistedConfiguration`.

etanshaul:
regarding CloudSdkStartupPolicy and AppEngineServerModel:

if we can get rid of that validation like you said then I still need to figure out how to make the sdk download link work. Typically this is down by getting a handle on the UI component and attaching an HREF listener. Since this message is passed to the framework in the form of an exception, I don't yet know how to do this.

etanshaul:
I have an idea:

Why not make the Cloud SDK a required field for the app server configuration. This way, we will not run into the situation where CloudSdkStartupPolicy and AppEngineServerModel have an invalid SDK and we can remove that duplicated validation logic.

This is exactly how deployment currently works: you cannot create a cloud without a valid SDK - it simply won't let you save the panel.

etanshaul:
To add to my point above: I don't see why we'd want to allow saving an app server without a valid sdk path since this is not a valid app server. By forcing it, like we do for clouds, we can ensure that they are always valid and runnable.

etanshaul:
Ok - as discussed offline, we'll keep the extra validation since it adds a little more guidance to the user.

I still need to figure out how to make the href work though before merging this.

etanshaul:
Another approach to get around this mess is to never allow saving an invalid Cloud SDK. We decided to allow this under the Google Settings, but I'm rethinking this now since it allows the run configs to get in this weird state forcing us to add extra validation. @patflynn what do you think of updating the Google Settings to not allow saving an invalid Cloud SDK?

patflynn:
don't you always need validation at runtime?  Presumably the validity of
the SDK could change after the user has configured it.

etanshaul:
Yeah good point.

etanshaul:
@joaoandremartins what do you think of the latest updates?

I can't find a way to make the href's (to the cloud sdk lownload) work when used via exceptions as they are in those classes. 

Either way, I think this new message is even clearer: it directs them to reconfigure the app server - and once they get into the app server UI they'll clearly see what the specific SDK problem is from there.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
```diff
@@ -164,28 +163,19 @@ public void checkConfiguration() throws RuntimeConfigurationException {
           "App Engine facet not found in '" + artifact.getName() + "' artifact");
     }
 
-    CloudSdkService cloudSdkService = CloudSdkService.getInstance();
```
joaoandremartins: Why aren't we checking for null here? I think we should because, otherwise, when you build the SDK object, the library is going to try and search for a Cloud SDK on a null path, and that's not what we want.

etanshaul:
image

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -99,6 +99,7 @@ appengine.deployment.error.invalid.cloudsdk=Invalid Cloud SDK directory path.
 appengine.deployment.error.cancelled=Deployment process was cancelled.
 appengine.facet.converter.description=Deprecated Google App Engine facets will be replaced with the latest version
 appengine.run.server.name=App Engine Development Server
+appengine.run.server.sdk.misconfigured.message=The Cloud SDK is misconfigured. Please reconfigure the application server above.
```
joaoandremartins: :+1:

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
```diff
@@ -67,18 +66,9 @@ public String getDisplayString() {
           @Override
           public OSProcessHandler createProcessHandler(
               String workingDirectory, Map<String, String> envVariables) throws ExecutionException {
-            CloudSdkService sdkService = CloudSdkService.getInstance();
-
-            if (sdkService.getSdkHomePath() == null
```
joaoandremartins: Same thing here -- we need this null validation.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -99,6 +99,7 @@ appengine.deployment.error.invalid.cloudsdk=Invalid Cloud SDK directory path.
 appengine.deployment.error.cancelled=Deployment process was cancelled.
 appengine.facet.converter.description=Deprecated Google App Engine facets will be replaced with the latest version
 appengine.run.server.name=App Engine Development Server
+appengine.run.server.sdk.misconfigured.message=The Cloud SDK is misconfigured. To fix, click configure next to the application server above.
```
joaoandremartins: Love this :+1:

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
etanshaul: I looked at the common lib code and it will check for null and still throw an AppEngineException (or a subclass of it).

etanshaul: ah! nice catch. I'll fix.

etanshaul:
@joaoandremartins ptal: https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1046/commits/4d2150710b299692e46c9b7a6176ae0ea350bfba

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
```diff
@@ -164,28 +163,25 @@ public void checkConfiguration() throws RuntimeConfigurationException {
...
       CloudSdk sdk = new CloudSdk.Builder()
-          .sdkPath(cloudSdkService.getSdkHomePath())
+          .sdkPath(sdkService.getSdkHomePath())
           .build();
       sdk.validateCloudSdk();
       sdk.validateAppEngineJavaComponents();
     } catch (AppEngineJavaComponentsNotInstalledException ex) {
       throw new RuntimeConfigurationError(
-          GctBundle.message("appengine.cloudsdk.java.components.missing") + " "
+          GctBundle.message("appengine.cloudsdk.java.components.missing") + "\n"
```
joaoandremartins: Did you check if this new line renders well in the UI?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/CloudSdkStartupPolicy.java
```diff
@@ -67,13 +66,11 @@ public String getDisplayString() {
...
-            CloudSdkService sdkService = CloudSdkService.getInstance();
 
-            if (sdkService.getSdkHomePath() == null
-                || sdkService.getSdkHomePath().toString().isEmpty()) {
+            CloudSdkService sdkService = CloudSdkService.getInstance();
+            if (sdkService.getSdkHomePath() == null) {
               throw new ExecutionException(
-                  CloudSdkPanel.createErrorMessageWithLink(
-                      GctBundle.message("appengine.cloudsdk.location.missing.message")));
+                  GctBundle.message("appengine.run.server.sdk.misconfigured.message"));
```
joaoandremartins: I don't think we should use the "application server above" message here because this message will actually appear in the event log/balloon notification.
In either of these environments, there should be a browser opening listener, did you try it? If so, we could probably use CloudSdkPanel.createErrorMessageWithLink

etanshaul: what would I attach the browser listener to? I typically do it by attaching to the UI component, which I don't seem to have access to here

etanshaul: I did try clicking the link and it didn't work. Regardless of that though - should we maybe make the message more generic then "application server above"? Guide them generically to fix the application server. This gets rid of the duplication of the download link (they'll eventually see it again once they configure the app server) and it moves the download message closer to the correct source.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
etanshaul: Will do

etanshaul:
The line break did not render properly in AppEngineServerModel so I restored it to a space like it was before. It does render correctly via CloudSdkStartupPolicy so I left it there.

etanshaul:
@joaoandremartins I added a more generic version of the message for when its shown outside of the panel.

In the panel:
image

Out of the panel:

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -99,6 +99,8 @@ appengine.deployment.error.invalid.cloudsdk=Invalid Cloud SDK directory path.
 appengine.deployment.error.cancelled=Deployment process was cancelled.
 appengine.facet.converter.description=Deprecated Google App Engine facets will be replaced with the latest version
 appengine.run.server.name=App Engine Development Server
+appengine.run.server.sdk.misconfigured.panel.message=The Cloud SDK is misconfigured. To fix, click configure next to the application server above.
```
etanshaul: Suggestion taken

PR #1055, opened 2016-10-20
Disambiguate deployment sources

### LINKED ISSUE #821: Race condition when running deploy/stop against the same project/module src.
Steps to reproduce:
1) Deploy a flex app to a project of your choice
2) Click 'undeploy' once deployment is complete
3) Redeploy the same project

observed result:
4) On completion of 'stop' it will overwrite the status of the deployment that's going on. 

expected result:
5) the new deployment should be a new version so conceptually different from the instance being stopped. I should continue to see the deployment happening even after my previous stop succeeds.

etanshaul:
fixes #821 

Ensures that unique deployments (those that occur at different times) are displayed and handled as separate line items in the deployment UI.

Here is what the sources now look like. This is showing each deployment source type twice - once with an auto-increment version, and once with a custom version id. Before this PR, the deployment source of the same type would have been overwritten and not displayed twice.

image

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSource.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSource.java
@@ -56,7 +56,7 @@ public class AppEngineArtifactDeploymentSource extends ArtifactDeploymentSourceI
   @NotNull
   @Override
   public String getPresentableName() {
-    return String.format("[%s] ", DateTime.now().toString("HH:mm:ss")) + name;
+    return String.format("[%s] ", DateTime.now().toString("yyyy-MM-dd HH:mm:ss")) + name;
   }
 
   @Override
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSource.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSource.java
@@ -30,12 +30,14 @@ public class AppEngineArtifactDeploymentSource extends ArtifactDeploymentSourceI
     implements AppEngineDeployable {
 
   private AppEngineEnvironment environment;
+  private String name;
 
   /**
    * Default constructor used instantiating plain Artifact Deployment sources.
    */
   public AppEngineArtifactDeploymentSource(@NotNull ArtifactPointer pointer) {
     super(pointer);
+    setName(getDefaultName());
   }
 
   /**
@@ -47,6 +49,23 @@ public class AppEngineArtifactDeploymentSource extends ArtifactDeploymentSourceI
       @NotNull ArtifactPointer pointer) {
     super(pointer);
     this.environment = environment;
+    setName(getDefaultName());
+  }
+
+  @NotNull
+  @Override
+  public String getPresentableName() {
+    return name;
+  }
+
+  @Override
+  public String getDefaultName() {
+    return getArtifactPointer().getArtifactName();
+  }
+
+  @Override
+  public void setName(String name) {
+    this.name = name;
   }
 
   @Override
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSource.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSource.java
@@ -21,6 +21,7 @@ import com.intellij.remoteServer.configuration.deployment.DeploymentSourceType;
 import com.intellij.remoteServer.impl.configuration.deployment.ArtifactDeploymentSourceImpl;
 
 import org.jetbrains.annotations.NotNull;
+import org.joda.time.DateTime;
 
 /**
  * An App Engine implementation of {@link ArtifactDeploymentSourceImpl} that provides its targeted
@@ -55,7 +56,7 @@ public class AppEngineArtifactDeploymentSource extends ArtifactDeploymentSourceI
   @NotNull
   @Override
   public String getPresentableName() {
-    return name;
+    return String.format("[%s] ", DateTime.now().toString("HH:mm:ss")) + name;
   }
 
   @Override
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSource.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSource.java
@@ -21,7 +21,7 @@ import com.intellij.remoteServer.configuration.deployment.DeploymentSourceType;
 import com.intellij.remoteServer.impl.configuration.deployment.ArtifactDeploymentSourceImpl;
 
 import org.jetbrains.annotations.NotNull;
-import org.joda.time.DateTime;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * An App Engine implementation of {@link ArtifactDeploymentSourceImpl} that provides its targeted
@@ -31,47 +31,44 @@ public class AppEngineArtifactDeploymentSource extends ArtifactDeploymentSourceI
     implements AppEngineDeployable {
 
   private AppEngineEnvironment environment;
-  private String name;
 
-  /**
-   * Default constructor used instantiating plain Artifact Deployment sources.
-   */
-  public AppEngineArtifactDeploymentSource(@NotNull ArtifactPointer pointer) {
-    super(pointer);
-    setName(getDefaultName());
-  }
+  private String projectName;
+  private String version;
 
   /**
    * Initialize the artifact deployment source given a target App Engine environment, and an
    * artifact pointer.
    */
   public AppEngineArtifactDeploymentSource(
-      @NotNull AppEngineEnvironment environment,
+      @Nullable AppEngineEnvironment environment,
       @NotNull ArtifactPointer pointer) {
     super(pointer);
     this.environment = environment;
-    setName(getDefaultName());
-  }
-
-  @NotNull
-  @Override
-  public String getPresentableName() {
-    return String.format("[%s] ", DateTime.now().toString("yyyy-MM-dd HH:mm:ss")) + name;
-  }
-
-  @Override
-  public String getDefaultName() {
-    return getArtifactPointer().getArtifactName();
-  }
-
-  @Override
-  public void setName(String name) {
-    this.name = name;
   }
 
   @Override
   public AppEngineEnvironment getEnvironment() {
     return environment;
+  }
+
+  @Override
+  public String getProjectName() {
+    return projectName;
+  }
+
+  @Override
+  public void setProjectName(String projectName) {
+    this.projectName = projectName;
+  }
+
+  @Override
+  public String getVersion() {
+    return version;
+  }
+
+  @Override
+  public void setVersion(String version) {
+    this.version = version;
   }
 
   @NotNull
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployable.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployable.java
@@ -29,18 +29,26 @@ public interface AppEngineDeployable extends DeploymentSource {
    */
   AppEngineEnvironment getEnvironment();
 
-  /**
-   * Sets the name of this deployable. Used for display in the Application Server deployment window
-   * and for disambiguation of multiple deployables. This field is mutable so that, at deploy time,
-   * we can append additional information to the name to ensure uniqueness (e.g. version).
-   */
-  void setName(String name);
 
   /**
-   * Returns the default name for this deployable. The default name represents the name of the
-   * deployment source prior to deploy time. For example, this would not include version information
-   * since it is not yet known.
+   * Returns the targeted cloud project name.
    */
-  String getDefaultName();
+  String getProjectName();
 
+  /**
+   * Sets the targeted cloud project. It is mutable because the cloud project is not known until
+   * deploy time where it is then set.
+   */
+  void setProjectName(String projectName);
+
+  /**
+   * Returns the targeted cloud project version.
+   */
+  String getVersion();
+
+  /**
+   * Sets the targeted cloud project version. It is mutable because the cloud project version is
+   * not known until deploy time where it is then set.
+   */
+  void setVersion(String version);
 }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployable.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployable.java
@@ -30,7 +30,7 @@ public interface AppEngineDeployable extends DeploymentSource {
   AppEngineEnvironment getEnvironment();
 
   /**
-   * Set the name of this deployable. Used for display in the Application Server deployment window
+   * Sets the name of this deployable. Used for display in the Application Server deployment window
    * and for disambiguation of multiple deployables. This field is mutable so that, at deploy time,
    * we can append additional information to the name to ensure uniqueness (e.g. version).
    */
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployable.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployable.java
@@ -29,4 +29,18 @@ public interface AppEngineDeployable extends DeploymentSource {
    */
   AppEngineEnvironment getEnvironment();
 
+  /**
+   * Set the name of this deployable. Used for display in the Application Server deployment window
+   * and for disambiguation of multiple deployables. This field is mutable so that, at deploy time,
+   * we can append additional information to the name to ensure uniqueness (e.g. version).
+   */
+  void setName(String name);
+
+  /**
+   * Returns the default name for this deployable. The default name represents the name of the
+   * deployment source prior to deploy time. For example, this would not include version information
+   * since it is not yet known.
+   */
+  String getDefaultName();
+
 }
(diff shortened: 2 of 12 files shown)


etanshaul:
TODO - disambiguate multiple deployments with auto-increment versioning (currently they will override each other). I think I will append a timestamp to these.

etanshaul:
@patflynn @joaoandremartins @aslo Looking for some input on the following:

For _non-auto-incremented_ versions, do we want to continue to allow multiple simultaneous deployment "actions" (deployments / stops)? If we prevent this - via some sort of error message - then together with my other fixes in the PR we should completely have solved the race condition problem (see #821).  I just want to make sure there are no valid use cases for doing this.

The other possible solution is to append a timestamp to all deployments so that even if the deployment is to the same project / version, we'll allow them to run simultaneously and battle it out. 

_Note to self: I need to add project to the deployment name since deployment with the same version id but different project should definitely be permitted._

aslo:
> For non-auto-incremented versions, do we want to continue to allow multiple simultaneous deployment "actions" (deployments / stops)?

I wonder what a user might expect/intend to happen when they try to double-deploy to the same project/version simultaneously.  I'd think a user would not intend for the two deployments to battle it out, so maybe the most friendly thing to do would be to warn them upfront and prevent them from double deploying.

etanshaul:
@aslo thats what I'm thinking too. But I want to make sure we cover all bases: for example say on of them is set to promote, but the other is not (although I don't think duplicate versions are allowed at all).

joaoandremartins:
I think concurrent deployments of the same version is a valid use case. The way App Engine resolves this internally is similar to what you're suggesting -- a hash is appended after the version name and this is called a minor version.

I think there are users out there who, e.g., use the same version ID for development. Version IDs also used to be limited to something like 20 or 50, not sure if that's still the case.

We could add the timestamp without the user seeing it, right? It's just a control mechanism, we wouldn't actually deploy that version string?

etanshaul:
@joaoandremartins when I just tried deploying the same project/version twice, it seems to have just overwritten the existing version (I see only one instance of that version in the web console). Is there something special I need to do to get that minor version behavior? 

Yes - the timestamp would not show up in gcp, its just a control mechanism like you said.

etanshaul:
Also to clarify - I wasn't suggesting that the user can't deploy the same project/version multiple times. Just that perhaps doing this simultaneously would be disallowed.

joaoandremartins:
Ah yes, you'll still only see one version -- minor versions are how GAE allows for deploying the same version repeatedly. You don't need to do anything.

That was meant to serve as motivation to go the timestamps way :)

I think if we can make it work using timestamps, and it's not too much work, it's worth a shot. Otherwise, maybe forbid concurrent deployments to the same version? Though I'd be worried of stalled deployments, or those slow Flex ones.

etanshaul:
Oh I see.

I think timestamps and forbidding concurrent deployments are not mutually exclusive. I need to use the timestamp mechanism to disambiguate auto-incremented versions anyway (I maybe have been a little confusing about this in my original question). 

If concurrent deployment of the same project/version is not useful then maybe we can forbid it (perhaps in another PR).

etanshaul:
The problem may be a bit hard to follow without actually seeing it happen. So just in case, here's an overview:

The way IJ handles deployments is by creating entries under a given server in App Server window:
image

The entries here are unique by the display string that you see in the above screenshot. If you attempt to deploy an app with the same display string, then it will _not_ create a new entry, but overwrite the existing one - and you loose control over your simultaneous deployments.

I largely solved this problem by appending project and version to this ID.

However, there is still a problem with auto-incremented versions. Currently here is what is shown:
image

Note that the version just says "auto". This is because we do not yet know the version that will be used (gcloud creates one for us later). So in order to avoid having this deployment get overriden when deployed again (using auto increment), I'd need to add a timestamp to this.

Lastly, there the open question of how to handle non-auto-incremented version id's deployed multiple times. Do disambiguate these too with a timestamp? Or do we let them override each other.

etanshaul:
@patflynn @joaoandremartins @aslo ready for review.

I've added a timestamp to the deployment display id:
image

This fixes the auto-increment issue I described above

etanshaul:
Minor update. added date to the deployment timestamp:

image

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployable.java
```diff
@@ -29,4 +29,18 @@
    */
   AppEngineEnvironment getEnvironment();
 
+  /**
+   * Set the name of this deployable. Used for display in the Application Server deployment window
```
joaoandremartins: nit: s/Set/Sets?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -346,19 +346,30 @@ private void resetOverridableFields(JCheckBox overrideCheckbox, JTextField field
...
+        + ". Version: " + getDisplayableVersion();
+
     if (isUserSpecifiedPathDeploymentSource()
         && !StringUtil.isEmpty(userSpecifiedArtifactFileSelector.getText())) {
-      ((UserSpecifiedPathDeploymentSource) deploymentSource).setName(
-          GctBundle.message(
-              "appengine.flex.user.specified.deploymentsource.name.with.filename",
-              new File(filePath).getName()));
+      deployable.setName(deployable.getDefaultName() + " - "
+          + new File(userSpecifiedArtifactFileSelector.getText()).getName() + projectVersionSuffix);
```
joaoandremartins: Can we use `Paths` here instead of `new File()`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/UserSpecifiedPathDeploymentSourceType.java
```diff
@@ -64,8 +64,8 @@ public ModuleDeploymentSource load(@NotNull Element tag, @NotNull Project projec
         userSpecifiedSource.setFilePath(filePath);
         userSpecifiedSource.setName(
             GctBundle.message(
-                "appengine.flex.user.specified.deploymentsource.name.with.filename",
-                new File(filePath).getName()));
+                "appengine.flex.user.specified.deploymentsource.name") + " - "
+                + new File(filePath).getName());
```
joaoandremartins: Can we use `Paths`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/UserSpecifiedPathDeploymentSource.java
```diff
@@ -59,13 +60,19 @@ public Icon getIcon() {
   @NotNull
   @Override
   public String getPresentableName() {
-    return name;
+    return String.format("[%s] ", DateTime.now().toString("yyyy-MM-dd HH:mm:ss")) + name;
```
joaoandremartins: Have you given any thought to extracting this method to an abstract class or so? This method and setName() seem to be the same throughout. I don't think this alone warrants creating an abstract class, just want to know what you think.

patflynn:
I like the way this is going.  I think removing the prefix of the artifact type (eg. maven vs filesystem war), makes sense. I'm a bit concerned, as we model the deployments as truly 'deployments' rather than a model of the running services, how users are going to understand the 'stop' command.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/UserSpecifiedPathDeploymentSource.java
etanshaul: I have been pondering this too. Its a little tricky given the inheritance hierarchy we are stuck with given the different deployment types. I am leaning towards saying this is ok because we can think of `presentableName` as being something unique to each individual deployment source, although I agree the duplication of this date format is less than ideal.

etanshaul:
@patflynn - maybe you caught this PR in an intermediate state, but I did not end up going down the route of removing the artifact prefix naming. Modeling the deployments are running services instead of deployments is something I hadn't considered. I think we can achieve this and still indicate where the deployment came from (a maven build, a manually selected jar etc.) by playing with the wording.

etanshaul:
@joaoandremartins I've addressed your comments (except for the one about extracting an abstract class - see my comment inline).

@patflynn @joaoandremartins @aslo Here is an overview of how things stand. Please take a look in addition to the code review to see if you have any comments on this:

Flex deployment sources when you first open up the deployment dialog:
image

The Application Servers panel showing all the running/completed deployments. Note that these labels exactly match the labels shown in the deployment source selector (they come from the same place) plus the extra deploy-time information like project and version:
image

Currently we allow multiple simultaneous deployments of the same project/version combination. We can disallow this in a separate PR.

aslo:
Apologies if this has already been covered, but can you explain the timestamps in the deployment artifact selection dropdown? Is that the time that the artifact was built? I was under the impression that the running deployments would be prefixed with a timestamp in the Application Servers panel to disambiguate them, but I'm not clear on what role the timestamps serve in the deployment artifact selection dropdown.

etanshaul:
Fair question.

I also don't like the timestamps showing up in the deployment source dropdown. This happens because the display ID used for running deployments is the same as the display ID used for the deployment sources in that dropdown (the app server window just lists the running deployment sources). I'd love to get around this but have not come up with a clean way, but I'll keep thinking.

The timestamp is meant to represent the time the deployment began. It does get updated when the user hits "run". Its kinda meaningless when seen in the deployment source dropdown.

etanshaul:
@aslo - I welcome suggestions on how to imporve this. The only thing I can think of doing to get around this is, when the deployment dialog is opened, to rewrite the display name of the deployment sources. This would have to be using something like a regex to remove the timestamp. I don't like this approach but given what we have to work with I'm not sure if we can do better.

etanshaul:
Nevermind - the suggested workaround in my previous comment won't work. We cannot rewrite the deployment source names since a new instance of the dialog is created for a single selected deployment source. The display of the other sources in the dropdown is out of our control

aslo:
Would it be possible to just prepend the deployment name with the timestamp [here](https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1055/files#diff-1ffab44e777b8735ecc22d775aeb8556R360), rather than baking it into the deployable's default name?

etanshaul:
Setting it there does change the name of the deployment source itself. This is how I get version/project to show up in the running deployment window. Adding time stamp to this will change what is shown in the dropdown as well (it still might make sense for me to move it to here just to get rid of the code duplication even though it doesn't fix the problem we are talking about).

joaoandremartins:
I hadn't realised that the timestamps would show in the deployment drop downs too. Is there any way at all we can hide it from there, and just add it on the application servers window?

I've been thinking about simultaneous deployments of the same app and version from the same IDE. Does it really make that much sense? How would this happen, if not by user accidental click or something? My only problem with disallowing concurrent deployments from the same IDE would be stuck deployments/deadlocks.

etanshaul:
Regarding simultaneous deployments - it can happen simply if the user executes a run config multiple times (before the previous ones finished). I don't think its valid really for the same project/version. I'm not sure what you mean by disallowing this causing stuck deployments or deadlocks (wouldn't this prevent them, not cause them)?

I'll get back to you about the first point, I'm still looking into it.

joaoandremartins:
I'm guessing that to disallow concurrent deployments, we'd implement some sort of lock where you can only run one deployment for the same app/version at a time. This can be troublesome if deployments fail or get stuck. I haven't seen that happening very much, though. And I guess the user could always cancel the deployment if it's taking too long?

etanshaul:
To summarize what we discussed offline:

The fixes this PR achieves with the disambiguated deployments is not worth the compromise of poorly named deployment sources in the dropdown. I'll investigate alternative solutions.

To prevent concurrent deployments of the same project/version, we'd ideally like to show a dialog warning the user than running this configuration will override their existing deployment (and probably stop the currently running duplicate if they choose to proceed).

patflynn:
oh yeah. I'd really love to avoid having the timestamp in the source
selector. I think that'll be very confusing.

We should discuss the UX for this offline.

etanshaul:
I think I solved all of our concerns with this after discovering a way to override the display label for the deployment (without affecting what is displayed in the source selector). 

Update coming soon.

etanshaul:
@patflynn @joaoandremartins @aslo Apologies, but this now needs a re-review.

Deployment source labels now do not change:
image

The deployment line item labels are now appropriately customized:
image

I hope this addresses all of the concerns. Some additional improvements:
- Notice above that the deployment label for jar/war files now just shows the name of the archive (previously it has "User specified JAR/WAR ...")
- Deserialization of these run configs now behaves predictably: the labels always appear as expected (previously, a newly deserialized config would initially be missing the project/version label).

etanshaul:
Also note that I am well aware that there is some code duplication. I am running into the limitations of Java here. I'm thinking of options and suggestions are welcome.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSource.java
```diff
@@ -31,19 +32,15 @@
...
+  private String projectName;
+  private String version;
 
   /**
    * Initialize the artifact deployment source given a target App Engine environment, and an
    * artifact pointer.
    */
   public AppEngineArtifactDeploymentSource(
-      @NotNull AppEngineEnvironment environment,
+      @Nullable AppEngineEnvironment environment,
```
joaoandremartins: @Nullable is the default option, right? Can we omit this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineRuntimeInstance.java
```diff
@@ -87,6 +89,38 @@ public void run(@NotNull ProgressIndicator indicator) {
...
+   */
+  @NotNull
+  @Override
+  public String getDeploymentName(@NotNull DeploymentSource source,
+      AppEngineDeploymentConfiguration configuration) {
+    String deploymentName = String.format("[%s] ", DateTime.now().toString("yyyy-MM-dd HH:mm:ss"));
+
+    // If its a user specified archive source then we want to label it by the name of the archive
+    if (source instanceof UserSpecifiedPathDeploymentSource && source.getFile() != null) {
+      deploymentName += source.getFile().getName();
```
joaoandremartins: I must be missing something, but how is the space being added here between the timestamp and the deployable name?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/MavenBuildDeploymentSourceType.java
```diff
@@ -82,13 +84,30 @@ protected BeforeRunTask createBuildTask(@NotNull Module module) {
   public MavenBuildDeploymentSource load(@NotNull Element tag, @NotNull Project project) {
     final String moduleName = tag.getAttributeValue(NAME_ATTRIBUTE);
 
-    return new MavenBuildDeploymentSource(
+    MavenBuildDeploymentSource source = new MavenBuildDeploymentSource(
         ModulePointerManager.getInstance(project).create(moduleName), project);
+    source.setProjectName(tag.getAttributeValue(PROJECT_ATTRIBUTE));
```
joaoandremartins: Should this happen inside the constructor?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/UserSpecifiedPathDeploymentSourceType.java
```diff
@@ -53,7 +50,10 @@ public UserSpecifiedPathDeploymentSourceType() {
   public ModuleDeploymentSource load(@NotNull Element tag, @NotNull Project project) {
     UserSpecifiedPathDeploymentSource userSpecifiedSource =
         new UserSpecifiedPathDeploymentSource(ModulePointerManager
-            .getInstance(project).create(tag.getAttributeValue(NAME_ATTRIBUTE)));
+            .getInstance(project).create(UserSpecifiedPathDeploymentSource.moduleName));
+
+    userSpecifiedSource.setProjectName(tag.getAttributeValue(PROJECT_ATTRIBUTE));
```
joaoandremartins: Should this also be inside the constructor?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineArtifactDeploymentSource.java
patflynn: @joaoandremartins it's actually the other way around typically, but we're exploiting the @NotNull runtime instrumentation that IJ has.  I don't think IJs null dereferencing checks work if you don't use @Nullable

etanshaul: Correct. we need this annotation to get the IDE checks.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineRuntimeInstance.java
etanshaul: Look at the format string: `"[%s] "`. Notice the space after the bracket.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/MavenBuildDeploymentSourceType.java
joaoandremartins: Ah yes, I didn't realise these were being set from an attribute of tag.

etanshaul:
Thanks for the review joao. In addition to this, @patflynn, @joaoandremartins, @aslo. How do you guys feel about the updated UX?

I'd be happy to demo it too if you prefer.

patflynn:
I feel great about it. Good job figuring it out.

aslo:
Yep, I agree. Looks great now!

joaoandremartins:
I have mixed feelings about showing that timestamp. It isn't very meaningful and we're using it just as a way to disambiguate deployments. It seems to be the only way to resolve that problem though, apart from disallowing concurrent deployments of the same app and version. So I think go for it.

etanshaul:
@joaoandremartins that's interesting. To me, that timestamp is very useful: imagine a user executing multiple deployments to some environment. Wouldn't it be great to be able to come back to the IDE, look at the deployment console, and be able to tell which deployment was executed at what time? Either way, like you said, its the best way to disambiguate deployments completely.

joaoandremartins:
Actually, I hadn't thought of that. We don't seem to have deployment information anywhere else, so that sounds good.

PR #1060, opened 2016-10-24
Remove the Managed VM feature flag and checks

### LINKED ISSUE #1032: remove feature flag checks from fully GA'd features

patflynn:
Closes #1032

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/GctFeature.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/GctFeature.java
@@ -28,6 +28,11 @@ import java.util.Set;
  * <p>Enum elements marked {@link Deprecated} are fully released and no longer hidden behind a flag.
  */
 public enum GctFeature implements Feature {
+
+  /**
+   * Deprecating but not removing this feature as it serves as a useful sample and it's kinda nice
+   * to see the {@code GctFeature} history of flags in this enum.
+   */
   @Deprecated
   APPENGINE_FLEX(null, "feature.appengine.flex", "ct4ij.feature.appengine.flex");
 
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/GctFeature.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/GctFeature.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2016 The Android Open Source Project
+ * Copyright 2016 Google Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


patflynn:
@aslo @etanshaul

joaoandremartins:
Should we also set to false, or remove the flag, where we were setting it before? Or was it set to true by default?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/GctFeature.java
```diff
@@ -24,16 +24,34 @@
 
 /**
  * The set of Google Cloud Tools features that we want to be able to conditionally enable.
+ *
+ * <p>Enum elements marked {@link Deprecated} are fully released and no longer hidden behind a flag.
  */
 public enum GctFeature implements Feature {
+  @Deprecated
   APPENGINE_FLEX(null, "feature.appengine.flex", "ct4ij.feature.appengine.flex");
```
joaoandremartins: Why can't we just remove this line? What could be broken?

patflynn: @joaoandremartins nothing. The main reason I left it is that it serves as an example, provides history, and the enum looks weird with no instances.  I'm on the fence about keeping the old ones around.

joaoandremartins: Ack. Could be good to add a clarifying "// Left here just to serve as an example. Don't use." comment. I think that's why you added the @Deprecated annotation for.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/CloudToolsPluginInitializationComponent.java
```diff
@@ -61,38 +47,13 @@ public void initComponent() {
...
 
     new ConflictingAppEnginePluginCheck().notifyIfConflicting();
   }
 
-  private void initAppEngineSupport(
-      CloudToolsPluginConfigurationService pluginConfigurationService) {
-    AppEngineCloudType appEngineCloudType = new AppEngineCloudType();
-    pluginConfigurationService.registerExtension(ServerType.EP_NAME, appEngineCloudType);
-    pluginConfigurationService.registerExtension(ConfigurationType.CONFIGURATION_TYPE_EP,
-        new DeployToServerConfigurationType(appEngineCloudType));
```
etanshaul: Is this extension point also handled by the additions to plugin.xml? I see the line item for `AppEngineCloudType` there, but not `DeployToServerConfigurationType`.

patflynn:
Yeah, I remember having to do a fair bit of spelunking back in the day to
figure out I had to register this manually when I don't use the
plugin.xml.  Still testing to make sure I haven't broken anything. but so
far looks good.

PR #1074, opened 2016-10-27
Hide local dev server as irrelevant if not an ae standard app

### LINKED ISSUE #1067: App Engine Local Development Server should be 'irrelevant' for non AE standard modules/projects
When creating a non AE standard project, the 'Google App Engine Local Development Server' should not show up as one of the 'relevant' run configs for local development.

etanshaul:
fixes #1067

### DIFF
--- a/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/AppEngineServerConfigurationType.java
+++ b/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/AppEngineServerConfigurationType.java
@@ -72,27 +72,27 @@ public class AppEngineServerConfigurationType extends J2EEConfigurationType {
 
   @Override
   public ConfigurationFactory[] getConfigurationFactories() {
-    return new ConfigurationFactory[] {
-       new ConfigurationFactory(this) {
-         @NotNull
-         @Override
-         public RunConfiguration createTemplateConfiguration(@NotNull Project project) {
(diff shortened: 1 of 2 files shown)


google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/AppEngineServerConfigurationType.java
```diff
@@ -72,27 +72,27 @@ public Icon getIcon() {
...
+            return createJ2EEConfigurationTemplate(this, project, true /*isLocal*/);
+          }
+
+          /**
+           * If false, then this run configuration shows up under the "irrelevant" section. If it's
+           * not an App Engine Standard project (doesn't have an appengine-web.xml) then the run
+           * config should show up as "irrelevant".
+           */
+          @Override
+          public boolean isApplicable(@NotNull Project project) {
```
patflynn: tests?

etanshaul: done. tests added

joaoandremartins:
Are we pursuing the other idea we discussed about showing the appengine-web.xml location in the run config UI and invalidating the run config based on that to reduce the amount of magic?

etanshaul:
I think so. But we can still mark it as irrelevant if its not an AE standard app right? This doesn't prevent the user from creating one, just hides it a bit.

joaoandremartins:
Yeah, sounds good. Was wondering if we wanted to do it in the same PR, but it's probably better not to.

PR #1075, opened 2016-10-27
Add various cloud actions to the tools menu

### LINKED ISSUE #1061: Cloud Tools plugin features are hard to discover
We are getting reports that people install the plugin, but have trouble discovering what they can do with it.

The recommendation is to add a 'Google' Sub-menu to the 'Tools' menu.

That menu can include:
- cloud debug 
- deploy to app engine
- checkout/import to GCR
- other stuff

etanshaul:
fixes #1061

**Not complete**. Actions are not implemented etc.  In place to get the conversation going.

image

To discuss:
- What actions do we want to include in the menu
- Menu structure: sub-groups, separators etc.
- We probably need more icons to differentiate between the actions. Missing unique icons for local run, cloud repo, debugger (this icon is hard to differentiate from the App Engine icon).


aslo:
These four options are pretty much what I was thinking should be included here.  I don't know exactly how the grouping could or should work, but it would be cool to group actions by service:
 
* App Engine
  * Deploy to App Engine
  * App Engine Standard Local Dev Server
* Stackdriver
  * Stackdriver Debug 
* Cloud Source Repositories
  * Upload to Google Cloud

etanshaul:
I like it.

joaoandremartins:
I'm not sure I can add much here. Agreed with the available actions, though I'm not sure if a menu/submenu structure makes much sense here, given that there are only a few choices and the menus might make browsing a bit more cumbersome.

etanshaul:
Yea I agree with that. Instead of submenus I think separators might make more sense.

etanshaul:
Something like this:

image

joaoandremartins:
Yeah I like that. Also not sure about the "..." suffix on every option, but I'm guessing we can talk about that later.

etanshaul:
I _think_ the convention for IJ here is to add "..." to actions that don't do something immediately (pop open a dialog or something).

joaoandremartins:
Ah, if that's the case, fair enough.

aslo:
Yep, I believe that's a pretty universal convention (even beyond intellij) with the `...` in menus like this.

patflynn:
I like where this has gone.  If we have an icon for each action we should
probably update the source repo one to be more specific.

etanshaul:
Agreed. It would be nice to also have a better one for Stackdriver (unless this is really the official one we should be using). At this size its hard to distinguish it from the AE icon.

patflynn:
There is no better Stackdriver Debugger icon. That is the official one.  I
agree that they are hard to distinguish in general, and in particular at
this small size.

etanshaul:
@joaoandremartins @aslo @patflynn 

This is ready for review. Here is how it looks in Ultimate:

image

And under Community (local run is not available):

image

Note that local run & stackdriver debug do not have ellipses. This is because clicking on the item will either execute the run configuration - if one exists - or open up the run config editor to create a new one. Deployment always opens up the deployment config window. 

The reason for this is that deployment config has a UI on top of the run config where you can execute the action directly from the panel, whereas local run and stackdriver debug do not - you create the run configuration, save it, then run it.

etanshaul:
Update wording based on feedback from ynr:

image

Note, the wording update to the cloud repo action also updates it in the VCS menu:

image

etanshaul:
todo: I need to hide the local dev server menu item for non-AE-standard apps

joaoandremartins:
Not sure if this is on your radar but, instead of hiding, could we instead disable it? (Less magic)

patflynn:
Wording seems a little off for the import. How about disambiguating that
it's source by using the full product name? e.g. "Upload to a Google Cloud
Source Repository..."

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/CloudToolsRunConfigurationAction.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    final RunnerAndConfigurationSettings settings = RunManagerImpl.getInstanceImpl(project)
+        .findConfigurationByName(configType.getDisplayName());
+
+    ApplicationManager.getApplication().invokeLater(new Runnable() {
+      @Override
+      public void run() {
+        if (settings == null) {
+          createNewConfiguration(project);
```
joaoandremartins: Am I missing anything, or does this option just create and not run the configuration? Should it create and run it?

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -132,7 +137,7 @@ clonefromgcp.parent.missing.error=The parent path must exist.
 clonefromgcp.repository=Cloning source repository {0}
 clonefromgcp.failed=Clone failed
 
-uploadtogcp.text=Upload Project to Google Cloud...
+uploadtogcp.text=Host the code on the Google Cloud Repository...
```
etanshaul: Yeah. I like it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/CloudToolsRunConfigurationAction.java
etanshaul: The way the run configs work in intellij is that after you create them, you simply save them. You then execute them later.

etanshaul:
@joaoandremartins I'm open to the idea of disabling the local dev server instead of hiding it for non ae standard apps.

etanshaul:
@patflynn Sure. Your suggested wording sounds good to me.

aslo:
:+1: to disabling the local dev server for non-standard apps.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/CloudToolsRunConfigurationAction.java
joaoandremartins: Isn't the else part running an existing config, though?
Maybe I'm missing something, but I thought this is when you select an action that runs a configuration, and creates a new one if none exists? Shouldn't the end result be to run a configuration in both cases?

etanshaul: To create a new configuration means to open the Run Config dialog for a given configuration type which only has "Ok", "Apply", and "Cancel". Clicking "Ok" does not run it. It simply instantiates the run config which you can then run later by clicking the play icon (or something similar in the menu).

So in other words, for this scenario the end result is instantiating a run config, not actually running it yet.

joaoandremartins: If I understand correctly, if a run configuration of that kind already exists, it is run instead, right? (by the else snippet)
And I'm guessing if there are multiple run configs of the same kind, only the first one is run?
Anyways, if you're aware of this behaviour and think it's the right thing to do, I trust you.

etanshaul: For instance: the auto created config for local run is named something like "App Engine local run". Say I rename that config to "My local dev server". Then if I use the new menu action it will create a new one because it could not find the existing one by name.

etanshaul: To me this was the best compromise: I didn't want to have to choose one arbitrarily from a list of configs (choosing the first one in the list for example).

joaoandremartins: I'm not so certain on what's more unnatural - to run the config immediately after you create it or to create it and then manually seek the generated config and run it, especially if the config isn't selected by default on the run configs dropdown right after it's created.

It is still possible for the user to create the config and not run it, by selecting Apply instead of OK, or to do neither and just press Cancel.

I don't feel strongly about this, but would actually be curious to know what others think.

etanshaul: The config actually is selected by default in the dropdown after its created (if your not seeing this then its a bug).

But regardless thats how run configs work in IJ like I said, you create them, then run them later. I don't think its a good idea to break this model just for our new menu.

I'm glad you brought all this up since this was one of the challenges with this PR.

etanshaul:
@patflynn @aslo @joaoandremartins 
Ready for final review.

The tools shortcut link to the dev app server now ensures that the project has an app engine standard module. If it does not, then it pops up a bubble notification:

image

The reason I did this instead of disabling the tools menu item is that the determination of if the project is AE standard can be done at the time of clicking the link (the user can re-configure the project at any time).

aslo:
Just wondering, why do you prefer the balloon notification to just disabling the menu item? Would it be possible to just set up a listener on the PSI files in order to disable/enable the menu item as needed?

etanshaul:
Yeah that would be possible (although I'd need to investigate how performant it would be).

But even if so, wouldn't the balloon be more informative + less magic than just disabling the link with no explanation?

aslo:
Good point. My concern was that it might be a bit surprising as a user, but now that I think about it, I suppose that is better than not knowing why something is disabled. I'm ok with it.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -28,7 +28,15 @@ appengine.name=Google App Engine
 appengine.environment.name.flexible=App Engine Flexible Environment
 appengine.environment.name.mvm=App Engine Managed VMs
 appengine.environment.name.standard=App Engine Standard Environment
-appengine.tools.menu.item.label=Deploy to App Engine...
+appengine.tools.menu.deploy.text=Deploy to App Engine...
+appengine.tools.menu.deploy.description=Launch App Engine Deployment Configuration Panel
+appengine.tools.menu.debug.text=Attach Stackdriver Debugger
```
etanshaul: I'm not sure I follow, can you clarify please? Do you mean using the "cloudbug.*" property instead? I think keeping them grouped together like I have will make it convenient if/when we adjust them.

PR #1077, opened 2016-10-31
Adds a Stackdriver facet.

joaoandremartins:
Helps with #450

I think this is ready for review. There isn't a lot of logic in this PR, I can't think of very good unit tests here. Any suggestions there?

Also, would you guys rather I merge this against a feature branch rather than master? If master, then we shouldn't release a new version until I finish the source context stuff.


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetType.java
```diff
@@ -56,12 +56,6 @@ public boolean isSuitableModuleType(ModuleType moduleType) {
     return moduleType instanceof JavaModuleType;
   }
 
-  @NotNull
-  @Override
-  public String getDefaultFacetName() {
```
joaoandremartins: getDefaultFacetName() returns FacetType.myPresentableName, which we're initializing in the constructor as "Google App Engine Standard" (through GctBundle).

PR #1078, opened 2016-10-31
Adds Stackdriver facet.

joaoandremartins:
Helps with #450

Adds a Stackdriver facet that can be selected at New Project time, or added to an existing project via Project Structure -> Facets.

It configures source context file generation so the Google Stackdriver Debugger works with App Engine apps.

It's available for every Java project (even for ones without an App Engine facet) and both on Community and Ultimate edition. This PR doesn't implement framework discovery -- similar to what we have in GAE to prompt the user if they want to add the facet to a project. Need to come up with a reasonable criteria to prompt the user.

screenshot from 2016-11-01 11 07 28

screenshot from 2016-11-01 11 07 50

I think this is ready for review. There isn't a lot of logic in this PR, I can't think of very good unit tests here. Any suggestions there?

This PR is to be merged to the feature branch srcctx-feat-branch, where the logic that hooks into IntelliJ's build workflow and that uses this facet will also be merged. When source context is fully working in that branch, I'll open another PR to merge it with master.

Also, I still haven't been able to find a way to access this facet from a build context. I'm hoping Nikolay/IntelliJ can provide helpful insights in https://intellij-support.jetbrains.com/hc/en-us/community/posts/203374204-Adding-dynamically-generated-file-to-artifact-root, and I'm hoping it's possible to access the facet from the build context.


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetType.java
```diff
@@ -56,12 +56,6 @@ public boolean isSuitableModuleType(ModuleType moduleType) {
     return moduleType instanceof JavaModuleType;
   }
 
-  @NotNull
-  @Override
-  public String getDefaultFacetName() {
```
joaoandremartins: getDefaultFacetName() returns FacetType.myPresentableName, which we're initializing in the constructor as "Google App Engine Standard" (through GctBundle).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverFrameworkType.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.cloud.tools.intellij.util.GctBundle;
+
+import com.intellij.framework.FrameworkTypeEx;
+import com.intellij.framework.addSupport.FrameworkSupportInModuleProvider;
+
+import org.jetbrains.annotations.NotNull;
+
+import javax.swing.Icon;
+
+public class StackdriverFrameworkType extends FrameworkTypeEx {
```
etanshaul: Can you explain why we need a framework type for Stackdriver? It was my understanding that the facet will hook onto the App Engine Standard framework initialization.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverPanel.java
```diff
@@ -0,0 +1,110 @@
...
+  private JPanel stackdriverPanel;
+  private JCheckBox ignoreErrors;
+  private FacetEditorContext editorContext;
+  private StackdriverFacetConfiguration configuration;
+
+  public StackdriverPanel() {
+    generateSourceContext.addActionListener(new ActionListener() {
+      @Override
+      public void actionPerformed(ActionEvent e) {
+        ignoreErrors.setEnabled(!ignoreErrors.isEnabled());
```
etanshaul: is this correct? Looks like this just taggles the ignoreErrors checkbox regardless of what generateSourceContext is set to

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverSupportProvider.java
```diff
@@ -0,0 +1,82 @@
...
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import javax.swing.JComponent;
+
+/**
+ * Adds Google Stackdriver support to IntelliJ IDEA Java projects.
+ */
+public class StackdriverSupportProvider extends FrameworkSupportInModuleProvider {
```
etanshaul: Similar to my question about framework types. Is this really needed?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverPanel.java
joaoandremartins: The problem here is that when you unselect the "Generate source context files" checkbox, `generateSourceContext.isSelected()` only returns `false` after this method returns.

Instead of hacking around that and using an unintuitive `ignoreErrors.setEnabled(!generateSourceContext.isSelected())`, I went for this. These checkboxes are correctly initialized everywhere, so toggling should always work here. I tested a bit and it works as expected.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverSupportProvider.java
joaoandremartins: Response above.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverPanel.java
etanshaul: I find this a bit confusing. What is the UX goal here? To disable ignoreErrors when generateSourceContext is off (and vice versa)? If so you can get the UI component from the ActionEvent param and it should be available immediately.

etanshaul: `event.getSource`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverPanel.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Override
+  public void reset() {
+    if (editorContext.getFacet().getConfiguration() instanceof StackdriverFacetConfiguration) {
+      generateSourceContext.setSelected(configuration.getState().isGenerateSourceContext());
+      ignoreErrors.setEnabled(generateSourceContext.isSelected());
+      ignoreErrors.setSelected(configuration.getState().isIgnoreErrors());
+      return;
+    }
+    // If not a StackdriverFacetConfiguration, use defaults, though this should never happen.
```
aslo: Out of curiosity, when would this happen?

aslo:
Since you asked about testing, I'd say it would be nice to unit test some of the logic in `StackdriverPanel`, but TBH I have no idea how to go about testing those classes that have UI elements magically injected into them. That might be automatically wired up for you if you extend `LightPlatformTestCase` but I'm not certain if it's even worth your time if you're just testing those few lines of code.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverPanel.java
etanshaul: @aslo I agree. I was suggesting a way to get around the checkbox state not updating until the callback returns. If you get the UI element from the event param, it should be fresh.

joaoandremartins:
@aslo indeed, I was also thinking about that tiny piece of logic in StackdriverPanel as the only testable code block here, but it's so basic that I'm not sure if it's worth the hassle of setting up tests just for it...

@patflynn added UI pics and more text to the PR's description.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverFacetType.java
```diff
@@ -0,0 +1,64 @@
...
+
+import javax.swing.Icon;
+
+public class StackdriverFacetType
+    extends FacetType<StackdriverFacet, StackdriverFacetConfiguration> {
+  public static final String STRING_ID = "stackdriver";
+  public static final FacetTypeId<StackdriverFacet> ID = new FacetTypeId<>(STRING_ID);
+
+  public StackdriverFacetType() {
+    super(ID, STRING_ID, GctBundle.getString("clouddebug.stackdriver.debugger"));
```
patflynn: this isn't the write wording here since it's not just for debugger.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverFrameworkType.java
patflynn: javadoc.  Looks like there's some interesting conversation going on here that could be captured.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverFrameworkType.java
```diff
@@ -0,0 +1,58 @@
...
+  @NotNull
+  @Override
+  public FrameworkSupportInModuleProvider createProvider() {
+    return new StackdriverSupportProvider();
+  }
+
+  @NotNull
+  @Override
+  public String getPresentableName() {
+    return GctBundle.getString("clouddebug.stackdriver.debugger");
```
patflynn: wrong name again.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverPanel.java
```diff
@@ -0,0 +1,110 @@
...
+ */
+public class StackdriverPanel extends FacetEditorTab {
+
+  private JCheckBox generateSourceContext;
+  private JPanel stackdriverPanel;
+  private JCheckBox ignoreErrors;
+  private FacetEditorContext editorContext;
+  private StackdriverFacetConfiguration configuration;
+
+  public StackdriverPanel() {
```
patflynn: do we need this no-arg constructor?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverPanel.java
```diff
@@ -0,0 +1,110 @@
...
+    generateSourceContext.addActionListener(new ActionListener() {
+      @Override
+      public void actionPerformed(ActionEvent e) {
+        ignoreErrors.setEnabled(((JCheckBox)e.getSource()).isSelected());
+      }
+    });
+  }
+
+  public StackdriverPanel(FacetEditorContext editorContext) {
+    this();
```
patflynn: FYI this is considered bad form. Constructor chaining should go the other way, from the zero-arg constructor, to the the unary constructor, and on.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverPanel.java
```diff
@@ -0,0 +1,110 @@
...
+      return;
+    }
+    // If not a StackdriverFacetConfiguration, use defaults, though this should never happen.
+    generateSourceContext.setSelected(true);
+    ignoreErrors.setSelected(true);
+  }
+
+  @Override
+  public void disposeUIResources() {
+
```
patflynn: // do nothing.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverPanel.java
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  @Nls
+  @Override
+  public String getDisplayName() {
+    return GctBundle.getString("clouddebug.stackdriver.debugger");
```
patflynn: text.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -193,7 +193,7 @@ clouddebug.errorset=The snapshot location could not be set.
 clouddebug.continuesession=Continue Session
 clouddebug.continueanyway=Continue Anyway
 clouddebug.text=Google Stackdriver Debug
-clouddebug.stackdriver.debugger=Stackdriver Debugger
+clouddebug.stackdriver.debugger=Google Stackdriver Debugger
```
joaoandremartins: Only one place, CloudDebugConfigType.getConfigurationTypeDescription().

Although the Javadoc of this method in ConfigurationType mentions we might want to us the same text as the display name of the configuration type, which is "Google Stackdriver Debug" (clouddebug.text), to keep in line with the existing convention, i.e., call it a Debug run config, not a Debugger run config.

I can do a quick PR into master after we decide: do we want to make getConfigurationTypeDescription() return the same thing as getDisplayName()? I think we should. In that case, clouddebug.stackdriver.debugger will solely be used by the facet.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverFacet.java
```diff
@@ -0,0 +1,33 @@
...
+
+package com.google.cloud.tools.intellij.debugger.facet;
+
+import com.intellij.facet.Facet;
+import com.intellij.facet.FacetType;
+import com.intellij.openapi.module.Module;
+
+import org.jetbrains.annotations.NotNull;
+
+public class StackdriverFacet extends Facet<StackdriverFacetConfiguration> {
```
joaoandremartins: I thought about adding Javadoc, but what to say in it?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverPanel.java
joaoandremartins: Yes, it's called from the New Project dialog, through `StackdriverSupportProvider`, which doesn't have a `FacetEditorContext` to pass, but also doesn't use the same `apply()` method that the facet way uses to persist the configuration.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverFrameworkType.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @NotNull
+  @Override
+  public Icon getIcon() {
+    return GoogleCloudToolsIcons.STACKDRIVER_DEBUGGER;
```
joaoandremartins: Ah, wasn't aware of that. Before I go on and make changes here, what source context usages should I also be aware of? Using it with App Engine (or GCP for that matter) is the only one I knew of.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverFacetType.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Override
+  public boolean isSuitableModuleType(ModuleType moduleType) {
+    return moduleType instanceof JavaModuleType;
+  }
+
+  @Nullable
+  @Override
+  public Icon getIcon() {
+    return GoogleCloudToolsIcons.STACKDRIVER_DEBUGGER;
```
joaoandremartins: How so? We're really only offering the Stackdriver functionality here. Unless there are plans to expand it to other Stackdriver features? Would be nice to know which ones they would be.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverPanel.java
joaoandremartins: Thanks, didn't know about that.

patflynn:
Any time you want to tie information about a running binary to code you
need source context. Cloud Logging, Trace, Errors (via logging), Debugging,
will all benefit from it.  Frankly I can imagine it being used for non
Stackdriver scenarios too.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverFacetType.java
joaoandremartins: Since this isn't just for the debugger, I'm also guessing we should relocate this from the debugger package to a new stackdriver one?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/facet/StackdriverFacetType.java
joaoandremartins: I requested one from Yann. In the meantime, I added a TODO so I don't get stuck because of this. It's not a PR to master, but to a feature branch, so I think it's OK to go like this for now.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetType.java
joaoandremartins: Don't know why these changes are still showing up here, when I've already submitted them in other PRs, rebased both branches with origin/master and pushed.

joaoandremartins:
@patflynn @aslo @etanshaul PTAL

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/stackdriver/facet/StackdriverFacetConfiguration.java
```diff
@@ -0,0 +1,89 @@
...
+public class StackdriverFacetConfiguration
+    implements FacetConfiguration, PersistentStateComponent<StackdriverProperties> {
+
+  private StackdriverProperties persistedProperties = new StackdriverProperties();
+
+  @Override
+  public FacetEditorTab[] createEditorTabs(FacetEditorContext editorContext,
+      FacetValidatorsManager validatorsManager) {
+    return new FacetEditorTab[]{
+        new StackdriverPanel(editorContext)
```
joaoandremartins: #1085 makes this constructor take a StackdriverProperties object everywhere.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/stackdriver/facet/StackdriverPanel.java
```diff
@@ -0,0 +1,122 @@
...
+
+  @NotNull
+  @Override
+  public JComponent createComponent() {
+    return stackdriverPanel;
+  }
+
+  @Override
+  public boolean isModified() {
+    return generateSourceContext.isSelected() != configuration.getState().isGenerateSourceContext()
```
etanshaul: looks like `getState()` is nullable (this comment applies to all usages of this in this class)

etanshaul: actually nm, looks like the properties class will never allow this to happen

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/stackdriver/facet/StackdriverPanel.java
```diff
@@ -0,0 +1,122 @@
...
+    configuration.getState().setIgnoreErrors(ignoreErrors.isSelected());
+  }
+
+  @Override
+  public void reset() {
+    if (editorContext.getFacet().getConfiguration() instanceof StackdriverFacetConfiguration) {
+      generateSourceContext.setSelected(configuration.getState().isGenerateSourceContext());
+      ignoreErrors.setEnabled(generateSourceContext.isSelected());
+      ignoreErrors.setSelected(configuration.getState().isIgnoreErrors());
+      return;
```
etanshaul: `else {` instead of `return`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/stackdriver/facet/StackdriverPanel.form
```diff
@@ -0,0 +1,36 @@
...
+          <grid row="2" column="0" row-span="1" col-span="1" vsize-policy="6" hsize-policy="1" anchor="0" fill="2" indent="0" use-parent-layout="false"/>
+        </constraints>
+      </vspacer>
+      <component id="c788" class="javax.swing.JCheckBox" binding="ignoreErrors">
+        <constraints>
+          <grid row="1" column="0" row-span="1" col-span="1" vsize-policy="0" hsize-policy="3" anchor="8" fill="0" indent="2" use-parent-layout="false"/>
+        </constraints>
+        <properties>
+          <selected value="true"/>
+          <text value="&amp;Ignore errors"/>
```
etanshaul: Maybe its just me but I didn't immediately know what "ignore errors" does. Maybe we can be more descriptive in the UI?

etanshaul:
From the code in the UI panel it looks like you intended "ignore errors" to be disabled when "source context" is unchecked. Although this isn't the behavior I see when loading up the facet editor.

etanshaul:
A couple more notes on the facet UI panel:

The facet itself is called Google Stackdriver. However, your settings appear to apply only to Stackdriver Debugger. Perhaps you could surround those settings with a bordered panel with a title of Stackdriver Debugger?

Also, while this may be best left to a separate PR, the experience could be improved with more wording and / or links to the documentation to help the user understand what these features are for.

etanshaul:
On second thought - source context could be used for more products than just debugger. I'm not sure about "ignore errors" though?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/stackdriver/facet/StackdriverFacet.java
```diff
@@ -0,0 +1,33 @@
...
+
+package com.google.cloud.tools.intellij.stackdriver.facet;
+
+import com.intellij.facet.Facet;
+import com.intellij.facet.FacetType;
+import com.intellij.openapi.module.Module;
+
+import org.jetbrains.annotations.NotNull;
+
+public class StackdriverFacet extends Facet<StackdriverFacetConfiguration> {
```
patflynn: javadoc

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/stackdriver/facet/StackdriverPanel.java
```diff
@@ -0,0 +1,122 @@
...
+import javax.swing.JComponent;
+import javax.swing.JPanel;
+
+/**
+ * Stackdriver facet's user interface.
+ */
+public class StackdriverPanel extends FacetEditorTab {
+
+  private JCheckBox generateSourceContext;
+  private JPanel stackdriverPanel;
```
patflynn: if this is the enclosing panel, I'd be tempted to put it at the top before I declare the elements that it's composed of.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/stackdriver/facet/StackdriverPanel.java
joaoandremartins: I should probably remove @Nullable from getState()

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/stackdriver/facet/StackdriverPanel.java
```diff
@@ -0,0 +1,122 @@
...
+
+  @Override
+  public void apply() throws ConfigurationException {
+    configuration.getState().setGenerateSourceContext(generateSourceContext.isSelected());
+    configuration.getState().setIgnoreErrors(ignoreErrors.isSelected());
+  }
+
+  @Override
+  public void reset() {
+    if (editorContext.getFacet().getConfiguration() instanceof StackdriverFacetConfiguration) {
```
joaoandremartins: Regarding this method, can you check the final version on #1085? I changed it around quite a bit and I think your points here became addressed.

patflynn:
I agree with @etanshaul wrgt the layout of the checkboxes.  They should be within a bordered titled panel that includes a description of what Stackdriver is and what the src ctxt is for.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/stackdriver/facet/StackdriverPanel.java
```diff
@@ -0,0 +1,122 @@
...
+    // Do nothing.
+  }
+
+  @Nls
+  @Override
+  public String getDisplayName() {
+    return GctBundle.getString("stackdriver.name");
+  }
+
+  public boolean isGenerateSourceContextSelected() {
```
joaoandremartins: good stuff.

joaoandremartins:
I agree with adding a link to documentation, but disagree on adding more text to that pane -- isn't that what documentation is for?

I added a tooltip for ignore errors, though I don't know what else to add for "Generate source context files"?

What's a bordered tilted box and what does it add to what's currently in the UI?

joaoandremartins:
@etanshaul I can't reproduce that behaviour with the Ignore errors box not disabling. I've tried it from the Facets menu and the Add Framework Support bit, in both places the box gets disabled.

joaoandremartins:
What do you guys think about this? I'm linking to https://cloud.google.com/stackdriver/ lacking a better page...

screenshot from 2016-11-17 16 06 33

joaoandremartins:
Actually, it will have to be this instead, since I couldn't find a way to wrap the text in this dialog.

screenshot from 2016-11-17 16 17 49

I also added a tooltip to Generate source context files that reads "Some features, like the Stackdriver Debugger, require source context information in order to function correctly"

patflynn:
assuming @etanshaul approves, let's merge this and we can revise the wording with Yann before we pull all this into master

joaoandremartins:
@etanshaul LMKWYT!

joaoandremartins:
Will merge this to the feature branch so we can make one final review there.

PR #1081, opened 2016-11-01
Validate java components installation at deploy config time

### LINKED ISSUE #1080: Warn the user if the java component is missing at deployment configuration time
before the deployment actually runs

etanshaul:
fixes #1080 

Now at deployment configuration, if the user is missing the component, this will be shown and deployment will be prevented:

image


etanshaul:
TODO: suppress this for flex deployments

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -384,6 +387,18 @@ private void validateConfiguration() throws ConfigurationException {
         throw new ConfigurationException(
             GctBundle.message("appengine.flex.config.custom.dockerfile.error"));
       }
+    } else if (environment.isStandard()) {
```
patflynn: test?

etanshaul:
Click run will show a popup with the same message. Working on tests now.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditorTest.java
```diff
@@ -60,6 +75,7 @@ public void testValidSelections() {
     AppEngineDeploymentConfiguration config = new AppEngineDeploymentConfiguration();
     config.setCloudProjectName("test-cloud-proj");
     config.setConfigType(ConfigType.AUTO);
+    when(cloudSdkService.hasJavaComponent()).thenReturn(true);
```
patflynn: Should we have a test that checks the validation error is shown when this is false?

etanshaul: Yep. About to commit one.

etanshaul:
Tests added. Also I refactored the java component validation so that the call to the CloudSdk is wrapped in a service so it can be mocked.

joaoandremartins:
Why do you want to suppress this for Flex deployments? gcloud app deploy still needs app-engine-java no?

joaoandremartins:
Also, since you're adding `CloudSdkService.hasJavaComponent()` in this PR, you could probably also modify `AppEngineStandardDeployTask` to check for the Java component where we're catching `AppEngineJavaComponentsNotInstalledException`, while keeping the same error flow/reporting.

etanshaul:
@joaoandremartins It's my understanding that flex deployment does not require that component.

I'll look into updating the other locations using that check, as you suggested.

aslo:
Will the user be notified if they configure their run configuration successfully, but then remove the required java component from gcloud? Is that in scope for this PR?

etanshaul:
Yes the user will be notified under that scenario because the validation happens each time the dialog is opened / updated

etanshaul:
@joaoandremartins actually I may have misunderstood your comment about `AppEngineStandardDeployTask`.  How did you suggest I update it?

If you meant that I should use the new method in CloudSdkService, I don't think it applies in this specific case: AppEngineStandardDeployTask (or its downstreem classes) doesn't directly call the validate method. It happens internally in the common lib when the appcfg command is executed for staging.

joaoandremartins:
Yes, I understand the validation happens internally, but it is still possible to directly call `CloudSdkService.getIsntance().hasJavaComponent()` before making the actual deploy call, right?

It would simplify the code there a little bit, and probably save us from doing some unnecessary setup work in case Cloud SDK doesn't contain the GAE Java SDK.

etanshaul:
Oh I see. I think that would be kind of redundant since we already perform the check in the UI and then again in the common lib right after calling staging. There's not much going on in between invoking staging and the internal validation check.

Also I'm not sure how you mean it would simplify the code because we'd still want to catch the exception thrown by staging (unless you mean just removing it since its unchecked..which I'd prefer not to so that we keep the check as "fresh" as possible).

joaoandremartins:
We perform the check in the UI, but it's still possible for the user to press "Continue anyway", and we observed some users tend to do that.
I would even call hasJavaComponent() before staging, so you wouldn't have to create anything there (logging handler, create staging directory, etc.). The first thing that would happen in `AppEngineStandardDeployTask.execute()`, after the metrics ping.

You're right, we still need to catch the `RuntimeException` so the code is maybe not that much simplified. I still think it's worth it to fail as early as possible. And, in general, I like to avoid catching exceptions and instead use control ifs. (Even though an exception would still be thrown and caught, but away from our eyes, inside `hasJavaComponent()`)

etanshaul:
I don't know if this changes your perspective here, but there's no "continue anyway" for deployment as there is for local run.

joaoandremartins:
Ah yes, that changes everything! I'm cool then.

etanshaul:
cool. thanks.

@patflynn are you good with the test coverage now?

PR #1082, opened 2016-11-01
Remove unnecessary method declaration.

joaoandremartins:
getDefaultFacetName() returns FacetType.myPresentableName, which we're initializing in the constructor as "Google App Engine Standard" (through GctBundle).


etanshaul:
this is a bit granular no? :)

joaoandremartins:
We love teh grainz

patflynn:
we do indeed. smaller is almost always better.

PR #1083, opened 2016-11-01
Fix naming consistency around Stackdriver Debugger

joaoandremartins:
Adds missing "Google" prefix to Stackdriver Debugger and adds consistency to naming, since `ConfigurationType.getConfigurationTypeDescription()` Javadoc suggests that the same text as the display name should be returned. The display name is currently "Google Stackdriver Debug" which is fine for a run config name, the same way the debug run config is called "Debug" and not "Debugger".


etanshaul:
where is ConfigurationType.getConfigurationTypeDescription() displayed in the UI?

joaoandremartins:
Bottom left (jiberish in caps).

screenshot from 2016-11-01 15 45 16

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -193,7 +193,7 @@ clouddebug.errorset=The snapshot location could not be set.
 clouddebug.continuesession=Continue Session
 clouddebug.continueanyway=Continue Anyway
 clouddebug.text=Google Stackdriver Debug
-clouddebug.stackdriver.debugger=Stackdriver Debugger
+clouddebug.stackdriver.debugger=Google Stackdriver Debugger
```
etanshaul: is this still being used now that you changed `getConfigurationTypeDescription`?

joaoandremartins: Good point. It's going to not be, but then be again by #1078 and anything that wants to use it in the meantime.

PR #1084, opened 2016-11-02
Implement cloud sdk version checks

aslo:
fixes https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1051

Warns the user if they are using an incompatible version of the Cloud SDK. Wherever possible, if an unsupported Cloud SDK version is found, the message will be displayed as a warning, not an error (it will not block the user from performing whatever action they were trying to do).

This check occurs in all of the following places: 
1. as an application-level balloon notification, triggered
   a. when a project is opened
   b. when a local dev server run is started
   c. when an app engine deployment run is started
2. in the cloud sdk panel (which appears in a couple places)
  a. when creating a new app engine project
  b. in the application servers menu
  c. in the Clouds menu

See screenshots below:

1. as an application-level balloon notification
balloon

2a. in the cloud sdk panel - new app engine project
cloudsdkpanel_newproject

2b. in the cloud sdk panel -  application servers menu
cloudsdkpanel_appservers

2c. in the cloud sdk panel -  clouds menu
cloudsdkpanel_clouds


aslo:
@etanshaul @joaoandremartins  @patflynn Can you think of anywhere else we might want to perform this check? Is gcloud `131.0.0` required for deployments as well?

etanshaul:
why not perform this check on the cloud sdk panel itself?

etanshaul:
Also, have you considered making this a warning instead of an error? Meaning they can still proceed at their own risk with local run / deployment? If you move this check to the cloud sdk panel I could see it functioning like this.

aslo:
I did experiment with adding it to the cloud sdk panel - though as I understand it, that means the user would not see the error/warning unless they open the panel, which they might not if these run configurations are automatically configured for them.

etanshaul:
Even if the configurations are automatically configured, they still need to select the cloud sdk path.

But I guess if its autodetected on their path then we run into some difficulty..so you're right in this case.

What if its added in all places? Cloud SDK panel, local run config, and deployment.

patflynn:
you guys don't like doing this as an IDE notification?  Feels like the
cleanest way to handle this to me.

etanshaul:
I like that idea. Whenever the CloudSdk path is set and is an old version then we could show a notification popup warning.

aslo:
What about if we trigger the IDE notification when they perform a local run or deploy run config? I like the idea of scoping the notification to the places where it's relevant.

aslo:
Something like "This version is not supported. Please upgrade or you may see weird behavior"

etanshaul:
It would be nice for them to be able to fail fast: e.g. not have to actually run a failing deployment before fixing their gcloud version.

Unless your implying the notification would block the action?

patflynn:
The benefit is when they do a plugin update and their local gcloud is out
of date, they know about it immediately, rather than when they want to go
deploy.

aslo:
@patflynn I see your point, though it would also be nice to only warn within the context where it's relevant. i.e. it might be weird to see a prompt to upgrade the cloud sdk when you are working on an unrelated non-appengine project.

aslo:
I don't feel strongly about it though - I think the "fail early" principle might take precedence over that.

etanshaul:
Since our plugin is not currently bundled with the ide I see your point about working on unrelated projects as less of a concern.

chanseokoh:
> why not perform this check on the cloud sdk panel itself? if you do that, we'll ensure we're covered in all cases.

I may be missing something here, but IMO, it's always good to check the version here and there whenever necessary. For example, you may have a second laptop at home which has an old gcloud version set up in the past. You could check out an App Engine project on that laptop and try to deploy. I think a version check before deploying is a good idea here.

@patflynn's idea about notification could be a solution too. However, even in that case, users could still try deploying without version upgrade, so I think having a check is good.

patflynn:
Yeah that's fair.  If the user ignores the notification, it makes sense to
remind them when they're performing an action that could fail due to this.

aslo:
I agree. Another reason that's necessary: the user might not have a cloud sdk configured at application/project initialization, so we would not be able to perform this check at that time anyway.

etanshaul:
Personally I like having this as a warning - not blocking deployment or local run.

If perform the check at project init _and_ at sdk setting time aren't we covered?

patflynn:
yeah I agree that it shouldn't block the execution. But you can have a
warning on a run config that does that without blocking the run right?  The
way I'm picturing it is that the 'server' or 'cloud' icon has a little red
'x' on it, and the validation for the version is on the cloud SDK config
screen.

etanshaul:
Yeah that is fine. Although I don't think we can add an icon next to the 'server' or 'cloud' ui elements since these are provided by framework components. The best we could do is something on bottom of the panel.

aslo:
@joaoandremartins @etanshaul @patflynn Requesting a code review. The PR description has been updated to list all of the places this warning will occur. 

I also have one last question here - in both the Local Dev run configuration and the App Engine Deployment run configuration, this check is necessarily called synchronously (because validation errors on the run configurations are shown by the framework when our code throws a `ConfigurationException`). Calling `CloudSdk.getVersion` under the covers can take over 200ms and blocks the UI thread and makes things feel rather slow.  (I was able to get around this in the CloudSdkPanel by performing this check in a background thread, but that isn't possible for the run configurations because of how the run configurations validation framework works.

I propose removing this check from the run configurations and instead just firing off the application-level balloon notification when they run the local dev server or a deployment. Thoughts?

aslo:
Another option to fix this performance issue would be to cache the cloud sdk version at the intellij application level, for a given cloud sdk directory, and bust the cache when we detect that the cloud SDK's RELEASE_NOTES file was updated. I think I actually like that better.

etanshaul:
I do prefer this check as a notification that doesn't block execution. I personally don't like having this throw a ConfigurationException in our UI panels. But I think there may have been some disagreement on this.

I'm thinking through your other comments.

aslo:
FYI there are classes of ConfigurationExceptions that intellij treats as warnings (which this PR currently uses here), that show up as an error state but don't block execution. It still might not be completely clear to the user that they can still run it, though. I'll add some screenshots to the PR description to make this discussion easier.

google-cloud-tools-plugin/resources/config.properties
```diff
@@ -19,3 +19,5 @@ feature.appengine.flex=true
 
 # Should be used in any Notifications explicitly produced by the plugin.
 notifications.plugin.groupdisplayid=Google App Engine plugin
+
+cloudsdk.required.version=131.0.0
```
etanshaul: `cloudsdk.min.version`?

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -49,6 +49,8 @@ appengine.cloudsdk.location.browse.window.title=Browse to Cloud SDK directory
 appengine.cloudsdk.location.invalid.message=No Cloud SDK was found in this directory.
 appengine.cloudsdk.location.missing.message=Cloud SDK home directory is not specified.
 appengine.cloudsdk.download.message={0}Click here{1} to download the Cloud SDK.
+appengine.cloudsdk.version.support.title=Google Cloud SDK Update Required
+appengine.cloudsdk.version.support.message=The configured Google Cloud SDK is out of date. Version {0} is required by the Google Cloud Tools Plugin. To update, run "gcloud components update".
```
etanshaul: Version ... or higher ... is required

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -70,46 +71,91 @@ public CloudSdkPanel() {
...
           }
         });
 
-    checkSdk();
+    checkSdkInBackground();
   }
 
-  private void checkSdk() {
-    String path = cloudSdkDirectoryField.getText();
+  private void checkSdkInBackground() {
```
etanshaul: what if the user closes / saves the panel before the thread completes?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -363,6 +367,22 @@ private String getDisplayableVersion() {
   }
 
   private void validateConfiguration() throws ConfigurationException {
+    CloudSdkService sdkService = CloudSdkService.getInstance();
+
+    // Perform cloud sdk validation
+    Path cloudSdkPath = sdkService.getSdkHomePath();
+    Set<CloudSdkValidationResult> results = sdkService.validateCloudSdk(cloudSdkPath);
```
etanshaul: Can the java components check (see the last else if) also be encapsulated into your new validation model?

aslo: Good idea - I don't see why not. I'll add it.

aslo: On second thought, this might not be worth doing - the java component is only required for AE standard deployments, and up until now, the validations have been treated like more general issues with a cloud sdk installation. 

Right now there isn't a great way for callers to customize which validations they care about. That's something we could consider in the future, but adding that would expand the scope of this PR further than I'd like to.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
aslo: I've manually tested, and this isn't an issue in that case. Since all UI updates we perform happen within calls to `Application.invokeLater`, intellij handles making sure we aren't updating anything that's in an inoperable state. Also, note that the method explicitly handles the case when the cloudsdk path is null.

I will do some small refactoring to hopefully make it more obvious that we won't run into problems if the panel is closed before the thread completes.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -70,46 +71,85 @@ public CloudSdkPanel() {
...
-      warningIcon.setVisible(true);
-      warningMessage.setVisible(true);
-      warningMessage.setText(
-          createErrorMessageWithLink(
-              GctBundle.message("appengine.cloudsdk.location.invalid.message")));
+    Set<CloudSdkValidationResult> validationResults =
+        CloudSdkService.getInstance().validateCloudSdk(Paths.get(path));
+
+    // if there are any validation errors, show the first one to the user
+    if (validationResults.size() > 0) {
```
etanshaul: minor: `!validationResults.isEmpty()`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -70,46 +71,85 @@ public CloudSdkPanel() {
...
+      CloudSdkValidationResult validationResult = validationResults.iterator().next();
+      String message;
+      switch (validationResult) {
+        case CLOUD_SDK_NOT_FOUND:
+          message = appendCloudSdkDownloadMessage(validationResult.getMessage());
+          break;
+        case CLOUD_SDK_VERSION_NOT_SUPPORTED:
```
etanshaul: since this is the same is that default case I'm pretty sure you can combine them:

```
case CLOUD_SDK_VERSION_NOT_SUPPORTED:
default:
 message = ....
```

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -70,46 +71,85 @@ public CloudSdkPanel() {
...
+        cloudSdkDirectoryField.getTextField().setForeground(JBColor.black);
+        warningIcon.setVisible(false);
+        warningIcon.setVisible(false);
+        warningMessage.setVisible(false);
+      }
+    });
+  }
+
+  private void invokePanelValidationUpdate(Runnable runnable) {
+    ApplicationManager.getApplication().invokeLater(runnable,
```
etanshaul: I'm a little lost here in this class. Aren't these actions already happening from the UI thread?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
```diff
@@ -39,6 +42,10 @@ public static CloudSdkService getInstance() {
 
   public abstract void setSdkHomePath(String path);
 
+  public abstract Set<CloudSdkValidationResult> validateCloudSdk(@NotNull CloudSdk sdk);
```
etanshaul: I couldn't see where this overloaded version is used?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploy.java
```diff
@@ -118,6 +120,11 @@ public void onOutputLine(String output) {
         deployLogListener,
         deployOutputListener,
         deployExitListener);
+
+    // show a warning notification if the cloud sdk version is not supported
+    new CloudSdkVersionCheck().performCloudSdkVersionCheck(
```
etanshaul: I wonder if there is a way to get a handle on the singleton, instead of instantiating a new one, since we are already using the intellij container to create this class as a StartupActivity.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
aslo: :+1:

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
aslo: No, because showWarning and hideWarning are called within `checkSdk` which is always called by a background thread. See `checkSdkInBackground`.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
aslo: Good catch, looks like it's not in use at all, except in tests. I'll remove.

google-cloud-tools-plugin/resources/config.properties
```diff
@@ -19,3 +19,5 @@ feature.appengine.flex=true
 
 # Should be used in any Notifications explicitly produced by the plugin.
 notifications.plugin.groupdisplayid=Google App Engine plugin
+
+cloudsdk.min.version=131.0.0
```
joaoandremartins: New line missing

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
```diff
@@ -39,6 +40,8 @@ public static CloudSdkService getInstance() {
 
   public abstract void setSdkHomePath(String path);
 
+  public abstract Set<CloudSdkValidationResult> validateCloudSdk(@NotNull Path pathToCloudSdk);
```
joaoandremartins: Pretty cool, even though I would really like to see something like this actually happening at the tools library. (and meant to do so whenever I had some free time)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -70,46 +71,82 @@ public CloudSdkPanel() {
...
-          createErrorMessageWithLink(
-              GctBundle.message("appengine.cloudsdk.location.invalid.message")));
+    Set<CloudSdkValidationResult> validationResults =
+        CloudSdkService.getInstance().validateCloudSdk(Paths.get(path));
+
+    // if there are any validation errors, show the first one to the user
+    if (!validationResults.isEmpty()) {
+      CloudSdkValidationResult validationResult = validationResults.iterator().next();
+      String message;
+      switch (validationResult) {
```
joaoandremartins: Isn't CLOUD_SDK_VERSION_NOT_SUPPORTED missing here?

google-cloud-tools-plugin/resources/config.properties
aslo: Is that part of the java style guide? (Though I'm not sure if that would apply to properties files).  This is something that I don't really feel strongly about, but I'm not sure this is really worth changing.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
aslo: I think we could use validations like this at both levels. For example, checking the sdk version is really specific to the intellij plugin. It would be nice to pull some of this validation into the common library in an extensible way.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
aslo: Yep, but it's covered by the default case

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -70,46 +71,82 @@ public CloudSdkPanel() {
...
 
-    checkSdk();
+    checkSdkInBackground();
   }
 
-  private void checkSdk() {
-    String path = cloudSdkDirectoryField.getText();
+  private void checkSdkInBackground() {
+    final String path = cloudSdkDirectoryField.getText();
+    ApplicationManager.getApplication().executeOnPooledThread(new CloudSdkChecker(path));
```
aslo: I'd prefer to rename. The main reason I created it was to make it more obvious that it is operating on the cloudSdkPath that is in the text box at the time that this runnable is submitted to the background thread executor.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploy.java
aslo: I could just register this class as an applicationService as well - that would make it straightforward to use it in a more idiomatic way from here. What do you think about that?

etanshaul: Yea sounds good. I'm wondering though if classes registered as a StartupActivity are already implicitly registered as services.

google-cloud-tools-plugin/resources/config.properties
joaoandremartins: I think it's Java best practices (as evidenced by the red sign at the end of the line GitHub displays).

It's just adding an enter and committing and pushing again, why is that an issue? :)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
joaoandremartins: Checking the version is common to every plugin that uses the library, since they all support the same operations. I'm pretty sure we could completely offload this to the library, but wouldn't recommend doing so in this PR since it can still be quite a workload.

aslo: Fair point, though each consumer of the common lib might want to require (or recommend) different versions so we would want to allow that flexibility. It would be cool to share this logic with the common lib nonetheless.

google-cloud-tools-plugin/resources/config.properties
aslo: No issue, if you feel strongly I'm happy to add the newline. (Though I'll point out that it's not mentioned in the style guide).

patflynn: @aslo @joaoandremartins +1 to new line at end of file.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -70,46 +71,82 @@ public CloudSdkPanel() {
...
-      cloudSdkDirectoryField.getTextField().setForeground(JBColor.red);
-      warningIcon.setVisible(true);
-      warningMessage.setVisible(true);
-      warningMessage.setText(
-          createErrorMessageWithLink(
-              GctBundle.message("appengine.cloudsdk.location.invalid.message")));
+    Set<CloudSdkValidationResult> validationResults =
+        CloudSdkService.getInstance().validateCloudSdk(Paths.get(path));
+
+    // if there are any validation errors, show the first one to the user
```
patflynn: @aslo I'm coming in a little late here.  Why only show one validation result?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -70,46 +71,82 @@ public CloudSdkPanel() {
...
 
-    checkSdk();
+    checkSdkInBackground();
   }
 
-  private void checkSdk() {
-    String path = cloudSdkDirectoryField.getText();
+  private void checkSdkInBackground() {
+    final String path = cloudSdkDirectoryField.getText();
+    ApplicationManager.getApplication().executeOnPooledThread(new CloudSdkCheckerRunnable(path));
```
aslo: This is true, but that's kind of the tradeoff of validating in non-blocking way. I'm not sure of how to get around that.

patflynn: could you only perform this validation post 'apply'?

google-cloud-tools-plugin/resources/config.properties
joaoandremartins: It seems that several tools panic over the lack of new lines at the end of text files, or add noise when you're closing them.

http://stackoverflow.com/questions/4700312/empty-new-line-at-the-end-of-the-java-source-files

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
aslo: I put this limit here because I was worried about visual overflows resulting from displaying an unbounded number of validation errors. Do you think that isn't something worth worrying about now? (Especially since currently, it's only possible for `validateCloudSdk` to return a single `ValidationResult`)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
aslo: I could, but wouldn't that mean the validation would only occur when the user clicks "save"? Right now, the validation takes place as soon as the user selects the cloud sdk directory.

google-cloud-tools-plugin/resources/config.properties
aslo: :+1: newline added.

patflynn:
I don't think the balloon style notification is appropriate for a field
level validation on something that hasn't been saved yet. I'd rather we
only apply the validation to the configured global SDK path.

aslo:
This actually is a field-level validation - it actually updates the UI in
the cloud sdk panel with the validation message as the user types. The
system notification (balloon) is used elsewhere (i.e. when starting a local
run or a deployment).

patflynn:
@aslo Aha. In that case it's quite alright. I misunderstood from our last discussion with @etanshaul

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
patflynn: What does it look like if you have 3-4 lines of content?

aslo: It could look something like this (though be aware that these two validation messages won't ever actually occur together. What do you think?

screenshot from 2016-11-16 15 13 44

aslo: FYI Here's another option that uses paragraphs instead of a list. I think this is better.
screenshot from 2016-11-16 15 22 28

patflynn:
I think I prefer it.  It seems like the code should be simpler too.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -95,35 +96,61 @@ protected void checkSdk(String path) {
...
+
+      int counter = 0;
+      for (CloudSdkValidationResult validationResult : validationResults) {
+        if (validationResult.isError()) {
+          containsErrors = true;
+        }
+        if (counter == 0) {
+          builder.append(getMessageForValidationResult(validationResult));
+        } else {
+          builder.append("<p>" + getMessageForValidationResult(validationResult) + "</p>");
```
patflynn: does it look bad if we always have the p tags, even with just one error?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -95,35 +96,61 @@ protected void checkSdk(String path) {
...
+        counter++;
+      }
+
+      showWarning(builder.toString(), containsErrors);
+    }
+  }
+
+  private String getMessageForValidationResult(CloudSdkValidationResult cloudSdkValidationResult) {
+    String message;
+    switch (cloudSdkValidationResult) {
```
patflynn: I feel like this method and the switch statement make the flow a little less obvious than it could be.

for instance would appendCloudSdkDownloadMessage ever be called with any other validation result other than CLOUD_SDK_NOT_FOUND?

aslo: Yep, it could be called with CLOUD_SDK_VERSION_NOT_SUPPORTED, in which case the message is displayed as in the default case.

patflynn:
LGTM

PR #1085, opened 2016-11-03
Adds source context generation logic

joaoandremartins:
Helps with #450.

Uses the Stackdriver facet to parameterize the source context generation process.

Two things still missing in this PR are tests (still thinking what I will test) and source directory to be passed to gen-repo-info-file. Currently, the latter is the first context root, but we want to change that to be defined by the user, and we give them a sensible default (like the module root).


google-cloud-tools-plugin/build.gradle
```diff
@@ -27,6 +27,7 @@ intellij {
     publish {
         pluginId '8079'
     }
+    systemProperties = ['compiler.process.debug.port':56666]
```
joaoandremartins: This isn't going on the final version, just for development.

google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -182,6 +182,7 @@
     <project.converterProvider implementation="com.google.cloud.tools.intellij.appengine.migration.AppEngineStandardFacetConverterProvider"/>
     <project.converterProvider implementation="com.google.cloud.tools.intellij.appengine.migration.AppEngineStandardFacetMigrationConverterProvider"/>
     <framework.detector implementation="com.google.cloud.tools.intellij.appengine.facet.AppEngineFrameworkDetector"/>
+    <compileServer.plugin classpath="google-cloud-tools-jps-plugin.jar;appengine-plugins-core-0.1.13.jar;cloudsdk-service.jar"/>
```
joaoandremartins: Not really sure how to resolve the issue around the static common lib version.

joaoandremartins:
@patflynn @aslo @etanshaul please review.

I'm also seeking your feedback on how to setup the test Git repository for the integration test. I'm thinking of creating a .git folder thin enough to run gen-repo-info-file and copying it to the test directory, but this doesn't feel very right. I also looked at using JGit, but I'm not sure if setting it up just for this test is worth the hassle.

google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -182,6 +182,7 @@
     <project.converterProvider implementation="com.google.cloud.tools.intellij.appengine.migration.AppEngineStandardFacetConverterProvider"/>
     <project.converterProvider implementation="com.google.cloud.tools.intellij.appengine.migration.AppEngineStandardFacetMigrationConverterProvider"/>
     <framework.detector implementation="com.google.cloud.tools.intellij.appengine.facet.AppEngineFrameworkDetector"/>
+    <compileServer.plugin classpath="google-cloud-tools-jps-plugin.jar;appengine-plugins-core-0.1.14.jar;"/>
```
joaoandremartins: Adding the tools lib to the class path this way means we'll have to manually change the jar name every time we bump the lib version, which is undesirable.
Any suggestions for improvement here? I think we can script some things in plugin.xml, but not custom stuff.

etanshaul:
Mind adding some short instructions (screenshots too if its helpful) on how to test these features out in the IDE?

google-cloud-tools-plugin/resources/META-INF/plugin.xml
aslo: Just arriving here - why does the tools lib jar need to be added to the classpath like this?

google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -182,6 +182,7 @@
     <project.converterProvider implementation="com.google.cloud.tools.intellij.appengine.migration.AppEngineStandardFacetConverterProvider"/>
     <project.converterProvider implementation="com.google.cloud.tools.intellij.appengine.migration.AppEngineStandardFacetMigrationConverterProvider"/>
     <framework.detector implementation="com.google.cloud.tools.intellij.appengine.facet.AppEngineFrameworkDetector"/>
+    <compileServer.plugin classpath="google-cloud-tools-jps-plugin.jar;appengine-plugins-core-0.2.0.jar;"/>
```
joaoandremartins: @aslo I think the build plugins use different class loaders to IJ, so they're much lighter to load and faster. This is how stuff is added to the build plugin classpath. Still at a miss as to how script the version number here...

joaoandremartins:
To test this feature, you should first add Stackdriver support through add facet/framework support/etc., select the Generate source config files option, and then go to Build -> Build artifacts...

joaoandremartins:
@patflynn @aslo @etanshaul I went ahead and removed the integration test.

I would like to merge this asap this week -- your reviews are welcome!

joaoandremartins:
Will merge this to the feature branch so we can make one final review there.

PR #1086, opened 2016-11-10
restore tutorial

elharo:
@patflynn @kmccormick I still don't understand what happened to these docs but luckily I found what looks like an up to date version on my laptop.


etanshaul:
@elharo I may have accidentally deleted the gh-pages branch when cleaning up our branches. Sorry about that. I would suggest to make that branch protected but I guess its not needed now that github allows it on the master branch.

lgtm

PR #1093, opened 2016-11-16
Check for forbidden path formats on Windows.

### LINKED ISSUE #1091: need to do some validation on invalid paths when checking the sdk directory
java.nio.file.InvalidPathException: Trailing char < > at index 41: C:\Users\elided\AppData\Local\Google\Cloud 
	at sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:191)
	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)
	at sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:255)
	at java.nio.file.Paths.get(Paths.java:84)
	at com.google.cloud.tools.intellij.appengine.sdk.CloudSdkPanel.checkSdk(CloudSdkPanel.java:94)
	at com.google.cloud.tools.intellij.appengine.sdk.CloudSdkPanel.access$000(CloudSdkPanel.java:45)
	at com.google.cloud.tools.intellij.appengine.sdk.CloudSdkPanel$1.textChanged(CloudSdkPanel.java:73)
	at com.intellij.ui.DocumentAdapter.removeUpdate(DocumentAdapter.java:32)
	at javax.swing.text.AbstractDocument.fireRemoveUpdate(AbstractDocument.java:259)
	at javax.swing.text.AbstractDocument.handleRemove(AbstractDocument.java:622)
	at javax.swing.text.AbstractDocument.remove(AbstractDocument.java:590)
	at com.intellij.openapi.editor.textarea.TextComponentDocument.deleteString(TextComponentDocument.java:121)
	at com.intellij.openapi.editor.actions.BackspaceAction.a(BackspaceAction.java:83)
	at com.intellij.openapi.editor.actions.BackspaceAction.access$100(BackspaceAction.java:35)
	at com.intellij.openapi.editor.actions.BackspaceAction$Handler.executeWriteAction(BackspaceAction.java:54)
	at com.intellij.openapi.editor.actionSystem.EditorWriteActionHandler$1.run(EditorWriteActionHandler.java:57)

joaoandremartins:
Fixes #1091 

Before I start spending time on unit tests, can you guys let me know what you think of this approach to avoid Paths.get() from blowing up?


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -90,6 +94,16 @@ private void checkSdk() {
       return;
     }
 
+    if (System.getProperty("os.name").contains("Windows")
+        && !path.matches(WIN_PATH_REGEX)) {
```
aslo: Rather than going this route, did you consider just calling Paths.get in a try...catch block, and catching the `InvalidPathException`? One advantage of going that way is you don't need to perform any special regex matching or platform-specific validation in our codebase.

aslo: Also, regardless of whether you use the regex or handle the InvalidPathException, you might want to consider adding this validation to the CloudSdkService.validateCloudSdk method that was added in https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1084. (When you rebase from master, it should be more apparent).

Though, I don't feel strongly that it necessarily belongs in there.

patflynn: +1 to Alex's suggestion.

joaoandremartins:
@aslo @etanshaul @patflynn PTAL

joaoandremartins:
FYI it's unlikely we can get over code review and merge this by EOD tomorrow, so #1091 will likely delay to 2016.12..

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -93,40 +92,38 @@ protected void checkSdk(String path) {
...
         }
 
-        if (counter == 0) {
+        if (isFirst) {
           builder.append(message);
         } else {
           builder.append("<p>" + message + "</p>");
         }
-        counter++;
+        isFirst &= false;
```
elharo: What? Isn't this always false? I.e. the same as isFirst = false?

joaoandremartins: Good point -- I can replace it for isFirst = false

patflynn:
I don't see much more to do apart from making the validation message more generic (remove the invalid path characters).  We should be able to get this in today. @aslo what do you think?

aslo:
Yep, looks mostly good. I'll give one more look and review

joaoandremartins:
I thought of adding the characters so users wouldn't have to guess which characters aren't OK. Although it might be confusing when we say paths can't contain colons and Windows paths always contain the drive colon...

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
```diff
@@ -40,7 +44,46 @@ public static CloudSdkService getInstance() {
 
   public abstract void setSdkHomePath(String path);
 
-  public abstract Set<CloudSdkValidationResult> validateCloudSdk(@NotNull Path pathToCloudSdk);
+  protected abstract Set<CloudSdkValidationResult> validateCloudSdk(Path path);
+
+  /**
+   * Checks if the stored path contains a valid Cloud SDK.
+   */
+  public Set<CloudSdkValidationResult> validateCloudSdk() {
```
aslo: Why did you choose to implement these in the abstract CloudSdkService class, and not in the DefaultCloudSdkService implementation?

joaoandremartins: They don't depend on any `DefaultCloudSdkService` members or methods, unlike `validateCloudSdk(Path)` which depends on `buildCloudSdkWithPath()` and `isCloudSdkVersionSupported()`.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
```diff
@@ -40,7 +44,46 @@ public static CloudSdkService getInstance() {
...
+    return validateCloudSdk(getSdkHomePath());
+  }
+
+  /**
+   * Checks if a given path is malformed.
+   *
+   * <p>Windows' implementation of Paths doesn't handle well converting strings with certain special
+   * characters to paths. This method should be called before {@code Paths.get(path)}.
```
aslo: I think this javadoc is a bit confusing - it implies that all this method does is validate that the path is not malformed, when in reality it does all validations on the cloud sdk at that location. 

It might be good to change it to just say something more similar to the other validateCloudSdk methods.

aslo: Also, same question as above regarding the DefaultCloudSdkService

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
```diff
@@ -40,7 +44,46 @@ public static CloudSdkService getInstance() {
...
+      return validateCloudSdk(Paths.get(path));
+    } catch (InvalidPathException ipe) {
+      return ImmutableSet.of(CloudSdkValidationResult.MALFORMED_PATH);
+    }
+  }
+
+  /**
+   * Checks if the provided path contains a valid Cloud SDK installation.
+   */
+  public boolean isValidCloudSdk(String path) {
```
aslo: same question as above

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
```diff
@@ -40,7 +44,46 @@ public static CloudSdkService getInstance() {
...
+    return validateCloudSdk(path).isEmpty();
+  }
+
+  /**
+   * Checks if the saved path contains a valid Cloud SDK installation.
+   */
+  public boolean isValidCloudSdk() {
```
aslo: same question as above

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
```diff
@@ -85,36 +81,25 @@ public JComponent createComponent() {
...
-          .validateCloudSdk();
-    } catch (AppEngineException aee) {
-      isSdkValid = false;
-    }
-
     // Forces a modify check so the user is unable to save an invalid Cloud SDK configuration from
     // Other Settings, on the Clouds menu.
-    return cloudSdkPanel.isModified() || !isSdkValid;
+    return cloudSdkPanel.isModified()
+        || !CloudSdkService.getInstance().isValidCloudSdk(cloudSdkPanel.getCloudSdkDirectory());
```
patflynn: doesn't the apply validation code protect against saving invalid cloud sdk path?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
joaoandremartins: Changed the first line to "Checks if a given path is malformed or if it contains a valid Cloud SDK."

The big motivation for this method really was this Windows issue we found, and it's the only use case for it right now, so I'd prefer to keep the second paragraph.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -181,8 +178,10 @@ JTextPane getWarningMessage() {
   public boolean isModified() {
     CloudSdkService sdkService = CloudSdkService.getInstance();
 
-    return !Paths.get(getCloudSdkDirectory() != null ? getCloudSdkDirectory() : "")
-        .equals(sdkService.getSdkHomePath() != null ? sdkService.getSdkHomePath() : Paths.get(""));
+    String typedDirectory = getCloudSdkDirectory() != null ? getCloudSdkDirectory() : "";
```
patflynn: cloudSdkDirectoryFieldValue
currentCloudSdkDirectory

patflynn: the getCloudSdkDirectory() method seems to primarily obfuscate what it represents. It should at least mirror its setter and be called getCloudSdkDirectoryText(). Or better yet remove the getter and setter entirely or rename them to getCloudSdkDirectoryFieldText() and setCloudSdkDirectoryFieltText().

Also is it possible for getCloudSdkDirectory to be null?

if not this code could be simplified to just cloudSdkDirectoryField.getText().equals(sdkService.getSdkHomePath()).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
aslo: I mentioned it because normally I've seen that we try to keep as little implementation in the abstract service classes as possible, and this seems to be a break from that pattern. @patflynn do you have a strong opinion on whether that's necessary here?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
joaoandremartins: I'm not 100% sure on what the right behaviour should be here, I just tried to keep backwards compatibility with what was there before. The comment is fuzzy, though, since it is possible to save invalid paths from Other Settings, but not from the Cloud menu.

patflynn:
I think as long as it's written in terms of methods on the interface it's
fine. Ideally I'd prefer to have this as an interface, but we can get there
once we move to Java 8 next year.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
patflynn: ok

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -46,6 +46,7 @@ appengine.cloudsdk.java.components.missing=Selected Cloud SDK does not contain t
 appengine.cloudsdk.java.components.howtoinstall=To install, run 'gcloud components install app-engine-java'.
 appengine.cloudsdk.location.label=Cloud SDK directory\:
 appengine.cloudsdk.location.browse.window.title=Browse to Cloud SDK directory
+appengine.cloudsdk.location.badchars.message=The specified Cloud SDK directory contains invalid characters (<>:"?*) or starts or ends with spaces.
```
patflynn: I'm lost a bit in the github comment stream.  I think you said that you removed this. 

We don't know what really generated the InvalidPathException. I don't want us defining what the platform specific rules of the file system are in our strings.   You can always log the IPE.getMessage if that's helpful.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
aslo: :+1: in that case

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
patflynn: makes sense. thanks

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
joaoandremartins: I actually wanted to start a dialog about the value of pointing out to the user which characters are actually considered invalid, instead of letting the user out on a witch hunt.

I'm sensitive to your point about not being sure of what can cause an IPE. By experience, we know that this is the only case when it is happening, but there might be other cases we're overlooking.

I'll go ahead and remove the special characters parenthesis.

PR #1096, opened 2016-11-28
Generates source context files according to parameters in a Stackdriver facet

### LINKED ISSUE #450: Source context not being correctly configured during MVM Deployment

joaoandremartins:
Resolves #450 

Source context files are generated when a native artifact is built.

The facet looks like this:
screenshot from 2016-11-28 14 55 33


google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -177,11 +177,14 @@
...
 
     <framework.type implementation="com.google.cloud.tools.intellij.appengine.facet.AppEngineFrameworkType"/>
+    <framework.type implementation="com.google.cloud.tools.intellij.stackdriver.facet.StackdriverFrameworkType"/>
     <fileTemplateGroup implementation="com.google.cloud.tools.intellij.appengine.facet.AppEngineTemplateGroupDescriptorFactory"/>
     <facetType implementation="com.google.cloud.tools.intellij.appengine.facet.AppEngineStandardFacetType"/>
+    <facetType implementation="com.google.cloud.tools.intellij.stackdriver.facet.StackdriverFacetType"/>
     <project.converterProvider implementation="com.google.cloud.tools.intellij.appengine.migration.AppEngineStandardFacetConverterProvider"/>
     <project.converterProvider implementation="com.google.cloud.tools.intellij.appengine.migration.AppEngineStandardFacetMigrationConverterProvider"/>
     <framework.detector implementation="com.google.cloud.tools.intellij.appengine.facet.AppEngineFrameworkDetector"/>
+    <compileServer.plugin classpath="google-cloud-tools-jps-plugin.jar;appengine-plugins-core-0.2.0.jar;"/>
```
joaoandremartins: Could still use suggestions for how to not set the version here manually. Couldn't find a way to script custom stuff into plugin.xml.

patflynn: @joaoandremartins resource filtering seems like an option.  @loosebazooka played around with doing this at one point I think.

google-cloud-tools-plugin/build.gradle
```diff
@@ -27,6 +27,7 @@ intellij {
     publish {
         pluginId '8079'
     }
+    systemProperties = ['compiler.process.debug.port':56666]
```
patflynn: what's this for?

joaoandremartins: The build flow happens in another process. When run in debug mode, it listens to a random port. This makes it always listen to the port 56666, so I don't have to constantly change the port in my run config every time I want to debug the build process.

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
```diff
@@ -0,0 +1,169 @@
...
+      Path outputDirectory = target.getOutputDir().toPath();
+
+      GenRepoInfoFile genAction = actionFactory.newAction(extension.getCloudSdkPath());
+
+      DefaultGenRepoInfoFileConfiguration configuration = new DefaultGenRepoInfoFileConfiguration();
+      configuration.setSourceDirectory(sourceDirectory.toFile());
+      configuration.setOutputDirectory(outputDirectory.toFile());
+      genAction.generate(configuration);
+
+      ExitCodeRecorderProcessExitListener exitListener = actionFactory.getExitListener();
```
patflynn: listeners are for an asynchronous API, but it seems like we're calling this in a synchronous fashion.  if 'generate(configuration)' is asynchronous, how do you know that your exit listener.getMostRecentExitCode has the code you're looking for?  Does it block?

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
```diff
@@ -0,0 +1,169 @@
...
+
+      if (exitListener.getMostRecentExitCode() != 0) {
+        LOG.warn("gen-repo-info-file command returned with status code "
```
patflynn: full gcloud command would be helpful here.

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
```diff
@@ -0,0 +1,169 @@
...
+
+  public static final String NAME = "Stackdriver source context generator";
+  private GenRepoInfoFileActionFactory actionFactory = new GenRepoInfoFileActionFactory();
+
+  public GenRepoInfoFileModuleBuilder() {
+    super(BuilderCategory.CLASS_POST_PROCESSOR);
+  }
+
+  @VisibleForTesting
+  GenRepoInfoFileModuleBuilder(GenRepoInfoFileActionFactory actionFactory) {
```
joaoandremartins: Does the convention still apply if one constructor is only used for tests? Feels a bit weird to enable a constructor that shouldn't be used in production...

patflynn: it can stay package protected.

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
```diff
@@ -0,0 +1,169 @@
...
+      if (extension.getCloudSdkPath() == null) {
+        LOG.warn("No Cloud SDK path specified. Skipping source context generation.");
+        // Cloud SDK path is a singleton for a project, so if it doesn't exist on the first module,
+        // we can return right away.
+        return ExitCode.NOTHING_DONE;
+      }
+
+      if (extension.getModuleSourceDirectory() == null) {
+        // Possible to happen on Windows, unlikely in Linux.
+        LOG.warn("Invalid module source directory. Check for special characters <>:\"|?*.");
```
joaoandremartins: This was intended to capture the case when Paths.get() throws an `InvalidPathException`, we catch it and return null.
I don't think it's possible for getModuleSourceDirectory() to return null in any other occasion since, for new projects, it's automatically fills this field with the parent directory of the module file.

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
joaoandremartins: GenRepoInfoFileModuleBuilder() can't call `this(new GenRepoInfoFileActionFactory());` because `GenRepoInfoFileActionFactory` can't be referenced until the supertype constructor is called.

I went ahead and duplicated the `super(BuilderCategory.CLASS_POST_PROCESSOR);` call in both constructors. LMKWYT

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
```diff
@@ -0,0 +1,169 @@
...
+        continue;
+      }
+
+      Path sourceDirectory = extension.getModuleSourceDirectory();
+      ModuleBuildTarget target =
+          new ModuleBuildTarget(jpsModule, JavaModuleBuildTargetType.PRODUCTION);
+      Path outputDirectory = target.getOutputDir().toPath();
+
+      GenRepoInfoFile genAction = actionFactory.newAction(extension.getCloudSdkPath());
```
joaoandremartins: Not quite sure if I understood you here..

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
```diff
@@ -0,0 +1,169 @@
...
+
+      DefaultGenRepoInfoFileConfiguration configuration = new DefaultGenRepoInfoFileConfiguration();
+      configuration.setSourceDirectory(sourceDirectory.toFile());
```
joaoandremartins: String -> Path -> File

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
joaoandremartins: We're calling generate() synchronously. ProcessExitListeners happen to be the only way to retrieve gcloud's exit code from the common lib.

Are you saying we should make `generate()` retrieve the status code? There might be other methods where we want to return something else, and we wouldn't be able to stick to this convention. In any case, I think any change here could be handled outside of this PR?

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
joaoandremartins: Done

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
```diff
@@ -0,0 +1,169 @@
...
+      }
+
+      if (jpsModule.getSourceRoots().isEmpty()) {
+        LOG.warn("Module " + jpsModule.getName() + " contains no source roots. Moving on to the "
+            + "next module for source generation.");
+        continue;
+      }
+
+      if (extension.getCloudSdkPath() == null) {
+        LOG.warn("No Cloud SDK path specified. Skipping source context generation.");
```
joaoandremartins: Done

google-cloud-tools-plugin/resources/META-INF/plugin.xml
joaoandremartins: Searching IJ's source code, there are other instances of compileServer.plugin classpath containing hardcoded version numbers.
I added a reminder on gradle.properties that, when the version number is changed there, it must also be changed in plugin.xml.

joaoandremartins: Another option would be to omit the version number from appengine-plugins-core jar name.
I'll also open a new thread on IJ's support board asking if there is any way to script this.

joaoandremartins: Actually, found a way around that works.
https://intellij-support.jetbrains.com/hc/en-us/community/posts/206104319-External-builder-classpath

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
patflynn: Can't you just construct the factory in the constructor?

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
patflynn: This seems unintuitive.  I wouldn't expect a null response on a getter of an object to indicate that it's invalid. Why is it even possible for it to be saved as an invalid path?

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
patflynn: I'm seeing now that the factory is scoped just to this class so disregard.

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
joaoandremartins: By the way, this null only happens if a path is invalid by having special characters. If it is a wellformed path without a Cloud SDK installation, this doesn't trigger (i.e., the return isn't null).

I must be missing something obvious, but why is it unintuitive? We're doing something similar for the Cloud SDK path.

Even if we didn't save an invalid Cloud SDK path, getModuleSourceDirectory() default return would still be null.

Using validators here, as opposed to a null return, is also not possible, since they're currently built into the IJ plugin. We could be on to something if the validators were implemented in appengine-plugins-core, which jps-plugin has access to.

What's your change suggestion here?

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
```diff
@@ -0,0 +1,169 @@
...
+import org.jetbrains.jps.model.module.JpsModule;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Collections;
+
+/**
+ * Generates source context files for a module using the Cloud SDK.
+ */
+public class GenRepoInfoFileModuleBuilder extends ModuleLevelBuilder {
```
joaoandremartins: @chashnikov would actually be great if you could take a quick look and validate if what we're doing here looks correct. Thanks!

google-cloud-tools-plugin/jps-plugin/src/com/google/cloud/tools/intellij/jps/GenRepoInfoFileModuleBuilder.java
patflynn: I think you're problem is that you inner class is non-static.

joaoandremartins:
Rightly so, because I want to mock it in unit tests.

I still had one action item to try and mock it anyways, somehow, after I
manage to alert for build issues on ignoreErrors=true.

patflynn:
I'm not sure I see how that effects the mockability of the class. Your
other option is to just pass null to your unary constructor and not set the
factory in that case.

joaoandremartins:
Huh, I was under the impression that Mockito wouldn't mock anything static.. I think what it is, is it won't mock static methods, since I'm being able to turn GenRepoInfoFileModuleBuilder to static and mock it..

joaoandremartins:
@patflynn @etanshaul @aslo 

I removed all the build and plugin.xml stuff so that the code can be pushed, but not be executed.

There is an outstanding issue with this PR, which is that outside of Debug mode, changes to the Cloud SDK path aren't picked up by StackdriverBuildProcessParametersProvider. @patflynn also let me know what you think about that null issue.

It would be great if I could merge this code by the end of today. Thanks!

joaoandremartins:
Asked why this is happening to the IJ team in https://intellij-support.jetbrains.com/hc/en-us/community/posts/206004124-BuildManagerListener-isn-t-being-triggered-outside-of-debug-mode

joaoandremartins:
Can we revisit this now that 16.11.6 is in? :grin:

patflynn:
I'm looking at that thread with JB and it looks like there are some
problems with the current approach. Were you able to resolve them?

joaoandremartins:
I haven't since I wanted to get started on the Flex doc right away. I was thinking the code could be merged with that outstanding issue?

There is the other extension.getModuleSourceDirectory() issue I was wondering how you feel about. Apart from that, there are no other open issues that I can think of.

joaoandremartins:
@patflynn FYI this now grabs the configuration from the Flexible facet, instead of the Stackdriver one.

Still some items pending:
- There is some delay after changing Cloud SDK's location, because of the build process cache,
- There are icons and messages left behind that I need to remove,
- We should probably also ask the user for actual module source root. We're just considering it to be the folder where the module's iml file is,
- UI probably needs some adjustments too (e.g., a Stackdriver header before the Stackdriver settings? Though it can become complicated once we add more app.yaml stuff to the facet.)

screenshot from 2017-03-31 17 33 03

Issue #451, opened 2016-02-18
Automatic deployment, and generated app.yaml and Dockerfile should come from gcloud app gen-config


patflynn:
We currently package template app.yaml and Dockerfiles ourselves to setup auto deployment and config generation.  This is because gcloud app gen-config creates AE2 style configs which are not available broadly. Once this is fixed we can move to using the gcloud command for generating these files.

chanseokoh:
'gcloud preview app gen-config --custom' currently generates an 'app.yaml' with 'env: 2', which is an expected behavior. Although this 'env: 2' should make 'gcloud' deploy an app to AE MVM, I can see that it currently deploys an app to AE standard. However, I think I can at least start implementing this and merge it later when 'gcloud' is eventually fixed to deploy an app to AE2.

patflynn:
the problem with env:2 is that it was whitelist only last time I checked.

Amir do you know if env:2 is off the whitelist?

patflynn:
@arouzrokh

arouzrokh:
Hey guys, 
we should not, at the moment, use gen-config because it does not support the duality that we are seeing for GAE standard and Flex. The code is also to some extend orphaned as the engineer whom first wrote this code has since departed. 

Please package the correct app.yaml and Dockerfile that is tested to work with the deployment at hand for now until we actually end up having reliable fingerprinting technology, if ever. 

Thanks,
Amir

chanseokoh:
@arouzrokh Can we now use `gcloud gen-config`? When do you think we can use it?

etanshaul:
blocked until gen-config fully supports our requirements

arouzrokh:
@rdayal Rajeev who's the person handling gen-config? it sits with gcloud right?

patflynn:
@meltsufin I believe Mike's last fix should have unblocked this.

meltsufin:
Yes, it has been fixed in gcloud, and should be there in the next release.

etanshaul:
@meltsufin do you know which version of gcloud has this? I'm trying to confirm if its been released or not.

meltsufin:
I'm not sure which version it was added to, but it is there now. Look for `vm: true` in this file `google-cloud-sdk/platform/ext-runtime/java/data/app.yaml.template`.

etanshaul:
closed as "won't fix" due to limited enhancements for ae flex

PR #1101, opened 2016-12-01
Generate env flex

aslo:
see also #451

This PR updates the app.yaml template to deploy to env: flex. Ideally, this would be fixed by using `gcloud gen-config` to generate app.yaml, but gen-config is currently in flux, and does not give us a way to specify the output directory for generated files.

I think we should make this change in order to start using env: flex in the upcoming 16.11 release, and then later use gcloud when it is ready.

### DIFF
--- a/google-cloud-tools-plugin/resources/generation/src/appengine/mvm/app.yaml
+++ b/google-cloud-tools-plugin/resources/generation/src/appengine/mvm/app.yaml
@@ -1,4 +1,4 @@
 runtime: custom
-vm: true
+env: flex
 env_variables:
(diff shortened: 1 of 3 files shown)


aslo:
cc @meltsufin

aslo:
They are generally the same, with a few exceptions: we only generate an app.yaml with runtime: custom and a provided Dockerfile, while gcloud gen-config will generate an app.yaml with runtime: java by default (works without a Dockerfile). 

The net result is the same, though we should also think about making it possible to generate an app.yaml with runtime:java

PR #1104, opened 2016-12-02
Cleanup debug watcher in unit test

etanshaul:
Testing unit test memory leak fix


etanshaul:
Looks like this took care of the problem. ptal.

PR #1107, opened 2016-12-02
Release v16.11.1

etanshaul:


patflynn:
Since it's now December we should probably change this to 16.12

joaoandremartins:
How about releasing 16.11? (No trailing .1)

I'm OK with 2016.11, it's only the beginning of December. We can release a 16.12 later in the month if we need to.

PR #1118, opened 2016-12-02
Fix cloud sdk validation error message flash

### LINKED ISSUE #1109: Cloud SDK validation errors flashed momentarily during form initialization
steps to reproduce:

1. new project
2. java ee -> web -> Google App Engine Srnadard
3. notice that the 'cloud sdk directory: " field appears with a validation error that quickly disappears when the right directory is found and initialized.

expected result:
1. no validation error is show until initialization is complete.

aslo:
fixes #1109

When `cloudSdkDirectoryField.setText` is called, it causes two DocumentEvents to be fired, one that removes the old text, and one that adds the new text. Since we validate on every DocumentEvent, this causes the message to flash.  Simply preventing unnecessary updates removes the issue.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
@@ -200,8 +200,14 @@ public class CloudSdkPanel {
     return cloudSdkDirectoryField.getText();
   }
 
+  /**
+   * Sets the Cloud SDK directory text. Performs no action if the current value of the Cloud SDK
+   * directory field is already equal to the path.
+   */
   public void setCloudSdkDirectoryText(String path) {
-    cloudSdkDirectoryField.setText(path);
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -200,8 +200,14 @@ public String getCloudSdkDirectoryText() {
     return cloudSdkDirectoryField.getText();
   }
 
+  /**
+   * Sets the Cloud SDK directory text. Performs no action if the current value of the Cloud SDK
+   * directory field is already equal to the path.
+   */
   public void setCloudSdkDirectoryText(String path) {
-    cloudSdkDirectoryField.setText(path);
+    if (!cloudSdkDirectoryField.getText().equals(path)) {
```
etanshaul: can getText be null?

aslo: I don't think so, but it can't hurt to be safe here since this is a third-party API. Good catch.

PR #1134, opened 2016-12-05
Fix Cloud Sdk panel validation messages

### LINKED ISSUE #1129: Cloud SDK panel missing java component error message is wrong
image

etanshaul:
fixes #1129

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
@@ -25,6 +25,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.intellij.execution.configurations.RuntimeConfigurationError;
 import com.intellij.openapi.options.Configurable;
 import com.intellij.openapi.options.ConfigurationException;
+import com.intellij.openapi.util.text.StringUtil;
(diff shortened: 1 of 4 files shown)


joaoandremartins:
How does this show a different message, for example, when the Cloud SDK validation shows that the Java components are missing?

Also, how do you feel about resolving the issue of the error message not appearing when the Settings window is open with an invalid SDK path in this PR? :X

etanshaul:
It shows a different message because the apply method in AppEngineCloudConfigurable now delegates to the cloudSdkPanel to construct the message. CloudSdkPanel already had logic to construct a specific message, except that it was previously couple to toggling UI components. I just refactored CloudSdkPanel to separate out the message construction part so it can be reused.

I haven't thought about that other issue, but I'll see.

joaoandremartins:
Ah, I see.

Well, this could be a good place for the other fix I mentioned. If you don't want it on this PR, feel free to merge.

etanshaul:
sure. see https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1135 for more details on that other issue.

joaoandremartins:
I'm not sure you understand what the issue I described is. :)

When you save an invalid SDK, close the settings window, then open it again, no error will be shown. You need to make a change to the path in order for the error message to show (if the new Cloud SDK is still invalid).

Maybe the best is for me to file a new issue.

etanshaul:
I did understand but may not have done a good job describing it in that issue. feel free to modify that issue directly since that is what I intended it to represent

PR #1136, opened 2016-12-06
Adds null control for Cloud SDK paths form check

joaoandremartins:
propertiesComponent.getValue(CLOUD_SDK_PROPERTY_KEY) is returning null, generating an NPE on CloudSdkService.isMalformedPath() at Paths.get().

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
@@ -86,9 +86,13 @@ public abstract class CloudSdkService {
   }
 
   /**
-   * Checks for invalid characters that trigger a {@link InvalidPathException} on Windows.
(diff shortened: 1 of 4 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
```diff
@@ -86,9 +86,13 @@ public boolean isValidCloudSdk() {
   }
 
   /**
-   * Checks for invalid characters that trigger a {@link InvalidPathException} on Windows.
+   * Checks for invalid characters that trigger an {@link InvalidPathException} on Windows.
    */
-  public boolean isMalformedCloudSdkPath(String sdkPath) {
+  public boolean isMalformedCloudSdkPath(@Nullable String sdkPath) {
+    if (sdkPath == null) {
```
etanshaul: would be cleaner if you just wrap the try catch in an if (sdkPath != null)

etanshaul:
:ship

PR #1137, opened 2016-12-06
Show invalid SDK warning when opening CloudSdkPanel

### LINKED ISSUE #1135: Opening Settings window with invalid Cloud SDK path doesn't produce a warning message
Some instantiations of the panel such as the Cloud Configuration, allow the user to save the panel when initially opening it in an error state. Modifying the field then triggers a validation when subsequently trying to save it.

Not revalidating on saving an unmodified panel is part of framework behavior. This issue is happening because we have some places where the cloud sdk can be invalid or missing (new project wizard, google settings) and others where a valid sdk is forced.

joaoandremartins:
Changes warning message modality.

Coupled with https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/277 so that a lingering empty message warning disappears.

Fixes #1135

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
@@ -167,7 +167,7 @@ public class CloudSdkPanel {
   }
 
   private void invokePanelValidationUpdate(Runnable runnable) {
-    ApplicationManager.getApplication().invokeLater(runnable, ModalityState.current());
+    ApplicationManager.getApplication().invokeLater(runnable, ModalityState.any());
   }
 
   @VisibleForTesting
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
@@ -148,7 +148,6 @@ public class CloudSdkPanel {
       public void run() {
         warningMessage.setText(message);
         warningMessage.setVisible(true);
-        warningMessage.updateUI();
         warningIcon.setVisible(true);
         cloudSdkDirectoryField.getTextField().setForeground(JBColor.red);
       }
@@ -162,16 +161,13 @@ public class CloudSdkPanel {
       public void run() {
         cloudSdkDirectoryField.getTextField().setForeground(JBColor.black);
         warningIcon.setVisible(false);
-        warningIcon.setVisible(false);
         warningMessage.setVisible(false);
-        cloudSdkPanel.updateUI();
       }
     });
   }
 
   private void invokePanelValidationUpdate(Runnable runnable) {
-    ApplicationManager.getApplication().invokeLater(runnable,
-        ModalityState.stateForComponent(cloudSdkPanel));
+    ApplicationManager.getApplication().invokeLater(runnable, ModalityState.current());
   }
 
   @VisibleForTesting


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -162,16 +161,13 @@ protected void hideWarning() {
...
         warningMessage.setVisible(false);
-        cloudSdkPanel.updateUI();
       }
     });
   }
 
   private void invokePanelValidationUpdate(Runnable runnable) {
-    ApplicationManager.getApplication().invokeLater(runnable,
-        ModalityState.stateForComponent(cloudSdkPanel));
+    ApplicationManager.getApplication().invokeLater(runnable, ModalityState.current());
```
etanshaul: can you explain what effect this has?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -148,7 +148,6 @@ protected void showWarning(final String message) {
       public void run() {
         warningMessage.setText(message);
         warningMessage.setVisible(true);
-        warningMessage.updateUI();
```
aslo: Why did you need to delete this call (and with cloudSdkPanel.updateUI() below)? I remember adding it to guarantee that the UI was properly redrawn when the message changed (specifically, to make sure that the window was automatically resized to fit the changed text).

joaoandremartins: Everywhere I tested CloudSdkPanel, the message gets properly updated without these calls.

In the case you described, the message just wraps (which I actually thought impossible in UI forms :o), which I think is the correct behaviour?

Are there any other cases I should be aware of?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
etanshaul: Aside from being drown when it's first opened, does it also prevent saving the dialog?

Similarly, what effect does this have on instances of this dialog where we don't require the SDK to be valid? The only one now I think is the Google settings panel in the IDE properties.

joaoandremartins: It doesn't prevent from saving the dialog -- that piece of logic doesn't belong here, it goes, for example, on apply().

joaoandremartins: One thing I'm now noticing is that the warning still doesn't get correctly drawn when cloudSdkPanel is opened from the deploy or local run run configs... Might have to do with this modality thing.

joaoandremartins: So, even though `ModalityState`'s documentation discourages the use of `any()` ("unless absolutely needed" they say), I found it was the only modality state where the message gets correctly drawn when the settings window is opened.

etanshaul: Just to make sure I understand:

You said that "It doesn't prevent from saving the dialog". So before your change, there was no warning and the user could save the dialog (even if it was in an invalid state). Now, the user sees the warning but can save the dialog anyway? I think the behavior we want is to see the warning and no be able to save from those dialogs that require a valid SDK, but allow saving from those that don't (I'm not sure if this is what you meant that is happening).

joaoandremartins:
Please approve when comfortable guys :smile:

PR #1145, opened 2016-12-06
Move to error prone

### LINKED ISSUE #1141: investigate migrating from findbugs to errorprone

patflynn:
preliminary. PTAL.  I won't check this in until after the release.  Also need to add instructions in the PR for setup IJ compiler.

fixes #1141 

Benefits are:
Less noise.
No config.
You never will push a PR and find that Jenkins has found an error, because those errors will fail your build and also show up as errors in your IJ editor.

@aslo @etanshaul @joaoandremartins


chanseokoh:
Also need to remove the `@SuppressFBWarnings` annotation. There are a couple of instances.

patflynn:
Good point. also removed the dependency in build.gradle.

.idea/compiler.xml
```diff
@@ -27,10 +27,6 @@
       <module name="google-account-plugin_test" target="1.7" />
       <module name="google-cloud-tools-plugin_main" target="1.7" />
       <module name="google-cloud-tools-plugin_test" target="1.7" />
-      <module name="jps-plugin_main" target="1.7" />
```
joaoandremartins: After the source context change, we'll need to leave this bit here I think (even though I'm not too sure what this does).

.idea/google-java-format.xml
```diff
@@ -2,5 +2,45 @@
...
+        <entry key="common-test-lib_main" value="ENABLED" />
+        <entry key="common-test-lib_test" value="ENABLED" />
+        <entry key="gcloud-intellij" value="ENABLED" />
+        <entry key="google-account-plugin" value="ENABLED" />
+        <entry key="google-account-plugin_main" value="ENABLED" />
+        <entry key="google-account-plugin_test" value="ENABLED" />
+        <entry key="google-cloud-tools-plugin" value="ENABLED" />
+        <entry key="google-cloud-tools-plugin_main" value="ENABLED" />
+        <entry key="google-cloud-tools-plugin_test" value="ENABLED" />
+        <entry key="ultimate" value="ENABLED" />
```
joaoandremartins: Mental note to also add jps-plugin here after the source context merge.

patflynn:
I'll wait until you merge your change first, and then rebase.

gradle.properties
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-ideaEdition = IC
+ideaEdition = IU
```
joaoandremartins: Is this intended?

patflynn:
nope.

gradle.properties
aslo: @patflynn Why not keep this? IU is our main use case anyway. I personally always manually change this to IU when running locally.

patflynn:
Mainly it's for the CI environment.

One convenient way I've found of handling this is adding the file to a
separate change list in the IntelliJ VCS panel. It then typically
intelligently stashes and reapplies as I switch between branches, and it's
not included in pushes (except accidentally in this case).

patflynn:
@aslo @etanshaul @joaoandremartins OK I think this is ready to merge.  PTAL and follow the CONTRIBUTING instructions after I merge.

.idea/compiler.xml
```diff
@@ -13,11 +13,6 @@
       <entry name="!?*.kt" />
       <entry name="!?*.clj" />
     </wildcardResourcePatterns>
-    <annotationProcessing>
```
joaoandremartins: Why is this removed?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/ui/CloudDebugHistoricalSnapshots.java
```diff
@@ -648,7 +648,7 @@ public void run() {
       for (int row = 0; row < getModel().getRowCount(); row++) {
         // todo: getModel should be newModel
         Breakpoint bp = getModel().getBreakpoints().get(row);
-        if (bp.getIsFinalState() != Boolean.TRUE) {
+        if (bp.getIsFinalState().equals(Boolean.TRUE)) {
```
joaoandremartins: I think you need a "!" before bp?

gradle.properties
```diff
@@ -13,8 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
-ideaEdition = IC
+ideaEdition=IC
```
joaoandremartins: Maybe unstage gradle.properties?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/ui/CloudDebugHistoricalSnapshots.java
patflynn: well shit! yeah. Good catch!

.idea/compiler.xml
patflynn: I didn't actually modify this file. It was the IJ settings editor that updated this when I switched the compiler.  Looking at our old and new settings UI (before and after the switch), it doesn't appear that annotation processors were ever on.

patflynn:
@joaoandremartins fixed. GoogleJavaFormat already doing it's magic.  I didn't realize it would rearrange method declarations.

patflynn:
any other comments?

gradle.properties
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-ideaEdition = IC
+ideaEdition=IC
```
joaoandremartins: nit: probably add spaces before and after the equals sign

patflynn:
Aha I had! But it turns out my IJ formatter is somehow messing this up.
fixed.

PR #1148, opened 2016-12-06
Fix for vanishing local dev run config

### LINKED ISSUE #1147: Local dev server run config vanishes
To reproduce:

1) create a new AE standard project. notice the local dev run config gets created
2) close the project, then reopen it

Notice the run config is gone.

etanshaul:
Fixes #1147 

The problem was that the getName method in the ConfigurationFactory was not overriden. This is needed on project re-open so that the previously created one can be restored. 

Also updating the overriden methods to delegate to the already existing ConfigurationFactory's methods.

### DIFF
--- a/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/AppEngineServerConfigurationType.java
+++ b/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/AppEngineServerConfigurationType.java
@@ -72,12 +72,19 @@ public class AppEngineServerConfigurationType extends J2EEConfigurationType {
 
   @Override
   public ConfigurationFactory[] getConfigurationFactories() {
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/AppEngineServerConfigurationType.java
```diff
@@ -72,12 +72,19 @@ public Icon getIcon() {
...
+          @Override
+          public String getName() {
+            return configurationFactory.getName();
+          }
+
           @NotNull
           @Override
           public RunConfiguration createTemplateConfiguration(@NotNull Project project) {
-            return createJ2EEConfigurationTemplate(this, project, true /*isLocal*/);
+            return configurationFactory.createTemplateConfiguration(project);
```
etanshaul: yes, ultimately it does the same thing. I think its better code to do the delegation like I updated it do instead of recreating it.

PR #1152, opened 2016-12-07
Rebump version number in release notes.

joaoandremartins:
And check if you want to add anything else.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 All notable changes to this project will be documented in this file.
 ## [unreleased]
 
-## [16.11.3]
+## [16.11.4]
 
 ### Added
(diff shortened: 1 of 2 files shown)


etanshaul:
why do we need to bump it again? lgtm in any case

joaoandremartins:
I think release 16.11.3 doesn't contain a lot of the latest fixes we've launched and a new release is necessary?

I actually thought the release would only get going from the moment your bump PR went through. Since that's not true, I think we need a new release.

joaoandremartins:
Also, I'll hold off on @patflynn since he might have something to add.

etanshaul:
oh ok. I thought you already released today with the new changes.

etanshaul:
btw check out the release page: https://github.com/GoogleCloudPlatform/google-cloud-intellij/releases
 to see what the release contains. Looks like there have been 9 commits since 16.11.3.

patflynn:
I was just saying it needs to be 4 since 3 is already cut, and 4 is going
to be the next RC with the latest bug fixes.

PR #1154, opened 2016-12-07
Add platform dimensions to usage tracking

### LINKED ISSUE #1124: Analytics pings should include platform data
As far as I know we originally included this but lost it when moving over to the new property and implementation.
Ideally:
1. ct4ij plugin version
1. IntelliJ version (including CE vs UE)
1. Java version
1. OS

patflynn:
Fixes #1124

### DIFF
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
@@ -17,14 +17,21 @@
 package com.google.cloud.tools.intellij.stats;
 
 import com.google.cloud.tools.intellij.AccountPluginInfoService;
+import com.google.common.base.Joiner;
+import com.google.common.base.Joiner.MapJoiner;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 
 import com.intellij.ide.util.PropertiesComponent;
+import com.intellij.openapi.application.ApplicationInfo;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.updateSettings.impl.UpdateChecker;
+import com.intellij.openapi.util.SystemInfo;
+import com.intellij.util.PlatformUtils;
 
 import org.apache.http.NameValuePair;
 import org.apache.http.StatusLine;
@@ -40,12 +47,14 @@ import org.jetbrains.annotations.Nullable;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Map;
 
 /** Google Usage Tracker that reports to Cloud Tools Analytics backend. */
 public class GoogleUsageTracker implements UsageTracker, SendsEvents {
 
+  private static final MapJoiner METADATA_JOINER =
+      Joiner.on(',').useForNull("null").withKeyValueSeparator("=");
   private static final Logger logger = Logger.getInstance(GoogleUsageTracker.class);
-
   private static final String ANALYTICS_URL = "https://ssl.google-analytics.com/collect";
   private static final String PROTOCOL_VERSION_KEY = "v";
   private static final String UNIQUE_CLIENT_ID_KEY = "cid";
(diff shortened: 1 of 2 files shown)


patflynn:
@etanshaul I think I want to land this for this release.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
```diff
@@ -40,12 +49,20 @@
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Map;
 
 /** Google Usage Tracker that reports to Cloud Tools Analytics backend. */
 public class GoogleUsageTracker implements UsageTracker, SendsEvents {
 
+  private static final MapJoiner METADATA_JOINER =
+      Joiner.on(',').useForNull("null").withKeyValueSeparator("=");
```
chanseokoh: This should be '&'.

patflynn:
I was looking at this:
https://github.com/google/cloud-reporting/blob/master/src/main/java/com/google/cloud/metrics/MetricsUtils.java#L159

Is that incorrect?

chanseokoh:
Interesting. Let me check.

chanseokoh:
Yeah, you're right. I got this wrong whole time. (Lucky I didn't really do
so yet.)

patflynn:
phew. :)

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
```diff
@@ -86,6 +112,18 @@ public GoogleUsageTracker() {
...
+    String cloudToolsPluginVersion = pluginInfo.getPluginVersion();
+    Map<String, String> systemMetaDataMap =
+        ImmutableMap.of(
+            PLATFORM_NAME_KEY, METADATA_ESCAPER.escape(intellijPlatformName),
+            PLATFORM_VERSION_KEY, METADATA_ESCAPER.escape(intellijPlatformVersion),
+            JDK_VERSION_KEY, METADATA_ESCAPER.escape(JDK_VERSION_VALUE),
+            OPERATING_SYSTEM_KEY, METADATA_ESCAPER.escape(OPERATING_SYSTEM_VALUE),
+            PLUGIN_VERSION_KEY, METADATA_ESCAPER.escape(cloudToolsPluginVersion));
+
+    metadataStringBuilder = METADATA_JOINER.appendTo(metadataStringBuilder, systemMetaDataMap);
```
etanshaul: this is a little weird since metadataStringBuilder is always empty here, yet you are still appending it to itself.

etanshaul: could you just use MapJoiner#join instead?

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
```diff
@@ -62,6 +79,15 @@
   private static final String PAGE_TITLE_KEY = "dt";
   private static final String STRING_FALSE_VALUE = "0";
   private static final String STRING_TRUE_VALUE = "1";
+  // Our plugin metadata keys.
+  private static final String PLATFORM_NAME_KEY = "applicationName";
+  private static final String PLATFORM_VERSION_KEY = "applicationVersion";
+  private static final String JDK_VERSION_KEY = "jdkVersion";
+  private static final String OPERATING_SYSTEM_KEY = "operationSystem";
```
patflynn: nice catch!

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
patflynn: join returns a string. If you want to continue appending to a list of key value pairs, this is the only API I found.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
```diff
@@ -114,11 +152,16 @@ public void sendEvent(
...
-          // Event metadata are passed as a (virtual) page title.
-          String virtualPageTitle = eventLabel + "=" + (eventValue != null ? eventValue : "null");
-          postData.add(new BasicNameValuePair(PAGE_TITLE_KEY, virtualPageTitle));
+          Map<String, Integer> eventMetaDataMap = Maps.newHashMap();
+          eventMetaDataMap.put(METADATA_ESCAPER.escape(eventLabel), eventValue);
+          metadataStringBuilder =
+              METADATA_JOINER.appendTo(
+                  // I feel like I shouldn't have to do the ',' append here but for some reason
+                  // I do.  Bug?
+                  metadataStringBuilder.append(','), eventMetaDataMap);
```
chanseokoh: How about merging the `systemMataDataMap` into `eventMetaDataMap` and then apply the joiner? That could simplify code noticeably. Concerned about performance?

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
etanshaul: ah yeah. too bad there's no way to initialize it

etanshaul:
lgtm on my end pending resolution to @chanseokoh's questions

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
patflynn: Yeah, I was avoiding recomputing the prefix for each analytics request. Probably a small cost. But I've seen a bug actually.

chanseokoh:
I don't know why, but I see the `eventLabel` appearing three times on Google Analytics. Looking into it.

patflynn:
@chanseokoh that's what I'm fixing.

patflynn:
@chanseokoh @etanshaul 
PTAL

etanshaul:
lgtm

patflynn:
FYI the bug was that the base metadata properties were getting appended to, so subsequent analytics events would include the full history.

I ended up simplifying things by just using a final string and managing the joining of the event and system metadata manually.

PR #1160, opened 2016-12-08
Fix deployment error messaging.

joaoandremartins:
When an invalid Cloud SDK path is specified but the deployment log output states the app-engine-java component is missing.


google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -47,8 +47,9 @@ appengine.cloudsdk.java.components.howtoinstall=To install, run "gcloud componen
 appengine.cloudsdk.location.label=Cloud SDK directory\:
 appengine.cloudsdk.location.browse.window.title=Browse to Cloud SDK directory
 appengine.cloudsdk.location.badchars.message=The specified Cloud SDK directory contains invalid characters or starts or ends with spaces.
-appengine.cloudsdk.location.invalid.message=No Cloud SDK was found in this directory.
+appengine.cloudsdk.location.invalid.message=No Cloud SDK was found in the specified directory.
 appengine.cloudsdk.location.missing.message=Cloud SDK home directory is not specified.
+appengine.cloudsdk.location.missing.message.deploy=The Cloud SDK is misconfigured. To fix, reconfigure the Server above.
```
patflynn: Configure the server above?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardDeployTask.java
```diff
@@ -63,6 +65,16 @@ public void execute(ProcessStartListener startListener) {
         .withLabel(isFlexCompat ? "flex-compat" : "standard")
         .ping();
 
+    if (CloudSdkService.getInstance().validateCloudSdk().contains(
+        CloudSdkValidationResult.CLOUD_SDK_NOT_FOUND)) {
```
aslo: Is `CloudSdkValidationResult.MALFORMED_PATH` something we should check for here as well?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -386,10 +387,17 @@ private void validateConfiguration() throws ConfigurationException {
...
             GctBundle.message("appengine.flex.config.custom.dockerfile.error"));
       }
-    } else if (environment.isStandard() && CloudSdkService.getInstance().validateCloudSdk()
-        .contains(CloudSdkValidationResult.NO_APP_ENGINE_COMPONENT)) {
-      throw new ConfigurationException(
-          CloudSdkValidationResult.NO_APP_ENGINE_COMPONENT.getMessage());
+    } else if (environment.isStandard()) {
+      Set<CloudSdkValidationResult> validationResults =
+          CloudSdkService.getInstance().validateCloudSdk();
+      if (validationResults.contains(CloudSdkValidationResult.CLOUD_SDK_NOT_FOUND)) {
```
aslo: same question here regarding MALFORMED_PATH

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -386,10 +387,17 @@ private void validateConfiguration() throws ConfigurationException {
         throw new ConfigurationException(
             GctBundle.message("appengine.flex.config.custom.dockerfile.error"));
       }
-    } else if (environment.isStandard() && CloudSdkService.getInstance().validateCloudSdk()
-        .contains(CloudSdkValidationResult.NO_APP_ENGINE_COMPONENT)) {
-      throw new ConfigurationException(
-          CloudSdkValidationResult.NO_APP_ENGINE_COMPONENT.getMessage());
+    } else if (environment.isStandard()) {
```
patflynn: why do we only care about cloud sdk not found for standard?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
aslo: Also, shouldn't this be checked regadless of whether the environment is standard or flex?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineStandardDeployTask.java
```diff
@@ -63,6 +65,16 @@ public void execute(ProcessStartListener startListener) {
...
+      deploy.getCallback().errorOccurred(
+          GctBundle.message("appengine.cloudsdk.location.invalid.message") + " "
+              + CloudSdkService.getInstance().getSdkHomePath());
+      logger.warn(GctBundle.message("appengine.cloudsdk.location.invalid.message") + " "
```
patflynn: do we need to log this? Also log messages should not come from the resource bundle.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
joaoandremartins: With the most recent changes, it is impossible to get here with a persisted malformed SDK path. For the minority of users who persisted a malformed SDK path, they should redeclare the SDK from settings. It won't blow up for them, though.

Spot on in the standard bit. We should always make this check for flex environments.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -120,6 +121,15 @@ public CancellableRunnable createDeployRunner(
       throw new RuntimeException("Invalid deployment source selected for deployment");
     }
 
+    if (CloudSdkService.getInstance().validateCloudSdk().contains(
```
joaoandremartins: @aslo @patflynn @etanshaul because AppEngineDeployStandardTask only handled the standard case, I moved the check here so it also works for Flex.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -120,6 +121,15 @@ public CancellableRunnable createDeployRunner(
...
+        CloudSdkValidationResult.CLOUD_SDK_NOT_FOUND)) {
+      callback.errorOccurred(GctBundle.message("appengine.cloudsdk.location.invalid.message") + " "
+          + CloudSdkService.getInstance().getSdkHomePath());
+      logger.warn(GctBundle.message("appengine.cloudsdk.location.invalid.message") + " "
```
patflynn: why does the first check throw an exception, but this one log a warning? Also you're still using the resource bundle in your log message.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -47,8 +47,9 @@ appengine.cloudsdk.java.components.howtoinstall=To install, run "gcloud componen
 appengine.cloudsdk.location.label=Cloud SDK directory\:
 appengine.cloudsdk.location.browse.window.title=Browse to Cloud SDK directory
 appengine.cloudsdk.location.badchars.message=The specified Cloud SDK directory contains invalid characters or starts or ends with spaces.
-appengine.cloudsdk.location.invalid.message=No Cloud SDK was found in this directory.
+appengine.cloudsdk.location.invalid.message=No Cloud SDK was found in the specified directory.
 appengine.cloudsdk.location.missing.message=Cloud SDK home directory is not specified.
+appengine.cloudsdk.location.missing.message.deployui=The Cloud SDK is misconfigured. To fix, reconfigure the Server above.
```
patflynn: Don't make positional references to a UI in a message.

patflynn: why is location.missing.message 'not specified', but ....message.deployui 'misconfigured'? Can't we just reuse appengine.cloudsdk.location.invalid.message for this error condition?

joaoandremartins: I thought it would make it more explicit when to use. The message is too specific, it doesn't make sense to use it anywhere else other than in the UI. Maybe whoever wants to use it can see that from the message itself, though?

Regarding misconfigured, I wanted that footer UI message to be in line with the other Misconfigured messages in the Devappserver UI. There is an issue in the message name -- appengine.cloudsdk.location.missing.message.deploy, s/missing/invalid

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
joaoandremartins: There is a second check which also returns null, like this one.

The error message is getting correctly printed and the deployment correctly aborted, so my goal is done without having to throw an exception. I don't know why the first check throws an exception, though, but I also wonder if we really need to there.

joaoandremartins: I got rid of the log message too.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentRunConfigurationEditor.java
```diff
@@ -386,10 +387,18 @@ private void validateConfiguration() throws ConfigurationException {
...
-      throw new ConfigurationException(
-          CloudSdkValidationResult.NO_APP_ENGINE_COMPONENT.getMessage());
+    } else {
+      Set<CloudSdkValidationResult> validationResults =
+          CloudSdkService.getInstance().validateCloudSdk();
+      if (validationResults.contains(CloudSdkValidationResult.CLOUD_SDK_NOT_FOUND)) {
+        throw new ConfigurationException(
+            GctBundle.message("appengine.cloudsdk.location.invalid.message.deploy"));
+      }
+      if (environment.isStandard()
```
etanshaul: I think this can be refactored a bit for clarity. e.g. move the standard check up one level.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -48,8 +48,8 @@ appengine.cloudsdk.location.label=Cloud SDK directory\:
 appengine.cloudsdk.location.browse.window.title=Browse to Cloud SDK directory
 appengine.cloudsdk.location.badchars.message=The specified Cloud SDK directory contains invalid characters or starts or ends with spaces.
 appengine.cloudsdk.location.invalid.message=No Cloud SDK was found in the specified directory.
+appengine.cloudsdk.location.invalid.message.deploy=The Cloud SDK is misconfigured. To fix, reconfigure the Server above.
```
etanshaul: since this is a context sensitive message (it references "Server above") how about including the context in the name? Something like:
`appengine.deploymentconfiguration.cloudsdk.location.invalid.message`

etanshaul: oh, I missed that. I was just thinking that since the message is already context sensitive, why not have the message name also indicate the context.

joaoandremartins: Yeah, that's where I wanted to go with adding deploy at the end. Though it might not be too much of a hint.

I'll rename it to appengine.cloudsdk.location.invalid.message.deploymentconfiguration, hopefully it's clearer?

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -47,7 +47,8 @@ appengine.cloudsdk.java.components.howtoinstall=To install, run "gcloud componen
 appengine.cloudsdk.location.label=Cloud SDK directory\:
 appengine.cloudsdk.location.browse.window.title=Browse to Cloud SDK directory
 appengine.cloudsdk.location.badchars.message=The specified Cloud SDK directory contains invalid characters or starts or ends with spaces.
-appengine.cloudsdk.location.invalid.message=No Cloud SDK was found in this directory.
+appengine.cloudsdk.location.invalid.message=No Cloud SDK was found in the specified directory.
+appengine.cloudsdk.location.invalid.message.deploy=The Cloud SDK is misconfigured. To fix, reconfigure the Server above.
```
patflynn: I still don't want us to put UI positional references in our messages. It's  aUI anti-pattern.

appengine.deploymentconfiguration.cloudsdk.location.invalid.message?

patflynn:
I gotta run. I'll let you guys duke this out. please address my remaining comments

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: after talking to @patflynn the concern is the positional reference "above". Can we just change it to "The Cloud SDK is misconfigured. To fix, reconfigure the Server." ?

etanshaul:
lgtm on my end if all of @patflynn comments were addressed

joaoandremartins:
Should be fine now. PTAL

PR #1162, opened 2016-12-08
Update changelog version

etanshaul:


patflynn:
need to think about how we batch together the release notes for multiple RCs. But this is good for now. thx.

PR #1166, opened 2016-12-09
Make the path string OS independent for verification.

### LINKED ISSUE #1165: Seeing unit test failures running from Windows
com.google.cloud.tools.intellij.appengine.cloud.CloudSdkAppEngineHelperTest > testCreateDeployRunnerInvalidSdk FAILED
    org.mockito.exceptions.verification.junit.ArgumentsAreDifferent at CloudSdkAppEngineHelperTest.java:143
                                                                               
com.google.cloud.tools.intellij.appengine.cloud.CloudSdkAppEngineHelperTest > testCreateApplicationDefaultCredentials FAILED
    java.nio.file.FileSystemException at CloudSdkAppEngineHelperTest.java:98

patflynn:
Close the credential file before trying to delete it.

closes #1165

### DIFF
--- a/google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelperTest.java
+++ b/google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelperTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2016 The Android Open Source Project
+ * Copyright 2016 Google Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelperTest.java
```diff
@@ -140,11 +150,13 @@ public void testCreateDeployRunnerInvalidSdk() {
 
     assertNull(runner);
     verify(callback, times(1))
-        .errorOccurred("No Cloud SDK was found in the specified directory. /this/path");
+        .errorOccurred(
+            "No Cloud SDK was found in the specified directory. " + path.toString());
```
joaoandremartins: It was suggested to me before that, by containing the integral path name instead of getting it from the reference, unit tests can be more assertive about our desired behaviour.
Doesn't make much difference in this case, though.

patflynn: Unfortunately in this case the path in the verification is incorrect on windows.

At a higher level, I think it's painful to verify exact messages.  Ideally we'd be verifying that a certain type of error occurred.

PR #1170, opened 2016-12-12
Application creation feature - services

aslo:
This PR contains the business logic / services for the application creation feature. Note that it does not target master, but instead https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1169, the parent PR for the application creation feature.

Notable changes:
- added new application service: **AppEngineAdminService** -  wraps the app engine admin api & provides some convenience methods for interacting with it. 
- added new application service: **GoogleApiClientFactory** - factory class for instantiating google API clients, since many of them have common instantiatoin logic and can benefit from sharing certain resources


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/DefaultAppEngineAdminService.java
```diff
@@ -0,0 +1,121 @@
...
+      @NotNull Credential credential) throws IOException {
+    try {
+      return GoogleApiClientFactory.getInstance().getAppEngineApiClient(credential)
+          .apps().get(projectId).execute();
+    } catch (GoogleJsonResponseException e) {
+      if (e.getStatusCode() == 404) {
+        // the application does not exist
+        return null;
+      }
+      throw e;
```
patflynn: Should we be throwing GoogleJsonReponseException?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/DefaultAppEngineAdminService.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  @Override
+  public Application createApplication(@NotNull String locationId, @NotNull final String projectId,
+      @NotNull final Credential credential) throws IOException, AppEngineOperationFailedException {
+
+    Application arg = new Application();
+    arg.setId(projectId);
+    arg.setLocationId(locationId);
+    Operation operation = GoogleApiClientFactory.getInstance().getAppEngineApiClient(credential)
```
patflynn: Dos the factory instance get cached?

patflynn: also same question about the App Engine API client.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/DefaultAppEngineAdminService.java
```diff
@@ -0,0 +1,121 @@
...
+        .apps().create(arg).execute();
```
patflynn: just curious does the AppEngineApiClient expose an async implementation?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineAdminService.java
```diff
@@ -0,0 +1,77 @@
...
+      @NotNull final String projectId, @NotNull final Credential credential)
+      throws IOException, AppEngineOperationFailedException;
+
+  /**
+   * Returns a list of all available App Engine Locations
+   *
+   * @param credential the authenticated user Credential
+   * @throws IOException if there was a transient error connecting to the API
+   */
+  public abstract List<Location> getAllAppEngineLocations(Credential credential) throws IOException;
```
aslo: I opted for using a String parameter in createApplication because this way callers don't need to construct a Location when all we need is the locationId. Also, the Location.locationId field is just a string as well - in my mind it's not really any safer.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineAdminService.java
```diff
@@ -0,0 +1,77 @@
...
+   */
+  @Nullable
+  public abstract Application getApplicationForProjectId(@NotNull String projectId,
+      @NotNull Credential credential) throws IOException;
+
+  /**
+   * Creates an Application for the given project in the given location. This is a long-running
+   * operation that can typically take up to one minute.
+   *
+   * @param locationId The GCP Location in which the application will be located
```
aslo: See my comment below on the decision to use a String type.

Also, FYI technically this is a different concept than a zone or a region - some Locations, such as `us-east` is multi-regional.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/DefaultAppEngineAdminService.java
```diff
@@ -0,0 +1,121 @@
...
+import com.google.cloud.tools.intellij.appengine.cloud.AppEngineOperationFailedException;
+import com.google.cloud.tools.intellij.resources.GoogleApiClientFactory;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.io.IOException;
+import java.util.List;
+
+public class DefaultAppEngineAdminService extends AppEngineAdminService {
```
aslo: GoogleApiClientAppEngineAdminService works for me

aslo:
@elharo I know @patflynn had some concerns about adding this to appengine-plugins-core. @patflynn, do you still think this type of logic would be inappropriate to add there?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/DefaultAppEngineAdminService.java
aslo: Not sure I understand what you're asking. Are you suggesting wrapping this in an unchecked exception or something like that? Or are you suggesting declaring `GoogleJsonReponseException` in the throws clause of the method signature?

FYI GoogleJsonResponseException inherits from IOException, so this is a way of allowing callers to deal with a transient exception (i.e. perhaps to retry).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/DefaultAppEngineAdminService.java
aslo: The factory is registered as an applicationService, so yep, it gets cached by the application ServiceManager (It's effectively a singleton). 

App Engine API Clients are not cached (that is, each call to a `GoogleApiClientFactory.get*Client` method will return a new instance of the api client.) Caching of api client instances within the GoogleApiClientFactory doesn't seem to make sense since according to the [javadoc](https://developers.google.com/resources/api-libraries/documentation/cloudresourcemanager/v1beta1/java/latest/) clients are not thread-safe.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/DefaultAppEngineAdminService.java
aslo: Doesn't appear to, unfortunately.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/DefaultAppEngineAdminService.java
```diff
@@ -0,0 +1,121 @@
...
+
+    while (response.getNextPageToken() != null) {
+      response = getAppEngineRegions(credential, response.getNextPageToken());
+      locations.addAll(response.getLocations());
+    }
+
+    return locations;
+  }
+
+  private ListLocationsResponse getAppEngineRegions(Credential credential, @Nullable String
```
joaoandremartins: Besides `createApplication()`, it looks like all of the methods would be better called straight from the API. i.e., This service isn't adding much value.

If that's the case, I would rather create a couple of utility methods somewhere that automate routines needed in more than one place and directly call the API, rather than creating this service abstraction.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/DefaultGoogleApiClientFactory.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.api.client.json.jackson.JacksonFactory;
+import com.google.api.services.appengine.v1.Appengine;
+import com.google.api.services.cloudresourcemanager.CloudResourceManager;
+import com.google.cloud.tools.intellij.CloudToolsPluginInfoService;
+
+import com.intellij.openapi.components.ServiceManager;
+
+import org.jetbrains.annotations.Nullable;
+
+public class DefaultGoogleApiClientFactory extends GoogleApiClientFactory {
```
joaoandremartins: The added value of this class is also not clear to me.

joaoandremartins:
Not sure I agree with the design. It sounds to me like we're adding abstraction layers unnecessarily. Depending on our usage of this, I'd much rather just call the API directly.

I have a strong preference for not adding this to appengine-plugins-core. As we discussed in the meeting, the common library provides a Java API for things that don't have a Java API, yet, which isn't the case for Cloud Resource Manager or App Engine.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/DefaultGoogleApiClientFactory.java
aslo: The main value of this class is that it holds static (global) references to `NetHttpTransport` and `JacksonFactory` instances, which are required for instantiation of several different Google API clients. Both of these classes are thread-safe, and both actually recommend sharing global instances (see javadoc for [NetHttpTransport](https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/http/javanet/NetHttpTransport) and [JacksonFactory](https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/json/jackson/JacksonFactory) for efficiency.

aslo:
:+1: I'm good on not including this is appengine-plugins-core

aslo:
Just pushed a new change to address some more comments from offline discussion with @patflynn: Modified the API of AppEngineAdminService to use more thoughtfully designed exception types. 

All methods in that service now can throw two exception types: `IOException`, when a transient or unexpected error occurs, and a `GoogleApiException`, when a request successfully completes but returns an error response. (A common case would be if we get a 400 response if user's credentials are revoked).

patflynn:
@aslo did you have a chance to check what gets thrown when the network is out?  It'd be nice for the client to know when to report an error to feedback vs transient network problems.

aslo:
Yep, if there's no network any API call throws an UnknownHostException, which extends IOException.  Currently, IOExceptions like this are being re-thrown by the GoogleApiClientAppEngineAdminService to be handled by the client, which I think is appropriate since this falls under the category of a transient exception. 

On the other hand, any issue relating to auth or misconfiguration on our part will result in us throwing a GoogleApiException. These can be handled differently at the UI level.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/DefaultGoogleApiClientFactory.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private static HttpTransport httpTransport = new NetHttpTransport();
```
patflynn: are these guys immitable? Should they be final and all caps?

patflynn:
I'd never considered whether an object's effect on external systems
qualifed as 'observable state'.  For instance the argument against Logger
being constant case has typically been that it has observable mutable
state, rather than that it performs IO mutations on disk.  However in this
case it's just not particularly obvious whether these classes are immutable
in the more narrow sense. JacksonFactory clearly seems to not be, and
NetHttpTransport I'm not really sure just by looking at it.  So fine with
leaving as is.

aslo:
Fair point, perhaps I was interpreting the phrase "observable side effects" too literally.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/GoogleApiClientAppEngineAdminService.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.List;
+
+/**
+ * An implementation of {@link AppEngineAdminService} that uses a Google API Client to communicate
+ * with the App Engine service.
+ */
+public class GoogleApiClientAppEngineAdminService extends AppEngineAdminService {
+
+  private final static String APP_ENGINE_RESOURCE_WILDCARD = "-";
+  private final long CREATE_APPLICATION_POLLING_INTERVAL_MS = 1000;
```
aslo: Good catch. I'm surprised our static analysis tools didn't pick this up. I seem to remember this sort of thing failing CI builds in the past...

patflynn:
related to #1232

PR #1175, opened 2016-12-14
Adds more error output to failed checkouts.

### LINKED ISSUE #1142: Checking out from Google Cloud Repositories errors out
Not a lot of detail in the error message.

We're either not sending the correct scope in the credential, or something else wacky is going on. I'm only getting this from my Windows VM, Linux is working fine.

screenshot from 2016-12-06 15 34 23

screenshot from 2016-12-06 15 33 55

joaoandremartins:
Fixes #1142 

Looks like this:

screenshot from 2016-12-14 14 38 07

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/GcpCheckoutProvider.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/GcpCheckoutProvider.java
@@ -16,12 +16,14 @@
 
 package com.google.cloud.tools.intellij.vcs;
 
+import com.google.api.client.repackaged.com.google.common.base.Joiner;
 import com.google.api.client.repackaged.com.google.common.base.Strings;
 import com.google.cloud.tools.intellij.stats.UsageTrackerProvider;
 import com.google.cloud.tools.intellij.util.GctBundle;
 import com.google.cloud.tools.intellij.util.GctTracking;
 
 import com.intellij.ide.util.PropertiesComponent;
+import com.intellij.notification.NotificationListener.UrlOpeningListener;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.progress.ProgressIndicator;
 import com.intellij.openapi.progress.Task;
@@ -46,6 +48,9 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.io.File;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/GcpCheckoutProvider.java
```diff
@@ -186,7 +186,8 @@ private static boolean doClone(@NotNull Project project, @NotNull ProgressIndica
       return true;
     }
     VcsNotifier.getInstance(project)
-        .notifyError(GctBundle.message("clonefromgcp.failed"), result.getErrorOutputAsHtmlString());
+        .notifyError(GctBundle.message("clonefromgcp.failed"), result.getErrorOutputAsHtmlString()
+            + " " + result.getOutputAsJoinedString());
```
etanshaul: what if you used getOutputAsHtmlString; could we then get the link clickable?

etanshaul: also, can we add a <br /> between the error and the output instead of just a space?

joaoandremartins: Unfortunately, that method doesn't exist :(

etanshaul: one thing I noticed, you may be able to use this overloaded method instead and provide a listener for the URL clicks:

`notifyError(@NotNull String title, @NotNull String message, @Nullable NotificationListener listener)`

etanshaul: oops you're right, I don't know what I was looking out, getOutputAsHtmlString indeed does not exist

joaoandremartins:
Absolutely.

On that, I was also thinking we could try to automate some of this, but I don't know how hard that would be. It would involve creating a ~/.netrc file, which the IDE might not have access to, and adding the output of the identifier creation to it, which is not trivial.

patflynn:
Our implementation uses a git credential helper and not .netrc. I think
that .netrc is an old gerrit way of doing things that may be obsolete at
this point.

joaoandremartins:
The web page users are directed to explicitly mentions a .netrc file.

I was using credential.helper gcloud.sh and could reproduce the bug.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/GcpCheckoutProvider.java
```diff
@@ -186,8 +190,24 @@ private static boolean doClone(@NotNull Project project, @NotNull ProgressIndica
...
+        .notifyError(GctBundle.message("clonefromgcp.failed"), result.getErrorOutputAsHtmlString()
+            + "<br>" + resultWithHtmlUrl(result.getOutput()), new UrlOpeningListener(true));
     return false;
   }
 
+  /**
+   * Searches for URLs inside a string list and adds a HTML "a" element, so they're openable from
+   * the notification.
+   */
+  private static String resultWithHtmlUrl(List<String> output) {
```
joaoandremartins: @etanshaul PTAL

I decided to try for URLs using startsWith instead of new URL(), so I wouldn't have to handle exceptions.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/GcpCheckoutProvider.java
```diff
@@ -187,8 +192,23 @@ private static boolean doClone(@NotNull Project project, @NotNull ProgressIndica
...
+    List<String> result = new ArrayList<>(output);
+    int position = 0;
+    for (String line : result) {
+      if (line.trim().startsWith("http://") || line.trim().startsWith("https://")) {
```
etanshaul: this requires the url to encompass the entire line right?

joaoandremartins: Actually, I think there's a far superior solution using regex. One sec

etanshaul:
lgtm assuming that regex is good

joaoandremartins:
Regex was tested and works. I got it from SO :X

PR #1176, opened 2016-12-19
Move to IJ 2016 as the lowest supported version of IJ

patflynn:
includes setting the project language level to java 8 since 2016 is java 8 based.

I had to upgrade the gradle-intellij-plugin to get this to build with the new config.

the random test case refactor is error-prone suddenly behaving slightly differently with the config change.

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -13,17 +13,22 @@
       <entry name="!?*.kt" />
       <entry name="!?*.clj" />
     </wildcardResourcePatterns>
+    <annotationProcessing>
+      <profile default="true" name="Default" enabled="false">
+        <processorPath useClasspath="true" />
+      </profile>
+    </annotationProcessing>
     <bytecodeTargetLevel>
-      <module name="common-lib_main" target="1.7" />
-      <module name="common-lib_test" target="1.7" />
-      <module name="common-test-lib_main" target="1.7" />
-      <module name="common-test-lib_test" target="1.7" />
-      <module name="google-account-plugin_main" target="1.7" />
-      <module name="google-account-plugin_test" target="1.7" />
-      <module name="google-cloud-tools-plugin_main" target="1.7" />
-      <module name="google-cloud-tools-plugin_test" target="1.7" />
-      <module name="ultimate_main" target="1.7" />
-      <module name="ultimate_test" target="1.7" />
+      <module name="common-lib_main" target="1.8" />
+      <module name="common-lib_test" target="1.8" />
+      <module name="common-test-lib_main" target="1.8" />
+      <module name="common-test-lib_test" target="1.8" />
+      <module name="google-account-plugin_main" target="1.8" />
+      <module name="google-account-plugin_test" target="1.8" />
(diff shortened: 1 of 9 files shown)


.idea/compiler.xml
```diff
@@ -13,17 +13,22 @@
...
+      <module name="common-test-lib_test" target="1.8" />
+      <module name="google-account-plugin_main" target="1.8" />
+      <module name="google-account-plugin_test" target="1.8" />
+      <module name="google-cloud-tools-plugin_main" target="1.8" />
+      <module name="google-cloud-tools-plugin_test" target="1.8" />
+      <module name="ultimate_main" target="1.8" />
+      <module name="ultimate_test" target="1.8" />
     </bytecodeTargetLevel>
   </component>
 </project>
```
joaoandremartins: Missing eol

etanshaul:
I'm happy to see that upgrading gradle-intellij-plugin didn't cause you any trouble with the nested module structure of our project. This was why I delayed the upgrade prevously.

patflynn:
Hopefully it hasn't. Still doing some manual testing.

google-account-plugin/resources/META-INF/plugin.xml
```diff
@@ -13,15 +13,13 @@
...
+
 <idea-plugin version="2">
   <id>com.google.gct.login</id>
   <name>Google Account</name>
   <vendor>Google</vendor>
   <!-- "version" set by gradle-intellij-plugin -->
   <!-- "idea-version since-build" set by gradle-intellij-plugin -->
-  <!-- Workaround for gradle-intellij-plugin limitation: https://github.com/JetBrains/gradle-intellij-plugin/issues/66 -->
-  <!-- Setting version to 15.0.6 and up. https://confluence.jetbrains.com/display/IDEADEV/IDEA+15+EAP -->
-  <idea-version since-build="143.2370.31" />
```
etanshaul: don't we still want to set a version here so that it shows up properly in the IJ repo page?

etanshaul: See https://github.com/JetBrains/gradle-intellij-plugin/issues/66 for the new way to specify version ranges.

patflynn: by default it's configured to the version we are running the build against.  Although I'll double-check

patflynn:
Yeah I saw that,  but it's only available in a snapshot release.

patflynn:
@etanshaul PTAL

etanshaul:
The only thing I see remaining is my comment about being explicit about our supported versions: e.g. specifying 2016+ in plugin.xml. Otherwise LGTM.

patflynn:
I'll reintroduce that as soon as we have a nice way of specifying it.  Right now I'm loathe to lookup the right build number every time we update the sdk version we build against.

etanshaul:
sgtm if you confirmed that it implicitly will be set to 2016+ if nothing is specified.

patflynn:
Thanks for pushing on that. I checked and it wasn't. pushed a change that
fixes that.

PR #1179, opened 2016-12-22
Application creation feature - UI and configuration model

aslo:
This PR implements the UI and configuration model changes required for the Application Creation feature. (See https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1169 for the main feature PR). Unit tests will be added shortly to this PR. 

See below for screenshots.

---

1. User opens the deploy run configuration and selects a project w/o an application. The deployment is invalid, and the user is prompted to create a project.
screen shot 2016-12-22 at 12 25 05 pm

2. User clicks the "click here" link, and this dialog appears. They select a region for their application.
screen shot 2016-12-22 at 12 25 12 pm

3. They click OK and the standard intellij loading bar appears.
screen shot 2016-12-22 at 12 25 34 pm

4. When app creation is complete, the dialogs disappear and the region of their newly created application is shown in the run config editor. The user can now deploy to their new application.
screen shot 2016-12-22 at 12 25 54 pm


patflynn:
Nice work! I doubt I''ll take a look at this today, so feel free to work
with Etan to get it merged next week.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -25,6 +25,19 @@ plugin.conflict.error.title=Google Plugin Conflict
 plugin.conflict.error.detail=You are running an old plugin ({0}) that conflicts with the installed Google Cloud Tools for Java plugin.
 plugin.conflict.error.action=Please {0}click here{1} to deactivate the old plugin.
 appengine.name=Google App Engine
+appengine.application.not.exist=An App Engine application does not exist for this project.
+appengine.application.required.deployment=An App Engine application must be created before you can deploy to App Engine.
+appengine.application.create={0}Click here{1} to create an application.
+appengine.application.create.instructions=Please select a region for your App Engine Application. Once you choose it, you will not be able to change it later.
+appengine.application.create.documentation=For more information about App Engine regions, see the {0}documentation{1}.
+appengine.application.create.error.transient=An unexpected error occurred while creating the application. Please try again later.
+appengine.application.create.loading=Creating App Engine application in {0}. This can take up to a minute.
```
patflynn: I think 'this can take a while' is a more googly voice and more accurate.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -25,6 +25,19 @@ plugin.conflict.error.title=Google Plugin Conflict
...
+appengine.application.region.supported.environments=The region {0} supports the following App Engine environments:
+appengine.application.region.supports.standard={0}App Engine Standard{1}
```
patflynn: I hadn't really thought about this.  The fact that different regions have different levels of support.

patflynn: I'm not sure if it's in here somewhere, but it should be clear to the user that they can't change the region after the fact.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.form
```diff
@@ -0,0 +1,64 @@
...
+        <constraints>
+          <grid row="2" column="0" row-span="1" col-span="1" vsize-policy="0" hsize-policy="1" anchor="0" fill="1" indent="0" use-parent-layout="false">
+            <minimum-size width="200" height="-1"/>
+          </grid>
+        </constraints>
+        <properties>
+          <enabled value="true"/>
+          <font/>
+          <model>
+            <item value="Loading available regions..."/>
```
patflynn: this needs to be pulled out to the resource bundle.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
```diff
@@ -0,0 +1,206 @@
...
+  private static final String HTML_OPEN_TAG = "<html><font face='sans' size='-1'>";
+  private static final String HTML_CLOSE_TAG = "</font></html>";
+
+  private JPanel panel;
+  private JTextPane instructionsTextPane;
+  private JComboBox<AppEngineLocationSelectorItem> regionComboBox;
+  private JTextPane statusPane;
+  private JTextPane regionDetailPane;
+
+  private final Credential userCredential;
```
patflynn: Why store this here instead of retrieving it lazily when you need it from the user service?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
```diff
@@ -0,0 +1,206 @@
...
+
+  public AppEngineApplicationCreateDialog(@NotNull Component parent, @NotNull String gcpProjectId,
+      @NotNull Credential userCredential) {
+    super(parent, false);
+
+    this.gcpProjectId = gcpProjectId;
+    this.userCredential = userCredential;
+
+    init();
+    setTitle(GctBundle.message("appengine.application.region.select"));
```
patflynn: I'm wondering why this stuff isn't part of init()

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
```diff
@@ -0,0 +1,206 @@
...
+
+      close(OK_EXIT_CODE);
+
+    } catch (IOException e) {
+      trackApplicationCreateFailure();
+      setStatusMessage(GctBundle.message("appengine.application.create.error.transient"), true);
+      setOKActionEnabled(true);
+      return;
+    } catch (GoogleApiException e) {
+      trackApplicationCreateFailure();
```
patflynn: do we want to get error reports for these?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineLocationSelectorItem.java
```diff
@@ -0,0 +1,79 @@
...
+  public boolean isStandardSupported() {
+    return parseMetadataBoolean(STANDARD_ENV_AVAILABLE_KEY, location.getMetadata());
+  }
+
+  private boolean parseMetadataBoolean(String key, Map<String, Object> metadata) {
+    if (!metadata.containsKey(key)) {
+      return false;
+    }
+    Object val = metadata.get(key);
+    if (val == null || !(val instanceof Boolean)) {
```
patflynn: If val is not an instance of boolean would that be surprising? Would we want to know if that's happening?

patflynn:
We seem to have lost test coverage updates on the PR. I'll look into that.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
```diff
@@ -0,0 +1,206 @@
...
+        appEngineRegions = AppEngineAdminService.getInstance()
+            .getAllAppEngineLocations(userCredential);
+      } catch (IOException | GoogleApiException e) {
+        setStatusMessageAsync(GctBundle.message("appengine.application.region.list.fetch.error"),
+            true);
+        return;
+      }
+
+      // perform the actual UI updates on the event dispatch thread
+      ApplicationManager.getApplication().invokeLater(() -> {
```
joaoandremartins: Why creating a new thread instead of doing this in the current pooled one?

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
aslo: 👍  I like that

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
aslo: yep, that's currently part of the instructions in the app engine application creation dialog

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.form
aslo: 👍

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
aslo: That might be nice, but is there a way to send an error report without throwing an uncaught exception to the user? Ideally I'd like to handle these gracefully and avoid showing the user those red exception popups.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineLocationSelectorItem.java
aslo: I think we would like to know, but if we can handle that case silently/gracefully I'd prefer to do that. (See my above comment regarding error reports). If there's a way to silently send error reports w/o bubbling up the exception, that would probably be a good idea here. I'll look into that.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -25,6 +25,19 @@ plugin.conflict.error.title=Google Plugin Conflict
 plugin.conflict.error.detail=You are running an old plugin ({0}) that conflicts with the installed Google Cloud Tools for Java plugin.
 plugin.conflict.error.action=Please {0}click here{1} to deactivate the old plugin.
 appengine.name=Google App Engine
+appengine.application.not.exist=An App Engine application does not exist for this project.
+appengine.application.required.deployment=An App Engine application must be created before you can deploy to App Engine.
+appengine.application.create={0}Click here{1} to create an application.
+appengine.application.create.instructions=Please select a region for your App Engine Application. Once you choose it, you will not be able to change it later.
```
aslo: 👍

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
aslo: UI updates need to be performed from the swing event dispatch thread. Since we want to update the list of regions from a background thread, we need to use `Application.invokeLater` in order to execute the code that modifies the UI within the event dispatch thread.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
aslo: I guess there's no good reason - I'll override init() in this class and move these things into there.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
aslo: Right now there isn't a service that provides access to a credential in that fashion (that I know of). The credential is originally retrieved from the project selector, since there can be multiple authed users - any project for any of those users is a valid deploy target.

The intended usage pattern here is to dispose of instances of this dialog class once it's closed, and create new instances on demand with the correct credential for a given project.  I'll add some comments that will hopefully clear that up.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * Dialog that allows creation of an App Engine Application
+ */
+public class AppEngineApplicationCreateDialog extends DialogWrapper {
+
+  private final static String LOCATIONS_DOCUMENTATION_URL
+      = "https://cloud.google.com/docs/geography-and-regions";
+  private final static String FLEX_DOCUMENTATION_URL
+      = "https://cloud.google.com/appengine/docs/flexible/";
```
etanshaul: really trivial, but you have a trailing "/" here but not on the other url's

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
```diff
@@ -0,0 +1,206 @@
...
+  private final static String STANDARD_DOCUMENTATION_URL
+      = "https://cloud.google.com/appengine/docs/about-the-standard-environment";
+  private static final String HTML_OPEN_TAG = "<html><font face='sans' size='-1'>";
```
etanshaul: Is this the IntelliJ recommended way? `<font>` is discouraged these days.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
```diff
@@ -0,0 +1,206 @@
...
+  protected void doOKAction() {
+    final Location selectedLocation
+        = ((AppEngineLocationSelectorItem) regionComboBox.getSelectedItem()).getLocation();
+
+    // show loading state
+    setOKActionEnabled(false);
+
+    try {
+      UsageTrackerProvider.getInstance()
+          .trackEvent(GctTracking.APP_ENGINE_APPLICATION_CREATE)
```
etanshaul: Just wondering if these event is necessary since ultimately won't either success or failure get logged?

etanshaul: Actually never mind; it is nice to have an aggregate metric like this

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
aslo: 👍

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
```diff
@@ -546,4 +596,36 @@ public void setModelNeedsRefresh(boolean modelNeedsRefresh) {
       this.modelNeedsRefresh = modelNeedsRefresh;
     }
   }
+
+  /**
+   * Interface that must be implemented in order to be informed of
+   * {@link ProjectSelectionChangedEvent} events.
+   */
+  public interface ProjectSelectionListener {
```
etanshaul: I like what you done here but I'm wondering if something should be refactored in the superclass `CustomizableComboBox` since its already setup for listening to selection changes?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
aslo: Good point, though adding metrics doesn't cost us anything, so I was thinking it would just be nice to include anyway. 

Thinking about it now, there are some situations where this could be valuable - since this is a long-running operation, it could be good to track requests made separately from completions, e.g. if the user cancels the request or kills the application while it's running.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
aslo: I agree that this does seem a bit weird. I'll see if I can set the font in a different way.

etanshaul:
I wonder if this is a regression (notice the white border around "To manually promote...":

screen shot 2016-12-27 at 11 19 20 am

This is in 2016.3

etanshaul:
I just tested out this scenario:

1) Open a Flex Project in IDEA
2) Create a new cloud project, select a region that supports only Standard
 -> There is no warning here (I'm not sure if its the right place for it yet though)
3) Attempt to deploy the app
 -> I get some weird errors that vaguely hint at a region error:

screen shot 2016-12-27 at 11 32 04 am

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
aslo: That's a good idea, though I'm not really sure what is worth pulling up into the parent class.  `CustomizableComboBox` already has a way of listening on selection changes to tree nodes, using swing `TreeSelectionListeners`.

The main value of the ProjectSelectionListener interface is that it provides events that are data-aware - meaning it's provides the `Project` object and the user's `Credential`directly to the listeners, which it is able to provide since it's aware of the tree structure of the project selector's data model. 

I don't think this would make sense to move to a higher level of abstraction.

aslo:
How do you think we should handle that case? Where should a warning like that live? Maybe a balloon notification when the user kicks off a deployment?

etanshaul:
The error I first showed was actually a billing error. I updated the screenshot to show the actually error which is much nicer. I think we can leave it as is for now.

aslo:
I can't seem to replicate the white border issue on 2016.3. Is it definitely a regression?

etanshaul:
I'm getting white borders around most of these UI elements. I'm not sure whats going on: 

screen shot 2016-12-27 at 11 42 46 am

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
etanshaul: sounds good

etanshaul:
Looks really good and the UI seems to work nicely. Pending any updates to my comments (I'm not sure whats up with those borders) feel free to merge.

aslo:
Regarding the borders, I've only tested on linux, so it could be something that only shows up on OSX. I'll look into it a bit more.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
aslo: After looking into this, I actually think the font tag is the right way to set a custom font on html element. ([This swing tutorial](http://docs.oracle.com/javase/tutorial/uiswing/components/html.html) seems to imply it anyway).

etanshaul: Sounds good

aslo:
@etanshaul FYI I solved the white border issue by just making all of the offending UI elements non-opaque, instead of just using the default background color.

patflynn:
#1232

PR #1169, opened 2016-12-12
Create App Engine Applications 

aslo:
This is the parent PR for the Application Creation feature.  **Please do not review or merge this PR.**
 
Development for this feature will be split into several smaller logical chunks, which will be referenced below. All reviews and comments should take place within those PRs. Once all sub-PRs are merged into this one, I'll ask for a final review of this pull request before merging it into master.

Topic | PR | Reviewed | Merged 
---|---|---|---
business logic and app engine API clients |https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1170 | <ul><li> [x] </li></ul> | <ul><li> [x] </li></ul>
UI changes |https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1179 | <ul><li> [x] </li></ul> | <ul><li> [x] </li></ul>

PR #1180, opened 2016-12-22
Post-filter deleted projects

aslo:
fixes https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1119. I decided to take care of this now, since this bug causes some issues for #1169. 

Since the Cloud Resource Manager API doesn't support filtering projects on their lifecycle status, we need to post-filter the results we get before displaying to the user. I also took the liberty of converting some code to take advantage of Java 8 features while I was in the neighborhood.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/GoogleUserModelItem.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/GoogleUserModelItem.java
@@ -37,7 +37,6 @@ import java.awt.Image;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
-import java.util.Comparator;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/GoogleUserModelItem.java
```diff
@@ -140,22 +137,24 @@ public void run() {
...
-          public int compare(Project p1, Project p2) {
-            return p1.getName().toLowerCase().compareTo(p2.getName().toLowerCase());
-          }
-        });
-        allProjects.addAll(response.getProjects());
+        // Create a sorted set to sort the projects list by project ID.
+        Set<Project> allProjects = new TreeSet<>((Project p1, Project p2) ->
+            p1.getName().toLowerCase().compareTo(p2.getName().toLowerCase()));
+
+        response.getProjects().stream()
```
patflynn: love it!

PR #1181, opened 2017-01-04
Application creation feature - caching

aslo:
Adds some caching to make the UI for application creation a bit snappier. 

Summary of changes:
- App engine application resources are cached within the `GoogleApiClientAppEngineAdminService` using a Guava Cache, with application IDs as keys. If the application does not exist, we will not cache a null result, as we cannot reliably invalidate the result when an application is eventually created. (i.e. if the user creates one in the web console we wouldn't know about it).  Cached applications are invalidated once per day.
- A list of all known app engine location resources are cached within the `GoogleApiClientAppEngineAdminService`. This list should change infrequently, and this cache is invalidated once per day. 
- Since all these caches are just in memory, they will also be invalidated when the user restarts the application


elharo:
Aside from a couple of JetBrains null annotations, I don't see a lot of IntelliJ specific functionality in  this work. Should this be going into appengine-plugins-core instead? I'd rather not reinvent it for Eclipse.

patflynn:
@aslo let's chat about this today.  A cursory view makes me think there should be a cleaner way to isolate cache management from the service code.

aslo:
@patflynn SGTM. One thought that occurs to me would be to move the caching code into the abstract class `AppEngineAdminService`, and change its abstract methods, such that extending classes only need to implement methods that make the remote calls. 

In that scenario, `GoogleApiClientAppEngineAdminService` would just be responsible for managing the API client and making the API calls.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/GoogleApiClientAppEngineAdminService.java
```diff
@@ -40,18 +47,62 @@
...
   private static final String APP_ENGINE_RESOURCE_WILDCARD = "-";
   private static final long CREATE_APPLICATION_POLLING_INTERVAL_MS = 1000;
+  private static final int APPLICATION_CACHE_MAX_SIZE = 10000;
+  private final long LOCATION_CACHE_REFRESH_PERIOD_MILLIS = 1000 * 60 * 60 * 24; // 1 day
+
+  // Cache of GCP application resources. This assumes that project IDs are globally unique. Null or
+  // missing applications are not cached because they cannot be reliably invalidated.
+  private final Cache<String, Application> appEngineApplicationCache = CacheBuilder.newBuilder()
+      // arbitrary size limit
+      .maximumSize(APPLICATION_CACHE_MAX_SIZE)
```
patflynn: Is this a required property?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/GoogleApiClientAppEngineAdminService.java
```diff
@@ -40,18 +47,62 @@
...
+      // Even though these values should never change, it won't kill us to refresh once per day.
+      .expireAfterWrite(1, TimeUnit.DAYS)
+      .build();
+
+  // Cache of all available GCP locations. This list is expected to change very infrequently.
+  private final List<Location> appEngineLocationCache = new CopyOnWriteArrayList<>();
```
patflynn: can we move to using Suppliers.memoizeWithExpiration here?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/GoogleApiClientAppEngineAdminService.java
aslo: Not required, but I thought a good thing to include.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/GoogleApiClientAppEngineAdminService.java
aslo: Yep!

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/GoogleApiClientAppEngineAdminService.java
patflynn: It seems like a very large default size.

patflynn: ah nm. **maximum**size. ignore me.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/GoogleApiClientAppEngineAdminService.java
aslo: Actually, that appears to not be an option, since an up-to-date credential is required in order to re-fetch the list of locations. I think it may actually just be better to use a regular map-based guava cache, that will just have a single value. 

I think I may also pull all of the caching logic into the parent of this class, the abstract service class. I think that logically makes a bit more sense and will help keep this code more readable.

patflynn: GoogleLoginService.getCredential()?

aslo: That's a good point; that would work. I played around with that approach, and while it does lead to cleaner code than we currently have here, I still vote against it, since by not requiring the user to call the `getAllRegions` method with a valid credential, we open ourselves up to a new class of issues, i.e. if that method is called without a logged in user. I really don't want to depend on that external state in order to have this call succed. 

I think the best compromise is to use a map-based cache with a single value, as that would support time-based eviction, and allows us to pass credentials at lookup time.

patflynn:
It seems kind of dumb that the API requires a credential to make this call, since I have to assume that app engine locations are public. Out of curiosity have you tried making the call without one?

Also what's the behavior of the form if the network is out?

aslo:
I just tried making the call without credentials, and we get a 401. Not sure why they require it either. Maybe something to do with rate limiting?

If the network is out, the project selector would fail to load available projects and shows an error message. Since there are no projects, we can't check for applications, so the "region" form field just stays blank.

joaoandremartins:
The location for a specific GAE app isn't public, though. Not sure why.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineAdminService.java
```diff
@@ -27,14 +30,35 @@
...
 import java.io.IOException;
 import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 
 /**
- * A Service that handles interactions with the App Engine Admin API. See
- * <a href="https://cloud.google.com/appengine/docs/admin-api/">
- *   https://cloud.google.com/appengine/docs/admin-api/</a>
+ * A Service that handles App Engine Administration. This class provides some general caching logic,
```
patflynn: 'Service' shouldn't be capitalized.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineAdminService.java
```diff
@@ -49,8 +73,54 @@ public static AppEngineAdminService getInstance() {
    * @throws IOException if there was a transient error connecting to the API
    */
   @Nullable
-  public abstract Application getApplicationForProjectId(@NotNull String projectId,
-      @NotNull Credential credential) throws IOException, GoogleApiException;
+  public Application getApplicationForProjectId(@NotNull String projectId,
```
patflynn: two thoughts: 
1) I thought we were only using an abstract class as a pseudo-interface that we could include a factory method on. (actually can we do that with an interface default method?)

2) It doesn't feel as though there are likely to be more than one subclasses of this abstract class.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineAdminService.java
```diff
@@ -49,8 +73,54 @@ public static AppEngineAdminService getInstance() {
...
+        handleExecutionException(e);
+        throw new RuntimeException(e);
+      }
+    }
+  }
+
+  /**
+   * Returns a list of all available App Engine Locations
+   *
+   * @param credential the authenticated user Credential
```
patflynn: lower case credential, or if you're referring to  the Credential type then {@link Credential}

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineAdminService.java
aslo: Fair points. The nice thing about this abstraction is that it lets us separate the caching logic from the actual business of communicating with the App Engine admin API. 

But perhaps this was a bad move since it breaks with the pattern we've been using throughout the codebase for these services. I'll move this code back into the implementation class.

patflynn: I'm fine with separating those concerns, but not using inheritance. Put it in the implementation or factor it out some other way.

aslo:
@patflynn I believe I've addressed all of your concerns - anything else before I merge?

PR #1182, opened 2017-01-05
Add error message to deploy metrics label

etanshaul:
The deploy error message (e.g. "error creating stage directory", or "no cloud sdk was found ...") is currently not being tracked via our metrics. The error message was being shown to the user, but we had no way of digging into the deployment failure to determine more details.

This PR appends the error message to the metrics label.

@chanseokoh do you mind taking a quick look to see if this makes sense to you on the metrics side? i.e is this an ok place to put the message?

### DIFF
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
@@ -22,7 +22,6 @@ import com.google.common.base.Joiner.MapJoiner;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
-import com.google.common.collect.Maps;
 import com.google.common.escape.CharEscaperBuilder;
 import com.google.common.escape.Escaper;
 
@@ -50,6 +49,7 @@ import org.jetbrains.annotations.Nullable;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 /** Google Usage Tracker that reports to Cloud Tools Analytics backend. */
 public class GoogleUsageTracker implements UsageTracker, SendsEvents {
@@ -131,9 +131,7 @@ public class GoogleUsageTracker implements UsageTracker, SendsEvents {
   public void sendEvent(
       @NotNull String eventCategory,
       @NotNull String eventAction,
-      @Nullable String eventLabel,
-      @Nullable Integer eventValue,
-      @Nullable String eventMessage) {
+      @Nullable Map<String, String> metadataMap) {
     if (!ApplicationManager.getApplication().isUnitTestMode()) {
       if (UsageTrackerManager.getInstance().isTrackingEnabled()) {
         // For the semantics of each parameter, consult the followings:
@@ -154,16 +152,14 @@ public class GoogleUsageTracker implements UsageTracker, SendsEvents {
         // I think 'virtual' indicates these don't correspond to real web pages.
         postData.add(new BasicNameValuePair(IS_VIRTUAL_KEY, STRING_TRUE_VALUE));
         String fullMetadataString = systemMetadataKeyValues;
-        if (eventLabel != null) {
-          // Not using ImmutableMap.of() because I want to handle null for values.
-          Map<String, Integer> eventMetadataMap = Maps.newHashMap();
-          eventMetadataMap.put(METADATA_ESCAPER.escape(eventLabel), eventValue);
-          fullMetadataString = fullMetadataString + "," + METADATA_JOINER.join(eventMetadataMap);
-        }
-        if (eventMessage != null) {
-          Map<String, String> eventMessageMap =
-              ImmutableMap.of("Message", METADATA_ESCAPER.escape(eventMessage));
-          fullMetadataString = fullMetadataString + "," + METADATA_JOINER.join(eventMessageMap);
+        if (metadataMap != null && !metadataMap.isEmpty()) {
+          Map<String, String> escapedMap = metadataMap.entrySet()
+              .stream()
+              .collect(Collectors.toMap(
+                  key -> METADATA_ESCAPER.escape(key.getKey()),
+                  val -> METADATA_ESCAPER.escape(val.getValue())));
+
+          fullMetadataString = fullMetadataString + "," + METADATA_JOINER.join(escapedMap);
         }
         postData.add(new BasicNameValuePair(PAGE_TITLE_KEY, fullMetadataString));
         sendPing(postData);
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
@@ -156,8 +156,8 @@ public class GoogleUsageTracker implements UsageTracker, SendsEvents {
           Map<String, String> escapedMap = metadataMap.entrySet()
               .stream()
               .collect(Collectors.toMap(
-                  key -> METADATA_ESCAPER.escape(key.getKey()),
-                  val -> METADATA_ESCAPER.escape(val.getValue())));
+                  entry -> METADATA_ESCAPER.escape(entry.getKey()),
+                  entry -> METADATA_ESCAPER.escape(entry.getValue())));
 
           fullMetadataString = fullMetadataString + "," + METADATA_JOINER.join(escapedMap);
         }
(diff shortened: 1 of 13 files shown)


aslo:
LGTM, as far as I can tell. However I'd also want to hear from @chanseokoh before merging.

chanseokoh:
To be correct, the error string should be escaped. Escaping isn't difficult; `GoogleUsageTracker` already has `METADATA_ESCAPER`. But I think we really want to extend the metrics reporting interface for this additional info, like what I said [here](https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1124#issuecomment-265496949).

etanshaul:
Yes good idea. I didn't realize we could send multiple key value pairs. I'll add this in.

chanseokoh:
Yeah, FYI, #1154 uses this format to send platform info.

etanshaul:
updated, PTAL.

I ended up getting rid of the `FluentTrackingEventWithValue` interface and instead creating an overloaded `withLabel` method that also accepts a scalar value. My reasoning is that a value is strongly associated with a label since we are sending these bundled as a key/value pair. Also this makes the interface more fluent in that you can continue setting metadata.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
```diff
@@ -159,14 +160,19 @@ public void sendEvent(
           eventMetadataMap.put(METADATA_ESCAPER.escape(eventLabel), eventValue);
           fullMetadataString = fullMetadataString + "," + METADATA_JOINER.join(eventMetadataMap);
         }
+        if (eventMessage != null) {
+          Map<String, String> eventMessageMap =
+              ImmutableMap.of("Message", METADATA_ESCAPER.escape(eventMessage));
```
etanshaul: @chanseokoh what do you think of this key value mapping? I'm just hardcoding the key to "Message" here.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/TrackingEventBuilder.java
```diff
@@ -39,19 +38,26 @@
...
   @Override
-  public FluentTrackingEventWithValue withLabel(String label) {
+  public FluentTrackingEventWithMetadata withLabel(String label) {
     this.label = Preconditions.checkNotNull(label);
     return this;
   }
 
   @Override
-  public PingsAnalytics setValue(Integer value) {
+  public FluentTrackingEventWithMetadata withLabel(String label, int value) {
```
chanseokoh: How about changing the `int` type of `value` to `String` instead of adding a new method `withMessage()`? That is, the metadata pair will be added as `<label>=<value>` whenever this method is called. You can then call `withLabel("message", errorMessage)`.

chanseokoh: And for this to work, you will store a list of label-value pairs so that clients can call `withLabel()` multiple times.

patflynn:
The current API is based off of the Android Studio model which was using an 'application' GA property type where - I think - the event label dimension was explicitly for scalar values.   

Since we've transition to concord that's no longer the case, and we now hack all our extra dimensions into this page title mess which clearly isn't going to do any special processing on the scalar values. So I tend to think that Chanseok's suggestion makes more sense of changing it to just accept string type is the better alternative.

patflynn:
Also if we want to improve the API to allow setting multiple key values I'd change the labelling to something more like .addMetadata(key, value), and get rid of with label all together.

chanseokoh:
> something more like .addMetadata(key, value)

I like it.

etanshaul:
Sounds good. I'll update this.

etanshaul:
After going down this path I see a complication with this suggested approach:

A key/value API doesn't really make much sense for clients wishing to set a label or a message. Most of the time they'd end up doing something like this:

```
.withMetadata(null, label)
.withMetadata(null, message)
```

I'm almost more inclined to leave it similar to how I have it now:
```
.withLabel(label)
.withMessage(message)
```
and internally set the key to either "label" or "message". Also getting rid of that scalar value altogether (since its not used anywhere) or moving it to a separate method (e.g. setValue(value)).

Thoughts?

etanshaul:
I think I'll settle for something like this:

```
.withMetadata(GctTracking.METADATA_LABEL_KEY, eventLabel)
.withMetadata(GctTracking.METADATA_MESSAGE_KEY, errorMessage)
```

patflynn:
I think the point here is that event label doesn't have any fixed meaning
any more in our reporting. So whether a caller decides to track a specific
key value or another, we're not going to try and assign any particular
category to it, since it all gets jammed into page title anyway.

so the actual calls should look like:

.addMetadata(someArbitraryLabel, labelValue)
.addMetadata(someOtherArbitraryLabe, otherLabelValue)

etanshaul:
@patflynn @chanseokoh updated. PTAL.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/TrackingEventBuilder.java
```diff
@@ -39,19 +39,13 @@
...
   @Override
-  public FluentTrackingEventWithValue withLabel(String label) {
-    this.label = Preconditions.checkNotNull(label);
-    return this;
-  }
-
-  @Override
-  public PingsAnalytics setValue(Integer value) {
-    this.value = Preconditions.checkNotNull(value);
+  public FluentTrackingEventWithMetadata withMetadata(String key, String value) {
```
patflynn: not 'with'Metadata. It should be clear that it appends.  I'd rename to addMetadata.

etanshaul:
good point. updated to `addMetadata`

google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
```diff
@@ -153,11 +152,14 @@ public void sendEvent(
...
         String fullMetadataString = systemMetadataKeyValues;
-        if (eventLabel != null) {
-          // Not using ImmutableMap.of() because I want to handle null for values.
-          Map<String, Integer> eventMetadataMap = Maps.newHashMap();
-          eventMetadataMap.put(METADATA_ESCAPER.escape(eventLabel), eventValue);
-          fullMetadataString = fullMetadataString + "," + METADATA_JOINER.join(eventMetadataMap);
+        if (metadataMap != null && !metadataMap.isEmpty()) {
+          Map<String, String> escapedMap = metadataMap.entrySet()
+              .stream()
+              .collect(Collectors.toMap(
```
patflynn: I'm a total stream noobs, but I would have expected stream().map() here.

etanshaul: with map() we'd still need to collect() the values so I believe this is more concise.

chanseokoh:
I wonder if the `errorMessage` may contain sensitive info.

chanseokoh:
Generally the new API looks good. I'll look into it a bit more.

About the actual format you'd want to use when providing metadata, it looks good too, but I just wanted to mention an alternative. For example, I see you would enforce using the two general keys "label" and "message" for all kinds of your event pings such as

"label=flex-compat,message=..."
"label=<runner ID>"

but you could also let the callers decide keys and values, e.g.,

"deployType=flex-compat,errorMessage=..."
"runnerId=<runner ID>"

I'll leave the design up to you.

etanshaul:
Yes, the caller can put in whatever they want for the key. I just created
those constants for convenience.

chanseokoh:
@etanshaul Great, I understood you intended it too.

etanshaul:
we control the error message. they are pre constructed strings from our
message bundle.

google-account-plugin/src/com/google/cloud/tools/intellij/stats/UsageTracker.java
```diff
@@ -31,7 +31,7 @@
    *               often prefixed with a domain such as 'appengine.' or 'clouddebugger.'
```
chanseokoh: An example call in this Javadoc needs to be updated.

patflynn:
@etanshaul Ideally the error message would not be a localized string.  I also agree that now that we don't need to use fixed key types it's probably more useful to use the key name to be more precise about the category of the value. 

Both things I'm fine fixing later.

patflynn:
I don't see the approve button for some odd reason.

etanshaul:
@patflynn I think you can approve if you go into the code tab and click 'review changes': https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1182/files

PR #1184, opened 2017-01-09
Multiple Cloud Source Repo support

etanshaul:
**NOT yet ready for review.** Still pending cleanup/refactoring and testing; in the meantime you can see the screenshots below.

This PR add support in IJ for the GCP support of multiple cloud repositories. Any reference or reliance on a 'default' repository is also removed.

Upload to Cloud Repo dialog:
image

Checkout from Cloud Repo dialog:
image

Selector widget from the upload dialog ("create new repo" button is available):
image

Selector widget from the checkout dialog ("create new repo" button _not_ available):
image

Selector widget in loading state:
image

Selector widget in error state:
image

Selector widget in empty state (no repos found):
image


etanshaul:
Thanks Rui for reviewing this. You can post feedback directly here on GitHub.

@patflynn - the 'create repo' button currently does not open a dialog. We have no API call for this yet - once the new Source API is release I plan up updating this. Instead, it opens up a browser window. Unfortunately the best I could do is open up the 'developer' page that lists all the repo's and has a "new repo" button at the top. This is because the web console's new repo button does not have a permanently linkable url (it just uses some javascript).

Regarding the spacing - I agree, I'll improve this.

etanshaul:
Rui, to install this build to check it out (but keep in mind things may change in the process of review) you can follow this steps:

1. Clone this repository
2. Switch to branch 'repository-selector'
3. Run IDEA with the plugin:
  `./gradlew clean buildPlugin runIdea`
4. Go to tools->Google Cloud Tools -> Upload to a Google Cloud Source Repository
    and
    VCS -> Checkout from Version Control -> Google Cloud

etanshaul:
@patflynn what are your thoughts on allowing a single GCP project to be uploaded to more than one Cloud Source Repo?

Context:

Previously, when there was only 1 'default' Cloud repo, IJ ran a check to see if there already existed a CSR remote. If the remote was there, then a dialog would come up:

screen shot 2017-01-10 at 4 17 29 pm

Now that we support multiple CSR repos, we have 2 options:

1) keep the logic from before and only allow a given project to be associated with a single CSR remote
2) allow multiple remotes and update the check to occur after selection of the CSR repo - after selection it would check to see if the remote exists. We'd have to update the remote naming scheme - probably to make the remote name equal to the name of the repository ID.

Personally I keep going back and forth on these. On the one hand, since GCP now supports multiple repos, we could just allow a single project to be pushed to multiple repos. Other the other hand, when were talking about a single project, the same thing can (and usually is) achieved with branches.

patflynn:
What does the generic git integration for IJ allow?

etanshaul:
If you have multiple remotes then the generic git integration allows you to select which remote to push to. I guess this answers my question.

Good point.

etanshaul:

Other then the enhancement to allow uploading to multiple CSR repos (as discussed above), this is ready for initial review.

etanshaul:
I've added support for multiple CSR remotes: https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1184/commits/b07b8a1324afbf11ae020d675b55d0407d97e951

Say I have a GCP project with the following repositories:

image

I can then upload my project to say 'test-repo-one', and locally, a git config with the following remote will be created:

image

I can then upload the same project to 'test-repo-2' as well, and 2 remotes will be present:
image

If I then try to upload the project again to 'test-repo-1' I'll get the following dialog:
image

Finally, if I manually update the remote url for 'test-repo-2' to some non GCP url to simulate a name collision and then try to upload to 'test-repo-2', I'll get the following dialog:
image

hray2003:
In "selector error state" and "selector empty state", what can user do to in order to move forward?

etanshaul:
@hray2003 - in both those cases, the user can either refresh, select a different project, or create a new repository. Perhaps the wording for the 'empty' state can be expanded to indicate that the user can create a new repository.

patflynn:
I don't fully understand the second case.

etanshaul:
Which second case? The 'empty state'? This is just when a selected cloud project has no CSR repos associated with it.

joaoandremartins:
1) Can we rename "Repository Selector:" to just "Repository:"?

2) Maybe I'm missing something, but why are we prepending "cloud-platform-" to the remote name? To make sure it doesn't collide with any previously existing remote with the same name? To mark it as a GCR remote? (and then we'd probably run into problems if there were already other remotes named "cloud-platform-*", unlikely as that is)

3) Finally, I also don't understand the second case (where you manually modify the remote URL so it's not a GCR URL). What are you accomplishing there? Validating/enforcing that a GCR remote has a valid URL? Why does renaming the URL result in a name collision? Why are "upload to a GCR remote" and "add a GCR remote" related?

Also, the title of the window needs to be changed from "Google" to something else like "Google Cloud Source Repositories" and the text on the second window could also use some improvements, e.g., s/GCP remote/Google Cloud Source Repository remote. (and why are we telling we couldn't add a remote when the user is trying to push to an already added remote?)

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -161,8 +161,9 @@ uploadtogcp.text=Upload to a Google Cloud Source Repository...
 uploadtogcp.description=Easily use Google as a source repository.
 uploadtogcp.giterror=Error Running Git
 uploadtogcp.git.unsupported.message=<html><tt>{0}</tt><br>This version is unsupported, and some plugin functionality could fail to work.<br>The minimal supported version is <em>{1}</em>.</html>
-uploadtogcp.alreadyexists=Project is already on Google Cloud Platform
-uploadtogcp.selecttext=Select a Google Cloud Project
+uploadtogcp.alreadyexists=Project is already uploaded to the selected Google Cloud Source Repository ''{0}''.
+uploadtogcp.nongcp.remotename.collision=Failed to add GCP remote because of an existing non GCP remote name collision.\nTo fix, remove or rename the remote named ''{0}''.
+uploadtogcp.selecttext=Upload to Google Cloud Repositories
```
joaoandremartins: s/Google Cloud Repositories/Google Cloud Source Repository

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectRepositoriesModelItem.java
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * TreeNode representation of the set of available Cloud Source Repositories for a given GCP
+ * project.
+ */
+public class ProjectRepositoriesModelItem extends DefaultMutableTreeNode {
+
+  private static final String CLOUD_SOURCE_API_ROOT_URL = "https://source.googleapis.com/";
+  private static final String CLOUD_SOURCE_API_LIST_URL = "v1/projects/{projectId}/repos";
+  static final String PANETHEON_CREATE_REPO_URL_PREFIX
```
joaoandremartins: Not being used anywhere else in this PR. Set to private?

etanshaul:
@joaoandremartins 

1. The "Repository Selector:" wording was a mistake and has since been updated (the screenshot is old). It now said "Cloud Repository".

2. You are right. I am prepending "cloud-platform-" to avoid naming collisions. If there happen to be collisions, then we show the following dialog:
image

3. This is just to similate the naming collision I just mentioned in (2). I specifically set it to a non GCP url because it if was a GCP url, then the user would see this dialog instead:
image
I'm not sure what you mean about "upload/add a GCR remote". Maybe lets talk more about this part offline.

I agree about your points about improving wording.

> and why are we telling we couldn't add a remote when the user is trying to push to an already added remote?

This is because our "Upload to GCP" functionality is _not_ for pushing to a GCP remote. Its for the initial project upload and for automatically creating the local git remote. Subsequently to push to the cloud repository the user is expected to use basic GIT functionality (either in IJ or otherwise).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectRepositoriesModelItem.java
```diff
@@ -0,0 +1,147 @@
...
+//      Thread.sleep(5000);
+//    } catch (InterruptedException e) {
+//      e.printStackTrace();
+//    }
+
+    try {
+      Credential credential = user.getCredential();
+      HttpRequestInitializer initializer = httpRequest -> {
+        HttpHeaders headers = new HttpHeaders();
+        httpRequest.setConnectTimeout(5000);
```
joaoandremartins: Parameterize these timeout values in config.properties?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectRepositoriesModelItem.java
```diff
@@ -0,0 +1,147 @@
...
+          // this ends up prefixed to user agent
+          .setApplicationName(userAgent)
+          .build();
+
+      ListReposResponse response = new CustomUrlSourceRequest(source, cloudProject).execute();
+
+      removeAllChildren();
+      List<Repo> repositories = response.getRepos();
+      if (!response.isEmpty() && repositories != null) {
+        for (Repo repo : repositories) {
```
joaoandremartins: Possibly looks better using Streams API?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositoryItem.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.intellij.ui.components.JBLabel;
+
+import javax.swing.BorderFactory;
+import javax.swing.SwingConstants;
+
+/**
+ * UI for node in the repository selector representing a single GCP repository.
+ */
+public class RepositoryItem extends JBLabel {
```
joaoandremartins: We're only using this class once. Should we inline all these settings where we declare and initialize its instance (RepositorySelectorRenderer)?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -0,0 +1,256 @@
...
+  private static final int SELECTOR_WIDTH = 400;
+
+
+  public RepositorySelector(@Nullable String cloudProject,
+      @Nullable CredentialedUser user, boolean canCreateRepository) {
+    this.cloudProject = cloudProject;
+    this.user = user;
+    this.canCreateRepository = canCreateRepository;
+
+    getTextField().getEmptyText().setText("Enter a cloud repository ID...");
```
joaoandremartins: Send text to resource bundle?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  @Nullable
+  public String getSelectedRepository() {
+    if (StringUtil.isEmpty(getText())) {
+      return null;
+    }
+
+    Enumeration repos = panel.getRepositories().children();
```
joaoandremartins: Is null check needed for getRepositories() return?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -0,0 +1,256 @@
...
+    while (repos.hasMoreElements()) {
```
joaoandremartins: If Streams API can be used with repos, it would greatly improve readability here.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  @Override
+  public boolean isPopupVisible() {
+    return popup != null && !popup.isDisposed() && popup.isVisible();
+  }
+
+  @Override
+  protected CustomizableComboBoxPopup getPopup() {
+    return this;
```
joaoandremartins: Do you want to return this or popup here?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  @Override
+  protected int getPreferredPopupHeight() {
+    return SELECTOR_HEIGHT;
+  }
+
+  private class RepositoryPanel extends JPanel {
```
joaoandremartins: Wouldn't it be better to create a .form file for this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelectorRenderer.java
```diff
@@ -0,0 +1,65 @@
...
+    errorItem = new ResourceSelectorErrorItem(JBColor.RED);
+  }
+
+  @Override
+  public Component getTreeCellRendererComponent(JTree tree, Object value, boolean selected,
+      boolean expanded, boolean leaf, int row, boolean hasFocus) {
+    if (value instanceof RepositoryModelItem) {
+      repositoryItem.setText(((RepositoryModelItem) value).getRepositoryId());
+      return repositoryItem;
+    } else if (value instanceof ResourceLoadingModelItem) {
```
joaoandremartins: Remove the elses for better readability?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ResourceLoadingModelItem.java
```diff
@@ -16,12 +16,46 @@
 
 package com.google.cloud.tools.intellij.resources;
 
+import com.sun.istack.NotNull;
```
joaoandremartins: Wrong NotNull?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ResourceLoadingModelItem.java
```diff
@@ -16,12 +16,46 @@
...
+ * This model item is shown when a resource manager call is outstanding.
  */
 class ResourceLoadingModelItem extends DefaultMutableTreeNode {
 
 }
 
+/**
+ * This model item is shown when a resource manager call returns an empty result.
+ */
+class ResourceEmptyModelItem extends DefaultMutableTreeNode {
```
joaoandremartins: package-protected visibility on purpose?

joaoandremartins: No custom behaviour seems to be tied to ResourceEmptyModelItem. Why not just use DefaultMutableTreeNode?

patflynn:
I was just talking with Etan about this offline and we're now thinking a
workflow might be more flexible with the following strategy:

- when 'uploading to GCP' always ask the user what they'd like to call the
remote we create.
 - For the first remote default it to 'origin'. (this is what the JB github
integration does)
 - For subsequent remotes possibly concatenate project name and repo name.
- don't try to map remote names to CSR repos.  Allow the user to create
multiple remotes pointing to the same uri if they want.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ResourceLoadingModelItem.java
```diff
@@ -16,12 +16,46 @@
...
+  public String getMessage() {
+    return message;
+  }
+}
+
+/**
+ * This model item represents a node shown in when an error occurs attempting to query a resource
+ * manager.
+ */
+class ResourceErrorModelItem extends DefaultMutableTreeNode {
```
joaoandremartins: Ditto as above.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ResourceSelectorEmptyItem.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.awt.Font;
+
+import javax.swing.BorderFactory;
+import javax.swing.SwingConstants;
+
+/**
+ * UI for the node that displays information if a resource manager call returns an empty result.
+ */
+public class ResourceSelectorEmptyItem extends JBLabel {
```
joaoandremartins: Also only used once in this PR. Inline?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ResourceSelectorErrorItem.java
```diff
@@ -29,9 +29,9 @@
 /**
  * UI for the node that displays error information if a resource manager call fails.
  */
-class ProjectSelectorErrorItem extends JBLabel {
+public class ResourceSelectorErrorItem extends JBLabel {
```
joaoandremartins: Inline?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneGcpDialog.form
```diff
@@ -51,20 +51,34 @@
...
           </grid>
         </constraints>
         <properties/>
       </component>
+      <component id="1645e" class="javax.swing.JLabel">
+        <constraints>
+          <grid row="1" column="0" row-span="1" col-span="1" vsize-policy="0" hsize-policy="0" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
+        </constraints>
+        <properties>
+          <text value="Cloud Repository:"/>
```
joaoandremartins: Rename "Cloud Repository" and "Cloud Project" to "Repository" and "Cloud" in this menu?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneGcpDialog.java
```diff
@@ -106,21 +108,21 @@ private void initComponents() {
...
-          if (text.length() == 0) {
-            VirtualFile file = project.getBaseDir();
-            if (file != null) {
-              return file;
-            }
-          }
-          return super.getInitialFile();
-        }
-      }
+                                          .BrowseFolderActionListener<JTextField>(
```
joaoandremartins: Spacing doesn't seem right according to style guide. Did google-java-format do this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneGcpDialog.java
```diff
@@ -106,21 +108,21 @@ private void initComponents() {
...
+                                          fcd.getTitle(), fcd.getDescription(), parentDirectory,
+                                          project, fcd, TextComponentAccessor.TEXT_FIELD_WHOLE_TEXT) {
```
joaoandremartins: Longer than 100 lines.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneGcpDialog.java
```diff
@@ -150,7 +152,22 @@ protected String getHelpId() {
    * Check fields and display error in the wrapper if there is a problem.
    */
   private void updateButtons() {
-    if (parentDirectory.getText().length() == 0 || directoryName.getText().length() == 0) {
+    if (!StringUtil.isEmpty(projectSelector.getText())
+        && projectSelector.getSelectedUser() == null) {
+      setErrorText("Invalid Cloud Project selected.");
+      setOKActionEnabled(false);
+      return;
+    } else if (!StringUtil.isEmpty(repositorySelector.getText())
```
joaoandremartins: Remove elses for readability?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneGcpDialog.java
```diff
@@ -150,7 +152,22 @@ protected String getHelpId() {
...
+        && StringUtil.isEmpty(repositorySelector.getSelectedRepository())) {
+      setErrorText("Invalid Cloud Repository selected.");
+      setOKActionEnabled(false);
+      return;
+    } else if(projectSelector.getSelectedUser() == null
+        || StringUtil.isEmpty(repositorySelector.getSelectedRepository())) {
+      setErrorText(null);
+      setOKActionEnabled(false);
+      return;
+    } else if (parentDirectory.getText().length() == 0 || directoryName.getText().length() == 0) {
```
joaoandremartins: Could parentDirectory().getText() ever return null?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneGcpDialog.java
```diff
@@ -167,34 +184,37 @@ private void updateButtons() {
...
   }
 
   private void createUIComponents() {
-    repositoryUrl = new ProjectSelector();
-    repositoryUrl.setMinimumSize(new Dimension(300, 0));
-    repositoryUrl.getDocument().addDocumentListener(new DocumentAdapter() {
+    projectSelector = new ProjectSelector();
+    projectSelector.setMinimumSize(new Dimension(300, 0));
+    projectSelector.getDocument().addDocumentListener(new DocumentAdapter() {
       @SuppressWarnings("ConstantConditions") // This suppresses an invalid nullref warning for
```
joaoandremartins: Is this still needed?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneGcpDialog.java
```diff
@@ -204,15 +224,29 @@ protected void textChanged(DocumentEvent event) {
...
+        repositorySelector.setText("");
+        updateButtons();
+      }
+    });
+    repositorySelector = new RepositorySelector(projectSelector.getText(),
+        projectSelector.getSelectedUser(), false /*canCreateRepository*/);
+
+    repositorySelector.getDocument().addDocumentListener(new DocumentAdapter() {
+      @Override
+      protected void textChanged(DocumentEvent e) {
```
joaoandremartins: s/e/event?

google-cloud-tools-plugin/build.gradle
```diff
@@ -54,6 +54,7 @@ dependencies {
         exclude group: 'com.google.guava', module: 'guava'
     }
     compile files('lib/google-api-services-source.jar')
+//    compile files('lib/libsource-v1-java.jar')
```
aslo: Is the idea that this line will be uncommented when the CSR API is ready?

etanshaul: thats not actually the correct API either. but yeah - this will be updated to use the correct API before merging.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/UploadSourceAction.java
```diff
@@ -136,31 +140,34 @@ private static void uploadProjectToGoogleCloud(@NotNull final Project project,
...
     if (!dialog.isOK() || dialog.getCredentialedUser() == null || Strings
         .isNullOrEmpty(dialog.getProjectId())) {
       return;
     }
 
+    Optional<GitRemote> remote = gitDetected
+        ? findRemote(gitRepository.getRemotes(), dialog.getRepositoryId())
+        : Optional.empty();
+    if (remote.isPresent() && !GcpHttpAuthDataProvider.hasGcpUrl(remote.get().getUrls())) {
+      Messages.showErrorDialog(project,
```
joaoandremartins: Bundleize "Google".

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/UploadSourceAction.java
```diff
@@ -136,31 +140,34 @@ private static void uploadProjectToGoogleCloud(@NotNull final Project project,
...
+    UploadSourceDialog dialog =
+        new UploadSourceDialog(project, GctBundle.message("uploadtogcp.selecttext"),
             GctBundle.message("uploadtogcp.oktext"));
     DialogManager.show(dialog);
     if (!dialog.isOK() || dialog.getCredentialedUser() == null || Strings
         .isNullOrEmpty(dialog.getProjectId())) {
       return;
     }
 
+    Optional<GitRemote> remote = gitDetected
```
joaoandremartins: If !gitDetected, can you return right away instead of having this assignment?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/UploadSourceAction.java
```diff
@@ -136,31 +140,34 @@ private static void uploadProjectToGoogleCloud(@NotNull final Project project,
...
+        ? findRemote(gitRepository.getRemotes(), dialog.getRepositoryId())
+        : Optional.empty();
+    if (remote.isPresent() && !GcpHttpAuthDataProvider.hasGcpUrl(remote.get().getUrls())) {
```
joaoandremartins: Have a single top level if (remote.isPresent()) ?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/UploadSourceAction.java
```diff
@@ -176,20 +183,17 @@ public void run(@NotNull ProgressIndicator indicator) {
...
-          SwingUtilities.invokeLater(new Runnable() {
-            @Override
-            public void run() {
-              Messages.showErrorDialog(project, GctBundle.message("uploadtogcp.failedtocreategit"),
-                  "Google");
-            }
-          });
+          SwingUtilities.invokeLater(() ->
+              Messages.showErrorDialog(project,
+                  GctBundle.message("uploadtogcp.failedtocreategit"), "Google"));
```
joaoandremartins: Bundleize Google and add something else, i.e., Google Cloud Source Repositories?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/UploadSourceDialog.java
```diff
@@ -73,18 +82,57 @@ public CredentialedUser getCredentialedUser() {
...
+        repositorySelector.setCloudProject(projectSelector.getText());
+        repositorySelector.setUser(projectSelector.getSelectedUser());
+        repositorySelector.setText("");
+        updateButtons();
       }
     });
+
+    repositorySelector.getDocument().addDocumentListener(new DocumentAdapter() {
+      @Override
+      protected void textChanged(DocumentEvent e) {
```
joaoandremartins: s/e/event?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/UploadSourceDialog.java
```diff
@@ -73,18 +82,57 @@ public CredentialedUser getCredentialedUser() {
...
+        updateButtons();
+      }
+    });
+
+    // todo figure out how to disable it when no project is selected
+//    repositorySelector.setEnabled(false);
+  }
+
+  private void updateButtons() {
```
joaoandremartins: Remove elses?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/UploadSourceDialog.java
```diff
@@ -73,18 +82,57 @@ public CredentialedUser getCredentialedUser() {
...
+    if (!StringUtil.isEmpty(projectSelector.getText())
+        && projectSelector.getSelectedUser() == null) {
+      setErrorText("Invalid Cloud Project selected.");
```
joaoandremartins: Bundleize text?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectRepositoriesModelItem.java
```diff
@@ -0,0 +1,147 @@
...
+    setUserObject(cloudProject);
+  }
+
+  public void loadRepositories() {
+    loadRepositories(false);
+  }
+
+  public void loadRepositories(boolean empty) {
+
+    // TODO remove
```
aslo: Don't forget to RM

etanshaul: Yea I should have mentioned prior to asking for review - I purposefully left in a couple of TODO's (will remove soon) and I haven't bundleized any messaged yet.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/UploadSourceDialog.java
```diff
@@ -73,18 +82,57 @@ public CredentialedUser getCredentialedUser() {
...
+        && StringUtil.isEmpty(repositorySelector.getSelectedRepository())) {
+      setErrorText("Invalid Cloud Repository selected.");
+      setOKActionEnabled(false);
+      return;
+    } else if(projectSelector.getSelectedUser() == null
+        || StringUtil.isEmpty(repositorySelector.getSelectedRepository())) {
+      setErrorText(null);
+      setOKActionEnabled(false);
+      return;
+    }
```
joaoandremartins: Is this the same updateButtons() as CloneGcpDialog? Can anything be done about that?

etanshaul: Part of it is the same. I think I can collapse the validation into the selector classes so the code can be reused.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneGcpDialog.form
etanshaul: I like the "Cloud" prefix here to disambiguate between the gcp project/repo and the local project/repo. Without it, it may be confusing.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ResourceLoadingModelItem.java
etanshaul: It is adding a message String state to the class. I'll see if there's a way to refactor this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ResourceLoadingModelItem.java
etanshaul: nice catch

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
etanshaul: Sadly it can't since the API forces Enumerations.

joaoandremartins: We could use Collections.list() to transform the enumeration into a collection. Would the extra pass be worth simpler code? I'm leaning to think yes, but I don't feel strongly about this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ResourceLoadingModelItem.java
joaoandremartins: You could use new DefaultMutableTreeNode(Object object)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneGcpDialog.java
```diff
@@ -52,22 +54,22 @@
  * The dialog that prompts the user to download (git clone) from a GCP project.
  */
 public class CloneGcpDialog extends DialogWrapper {
+
   private static final Logger LOG = Logger.getInstance(CloneGcpDialog.class);
   public static final String INVALID_FILENAME_CHARS = "[/\\\\?%*:|\"<>]";
```
joaoandremartins: This variable doesn't seem to be used anywhere else other than this class.
LOG is also showing as unused

etanshaul:
@patflynn I've updated the experience to reflect our discussion:

There is now always a "remote name" selector:
image

The remote is auto-populated with a suggestion:
- "origin" if there is no remote named "origin" already
- "google-cloud-platform-{repository_name}" if there is already a remote named "origin"

If the user forces a conflict (by manually typing in a remote name that already exists) the following will be shown:
image

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectRepositoriesModelItem.java
etanshaul: Its being used by the RepositorySelector class

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositoryItem.java
etanshaul: I think I will leave them in separate class as these can (and some of them are) be reused by multiple renders.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
etanshaul: It needs a CustomizableComboBoxPopup, so "this" is correct.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ResourceLoadingModelItem.java
etanshaul: yes

etanshaul: The bigger point is that if you look at RepositorySelectorRenderer#getTreeCellRendererComponent its checking the type of DefaultMutableTreeNode so it can then decide which UX to show.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectRepositoriesModelItem.java
joaoandremartins: It's the only place it's being used at. Why not moving it over there?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
joaoandremartins: I wonder what is the use of a method that returns the reference of the object it's being called on?

etanshaul: yeah its a strange remnant of legacy code. I'll look to see what can be done

etanshaul:

What are your thoughts on the wording of this dialog, specifically the "Remote name" section:
image

I was going through the process of testing out the flow as a normal user and realized it may not be totally clear that this dialog will generate a new remote for the user. I'm concerned that a user may confuse this with us requiring an existing remote to exist and that its up to the user to type in the matching name.

Here are a few options:

- Change the label from "Remote name" to "New remote name"
- Have placeholder text that says something like "enter a remote name to be created for this repository"
- Have some labeling for the dialog explaining that this creates a new git repo (if one doesn't exist), and a new remote

joaoandremartins:
Does it have to be a new remote? If it doesn't, then I'd be OK with not extra wording -- if the remote doesn't exist yet, we create it for the user.

etanshaul:
Yes, it does have to be a new remote. Otherwise its an error.

joaoandremartins:
I think "New remote name" looks good. The only problem is if the user doesn't understand why it needs to be a new repo (e.g., if they haven't checked out the remote repo but want to push there right away. not sure how often/if this would happen at all?). Then we could need some text.

etanshaul:
> e.g., if they haven't checked out the remote repo but want to push there right away. not sure how often/if this would happen at all?

I'd like to understand this scenario. Not having checked out the remote repo is the majority of use cases. This feature is precisely for "intializing" the remote repo by pushing the project to it and establishing the local remote (sounds like an oxymoron).

joaoandremartins:
I think the case you're referring to is when the user doesn't _already_ have a remote repo with that code, and is starting fresh.

If the code is already in the remote, but the remote somehow isn't synced locally (does this happen at all?), the user might want to use that "Upload to GCSR" dialog to upload it there. Though, thinking about it, I don't think this happens frequently?

Also, what's the story when you check out the remote with git or gcloud? Will we still require the users to create a new remote?

etanshaul:
> If the code is already in the remote, but the remote somehow isn't synced locally (does this happen at all?), the user might want to use that "Upload to GCSR" dialog to upload it there. Though, thinking about it, I don't think this happens frequently?

This dialog will always create a new remote. Whether or not we allow uploading to a cloud repo that already has code is up for debate: previously it would _not_ allow this. After talking to Patrick we decided that it would (a new remote would still be created pointing to the same cloud url). However, I am going back on this now because there are cases where this could fail: e.g. non-fast-foward changes, which is tough to detect before hand.

> Also, what's the story when you check out the remote with git or gcloud? Will we still require the users to create a new remote?

Yes, if they use this dialog in this case (and they shouldn't really) a new remote will be forced.

If I had to summarize what this dialog does it would be as follows:

The dialog creates a new local git repository if one does not yet exist, creates a new remote pointing to the Google Cloud Repository, and finally pushes the code to this remote.

joaoandremartins:
I wonder if it would make sense to add a new issue to iron that bit out, i.e., if the user already has a remote with the code that isn't mirrored in a way our IJ plugin can't use it out of the bat?

In any case, since that dialog will always create a new remote and users using this feature who don't have a correctly set up remote, I think we can do without the extra text? Patrick or Rui might have a different take on that, though.

patflynn:
I'm kind of OK with either.

Could hover text or some extra wording above the fields clarify things?

etanshaul:
Here is what I came up with, with help from @joaoandremartins:

Updated action text (to make it clear that this is for "Setup" and not for normal "uploading / pushing"):
image

Updated remote name label to make it clear that its creating a new one:
image

etanshaul:
I've updated the error messages to be notification popups (instead of an error dialog) and to show the contents of the git error output so the user can debug the problem.

Here is an example of the user attempting to use the dialog to a project to an existing git repo:

image

And after expanding, you can see the full message in a scrollable pane:
image

etanshaul:

This is ready for review. All functionality has been implemented. Only outstanding task is test coverage.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -0,0 +1,292 @@
...
+    this.user = user;
+    this.canCreateRepository = canCreateRepository;
+
+    getTextField().getEmptyText()
+        .setText(GctBundle.message("cloud.repository.selector.placeholder.text"));
+  }
+
+  @Nullable
+  public String getSelectedRepository() {
+    if (StringUtil.isEmpty(getText())) {
```
joaoandremartins: What does this condition assure? Seems like we could do without it, unless the callers treat nulls and empty strings differently. Or if the assumption is being made that if the text is empty, getRepositories()/children() returns null or an empty collection, in which case, the condition should probably target those returns.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -0,0 +1,292 @@
...
+      return null;
+    }
+
+    Enumeration repos = ((RepositoryPanel) panel).getRepositories().children();
```
joaoandremartins: Is a null check for getRepositories() needed?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -0,0 +1,292 @@
...
+
+  private class RepositoryPanel extends JPanel {
+
+    private JTree repositoryTree;
+    private DefaultMutableTreeNode projectRootNode;
+    private DefaultTreeModel treeModel;
+    private ProjectRepositoriesModelItem repositories;
+
+    RepositoryPanel(String cloudProject) {
+      repositories
```
joaoandremartins: Is the new line needed here?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -0,0 +1,292 @@
...
+          = new ProjectRepositoriesModelItem(cloudProject, user);
+      projectRootNode = new DefaultMutableTreeNode("root");
```
joaoandremartins: Is "root" a special tree node?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -0,0 +1,292 @@
...
+      repositoryTree = new JTree(treeModel);
+      repositoryTree.setRowHeight(0);
+      repositoryTree.setRootVisible(false);
+      repositoryTree.setOpaque(false);
+      repositoryTree.getSelectionModel().setSelectionMode(TreeSelectionModel.SINGLE_TREE_SELECTION);
+      repositoryTree.expandRow(0);
+
+      RepositorySelectorRenderer renderer = new RepositorySelectorRenderer();
+      renderer.setOpenIcon(GoogleCloudToolsIcons.CLOUD);
+      renderer.setClosedIcon(GoogleCloudToolsIcons.CLOUD);
```
joaoandremartins: Do we want to have a different icon here? What does the closed icon represent anyway?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -0,0 +1,292 @@
...
+      renderer.setLeafIcon(null);
+      repositoryTree.setCellRenderer(renderer);
+
+      repositoryTree.addTreeSelectionListener(event -> {
+        DefaultMutableTreeNode node
+            = (DefaultMutableTreeNode) repositoryTree.getLastSelectedPathComponent();
+        if (node != null && node instanceof RepositoryModelItem) {
+          RepositoryModelItem repoNode = (RepositoryModelItem) node;
+          RepositorySelector.this.setText(repoNode.getRepositoryId());
+          ApplicationManager.getApplication().invokeLater(RepositorySelector.this::hidePopup);
```
joaoandremartins: wouldn't this::hidePopup suffice?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -0,0 +1,292 @@
...
+
+      JPanel bottomPane = new JPanel();
+      JPanel buttonPanel = new JPanel();
+
+      bottomPane.setLayout(new BorderLayout());
+      bottomPane.setBorder(BorderFactory.createMatteBorder(1, 0, 0, 0, Color.LIGHT_GRAY));
+      buttonPanel.setLayout(new BoxLayout(buttonPanel, BoxLayout.LINE_AXIS));
+      buttonPanel.setBorder(BorderFactory.createEmptyBorder(5, 5, 5, 5));
+
+      if (canCreateRepository) {
```
joaoandremartins: If you can't create a repository, what happens?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -0,0 +1,292 @@
...
+        ApplicationManager.getApplication().invokeAndWait(() -> {
+          treeModel.insertNodeInto(repositories, projectRootNode, 0);
+          treeModel.reload();
+          repositoryTree.expandRow(0);
+        }, ModalityState.stateForComponent(RepositorySelector.this));
+      });
+    }
+
+    private void setLoader() {
+      ApplicationManager.getApplication().invokeAndWait(() -> {
```
joaoandremartins: Why does this need to be done in a separate thread?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelectorRenderer.java
```diff
@@ -0,0 +1,71 @@
...
+    loadingItem = new ResourceSelectorLoadingItem(defaultRenderer.getBackgroundNonSelectionColor(),
+        defaultRenderer.getTextNonSelectionColor());
+    emptyItem = new ResourceSelectorEmptyItem();
+    errorItem = new ResourceSelectorErrorItem(JBColor.RED);
+  }
+
+  @Override
+  public Component getTreeCellRendererComponent(JTree tree, Object value, boolean selected,
+      boolean expanded, boolean leaf, int row, boolean hasFocus) {
+    if (value instanceof RepositoryModelItem) {
```
joaoandremartins: This still strikes me as a bit weird style of programming. Is there no better alternative to these class checks? And is there a reason we can't simply initialize the *Item's here instead of in the constructor?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ResourceLoadingModelItem.java
```diff
@@ -16,12 +16,46 @@
...
+  public String getMessage() {
+    return message;
+  }
+}
+
+/**
+ * This model item represents a node shown when an error occurs attempting to query a resource
+ * manager.
+ */
+class ResourceErrorModelItem extends DefaultMutableTreeNode {
```
joaoandremartins: nit: Create new file for this class?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneGcpDialog.java
```diff
@@ -150,51 +150,77 @@ protected String getHelpId() {
...
+    }
+
+    if (StringUtil.isEmpty(parentDirectory.getText())
+        || StringUtil.isEmpty(directoryName.getText())) {
       setErrorText(null);
       setOKActionEnabled(false);
       return;
     }
+
     File file = new File(parentDirectory.getText(), directoryName.getText());
```
joaoandremartins: Would it make sense to refactor here to use java.nio? It would come with the annoyance of having to catch InvalidPathExceptions from paths with special characters on Windows...

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneGcpDialog.java
```diff
@@ -150,51 +150,77 @@ protected String getHelpId() {
...
-    CredentialedUser selectedUser = repositoryUrl.getSelectedUser();
+    CredentialedUser selectedUser = projectSelector.getSelectedUser();
 
-    if (selectedUser == null || Strings.isNullOrEmpty(repositoryUrl.getText())) {
+    if (selectedUser == null || Strings.isNullOrEmpty(projectSelector.getText())) {
       return null;
     }
 
-    return GcpHttpAuthDataProvider.getGcpUrl(repositoryUrl.getText());
+    // TODO what if the repo is null??
```
joaoandremartins: Do you want to address this TODO in this PR?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/GcpCheckoutProvider.java
```diff
@@ -183,6 +183,7 @@ private static boolean doClone(@NotNull Project project, @NotNull ProgressIndica
       @NotNull String sourceRepositoryUrl) {
     indicator.setIndeterminate(false);
     GitLineHandlerListener progressListener = GitStandardProgressAnalyzer.createListener(indicator);
+
```
joaoandremartins: heh nice

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/GcpHttpAuthDataProvider.java
```diff
@@ -113,41 +111,18 @@ public void run() {
...
-    for (GitRemote gitRemote : repository.getRemotes()) {
-      for (String remoteUrl : gitRemote.getUrls()) {
-        if (isUrlGcp(remoteUrl)) {
-          return Pair.create(gitRemote, remoteUrl);
-        }
-      }
-    }
-    return null;
+  public static String getGcpUrl(String projectId, String repositoryId) {
+    return "https://source.developers.google.com/p/" + projectId + "/r/" + repositoryId + "/";
```
joaoandremartins: Parameterize this URL somewhere?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/UploadSourceDialog.java
```diff
@@ -0,0 +1,187 @@
...
+import javax.swing.JPanel;
+import javax.swing.event.DocumentEvent;
+
+import git4idea.repo.GitRepository;
+
+/**
+ * Shows the upload to Google Cloud Source Repositories dialog for initializing a project on
+ * a GCP repository.
+ */
+public class UploadSourceDialog extends DialogWrapper {
```
joaoandremartins: Rename to SetupGCSRepositoryDialog?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/UploadSourceDialog.java
```diff
@@ -0,0 +1,187 @@
...
+        repositorySelector.setCloudProject(projectSelector.getText());
+        repositorySelector.setUser(projectSelector.getSelectedUser());
+        repositorySelector.setText("");
+        updateButtons();
+      }
+    });
+
+    repositorySelector.getDocument().addDocumentListener(new DocumentAdapter() {
+      @Override
+      protected void textChanged(DocumentEvent e) {
```
joaoandremartins: s/e/event

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositoryRemotePanel.java
```diff
@@ -0,0 +1,88 @@
...
+   */
+  private String getRemoteNameSuggestion(String cloudRepository) {
+    if (gitRepository != null && hasOriginRemote()) {
+      return CLOUD_SOURCE_REPO_REMOTE_PREFIX + cloudRepository;
+    } else {
+      return "origin";
+    }
+  }
+
+  private boolean hasOriginRemote() {
```
etanshaul: since there wont be any performance improvement (not that its a concern even), I'll leave it since I like tiny methods.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
etanshaul: If getText() is empty, this means that the user hasn't selected a repo from the tree, making all the subsequent checks unnecessary.

etanshaul: if the text is empty it doesn't mean that the repository tree is empty, it just means that the user hasn't made a selection from the tree so checking each node is pointless

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
etanshaul: no, added a `@NotNull` to make it explicit

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
etanshaul: no, the string "root" has no significance, its there just for clarity. I think the tree will behave differently if I were to pass in null.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
etanshaul: no, we are in an inner class here, so I need a reference to the enclosing class

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
etanshaul: then it just omits the "create repository" button. this is as expected - there are 2 dialogs where this widget is used, 1 where we need the button, and 1 where we don't.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
etanshaul: because it modifies the UI

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelectorRenderer.java
etanshaul: yea its super weird, but its because of how the renderers work:

`add`ing to a tree requires you to pass a TreeNode. But using a custom renderer then allows you to "override" this with a custom component for the node (like a JLabel or JPanel). This overriding happen in the renderer - its caught by getTreeCellRendererComponent

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/GcpCheckoutProvider.java
etanshaul: lol...sometimes you just need some space

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
etanshaul: yea, its for the icon that is used when the tree is expanded vs collapsed. For whatever reason theres no way to set both at once.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ResourceLoadingModelItem.java
etanshaul: will leave as is since all these classes are very trivial so I'd prefer to just group them all in the same file (there are 3 now)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectRepositoriesModelItem.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.List;
+
+import javax.swing.tree.DefaultMutableTreeNode;
+
+/**
+ * TreeNode representation of the set of available Cloud Source Repositories for a given GCP
+ * project.
+ */
+public class ProjectRepositoriesModelItem extends DefaultMutableTreeNode {
+
```
joaoandremartins: Nit: extra space?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -0,0 +1,301 @@
...
+    this.cloudProject = cloudProject;
+    this.user = user;
+    this.canCreateRepository = canCreateRepository;
+
+    getTextField().getEmptyText()
+        .setText(GctBundle.message("cloud.repository.selector.placeholder.text"));
+  }
+
+  @Nullable
+  public String getSelectedRepository() {
```
joaoandremartins: How about we change the return type to Optional<> here, to avoid returning null?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/GcpHttpAuthDataProvider.java
joaoandremartins: What do you think about this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/SetupCloudRepositoryDialog.java
```diff
@@ -0,0 +1,187 @@
...
+  private String repositoryId;
+  private String remoteName;
+  private CredentialedUser credentialedUser;
+  private GitRepository gitRepository;
+
+  public SetupCloudRepositoryDialog(@NotNull Project project, @Nullable GitRepository gitReository,
+      @NotNull String title, @NotNull String okText) {
+    super(project, true);
+
+    this.gitRepository = gitReository;
```
joaoandremartins: s/gitReository/gitRepository?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/SetupCloudRepositoryDialog.java
```diff
@@ -0,0 +1,187 @@
...
+    return repositoryId;
+  }
+
+  @NotNull
+  public String getRemoteName() {
+    return remoteName;
+  }
+
+  /**
+   * Return the credentialeduser that owns the ID returned from {@link #getProjectId()}.
```
joaoandremartins: nit add a space to credentialeduser?

google-cloud-tools-plugin/build.gradle
joaoandremartins: Are you planning on removing this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/GcpHttpAuthDataProvider.java
etanshaul: I'll extract the prefix to member variable

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
etanshaul: In theory this is good. But we are still using nullables all over the place. Also if the API itself called from `getSelectedRepository` used Optionals I wouldn't hesitate to just return them, but I don't see much value in wrapping in an Optional now.

etanshaul:
@joaoandremartins see https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1184/commits/d6b25811ee2549176acefefa47f1ff02de68b1a6

I made some non-trivial changes to how we are executing asynchronous code. This is mainly for testability (there was no good way to test the asynchronous code previously without injecting a Thread.sleep into the unit test; now the asynchronous part is easily mocked away), but also for better design (hopefully) and more idiomatic java 8.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
joaoandremartins: If we never stop using nullables, then we're never going to stop using them. :)
We control the API that's calling this method, so we can adapt it to use Optional.
I feel strongly we should start replacing every null return to Optional wherever we see them.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectRepositoriesModelItem.java
```diff
@@ -53,26 +50,36 @@ public ProjectRepositoriesModelItem(@NotNull String cloudProject,
...
-      if (!response.isEmpty() && repositories != null) {
-        repositories.forEach(repo -> {
-          Object name = repo.get("name");
-          if (name != null) {
-            add(new RepositoryModelItem(name.toString()));
+  public void loadRepositories(Runnable onComplete) {
+    cloudRepositoryService
+        .listAsync(user, cloudProject)
+        .thenAccept(response -> {
+          if (response == null) {
```
joaoandremartins: Can response ever be null?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectRepositoriesModelItem.java
```diff
@@ -53,26 +50,36 @@ public ProjectRepositoriesModelItem(@NotNull String cloudProject,
...
+          }
+
+          onComplete.run();
+        })
+        .exceptionally(response -> {
+          removeAllChildren();
+          add(new ResourceErrorModelItem(GctBundle.message("cloud.repository.list.error")));
+
+          onComplete.run();
+          return null;
```
joaoandremartins: Is this the method return? If so, should we make it return; only?

etanshaul: it requires a return value. I couldn't find a cleaner way to handle the exception case. The exceptionally method requires a lambda returning another CompletableFuture since its a fluent API.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloudRepositoryService.java
```diff
@@ -45,29 +46,38 @@
...
+      Credential credential = user.getCredential();
+      HttpRequestInitializer initializer = httpRequest -> {
+        HttpHeaders headers = new HttpHeaders();
+        httpRequest.setConnectTimeout(LIST_TIMEOUT_MS);
+        httpRequest.setReadTimeout(LIST_TIMEOUT_MS);
+        httpRequest.setHeaders(headers);
+        credential.initialize(httpRequest);
+      };
+
+      HttpTransport httpTransport = GoogleNetHttpTransport.newTrustedTransport();
```
joaoandremartins: The API makes us go this low? I wonder if the new one takes care of this for us..

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloudRepositoryService.java
```diff
@@ -88,3 +98,6 @@ public ListReposResponse list(CredentialedUser user, String cloudProject)
     }
   }
 }
+
+class CloudRepositoryServiceException extends RuntimeException {
```
joaoandremartins: Why did you create this new exception?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloudRepositoryService.java
etanshaul: hopefully

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectRepositoriesModelItem.java
joaoandremartins: Thanks for the explanation. Then I'd return Optional.empty().

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectRepositoriesModelItem.java
etanshaul: good question. I think we can remove this, but I'm checking.

etanshaul: removed the null check. it will either return a response or throw an exception, never null.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
etanshaul: I started going down this path but it led down a rabbit hole of changing too much code in this PR (the consumers of this method are also consuming other code that uses nullables and I would want to change it all simultaneously so we don't have coding mixing and matching).

I'd rather not make this PR even bigger. Thoughts?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectRepositoriesModelItem.java
etanshaul: I can't return that. The return value is a CompletableFuture so it needs to be of that type (or null).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
joaoandremartins: Not sure if adding a TODO is worth in this case. I think it's a fact that at some point, we'll go over all these instances and make them Java 8 friendly, so it should be OK.

PR #1187, opened 2017-01-19
Add appengine web xml api

patflynn:
@joaoandremartins 
@etanshaul 

PTAL let me know what you think. It could potentially simplify a few code paths that check these values (mutate also works which is very handy). 

I'd like to take this approach with the app.yaml for flex as well.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/descriptor/dom/AppEngineStandardWebApp.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/descriptor/dom/AppEngineStandardWebApp.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.intellij.appengine.descriptor.dom;
+
+import com.intellij.util.xml.DomElement;
+import com.intellij.util.xml.GenericDomValue;
+
+/**
+ * This is the Dom for the App Engine config file.
+ */
+public interface AppEngineStandardWebApp extends DomElement {
+
+  /**
+   * Returns the App Engine service name of this application
+   */
+  GenericDomValue<String> getService();
+
+  /**
+   * Returns the 'runtime' type of the application. For Java this is primarily empty for Java 7 and
+   * 'java8' for Java 8.
+   */
+  GenericDomValue<String> getRuntime();
+
+  /**
+   * Returns the App Engine environment type this application will run in. Currently this is empty
+   * for Standard Environment or 'flex' for Flexible Environment.  appengine-web.xmls support for
+   * 'flex' will be going away in 2018.
+   */
+  GenericDomValue<String> getEnv();
+
+  /**
+   * This field has been deprecated in the appengine-web.xml but we still need to be able to
+   * read it for older projects.
+   */
+  GenericDomValue<String> getVm();
+}
+
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacet.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacet.java
@@ -1,11 +1,11 @@
 /*
- * Copyright 2000-2016 JetBrains s.r.o.
+ * Copyright 2017 Google Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
+ *      http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -16,24 +16,28 @@
 
 package com.google.cloud.tools.intellij.appengine.facet;
 
+import com.google.cloud.tools.intellij.appengine.descriptor.dom.AppEngineStandardWebApp;
+import com.google.cloud.tools.intellij.appengine.project.AppEngineAssetProvider;
+import com.google.common.collect.ImmutableList;
 import com.intellij.facet.Facet;
 import com.intellij.facet.FacetManager;
 import com.intellij.facet.FacetType;
 import com.intellij.facet.FacetTypeId;
 import com.intellij.facet.FacetTypeRegistry;
 import com.intellij.openapi.module.Module;
-
+import com.intellij.psi.xml.XmlFile;
+import com.intellij.util.xml.DomFileElement;
+import com.intellij.util.xml.DomManager;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-/**
- * @author nik
- */
+/** @author nik */
 public class AppEngineStandardFacet extends Facet<AppEngineFacetConfiguration> {
 
   public static final FacetTypeId<AppEngineStandardFacet> ID = new FacetTypeId<>("appEngine");
 
-  public AppEngineStandardFacet(@NotNull FacetType facetType,
+  public AppEngineStandardFacet(
+      @NotNull FacetType facetType,
       @NotNull Module module,
       @NotNull String name,
       @NotNull AppEngineFacetConfiguration configuration) {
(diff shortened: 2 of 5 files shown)


etanshaul:
looks good I like centralizing this. 

what do you mean by mutate also works?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacet.java
```diff
@@ -52,4 +56,51 @@ public static AppEngineStandardFacet getAppEngineFacetByModule(@Nullable Module
...
   }
 
+  /**
+   * Returns {@code true} if this app is targeting the Flexible Environment instead of the Standard
+   * Environment.
+   */
+  public boolean isFlexEnvironment() {
+    AppEngineStandardWebApp appEngineStandardWebApp = getAppEngineStandardWebXml();
+    return appEngineStandardWebApp != null
+        && "flex".equalsIgnoreCase(appEngineStandardWebApp.getEnv().getStringValue());
```
etanshaul: what do these getters return if the xml element is missing? do we need null checking?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacet.java
```diff
@@ -52,4 +56,51 @@ public static AppEngineStandardFacet getAppEngineFacetByModule(@Nullable Module
...
+    AppEngineStandardWebApp appEngineStandardWebApp = getAppEngineStandardWebXml();
+    return appEngineStandardWebApp != null
+        && "true".equalsIgnoreCase(appEngineStandardWebApp.getVm().getStringValue());
+  }
+
+  /**
+   * Returns {@code true} if this app is targeting some environment other than the Standard
+   * Environment.
+   */
+  public boolean isNonStandardEnvironment() {
```
etanshaul: could we negate this so its easier to read when used in conditionals? e.g. `isStandardEnv()`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacet.java
patflynn: @etanshaul no need for null checking.  the GenericDomValue element deals with missing elements.

etanshaul:
Can a few of the methods in `AppEngineProjectService` now be removed due to these changes?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/descriptor/dom/AppEngineStandardWebApp.java
```diff
@@ -0,0 +1,51 @@
...
+
+package com.google.cloud.tools.intellij.appengine.descriptor.dom;
+
+import com.intellij.util.xml.DomElement;
+import com.intellij.util.xml.GenericDomValue;
+
+/**
+ * This is the Dom for the App Engine config file.
+ */
+public interface AppEngineStandardWebApp extends DomElement {
```
joaoandremartins: Do we really need another interface? I thought AppEngineProjectService would be a good place for functionality like this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacet.java
```diff
@@ -52,4 +56,51 @@ public static AppEngineStandardFacet getAppEngineFacetByModule(@Nullable Module
...
+    AppEngineStandardWebApp appEngineStandardWebApp = getAppEngineStandardWebXml();
+    return appEngineStandardWebApp != null
+        && "flex".equalsIgnoreCase(appEngineStandardWebApp.getEnv().getStringValue());
+  }
+
+  /**
+   * Returns {@code true} if this app is targeting the Managed VM environment instead of the
+   * Standard Environment.
+   */
+  public boolean isManagedVmEnvironment() {
```
joaoandremartins: I renamed this in my PR to isFlexEnvironment()

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacet.java
patflynn: @etanshaul funny you should ask.  I thought about this for a bit and my two rationales for doing it this way is that the is an AppEngineStandardFacet so I kind of assume that it's a standard environment.  The higher-order question I want to know is whether it's targeting some unusual non-standard environment, and since those environments are eventually going away (for standard) I can just remove the code paths that take the non-standard routes later.

It's a bit metaphysical I grant you. So I'm not deeply entrenched to this approach.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/descriptor/dom/AppEngineStandardWebApp.java
patflynn: This interface is automatically populated with values from the appengine-web.xml document.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacet.java
joaoandremartins: Oh, nvm. This is actually checking if isCompatEnvironment()?

patflynn: That's inaccurate. Flex and MVM are different environments.

patflynn:
@etanshaul by mutable I mean you can call setStringValue on AppEngineStandardWebApp nodes if the underlying XmlFile is opened for write.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacet.java
etanshaul: lol. I hear you. sounds good.

etanshaul:
oh I see - cool, that will be useful later

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacet.java
joaoandremartins: You mean env: flex vs vm: true? I don't know if the name Managed VMs is still being used for anything.

patflynn:
@etanshaul yeah I would refactor the AppEngineProjectService if you guys agree to this PR.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacet.java
patflynn: @joaoandremartins well at least they **were**. I'm double-checking

joaoandremartins:
Regarding following this approach for app.yaml for flex, I'll have to find a Java yaml parser if IJ's yaml package won't do the trick. However, we agreed to do this after the flex boilerplate is merged.

patflynn:
yep.  that's fine. I've found equivalent PSI YamlElements. But I haven't
found the same entry points to constructing the document representation.
Needs investigating.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacet.java
```diff
@@ -52,4 +56,51 @@ public static AppEngineStandardFacet getAppEngineFacetByModule(@Nullable Module
...
+  }
+
+  public boolean isJava8Runtime() {
+    AppEngineStandardWebApp appEngineStandardWebApp = getAppEngineStandardWebXml();
+    return appEngineStandardWebApp != null
+        && "java8".equalsIgnoreCase(appEngineStandardWebApp.getRuntime().getStringValue());
+  }
+
+  @Nullable
+  public AppEngineStandardWebApp getAppEngineStandardWebXml() {
```
etanshaul: Do you think it would make sense to modify the asset provider to return AppEngineStandardWebApp instead of XmlFile given these changes?

patflynn:
Probably but I'd do that after this change to keep this small.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacet.java
etanshaul: I remember that I exposed separate methods for env:flex and vm:true previously because the name of the environment is displayed differently on the web console and we wanted to reflect this in our deployment dialog too.

patflynn: @joaoandremartins @etanshaul confirmed that these environments are very different. I think for now it's ok to continue differentiating them in our UIs.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacet.java
etanshaul: just thinking about this. what if we called this method isCompatEnvironment() (or something similar) instead?

etanshaul: the reason I suggest that is that a "normal" flex setup is also "non-standard", but this method (since it lives in the standard facet class) really only applies to compat situations.

patflynn:
@joaoandremartins any further comments?

PR #1192, opened 2017-01-25
Fix layout overflow issue in AppEngineApplicationInfoPanel

aslo:
This fixes a UI regression when using small deployment configuration window sizes. With this change, text is able to overflow to the next line, while still keeping the `messageText` component to the right of the error icon. 

screenshot from 2017-01-25 15 21 04


joaoandremartins:
Actually, this isn't the fix I was thinking of.

screenshot from 2017-01-25 15 34 33

It's not very visible in the image but, if you look closely, the region is dislocated to the right.
I spent some time looking into it but couldn't discover why this happens. Do you have any idea why?

aslo:
Not at first glance but I can take a look.

joaoandremartins:
The misalignment becomes more visible if you add more aligned labels below the app info panel.

aslo:
@joaoandremartins I believe my latest commit should fix the issue: 

screenshot from 2017-01-25 15 46 02

joaoandremartins:
Wonderful, thanks!

PR #942, opened 2016-10-03
Use project jdk for dev_appserver local run

aslo:
fixes https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/927

Sets the dev_appserver's configured jdk to use the current project jdk. The ability for the user to manually configure the jdk was purposely omitted. This enhancement will come in a future PR (see https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/926).

TODO before merging:
- [x] perform 0.1.11 release of appengine-plugins-core
- [x] merge https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/943 (updates to the latest appengine-plugins-core release)

PR #1197, opened 2017-01-30
don't persist the java home path for shared local dev run configs

### LINKED ISSUE #1094: Shared run configurations include Java SDK path
We are using Google Cloud Tools plugin with latest Intellij IDEA (2016.2.5) and Java (1.8.0_112) to create "Google App Engine Local Development Server" run configurations used in our local development environment. The run configurations are shared, i.e. the "Share" checkbox is enabled and the resulting XML files are version controlled.

Inside the XML the `/component/configuration/server-settings/java_home_directory` element contains the path to Java SDK used in the IntelliJ IDEA project. When the run configuration is used, the plugin automatically adjusts the path in the XML file. Since Java SDK directory can differ from a developer to developer, this behaviour results in occasional version control commits and unnecessary conflicts.

We tried aligning the Java SDK location, but this is not realistic for most developer teams since the path depends on OS and Java version used. Most other Java-based plugins (e.g. Maven) do not include Java SDK path into the XML file and just use the one defined in IntelliJ IDEA project settings. It would be helpful if Google Cloud Tools would follow the same behaviour.

etanshaul:
fixes #1094

### DIFF
--- a/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
+++ b/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
@@ -434,16 +434,7 @@ public class AppEngineServerModel implements ServerModel, DeploysArtifactsOnStar
 
   @Override
   public String getJavaHomeDir() {
-    return settings.getJavaHomeDir();
-  }
-
-  public Sdk getDevAppServerJdk() {
-    return devAppServerJdk;
-  }
-
-  public void setDevAppServerJdk(Sdk devAppServerJdk) {
-    this.devAppServerJdk = devAppServerJdk;
-    settings.setJavaHomeDir(devAppServerJdk.getHomePath());
+    return devAppServerJdk.getHomePath();
   }
 
   public Boolean getClearDatastore() {
@@ -452,6 +443,10 @@ public class AppEngineServerModel implements ServerModel, DeploysArtifactsOnStar
 
   public void setClearDatastore(Boolean clearDatastore) {
     settings.setClearDatastore(clearDatastore);
+  }
+
+  private void setDevAppServerJdk(Sdk devAppServerJdk) {
+    this.devAppServerJdk = devAppServerJdk;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


patflynn:
@etanshaul can you provide more detail as to how that property is used.  What impact does not persisting the path have?  Shouldn't java home path always be provided by the module?

etanshaul:
@patflynn 

The project JDK is fetched when the server instance is run. This is so that the JAVA_HOME environment variable can be set for the dev app server.

Persisting it only has the effect of saving it when the user chooses to "share" the run configuration - if "share" is not checked, then no xml file is created. When shared, this has the incorrect outcome of saving a user's path to the JDK which would cause problems when the shared config is version controlled (as described in the referenced bug).

Not persisting it does not have any effect as the JDK value will be fetched on each server run.

See https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/942 for details on the initial implemation of this. + @aslo (the auther of that PR) - does this sound correct to you?

etanshaul:
Worth pointing out too:

Weather this is persisted or not has no effect on if the dev app server will run correctly: as I mentioned, the JDK value is retrieved on each run of the dev app server.

Looking at the user reported bug more closely it seems that its merely a problem of unnecessary commits:

> this behaviour results in occasional version control commits and unnecessary conflicts.

I verified this too. The correct env variable is set regardless of if this is persisted or not, so this fix will just prevent excess commits for those who version control their shared run configs.

joaoandremartins:
Actually, if we're not persisting javaHomeDir, is it still required in the class, or can we get rid of the member completely?

etanshaul:
I wondered that too. But looking at @aslo's related PR (#942) I think this is in preparation for supporting a JDK chooser from within the server configuration. @aslo can you further comment on this?

aslo:
My memory of https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/942 is a bit hazy, but I don't believe it should be necessary to keep `javaHomeDir` as a member of `AppEngineModelSettings` if we're no longer persisting it. It's also not necessary to keep that field in order to implement a jdk chooser.

I think `AppEngineModelSettings.javaHomeDir` is read elsewhere though, so you'd need to update any references to grab that setting from elsewhere if you do remove it.

etanshaul:
@joaoandremartins @patflynn @aslo 

ptal. I removed the javaHomeDir var from the persisted model settings class altogether.

joaoandremartins:
Go for it

PR #1198, opened 2017-01-31
update common lib to get clear data store flag functionality for local dev server

### LINKED ISSUE #1188: clear datastore checkbox for app engine standard local server does not work
It's not hooked up to anything it seems.

Probably blocked on https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/303

etanshaul:
fixes #1188 

The functionality on our end to pass in this flag was already in place. I just had to update the core lib version to get the feature.

I'm still verifying that there aren't any regressions due to the upgrade.

### DIFF
--- a/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
+++ b/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
@@ -446,6 +446,7 @@ public class AppEngineServerModel implements ServerModel, DeploysArtifactsOnStar
     settings.setJavaHomeDir(devAppServerJdk.getHomePath());
   }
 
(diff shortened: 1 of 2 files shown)


etanshaul:
@patflynn @joaoandremartins 

No regressions found with upgrade to latest common lib.

patflynn:
@etanshaul I just noticed that this PR closes out the bug. But don't we still need to make sure we take the flag value from the model and pass it to the devappserver run config?

etanshaul:
@patflynn it is. See CloudSdkAppEngineDevServer#101:

`arguments.addAll(DevAppServerArgs.get("clear_datastore", config.getClearDatastore()));`

etanshaul:
Actually that is in the core lib. To see how its passing it to there, you can see AppEngineServerModel#441 where we override the getter in the core lib's RunConfiguration class.

patflynn:
Aha. I see now. Thanks!

PR #1209, opened 2017-02-07
the lack of a 'dh' key/value in our analytics pings is causing our data to be filtered out. 

patflynn:


google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
```diff
@@ -96,7 +98,8 @@
           new BasicNameValuePair(IS_NON_INTERACTIVE_KEY, STRING_FALSE_VALUE),
           new BasicNameValuePair(
               UNIQUE_CLIENT_ID_KEY,
-              UpdateChecker.getInstallationUID(PropertiesComponent.getInstance())));
+              UpdateChecker.getInstallationUID(PropertiesComponent.getInstance())),
```
etanshaul: I may just not understand how this works, but I notice the already existing "dp" value is added in the sendEvent method and guarded by isTrackingEnabled, but for this new "dh" value, you are adding it on class init. Should we update them to make them consistent?

patflynn: @etanshaul 

Do you mean the 'dt' value?

DH never changes so it can be intialized into the base data. DT has data about the current ping which is only known at the time the event is called.

patflynn:
@chanseokoh PTAL.

patflynn:
@chanseokoh I'm merging so that we can prepare the release from master.  LMK if you see any issues.

chanseokoh:
LGTM

PR #1213, opened 2017-02-07
Fix CSR cloning directory selection validation error state

### LINKED ISSUE #1206: Misleading "directory already exists" error in GCR clone dialog
When attempting to clone a GCR repo to an already existing directory, 1) Parent directory gets highlighted instead of Directory Name and 2) the error caption mentions that parent path exists, where the real problem is that the specified directory already exists (the parent existing is actually a good thing).

screenshot from 2017-02-07 11 03 27

etanshaul:
fixes #1206


google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -156,7 +156,7 @@ clonefromgcp.title=Clone from Google Cloud
 clonefromgcp.button=Clone
 clonefromgcp.destination.directory.description=Select a parent directory destination directory for clone
 clonefromgcp.destination.directory.title=Parent Directory
-clonefromgcp.destination.exists.error=The parent path already exists. Repository cannot be cloned to an existing directory.
+clonefromgcp.destination.exists.error=The directory already exists. Repository cannot be cloned to an existing directory.
```
joaoandremartins: Maybe add "selected" or "specified"?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneCloudRepositoryDialog.java
```diff
@@ -267,15 +268,25 @@ protected JComponent createCenterPanel() {
...
     parentDirectory.setBackground(Color.RED);
     parentDirectoryLabel.setForeground(Color.RED);
   }
+
+  /**
+   * Paints the directory selector red when in error.
+   */
+  private void paintDirectorySelectionError() {
+    paintSelectionOk();
+    directoryNameLabel.setForeground(Color.RED);
```
joaoandremartins: nit: maybe you also want to set the directoryName background to red, so the error renders similarly to the parent directory one?

etanshaul: i think I'll remove the background setting from the parent directory widget. setting the background doesn't have the same effect (for the parent directory it outlines the border, while for the directory it actually colors in the background).

joaoandremartins: Yes, I meant coloring the text box border red (which is accomplished by setting the background to red).
You can also remove the border colouring altogether, as long as both errors render the same way.

etanshaul:
@joaoandremartins - updated based on your comments

PR #1216, opened 2017-02-07
Recheck application & region when manually typing in a cloud project

### LINKED ISSUE #1208: App Engine project selector region warning does't go away
To reproduce:

1) open up a flex project in IJ
2) open up the deployment configuration
3) select a project that is missing a region:
screen shot 2017-02-07 at 11 57 13 am
4) select a valid project that does have a region **by typing the project id**:
screen shot 2017-02-07 at 11 58 19 am

Notice the warning does not go away. If I were to manually select the project from the dropdown instead of typing it, it would disappear as expected.

etanshaul:
fixes #1208

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
@@ -44,6 +44,8 @@ import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.awt.event.ComponentAdapter;
 import java.awt.event.ComponentEvent;
+import java.awt.event.FocusEvent;
+import java.awt.event.FocusListener;
 import java.awt.event.HierarchyEvent;
 import java.awt.event.HierarchyListener;
 import java.util.ArrayList;
@@ -389,6 +391,25 @@ public class ProjectSelector extends CustomizableComboBox implements Customizabl
     if (!popup.isVisible()) {
       popup.show(showTarget);
     }
+  }
+
+  @Override
+  public void addFocusListener() {
+    super.addFocusListener();
+
+    getTextField().addFocusListener(new FocusListener() {
+      @Override
+      public void focusGained(FocusEvent e) {
+        // do nothing
+      }
+
+      @Override
+      public void focusLost(FocusEvent event) {
+        if (!event.isTemporary()) {
+          onSelectionChanged(getCurrentModelItem());
+        }
+      }
+    });
   }
 
   private class PopupPanel extends GoogleLoginEmptyPanel {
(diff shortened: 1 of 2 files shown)


etanshaul:
Previously, no event was fired when the user manually typed in a cloud project, causing any application / region errors to not go away.

This forces an event (on losing focus) which will then cause the region to be revalidated.

@aslo can you also ptal (since you wrote the pr dealing with app creation)?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
```diff
@@ -391,6 +393,25 @@ public void showPopup(RelativePoint showTarget) {
...
+
+    getTextField().addFocusListener(new FocusListener() {
+      @Override
+      public void focusGained(FocusEvent e) {
+        // do nothing
+      }
+
+      @Override
+      public void focusLost(FocusEvent event) {
+        if (!event.isTemporary()) {
```
aslo: What does it mean for an event to be temporary?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
```diff
@@ -391,6 +393,25 @@ public void showPopup(RelativePoint showTarget) {
...
+          onSelectionChanged(getCurrentModelItem());
```
aslo: This means that `onSelectionChanged` could be called with a null argument. Did you verify that this won't break anything if that happens?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
```diff
@@ -391,6 +393,25 @@ public void showPopup(RelativePoint showTarget) {
     }
   }
 
+  @Override
+  public void addFocusListener() {
```
aslo: Small nit, but I'd argue that it might be cleaner to just call getTextField.addFocusListener(...) directly in this class's constructor rather than overriding this method. That way this class has control of how/when the listener is added.

etanshaul: yeah, totally.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
etanshaul: a lose focus event is permanent if it remains unfocused after everything is done. 

I noticed that this event was getting fired twice always: once with temporary=true and once with temporary=false. I did a little digging and saw that we can ignore the temporary once.

aslo: :+1:

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
etanshaul: good catch!

etanshaul:
@aslo - updated from your comments

PR #1229, opened 2017-02-08
Load CSRs after selecting a valid project instead of only on opening the widget

### LINKED ISSUE #1228: Manually typing a CSR repo prior to opening dropdown fails
The cloud repos are loaded once the user opens up the dropdown.

If the user tabs into the field (without opening the dropdown) and starts typing a valid repo, it will not be able to match it and the dialog will complain.

Need to load the CSR repo's earlier.

etanshaul:
fixes #1228

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
@@ -34,7 +34,6 @@ import com.intellij.ui.awt.RelativePoint;
 import com.intellij.ui.components.JBLabel;
 import com.intellij.ui.components.JBScrollPane;
 
-import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.awt.BorderLayout;
@@ -71,6 +70,7 @@ public class RepositorySelector extends CustomizableComboBox implements Customiz
       = "https://console.cloud.google.com/code/develop/repo?project=";
   private JBPopup popup;
   private JPanel panel;
+  private ProjectRepositoriesModelItem repositories;
   private String cloudProject;
   private CredentialedUser user;
   private boolean canCreateRepository;
@@ -85,21 +85,21 @@ public class RepositorySelector extends CustomizableComboBox implements Customiz
         .setText(GctBundle.message("cloud.repository.selector.placeholder.text"));
   }
 
+
+
   @Nullable
   public String getSelectedRepository() {
-    if (StringUtil.isEmpty(getText())) {
+    if (StringUtil.isEmpty(getText()) || repositories == null) {
       return null;
     }
 
-    if (panel instanceof RepositoryPanel) {
-      Enumeration repos = ((RepositoryPanel) panel).getRepositories().children();
-      while (repos.hasMoreElements()) {
-        TreeNode repo = (TreeNode) repos.nextElement();
+    Enumeration repos = repositories.children();
+    while (repos.hasMoreElements()) {
+      TreeNode repo = (TreeNode) repos.nextElement();
 
-        if (repo instanceof RepositoryModelItem
-            && getText().equals(((RepositoryModelItem) repo).getRepositoryId())) {
-          return getText();
-        }
+      if (repo instanceof RepositoryModelItem
+          && getText().equals(((RepositoryModelItem) repo).getRepositoryId())) {
+        return getText();
       }
     }
 
@@ -112,6 +112,18 @@ public class RepositorySelector extends CustomizableComboBox implements Customiz
 
   public void setUser(CredentialedUser user) {
     this.user = user;
+  }
+
+  public void loadRepositories(Runnable onComplete) {
+    if (user == null || cloudProject == null) {
+      return;
+    }
+
+    if (repositories == null) {
+      repositories = new ProjectRepositoriesModelItem(cloudProject, user);
+    }
+
+    repositories.loadRepositories(onComplete);
   }
 
   @Override
@@ -159,7 +171,7 @@ public class RepositorySelector extends CustomizableComboBox implements Customiz
    * and then expand with the repository textfield if the user manually stretches the dialog.
    */
   private int getPopupWidth() {
-    int actualWidth = RepositorySelector.this.getTextField().getWidth();
+    int actualWidth = this.getTextField().getWidth();
     return Math.max(SELECTOR_WIDTH, actualWidth);
   }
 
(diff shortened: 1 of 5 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneCloudRepositoryDialog.java
```diff
@@ -235,6 +235,7 @@ protected void textChanged(DocumentEvent event) {
         repositorySelector.setCloudProject(projectSelector.getText());
         repositorySelector.setUser(projectSelector.getSelectedUser());
         repositorySelector.setText("");
+        repositorySelector.loadRepositories(null /*onComplete*/);
```
joaoandremartins: nit: is it worth it to create an overloaded no-arg loadRepositories() method?

etanshaul: not sure. I'm on the fence about if I prefer being explicit about whether the callback is passed in or not. So I'll leave it unless you feel strongly.

joaoandremartins: IMO there's no better way to convey that no callback is being passed than using a no-arg value.
Conventions on passing null's around can be very lax and hard to understand without context. Not to mention that explicitly passing null into a method smells like NPE. The gradual move to Optionals would be another argument against explicitly passing in null.

In your place, I'd definitely create a new method. Up to you!

joaoandremartins:
I don't understand why this fix works. What event triggers the refresh?
Also, I noticed that if I type in a repo name before, then start typing the project name, the repo name gets deleted. It's probably a use case that's never going to get triggered, but it might hint at some other issues, and there might be an easy fix? What do you think?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -114,6 +112,18 @@ public void setUser(CredentialedUser user) {
     this.user = user;
   }
 
+  public void loadRepositories(Runnable onComplete) {
```
patflynn: Should onComplete be @Nullable here?

etanshaul: yeah, definitely

etanshaul:
@joaoandremartins it works because now the repositories are loaded when a valid project is selected. If you look at the updated lines in `CloneCloudRepositoryDialog` and `SetupCloudRepositoryDialog` you'll see that `loadRepositories` is called in the project selector listener.

Regarding your second point about typing a repo name and then it disappearing. This isn't indicating another issue, but is as expected: if the user changes the project, the entire set of repo's changes. In my opinion (though I can sort of see the counter argument) it makes sense to clear out the repo name.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneCloudRepositoryDialog.java
etanshaul: I need to read more about the arguments to form a full opinion here, but there seems to be a prevailing school of thought that says that using Optionals for method arguments is bad practice.

Regarding the overloaded method - if I do it here, you'll notice that I may have to do it again in `ProjectRepositoriesModelItem#loadRepositories` too. It may make sense, but my initial feeling was that its cleaner to allow @Nullable arguments. Ill consider it more.

etanshaul:
To elaborate on why the fix works:

Previously, the repositories would only be loaded once the user actually opened up the dropdown widget. Tabbing into the textfield and typing would not cause any repos to be fetched.

This PR now also fetches repos once a valid project is selected. This way, once the user starts typing into the repo textfield, they will already have been loaded (assuming a valid project was selected first).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneCloudRepositoryDialog.java
joaoandremartins: Did you mean the Brian Goetz preso? Wasn't he saying there should be no members of type Optional? Can't remember what his point of view on method arguments was but, interestingly enough, the authors of Java 8 make up good points about using Optionals everywhere.

Indeed, you'd still have to pass null in in the body of the arg-less method. But the API wouldn't require you to pass the null in, which is what I like about it. I don't see how the @Nullable way is cleaner, I actually think it's the opposite -- loadRepositories() is so much cleaner than loadRepositories(null /* onComplete */) :)

joaoandremartins:
The ideal behaviour to me would be that nothing disappears from the UI, only the dialog state changes to error if, e.g., you change to an non-existing project. However, we're dealing with such a farfetched and unusual situation that I don't mind whichever way we end up going.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneCloudRepositoryDialog.java
etanshaul: Didn't see the Goetz pres. yet, but plan to. Many of the places I've randomly read online talk about it being bad practice. E.g.:

http://stackoverflow.com/questions/31922866/why-should-java-8s-optional-not-be-used-in-arguments
https://dzone.com/articles/java-8-optional-how-use-it

But they could be totally wrong. I need to read more about it.

joaoandremartins: Ah yes, this is why the solution works. I missed this bit of code, so feel free to ignore my question about why the solution works.

etanshaul:
I actually found a bigger bug with this while testing out your suggestion of not clearing out the cloud repo.

It seems the repo selector isn't getting updated when the user chooses a different project.

I'll update again once fixed.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/CloneCloudRepositoryDialog.java
patflynn: > IMO there's no better way to convey that no callback is being passed than using a no-arg value.
👍 

WRGT to passing Optional as a parameter I am strongly against it. It's very ungainly from a callers point of view and I don't see a lot of benefits.  Generally optional/nullable parameter types are like flag parameter types. An anti-pattern that should be avoided.

etanshaul:
@joaoandremartins @patflynn ready for review again

PR #1240, opened 2017-02-13
Committing project structure changes causing local vcs conflicts

### LINKED ISSUE #1239: Opening our project locally creates VCS changes
for example, just opening the project and doing nothing causes a local change in `compiler.xml`

We should figure out why and fix it (either commit the change or change the project somehow).

etanshaul:
fixes #1239

### DIFF
--- a/.idea/.name
+++ b/.idea/.name
@@ -1,1 +1,0 @@
-gcloud-intellij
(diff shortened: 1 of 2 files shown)


patflynn:
Does this turn off error prone?

etanshaul:
Error prone still runs when building the plugin. I'm check to see if this caused any IDE inspections to be removed.

joaoandremartins:
These are the changes the IDE always applies by default, which make me have to stash all the time.
They're always on in my local environment and I haven't seen any issues.

etanshaul:
@patflynn I'm not seeing any changes with regards to inspections within the IDE (there were none before), and as I mentioned error prone still runs.

Issue #1241, opened 2017-02-13
High rates of user reported deployment failure.


patflynn:
Over half of app engine standard deployment failures are reporting:

"Deployment failed due to an unexpected error while staging the project."

as the cause.   Staging should very rarely fail. 

We're also seeing a consistently high rate of:

"Deployment failed with exit code: 1 Please make sure that you are using the latest version of the Google Cloud SDK. Run ''gcloud components update'' to update the SDK. (See: https://cloud.google.com/sdk/gcloud/reference/components/update.)"

This requires some digging since I would expect the rate of this error to go down over time.

etanshaul:
@patflynn to check analytics again

nkibler:
Duplicate of #1509.

AdamSHurwitz:
I'm receiving this error "**Deployment failed with exit code: 1...**" How do I identify the root of the issue?

patflynn:
Hi @AdamSHurwitz, could you file a [new issue](https://github.com/GoogleCloudPlatform/google-cloud-java/issues/new) including the information mentioned in the template?

thx.

PR #1242, opened 2017-02-13
Improve deployment and staging error metrics

etanshaul:
#1241


etanshaul:
Still looking at more ways to improve this

patflynn:
We should update the common library to do staging using future's the same way the experimental deploy does.  I'd like to serialize all this code as much as possible. We've got node like code soup here which is hard to read.

etanshaul:
yes, we indeed have callback hell here. I'll look at this in a different PR.

A common pattern we have with handling deployment/staging exception is something like this:

```
catch (SomeException se) {
  callback.errorOccured(someLocalizedMessage); // cancels the operation and sends the message to metrics
  logger.error(se); // IDE error and notifies us
}
```

@patflynn, @joaoandremartins  what are your thoughts on including the localized message in the logger.error and / or including the exception message in the callback errorOccured?

What I'm aiming for is a way to associate the localize message (e.g. "staging failed due to an unknown error") with the actual exception that was thrown. @patflynn I do remember you were against including localized messages in the logging statements.

Note: this will not help for cases where staging/deploy fails due to the process failing unsuccessfully (which is the majority of our failures). These cases still require updating gcloud's error messaging.

patflynn:
@etanshaul why include localized message when you can just add your own more specific message to the error logging statement?

etanshaul:
I don't think we have anything more specific to give. But on second thought, I don't think this will be helpful. We can already cross-reference the exception with the reported metric error with my current changes.

etanshaul:
@patflynn, before I merge - did you have anything else in mind for this PR?

patflynn:
nope.

PR #1247, opened 2017-02-16
improves the codeflow for cases where the user has to auth when missing creds for a project they're trying to deploy.

patflynn:
scratching an itch I came across when reviewing the flex changes.

### DIFF
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/login/IntellijGoogleLoginService.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/login/IntellijGoogleLoginService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2016 The Android Open Source Project
+ * Copyright 2017 Google Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -61,6 +61,7 @@ import java.awt.Window;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 
 /**
@@ -68,12 +69,11 @@ import java.util.Map;
  */
 public class IntellijGoogleLoginService implements GoogleLoginService {
 
+  private static final Logger LOG = Logger.getInstance(IntellijGoogleLoginService.class);
   private ClientInfo clientInfo;
   private AndroidUiFacade uiFacade;
   private AndroidPreferencesOAuthDataStore dataStore;
   private CredentialedUserRoster users;
-
-  private static final Logger LOG = Logger.getInstance(IntellijGoogleLoginService.class);
 
   private IntellijGoogleLoginService() {
     this.clientInfo = getClientInfo();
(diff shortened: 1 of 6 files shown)


google-account-plugin/src/com/google/cloud/tools/intellij/login/IntellijGoogleLoginService.java
```diff
@@ -83,6 +83,37 @@ private IntellijGoogleLoginService() {
     addLoginListenersFromExtensionPoints();
   }
 
+  @Nullable
+  private static Project getCurrentProject() {
+    Window activeWindow = WindowManagerEx.getInstanceEx().getMostRecentFocusedWindow();
+    if (activeWindow == null) {
+      return null;
```
joaoandremartins: I'd rather we return Optional.empty() here and change the return to Optional<Project>?

joaoandremartins: Ah, I see this code was already existing. I leave it to you if you want to make this change or not.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelperTest.java
```diff
@@ -89,8 +89,9 @@ public void testCreateApplicationDefaultCredentials() throws Exception {
     String clientSecret = "clientSecret";
     String refreshToken = "refreshToken";
     when(deploymentConfiguration.getGoogleUsername()).thenReturn(username);
-    when(googleLoginService.getAllUsers())
-        .thenReturn(ImmutableMap.of(username, credentialedUser));
+    when(googleLoginService.ensureLoggedIn(username)).thenReturn(true);
```
joaoandremartins: Is it worth adding a test when this returns false?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -231,42 +235,20 @@ public CloudSdk createSdk(
...
   private Path doStageCredentials(String googleUsername) {
-    CredentialedUser projectUser = Services.getLoginService().getAllUsers()
-        .get(googleUsername);
+    Optional<CredentialedUser> projectUser = Services.getLoginService().getLoggedInUser(googleUsername);
 
     GoogleLoginState googleLoginState;
-    if (projectUser != null) {
-      googleLoginState = projectUser.getGoogleLoginState();
+    if (projectUser.isPresent()) {
+      googleLoginState = projectUser.get().getGoogleLoginState();
```
joaoandremartins: How do you feel about a different notation? Something in the likes of

projectUser.map(CredentialedUser::getGoogleLoginState).orElse(null);

Feel free to replace null by anything else better, or even throw an exception.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelperTest.java
patflynn: done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
patflynn: neat. works for me. I'm going to avoid changing the behavior to minimize scope.

patflynn:
The benefit of the PR is that it reduces complexity of the stageCredentials command and handles the case where the user logs in with the wrong account.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -231,58 +235,36 @@ public CloudSdk createSdk(
...
   }
 
   private Path doStageCredentials(String googleUsername) {
-    CredentialedUser projectUser = Services.getLoginService().getAllUsers()
-        .get(googleUsername);
+    Optional<CredentialedUser> projectUser =
+        Services.getLoginService().getLoggedInUser(googleUsername);
+
+    GoogleLoginState googleLoginState =
+        projectUser.map(CredentialedUser::getGoogleLoginState).orElse(null);
```
joaoandremartins: I think a null check is in order, here, since orElse(null) won't make the method return null.

patflynn: It'll NPE on line 253 of it's null which I believe preserves the current behavior.

joaoandremartins: Maybe leave it at `Optional<GoogleLoginState> googleLoginState = projectUser.map(CredentialedUser::getGoogleLoginState);` and then `if (!googleLoginState.isPresent) return null;`?

joaoandremartins: The current behaviour is the method returning null, not throwing an NPE

patflynn: I changed this back to the original logic.

PR #1248, opened 2017-02-17
Reverts to parsing CDB breakpoint dates with only one format

joaoandremartins:
Unfortunately the new Java 8 DateTime API doesn't parse the "yyyy-MM-dd'T'HH:mm:ss.SSSZ" format out of the box (ISO_INSTANT looks like "yyyy-MM-dd'T'HH:mm:ssZ" and explodes if millis are in), so I kept using joda.time.


patflynn:
Just curious if this is fixing any existing bug or if it's just a refactoring.

joaoandremartins:
I guess you could call it a refactor.

joaoandremartins:
In case you're not aware, I should point out this is kind of a rollback of #919, when, all of a sudden, CDB started sending us timestamps in an unexpected format. At the time, I recall someone had concerns about always trying to parse timestamps with two different formats, even though, in practice, all of them should be parsed by the first tried format.

The test would have to be an integration test, since this depends on the timestamps returned by CDB. I could add a TODO somewhere, so that when we tackle integration tests properly, this gets picked up?

joaoandremartins:
I could also add unit tests for a good and bad format, ensuring that bad formats don't cause explosions.

joaoandremartins:
Actually, I don't see a very good place to add a unit test for this. I think it would be best to have an integration test. I'll file an issue here instead of adding TODO to the code.

joaoandremartins:
I also got approval from the CDB team on b/31648200

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/BreakpointComparer.java
```diff
@@ -75,14 +77,16 @@ public int compare(Breakpoint o1, Breakpoint o2) {
       }
       return s1.getPath().compareTo(s2.getPath());
     }
-    Date d1 = BreakpointUtil.parseDateTime(o1.getFinalTime());
-    Date d2 = BreakpointUtil.parseDateTime(o2.getFinalTime());
-    if (d1 == null) {
+
+    Date d1, d2;
+    try {
+      d1 = ISODateTimeFormat.dateTime().parseDateTime(o1.getFinalTime()).toDate();
```
patflynn: there is a slight behavior change here in the sense that if parsing the first fails we don't parse the second. Not sure if that's important though.

joaoandremartins: Yes, this was done in purpose, though. If the first timestamp fails to parse, it's because the server is crazy again and the second timestamp will almost surely also fail to parse.

PR #1257, opened 2017-02-21
fix reference equality call

patflynn:


etanshaul:
could you please point me to the part of the code that wasn't formatting changes? I'm having a hard time finding the reference fix.

patflynn:
Yeah. I know that's annoying.  Something seems to have gone wrong with the
googleJavaFormatter configured in Gradle because all of our code is
supposed to be formatted correctly already.  I'm looking into it.

The behavior change is on line 95.

PR #1258, opened 2017-02-21
Make errorprone happy and silent

joaoandremartins:
- Makes errorprone happy and silent,
- Several Java8 compliant changes.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
@@ -25,7 +25,6 @@ import com.google.common.collect.Sets;
 import com.intellij.facet.Facet;
 import com.intellij.facet.ui.FacetEditorContext;
 import com.intellij.facet.ui.FacetEditorTab;
-import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.application.Result;
 import com.intellij.openapi.application.WriteAction;
 import com.intellij.openapi.roots.DependencyScope;
@@ -40,9 +39,7 @@ import com.intellij.packaging.artifacts.Artifact;
 
 import org.jetbrains.annotations.NotNull;
 
-import java.util.Arrays;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.Set;
 
 import javax.swing.JComponent;
@@ -108,7 +105,8 @@ public class AppEngineStandardFacetEditor extends FacetEditorTab {
 
       if (!libsToAdd.isEmpty()) {
         for (AppEngineStandardMavenLibrary library : libsToAdd) {
-          AppEngineSupportProvider.loadMavenLibrary(context.getModule(), library);
+          MavenRepositoryLibraryDownloader.getInstance().downloadLibrary(
+              context.getModule(), library);
         }
       }
 
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
@@ -25,6 +25,7 @@ import com.google.common.collect.Sets;
 import com.intellij.facet.Facet;
 import com.intellij.facet.ui.FacetEditorContext;
 import com.intellij.facet.ui.FacetEditorTab;
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.application.Result;
 import com.intellij.openapi.application.WriteAction;
 import com.intellij.openapi.roots.DependencyScope;
@@ -39,7 +40,9 @@ import com.intellij.packaging.artifacts.Artifact;
 
 import org.jetbrains.annotations.NotNull;
 
+import java.util.Arrays;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 
 import javax.swing.JComponent;
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
@@ -63,8 +63,6 @@ import com.intellij.util.PlatformUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.jetbrains.annotations.TestOnly;
-import org.jetbrains.idea.maven.utils.library.RepositoryLibraryDescription;
-import org.jetbrains.idea.maven.utils.library.RepositoryLibraryProperties;
 
 import java.io.IOException;
 import java.util.Collection;
@@ -186,7 +184,8 @@ public class AppEngineSupportProvider extends FrameworkSupportInModuleProvider {
       protected void run(@NotNull Result result) throws Throwable {
         if (librariesToAdd != null && !librariesToAdd.isEmpty()) {
           for (AppEngineStandardMavenLibrary libraryToAdd : librariesToAdd) {
-            Library mavenLibrary = loadMavenLibrary(module, libraryToAdd);
+            Library mavenLibrary = MavenRepositoryLibraryDownloader.getInstance()
+                .downloadLibrary(module, libraryToAdd);
             if (mavenLibrary != null) {
               rootModel.addLibraryEntry(mavenLibrary).setScope(libraryToAdd.getScope());
               AppEngineStandardWebIntegration.getInstance()
@@ -218,8 +217,7 @@ public class AppEngineSupportProvider extends FrameworkSupportInModuleProvider {
           @Override
           protected void run(@NotNull Result result) throws Throwable {
             for (AppEngineStandardMavenLibrary libraryToRemove : librariesToRemove) {
-              final String displayName = AppEngineStandardMavenLibrary
-                  .toMavenDisplayVersion(libraryToRemove.getLibraryProperties());
+              final String displayName = libraryToRemove.toMavenDisplayVersion();
               final Library library = libraryTable.getLibraryByName(displayName);
               if (library != null) {
                 libraryTable.removeLibrary(library);
@@ -237,14 +235,6 @@ public class AppEngineSupportProvider extends FrameworkSupportInModuleProvider {
         }.execute();
       }
     }, ModalityState.NON_MODAL);
-  }
-
-  static Library loadMavenLibrary(final Module module, AppEngineStandardMavenLibrary library) {
-    RepositoryLibraryProperties libraryProperties = library.getLibraryProperties();
-
-    return MavenRepositoryLibraryDownloader.getInstance().downloadLibrary(module,
-        RepositoryLibraryDescription.findDescription(libraryProperties), libraryProperties,
-        libraryProperties.getVersion());
   }
 
   @NotNull
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/login/GoogleLoginUtils.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/login/GoogleLoginUtils.java
@@ -17,7 +17,6 @@
 package com.google.cloud.tools.intellij.login;
 
 import com.google.api.client.auth.oauth2.Credential;
-import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
 import com.google.api.client.http.javanet.NetHttpTransport;
 import com.google.api.client.json.jackson2.JacksonFactory;
 import com.google.api.services.oauth2.Oauth2;
@@ -131,28 +130,5 @@ public class GoogleLoginUtils {
         }
       }, ModalityState.defaultModalityState());
     }
-  }
-
-  /**
-   * Returns a {@link Credential} object for a fake user. Used for testing.
-   *
-   * @return a {@link Credential} object for the fake user.
-   */
-  @NotNull
-  public static Credential makeFakeUserCredential() {
-    String clientId = System.getenv().get("ANDROID_CLIENT_ID");
-    String clientSecret = System.getenv().get("ANDROID_CLIENT_SECRET");
-    String refreshToken = System.getenv().get("FAKE_USER_REFRESH_TOKEN");
-    String accessToken = System.getenv().get("FAKE_USER_ACCESS_TOKEN");
-
-    Credential cred =
-        new GoogleCredential.Builder()
-            .setJsonFactory(new JacksonFactory())
-            .setTransport(new NetHttpTransport())
-            .setClientSecrets(clientId, clientSecret)
-            .build();
-    cred.setAccessToken(accessToken);
-    cred.setRefreshToken(refreshToken);
-    return cred;
   }
 }
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/stats/GoogleUsageTracker.java
@@ -25,13 +25,11 @@ import com.google.common.collect.Lists;
 import com.google.common.escape.CharEscaperBuilder;
 import com.google.common.escape.Escaper;
 
-import com.intellij.ide.util.PropertiesComponent;
 import com.intellij.openapi.application.ApplicationInfo;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.application.PermanentInstallationID;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.diagnostic.Logger;
-import com.intellij.openapi.updateSettings.impl.UpdateChecker;
 import com.intellij.openapi.util.SystemInfo;
 import com.intellij.util.PlatformUtils;
 
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardMavenLibrary.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardMavenLibrary.java
@@ -24,7 +24,6 @@ import com.intellij.openapi.roots.DependencyScope;
 
 import org.apache.commons.lang.WordUtils;
 import org.jetbrains.annotations.Nullable;
-import org.jetbrains.idea.maven.utils.library.RepositoryLibraryProperties;
 import org.jetbrains.idea.maven.utils.library.RepositoryUtils;
 
 import java.util.Arrays;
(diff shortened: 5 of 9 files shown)


patflynn:
it looks like you need to rebase.

joaoandremartins:
Seems like you and I spent yesterday working on the same things.

patflynn:
@joaoandremartins are you watching the repo?  You should have gotten notifications of the PRs yesterday.

patflynn:
are you using a tool to generate these refactorings or doing them manually?

joaoandremartins:
I wasn't, but I will now.

The main motivation for this PR was getting rid of all the errorprone warnings, including the futures one and that enum mutable field. That required changes to a few files, and I did the rest of the refactors manually. I doubt there is a tool that can perform most of the refactors automatically, since they required changing the logic a little bit.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
```diff
@@ -166,50 +167,42 @@ private boolean isManagedLibrariesEnabled() {
...
-          ModuleRootManager manager = ModuleRootManager.getInstance(context.getModule());
-          ModifiableRootModel model = manager.getModifiableModel();
-
-          model.addLibraryEntry(addedLibrary);
-
-          for (OrderEntry orderEntry : model.getOrderEntries()) {
-            if (orderEntry.getPresentableName().equals(addedLibrary.getName())) {
-              ((ExportableOrderEntry) orderEntry).setScope(scope);
-            }
+      AppEngineStandardMavenLibrary.getLibraryByMavenDisplayName(addedLibrary.getName()).ifPresent(
```
etanshaul: Can you explain a bit why you added this check?

etanshaul:
this may be an annoying request, but since these changes are easily submit separately, could you split them up? perhaps one pr to update the library stuff, another to change to lambdas, another to change to Optionals.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
etanshaul: Specifically I guess - how could it be that the added library is not already an AppEngineStandardMavenLibrary?

joaoandremartins: This is just the translation of the previous code. Notice that you're checking if library is null.

I don't think it will ever happen that library isn't present. But, since getLibraryByMavenDisplayName() returns an Optional, I think this is the best way to work with that.

etanshaul: Oh yea - I missed that part because of the refactoring.

Could you restore it to the format I had it to prevent such deep nesting? I think it makes it easier to read.

joaoandremartins:
I thought the way we were going about this was to make Java8 changes whenever we came across the files.

It would be a lot of work to split this into separate PRs, and I don't think it's worth the trouble. All the code here, besides replacing RepositoryLibraryProperties in the enum, should just be a direct translation of Java 7 - Java 8. I didn't add/remove anything.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
joaoandremartins: I try to avoid using isPresent() has hard as I can, since it throws away the benefit of using Optionals. I prefer my approach, it's more functional style, and I think it's more readable too.

etanshaul: fair enough. i do like the use of ifPresent, but the nesting is really deep. I can live with it. you could extract the lambda inside of it to another method too.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
```diff
@@ -147,8 +147,8 @@ private void trackApplicationCreateFailure() {
         .ping();
   }
 
-  private void setStatusMessageAsync(final String message, final boolean isError) {
```
etanshaul: does errorProne really complain about this? even though its always true (since we don't have other use cases yet), it feels weird to be hardcoding it to true in this method. Not a big deal though

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacetConfiguration.java
```diff
@@ -69,19 +71,11 @@ public void writeExternal(Element element) throws WriteExternalException {
...
-      if (mavenLibrary != null) {
-        configuredLibraries.add(mavenLibrary);
-      }
-    }
-
-    return configuredLibraries;
+    return Arrays.stream(
+        LibraryTablesRegistrar.getInstance().getLibraryTable(project).getLibraries())
+        .map(library ->
+            AppEngineStandardMavenLibrary.getLibraryByMavenDisplayName(library.getName()).get())
```
etanshaul: is this null safe? what about only adding it to the set if its non null, does this still do that?

etanshaul: the null safe part of my question is referring to the direct get() you are doing on the Optional.

patflynn:
@joaoandremartins  
Keep separate concerns in separate PRs whenever possible.  

You should have broken this PR up as:
 PR #1: Fix unused future warnings
 PR #2: Fix mutable enum
 PR #3: Refactor to Java8 closures

For PR #3, we should discuss our approach to Java 8 migration.  I haven't personally come to a conclusion on what the right approach is here, and I'm open to suggestions.  

I had taken the attitude that flyby formatting changes and copyright updates was OK, but I know they annoyed @etanshaul  recently and I see his point.  Probably something worth discussing.. possibly in the retrospective this afternoon?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
```diff
@@ -65,67 +60,47 @@ public AppEngineStandardLibraryPanel(boolean enabled) {
...
+      JCheckBox checkbox = (JCheckBox) event.getItem();
+
+      if (!checkbox.isSelected()) {
+        objectifyCheckBox.setSelected(false);
+        endpointsCheckBox.setSelected(false);
       }
     });
-    objectifyCheckBox.addItemListener(new AppEngineApiDependencyCheckboxListener());
-    endpointsCheckBox.addItemListener(new AppEngineApiDependencyCheckboxListener());
+    objectifyCheckBox.addItemListener(this::listenAppEngineApiDependencyCheckbox);
```
etanshaul: I like the refactoring, but not the name choice for the method. Maybe handleAppEngineApiDependencyCheckboxEvent?

joaoandremartins:
Yes, I understand where you guys are coming from. However, as I mentioned before, all these changes are direct translations to Java 8 idioms, without any actual behavioural changes, except for removing the mutable field from the enum.

I feel that your suggested changes highly affect productivity, and I meant to approach this problem some time in conversation.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
```diff
@@ -65,67 +60,47 @@ public AppEngineStandardLibraryPanel(boolean enabled) {
...
-                    ? AppEngineStandardMavenLibrary
-                    .getLibraryByDisplayName(libraryCheckbox.getText())
-                    : null;
-              }
-            }),
-        Predicates.notNull()
-    ));
+    return libraries.stream()
+        .filter(JCheckBox::isSelected)
+        .map(
```
etanshaul: what about the filtering out of the nulls?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
```diff
@@ -65,67 +60,47 @@ public AppEngineStandardLibraryPanel(boolean enabled) {
...
+            library ->
+                AppEngineStandardMavenLibrary.getLibraryByDisplayName(library.getText()).get())
```
etanshaul: again, isn't this not-null safe to do a get() on an Optional like this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
```diff
@@ -65,67 +60,47 @@ public AppEngineStandardLibraryPanel(boolean enabled) {
...
-    for (JCheckBox libraryCheckbox : libraries) {
-      libraryCheckbox.setSelected(availableLibraryNames.contains(libraryCheckbox.getText()));
-    }
+    Set<String> availableLibraryNames = mavenLibraries.stream()
+        .map(AppEngineStandardMavenLibrary::getDisplayName)
+        .collect(toSet());
+
+    libraries.stream()
+        .filter(library -> availableLibraryNames.contains(library.getText()))
+        .forEach(checkbox -> checkbox.setSelected(true));
```
etanshaul: I don't think this is the same behavior. Previously the logic was also unselecting libraries. Now, it looks like you are only ticking them.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacetConfiguration.java
chanseokoh: > does this still do that?

I don't think it will. `map()` in any functional language never removes an element, i.e., the resulting collection has the same size.

etanshaul: @chanseokoh - agreed, I have a bad habit of phrasing my observations as questions :)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
patflynn: I imagine that this is going to take some time to get used to, but for now I'm finding this style to be a significant step back in readability.

patflynn: @elharo any thoughts on this?  This seems related to our discussion this morning.

etanshaul: my opinion on this is that I like the style in general but a couple things could improve readability:

1) the nesting due to the lambda plus the write action etc. is hurting readability. If the contents of the ifPresent lamdba were extracted out to logically named methods I think it would help greatly.
2) Extract out AppEngineStandardMavenLibrary.getLibraryByMavenDisplayName(addedLibrary.getName()) to a named variable.

chanseokoh: > If the contents of the ifPresent lamdba were extracted out to logically named methods

If it becomes a method, I'm guessing just passing the method name would work too (http://stackoverflow.com/a/24228366/1701388).

etanshaul: true, but I didn't mean to imply that it necessarily should be split that way into 1 method

joaoandremartins: I think factoring out the lambda contents into a new method would actually decrease readability, since that lambda essentially contains the bulk of logic of the method. i.e., `afterLibraryAdded()` would be a one-liner with a method reference, which voids the utility of the reference method.

We could get rid of the ifPresent() by making sure that getLibraryByMavenDisplayName() can always return something, though I'm not familiar enough with the code to say if that can be done.

Finally, I think the following is actually pretty readable, but feel free to disagree.

screenshot from 2017-02-22 15 25 46

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
joaoandremartins: errorprone doesn't, but the IJ linter does, so I thought I'd remove it. I was wondering if this method had been called before actually using that argument, but then that call was deleted. If there aren't any plans to start using this with a variable isError, then I'd like to delete based on YAGNI.

`setStatusMessage()` is called with variable isError, which is why I didn't cut it there as well.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
etanshaul: its ok.

I think my suggestion (2) would help a bit.

Also (and this was there before any of your changes), you could update the for loop in the WriteAction to be java8'y

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacetConfiguration.java
joaoandremartins: What I think happens is map maps to Optional.empty(), in case of empty optionals returned by the mapping method (i.e., no nulls involved), and then later, the collector skips empty optionals.

The example in http://www.oracle.com/technetwork/articles/java/java8-optional-2175753.html seems to indicate that, and I think the Java 8 in Action book covers that, too.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
joaoandremartins: Done.

patflynn:
The fact that there are multiple points of debate about behavior preservation leads me to believe that this PR skews to the wrong side of risk/cost/benefit. 

Even if it were worth it the PR needs to be broken up, with the benefit that the safe and non-controversial changes will be approved very quickly.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineFacetConfiguration.java
chanseokoh: Below gives me `[Optional.empty, Optional.empty, Optional.empty]`.

```java
  public static void main(String[] args) {
    String[] arr = new String[]{ null, null, null };
    System.out.println(Arrays.stream(arr).map(e -> Optional.empty()).collect(Collectors.toList()));
  }
```

joaoandremartins: Yes. Turns out that the recommended way appears to be:

libraries.stream()
        .filter(JCheckBox::isSelected)
        .map(
            library ->
                AppEngineStandardMavenLibrary.getLibraryByDisplayName(library.getText()))
        .filter(Optional::isPresent)
        .map(Optional::get)
        .collect(toSet());

The Optional creator discusses it in http://stackoverflow.com/questions/22725537/using-java-8s-optional-with-streamflatmap. Java 9 will enable Optional.stream(), which makes flatMap here work as intended.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardFacetEditor.java
joaoandremartins: So, if I understand (2) correctly, I'd extract the whole method to a separate method, and then just call it from afterLibraryAdded()? If so, why not just call the whole thing there?

Yes, I like getting rid of the for loop. Also, to get rid of that WriteAction boilerplate, I'm considering using `ApplicationManager.getApplication().runWriteAction()`;. Should be equivalent?

etanshaul: For 2 thats not what I meant. I just meant something simply like this for readability:

```
AppEngineStandardMavenLibrary libary = 
  AppEngineStandardMavenLibrary.getLibraryByMavenDisplayName(addedLibrary.getName());

library.ifPresent(.. => ..
```

patflynn:
@joaoandremartins if you still want to make these changes let's move the conversation to multiple PRs addressing the separate concerns that are contained in this one.

I just want to reiterate that this PR should not be merged.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardLibraryPanel.java
joaoandremartins: Ah, good catch!

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineSupportProvider.java
```diff
@@ -211,42 +212,30 @@ static void removeMavenLibraries(final Set<AppEngineStandardMavenLibrary> librar
...
-              final Library library = libraryTable.getLibraryByName(displayName);
-              if (library != null) {
-                libraryTable.removeLibrary(library);
-
-                for (OrderEntry orderEntry : model.getOrderEntries()) {
-                  if (orderEntry.getPresentableName().equals(library.getName())) {
-                    model.removeOrderEntry(orderEntry);
-                  }
+    ApplicationManager.getApplication().invokeLater(() -> {
+      new WriteAction() {
```
joaoandremartins: I wonder if we could just use a single thread here. We're running invokeLater, which then runs writeAction. I wonder would simply running ApplicationManager.getApplication().runWriteAction() cut it here?

etanshaul: I don't know the answer off hand. runWriteAction does not invoke the Runnable asynchronously so it may be a behavior change. Tt would need to be thoroughly tested.

Regardless, along the lines of Patrick's comment, this change alone could be its own PR.

etanshaul:
You may be on it already, but I'm seeing files unrelated to this PR. I don't know why these merge issues keep happening assuming you are doing an actual merge, not a rebase? Maybe we can compare workflows since I never have to do this (I merge directly from IJ and do not rebase).

joaoandremartins:
Java 8 changes removed. PTAL.

@etanshaul that would be most useful! I usually run git rebase origin/master from the command line. This is what happens when I attempt a Push -> Merge from IntelliJ. It has never worked for me, in the half dozen of times I've tried it.

patflynn:
LGTM on the enum refactor. @etanshaul owns this code so I'll let him be the approver.

PR #1265, opened 2017-02-23
default compiler to error prone.

patflynn:
I know etan changed removed this recently.

Can you check that your compiler is set to error prone in settings?

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
+    <option name="DEFAULT_COMPILER" value="error-prone" />
     <wildcardResourcePatterns>
       <entry name="!?*.java" />
       <entry name="!?*.form" />


etanshaul:
It is not set to error prone:

image

I do still get error prone messages when building the project.

Are you not?

patflynn:
You need to set the internal compile to error prone to get the 'in-editor'
(red squiggly lines) error-prone errors.

Have you installed the plugin?

patflynn:
capture

patflynn:
BTW I think the plugin is installed by default on our work intellij

etanshaul:
I do not have the error prone plugin installed. 

And I did test out getting in-editor validation before and after removing the compiler setting and never got them, which is why I was comfortable with the change. I'll try it out again.

joaoandremartins:
I don't see that errorprone option in the Java compiler settings.
I also can't find the error prone plugin in the plugins list.

etanshaul:
oh..I guess the plugin would be required for this right? hence the vcs conflicts when the setting is there, but the plugin is not installed. this makes more sense. my only hesitation is that now developers are required to have the errorprone plugin to avoid these conflicts

etanshaul:
@joaoandremartins the error prone plugin is there if you go to "browse repositories..."

etanshaul:
On second thought, I think its reasonable as long as we document it for developers on the github page.

patflynn:
It's in our setup instructions:

etanshaul:
oh snap.

lgtm then.

PR #1273, opened 2017-02-27
Adds YAML parsing for app.yaml.

joaoandremartins:
There is a way to get rid of the unchecked cast `Map<String, String> yamlMap = (Map<String, String>) parseResult;`.

However, it requires a Java object that mirrors an app.yaml which, at this point, I think requires a bit more discussion about whether we want to have and maintain.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
@@ -41,8 +41,11 @@ import org.jetbrains.idea.maven.project.MavenProjectsManager;
 import org.jetbrains.plugins.gradle.util.GradleConstants;
 import org.yaml.snakeyaml.Yaml;
 
-import java.io.FileNotFoundException;
-import java.io.FileReader;
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.InvalidPathException;
+import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
@@ -229,7 +232,8 @@ public class DefaultAppEngineProjectService extends AppEngineProjectService {
       @NotNull String key) {
     Yaml yamlParser = new Yaml();
     try {
-      Object parseResult = yamlParser.load(new FileReader(appYamlPathString));
+      Object parseResult = yamlParser.load(
+          Files.newBufferedReader(Paths.get(appYamlPathString), Charset.defaultCharset()));
 
       if (!(parseResult instanceof Map)) {
         return Optional.empty();
@@ -240,7 +244,7 @@ public class DefaultAppEngineProjectService extends AppEngineProjectService {
       Map<String, String> yamlMap = (Map<String, String>) parseResult;
 
       return yamlMap.containsKey(key) ? Optional.of(yamlMap.get(key)) : Optional.empty();
-    } catch (FileNotFoundException fnfe) {
+    } catch (InvalidPathException | IOException ioe) {
       return Optional.empty();
     }
   }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
@@ -235,6 +235,8 @@ public class DefaultAppEngineProjectService extends AppEngineProjectService {
         return Optional.empty();
       }
 
+      // It's possible to get rid of this unchecked cast using a loadAs(file,
+      // AppEngineYamlWebApp.class) sort of approach.
       Map<String, String> yamlMap = (Map<String, String>) parseResult;
 
       return yamlMap.containsKey(key) ? Optional.of(yamlMap.get(key)) : Optional.empty();
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
@@ -45,6 +45,7 @@ import java.io.IOException;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.InvalidPathException;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.Map;
@@ -232,8 +233,13 @@ public class DefaultAppEngineProjectService extends AppEngineProjectService {
       @NotNull String key) {
     Yaml yamlParser = new Yaml();
     try {
-      Object parseResult = yamlParser.load(
-          Files.newBufferedReader(Paths.get(appYamlPathString), Charset.defaultCharset()));
+      Path appYamlPath = Paths.get(appYamlPathString);
+      if (!Files.isRegularFile(appYamlPath)) {
+        return Optional.empty();
+      }
+
+      Object parseResult =
+          yamlParser.load(Files.newBufferedReader(appYamlPath, Charset.defaultCharset()));
 
       if (!(parseResult instanceof Map)) {
         return Optional.empty();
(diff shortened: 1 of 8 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
```diff
@@ -228,19 +227,20 @@ public AppEngineStandardRuntime getAppEngineStandardDeclaredRuntime(
 
   private Optional<String> getValueFromAppYaml(@NotNull String appYamlPathString,
       @NotNull String key) {
-    Path appYamlPath;
+    Yaml yamlParser = new Yaml();
```
etanshaul: do you need to create a new parser each time here?

etanshaul:
thanks for doing another pr like this

etanshaul:
My preference would be for a "DOM"-like class for app.yaml similarly to how to we do for appengine-web.xml. Doing this almost always simplifies the code in my experience. But we can enhance it later if you're not convinced.

google-cloud-tools-plugin/testData/project/custom.yaml
```diff
@@ -0,0 +1,25 @@
+runtime: custom
+service: customService
+api_version: 1
+threadsafe: true
+
+handlers:
```
patflynn: don't include any of the below stuff in tests for now.  most of it is unsupported in app engine flex

google-cloud-tools-plugin/testData/descriptor/appengine-web_runtime-java8-serviceandmodule.xml
```diff
@@ -0,0 +1,24 @@
...
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<appengine-web-app xmlns="http://appengine.google.com/ns/1.0">
+    <application>app</application>
+    <version>1</version>
+    <runtime>java8</runtime>
+    <service>java8Service</service>
```
patflynn: an application shouldn't have both service and module

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
joaoandremartins: I don't know the API that well. More specifically, I don't know if the parser carries any state between successive calls. We're only calling it twice, so it should be fine.
If we end up going down the parsing into an AppEngineAppYamlFile object, we'll only need to call it once.

joaoandremartins:
Yes, that's my preference too, as I mentioned in the PR description.

Though that needs more discussion and work. For example, in my proof of concept, I only added a service and runtime fields, and the parser complained that the object didn't contain an api_version field, when I tried to parse the hello world app.yaml.

We would need to have an object that is the exact schema of an app.yaml file, and I don't know if we have that, or if we want to maintain it (e.g., whenever we add a new file to app.yaml and a user tries to use an app.yaml file with it, things break or malfunction).

google-cloud-tools-plugin/testData/descriptor/appengine-web_runtime-java8-serviceandmodule.xml
joaoandremartins: The schema allows it, though. I thought a unit test for that couldn't hurt, since that's the behaviour of the method -- service has precedence over module.

etanshaul:
Oh - I had just assumed that it would just ignore missing fields, like the DomElement interface does for xml.

google-cloud-tools-plugin/testData/descriptor/appengine-web_runtime-java8-serviceandmodule.xml
etanshaul: module isn't even mapped out though? (in AppEngineStandardWebApp)

patflynn:
I would hope that the API allows you to parse documents loosely and ignore
undefined fields. yaml tends to be very loose so I'm surprised it insists
on being strict.

Defining the app.yaml strictly is a non-starter since the App Engine team
haven't even done that themselves.

google-cloud-tools-plugin/testData/descriptor/appengine-web_runtime-java8-serviceandmodule.xml
joaoandremartins: Services used to be called Modules. The module tag is still supported.

https://cloud.google.com/appengine/docs/standard/java/config/appref

etanshaul: I meant in AppEngineStandardWebApp. I'm not hung up on this either way.

joaoandremartins:
This is the way you can parse documents in that way -- that I could find.

google-cloud-tools-plugin/testData/descriptor/appengine-web_runtime-java8-serviceandmodule.xml
joaoandremartins: Yes, I understood. This change is totally independent from AppEngineStandardWebApp, though.

etanshaul:
I'd be surprised if there wasn't a way, but approving on my end to not hold this up.

loosebazooka:
Just as a comment here (we don't do a hard schema based parsing for app.yaml in the common lib either) : https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/b285bfff9576da6aa35c0eb4b3530de5911d3861/src/main/java/com/google/cloud/tools/project/AppYaml.java

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
```diff
@@ -228,19 +227,20 @@ public AppEngineStandardRuntime getAppEngineStandardDeclaredRuntime(
...
-          .map(line -> line.substring(key.length()).trim())
-          .findFirst();
-    } catch (IOException ioe) {
+      if (!(parseResult instanceof Map)) {
+        return Optional.empty();
+      }
+
+      // It's possible to get rid of this unchecked cast using a loadAs(file,
+      // AppEngineYamlWebApp.class) sort of approach.
+      Map<String, String> yamlMap = (Map<String, String>) parseResult;
```
loosebazooka: I believe you can explicitly ignore the unchecked on just this conversion. 

```
@SuppressWarnings("unchecked")
```

joaoandremartins:
@loosebazooka you're doing the same as myself in this PR.

I wanted to avoid the unchecked cast and us loadAs(). But that requires that an object type that is basically the app.yaml schema exists.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
```diff
@@ -228,19 +230,21 @@ public AppEngineStandardRuntime getAppEngineStandardDeclaredRuntime(
...
+
+      return yamlMap.containsKey(key) ? Optional.of(yamlMap.get(key)) : Optional.empty();
+    } catch (InvalidPathException | IOException ioe) {
```
patflynn: don't we want to differentiate from a missing file or other IO exception from a missing yaml element?

joaoandremartins: This block isn't aimed at catching a missing yaml element. InvalidPathException is thrown on Windows when the Path contains weird characters. IOException is thrown when the file can't be read.

google-cloud-tools-plugin/testData/descriptor/appengine-web_runtime-java8-serviceandmodule.xml
patflynn: I'm kind of confused why XML is part of this PR at all

google-cloud-tools-plugin/testData/project/java.yaml
```diff
@@ -0,0 +1,15 @@
...
+- url: /.*
+  script: this field is required, but ignored
+  secure: always  # Require HTTPS
+
+runtime_config:  # Optional
+  jdk: openjdk8
+  server: jetty9
+
+manual_scaling:
+  instances: 1
```
patflynn: missing new line. just curious, is this the yaml generated by gen-config?  I'm asking because there's an arbitrary amount of detail/noise in this test data that we don't really care about.

joaoandremartins: This was copied from https://cloud.google.com/appengine/docs/flexible/java/quickstart

Using an elaborate app.yaml already let me find a bug. And I guess it can stay there for future tests?

joaoandremartins:
@etanshaul @patflynn had to push another commit. Let me know if you have any issues

google-cloud-tools-plugin/testData/descriptor/appengine-web_runtime-java8-serviceandmodule.xml
joaoandremartins: Oh, hah, that's because I realised I hadn't added unit tests for the getServiceNameFromAppEngineWebXml() method.

I guess it won't matter after I merge this with the flexible-facet branch.

PR #1280, opened 2017-03-01
Migrate legacy run configurations to work with this plugin

### LINKED ISSUE #1040: Investigate migrating run configs and libraries brought in by the old plugin
If you create an AE standard project using the old plugin, there will be a local dev server run config created for the user.

Then, when switching to our new tooling and opening the above project, the run config will be in an invalid state: it will have a legacy name (referencing the app engine sdk) and the sdk path will be empty. 

Similarly, the module's libaries will have a provided dependency on the App Engine SDK.

etanshaul:
fixes #1040 

with this, deployment run configs created by the old plugin will no longer be in a broken unrecoverable state. See the issue for more details and link to a doc with screenshots.

I'm looking to see if we want to do anything similar for the local dev run configs. But I may split that out into a separate PR if so.


google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -132,6 +132,7 @@ appengine.deployment.error.invalid.cloudsdk=Invalid Cloud SDK directory path.
 appengine.deployment.flexible.nogaesupport=No modules with Google App Engine support found. Please provide your configuration files
 appengine.deployment.error.cancelled=Deployment process was cancelled.
 appengine.facet.converter.description=Deprecated Google App Engine facets will be replaced with the latest version
+appengine.deploy.run.config.converter.description=Convert deprecated plugin deployment run configurations to work with the new plugin
```
joaoandremartins: "deprecated plugin deployment run configurations" is a mouthful. I wonder if there's a better way to phrase it.

"Convert deployment run configurations from deprecated plugin to be compatible with the new plugin"?

etanshaul: its not shown anywhere, but sure. done

etanshaul: actually it might be shown in the bottom left messaging in the ide when the converter window pops up. either way, I updated it

etanshaul:
theres an issue where this migration is not being executed if there are other migrations that are needed (e.g. the facet converter). investigating.

etanshaul:
fixed. the multiple converters we had for this project needed to share a single converter provider (defining multiple converter providers was invalid).

Please re-review.

etanshaul:
@joaoandremartins

PR #1282, opened 2017-03-02
Generate configuration files at add framework support time

### LINKED ISSUE #1276: Generate app.yaml and Dockerfile on add Flexible framework support
This will allow us to make the assumption that these files should always exist.

We could also remove the Generate buttons from the facet?

joaoandremartins:
Fixes #1276

On New Project and Add Framework Support, a new Runtime option appears with the options "java" and "custom" for App Engine Flexible.

screenshot from 2017-03-02 17 50 49

If files already exist at the default locations, the user is asked whether they want to keep them or generate new ones. If new ones are generated, the old ones are backed up (moved to filename.bak).

screenshot from 2017-03-02 17 51 11


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -118,7 +118,8 @@ protected void textChanged(DocumentEvent event) {
     );
 
     genAppYamlButton.addActionListener(new GenerateConfigActionListener(project, "app.yaml",
-        appEngineHelper::defaultAppYaml, appYaml, this::validateConfiguration));
+        () -> appEngineHelper.defaultAppYaml(FlexibleRuntime.java), appYaml,
```
joaoandremartins: A decision needs to be made here: do we want to add a runtime picker to the facet as well, or always generate runtime: java app.yamls?

joaoandremartins:
Went ahead and implemented the checkbox that generates app.yaml with runtime: java.

I left the support to generate both runtime java and custom app.yamls from the previous iteration, in case we want it in the future. But I can remove it if you guys think it clutters the PR.

joaoandremartins:
@patflynn I know were opinionated about this change. I ended up going for the sole "generate app.yaml" checkbox, that generates runtime: java app.yaml.
Would be good to get a thumbs up from you before @etanshaul reviews.

patflynn:
that functionality sounds right to me. :+1

etanshaul:
I didn't get the warning when an app.yaml already existed on the filesystem. Instead, it just created the new one and backed up the old one.

google-cloud-tools-plugin/resources/generation/src/appengine/flexible/java/app.yaml
```diff
@@ -0,0 +1,4 @@
+runtime: java
+env: flex
+env_variables:
+  'DBG_ENABLE': 'true'
```
etanshaul: @chanseokoh is this still needed for debugger?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
```diff
@@ -103,14 +113,40 @@ public void addSupport(@NotNull Module module, @NotNull ModifiableRootModel root
...
+        Path appYamlPath = Paths.get(
             appEngineProjectService.getDefaultAppYamlPath(contentRoots[0].getPath()));
+
+        // Allows suggesting app.yaml and Dockerfile locations in facet and deployment UIs.
+        facet.getConfiguration().setAppYamlPath(appYamlPath.toString());
         facet.getConfiguration().setDockerfilePath(
             appEngineProjectService.getDefaultDockerfilePath(contentRoots[0].getPath()));
+
+        if (generateConfigurationFilesCheckBox.isSelected()) {
+          if (Files.exists(appYamlPath)) {
```
etanshaul: for only 2 conditions like this my preference is combine with && instead of nested if's.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
```diff
@@ -103,14 +113,40 @@ public void addSupport(@NotNull Module module, @NotNull ModifiableRootModel root
...
+            try {
+              Files.move(appYamlPath,
+                  appYamlPath.getParent().resolve(
+                      appYamlPath.getFileName().toString() + ".bak"));
+            } catch (IOException e) {
+              // Do nothing for now.
```
etanshaul: shouldn't we abort the generation if backup fails?

joaoandremartins: I was thinking we should maybe log something, not sure about aborting. We'd need to make sure that add project is transactional, and undo everything if this fails here. Realistically, I also don't see this happening very often, if ever?

joaoandremartins:
I removed the warning so the process would be simpler.

Do we want to add it back?

etanshaul:
Oh. I did like the idea of warning the user instead of overwritting their file immediately.

etanshaul:
Also (and I don't know if theres a way around this), but when generating the configuration the IDE would pop up a message notifying about "external modifications" on the filesystem. I wonder if you integrated into the IDE's file mechanisms (PSI files etc) we can prevent this.

patflynn:
yea, I think I preferred your initial idea of prompting to overwrite in the
case the file already exists.

joaoandremartins:
PTAL, if an app.yaml file exists at the suggested location, a popup asks if the user wants to override it.
If yes, a backup file is created and a new app.yaml is generated. If a backup file with the same name (e.g., app.yaml.bak) already exists, another backup won't be created.
If no, then the old file is reused and no file is generated.

etanshaul:
Did you address the issue of the IDE warning the user of external file system modifications when the app.yaml is generated? I suspect this can be worked around.

joaoandremartins:
I'm not sure how or if that would work. I think the user being notified that the file has changed is the correct behaviour. That improvement sounds like something that could be solved later in another PR.

etanshaul:
I'm ok with doing it another PR. But I do think its the correct behavior: its the IDE thats guiding the creation of these new configuration file (after prompting the user). The experience, imo, should not be as if the file was created externally. You could check what other plugins do here, and how they achieve this (if they do it).

etanshaul:
also, did you test it out on Win? I remember we've had some issues in the past with file manipulation not working there. I didn't check myself if the calls you are using are system independent.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
etanshaul: agree that we should log something. but why not abort if the backup fails? 

I think we can still generate the project if the generation of configs fails, but it seems incorrect to just override the file if the backup fails.

google-cloud-tools-plugin/resources/generation/src/appengine/flexible/java/app.yaml
etanshaul: after offline discussion it seems that it _should_ work without it, but we'll leave it until we can verify.

patflynn:
I 'm thinking we shouldn't create a backup file. I've never seen intellj do
this. If they choose to overwrite and change their mind they can always
ctrl+z, or use IJ history to revert, or use their version control system.

patflynn:
Also ctrl+z is a good test scenario.  If it doesn't work you know you're
not creating the file correctly.

joaoandremartins:
The file is created by Add Framework Support, and I just tried to ctrl + z Add Framework Support and it doesn't work.

Unless you mean undoing only the file creation without undoing Add Framework Support?

joaoandremartins:
I was just able to add framework support, overwrite app.yaml and not get the warning. But I got it when I tried before, so not sure what's triggering it.

joaoandremartins:
OK, I changed it so a PsiFile is used inside a WriteCommandAction. I didn't get any more of those "file-system file changed" warnings and it is now ctrl + z-able.
Let me know if you have more feedback.

patflynn:
I defer to Etan, I haven't actually looked at this code, just lurking on
the comments.

joaoandremartins:
FYI there's an open issue re-writing files on Windows due to the separators I'm using, but creating works there.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
```diff
@@ -87,24 +110,81 @@ public boolean isSupportAlreadyAdded(@NotNull Module module,
...
+                      if (appYamlVirtualFile != null) {
+                        PsiFile appYamlPsiFile =
+                            PsiManager.getInstance(project).findFile(appYamlVirtualFile);
+                        if (appYamlPsiFile != null) {
+                          Document appYamlDocument =
+                              PsiDocumentManager.getInstance(project)
+                                  .getDocument(appYamlPsiFile);
+                          if (appYamlDocument != null) {
+                            try {
+                              appYamlDocument.setText(
```
etanshaul: any specific reason why you are doing it this way? i.e. copying the text over instead replacing the file? or am I misunderstanding whats going on here.

Also if possible to refactor all this logic a bit for readability that would be great. As is, its a little long with a lot of nesting.

joaoandremartins: I thought this was the recommended way of working with Documents. I didn't see any method to replace Documents -- did you have anything specific in mind?

Same for the nesting, I thought of using Optionals, I would only be using them for readability purposes, and some of you have mentioned you're not sure about using Optionals. I don't see a better way to check for nulls, though. Do you have anything in mind?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
```diff
@@ -87,24 +110,81 @@ public boolean isSupportAlreadyAdded(@NotNull Module module,
...
+                        }
+                      }
+                    }));
+          }
+        } else { // !Files.exists(appYamlPath)
+          // Just copy the file.
+          defaultAppYaml.ifPresent(
+              appYaml -> {
+                try {
+                  FileUtil.copy(appYaml.toFile(), appYamlPath.toFile());
```
etanshaul: Oh, so this part wasn't causing the "External filesystem modification" warning?

etanshaul: ah I see, this is an IntelliJ utility

joaoandremartins: Not when the file didn't exist before, but I'm glad you mention it because it's in my immediate plans to look into using Documents/PsiFiles to generate this file too.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
etanshaul: Regarding the Document question - can't you use some of the methods in FileUtil: I'm seeing things like `copyContent`, `createIfDoesntExist`. I'm thinking you may be able to simplify this logic to have a single method that writes (or overwrites the file) regardless of if it exists or not (of course keeping the confirmation dialog if it does exist).

joaoandremartins: I think the problem with FileUtil is that it performs the kind of low level file-system IO operation that we don't want. For example, it makes use of Files/Paths, not Documents/PsiFiles. I think going down that route would lead to similar issues with warnings, and isn't the best practice, but correct me if I'm wrong.

etanshaul: perhaps you are right - but how did that warning not show up when using FileUtil for the case where app.yaml didn't previously exist? Doesn't that count as external fileysystem modifications too?

joaoandremartins: I was able to trigger the changed file warning when I had the file open in the editor and the its file-system contents changed. I never got the warning when the file wasn't open in the editor. What I observed is that it takes a while for the file to show up in the Project view (i.e., it appears when synchronize is called, either explicitly or via cron).

Did you ever get the warning if the file wasn't open in the editor?

etanshaul: ah good point. should be easy to test with any file in the project, just don't have it open and edit it externally and see. I can try too.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
joaoandremartins: You're probably tired of hearing this :) but I would also suggest handling this together with #1298, since they should essentially be the same fix.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
joaoandremartins: I've been generating files with the generate button, for example, and never came across this issue before. I was not able to reproduce the warning issue, let me know if you are.

joaoandremartins:
The separators issue is solved, but afterwards I uncovered another weird IJ issue on Windows only -- adding Flexible support to a project that already contains an app.yaml, and then trying to manually edit the file, results in the following error:

Cannot Save Files
Cannot save _file name_
The file left unchanged.
Your changes were written to app.yaml__jb_tmp__.

screenshot from 2017-03-17 11 03 59

In fact, this happens for every file which is generated by the IDE (or, in the case of add flexible support, modified by the IDE with Document.setText()), e.g., for files and directories created by the generate button.

There are a couple of available workarounds. Restarting the IDE makes the file editable again, and disabling Safe Write from Settings -> Appearance and Behaviour -> System Settings also works.

http://stackoverflow.com/questions/19513779/what-is-this-bug-in-intellij-idea
https://youtrack.jetbrains.com/issue/IDEA-78480
http://stackoverflow.com/questions/23864827/jb-bak-and-jb-old-files-in-pycharm

Since this problem doesn't appear to be specific to this PR, I suggest filing an issue about this and not consider it a blocker for this PR.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
patflynn: for now always generate runtime java.

etanshaul:
Maybe you misstated, but why would that bug happen to a project that _already_ has an app.yaml? If the app.yaml already exists shouldn't the file just be left alone? Or do you mean when the user chooses to overwrite it?

joaoandremartins:
Yeah, this happens when the file is overwritten or created by IJ.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
etanshaul: Have you tried using the combination of FileUtil.createIfDoesntExist and FileUtil.copyContent? I'm wondering if you did this then you can unify these logic paths (the case was the file file already exists, and the case were it does not) and simplify this method a lot. Also, this _might_ sidestep the line separator issue.  As is, I find this method a little too messy.

etanshaul: Actually, this might re-introduce those external file system warnings like you mentioned. But I'd still like some refactoring here if using FileUtil doesn't work

joaoandremartins: Yeah, that's how I had this method in the beginning and you suggested this change.

I think the complexity (i.e., checking nulls) is required in this case.

etanshaul: Ok, then at least extracting things out (like the contents of the override logic lamda) into separate named methods would help with readability.

joaoandremartins: I don't think that really improves readability, on the contrary. I think it comes down to personal styles. But, to unblock things here, I'll do it.

etanshaul: Feel free to refactor it in a different way if you like, however as is the method is too long with hard to follow conditionals. I personally think any time you have add a comment next to an `else {` to indicate what the condition does then refactoring is in order.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
```diff
@@ -141,31 +141,10 @@ public static void addSupport(@NotNull AppEngineFlexibleFacet facet,
...
-                                      new String(Files.readAllBytes(appYaml),
-                                          Charset.defaultCharset())));
-                            } catch (IOException ioe) {
-                              logger.debug("Could not copy app.yaml text. " + ioe.getMessage());
-                            }
-                          }
-                        }
-                      }
-                    }));
+            defaultAppYaml.ifPresent(appYaml ->
```
joaoandremartins: @etanshaul PTAL. You might be right that it improves readability of the addSupport() method.

PR #1288, opened 2017-03-03
Fills the app.yaml text field at UI construction time

### LINKED ISSUE #1285: app.yaml override text box not being filled with suggestion for runtime: java, when override checkbox is pressed
screen shot 2017-03-03 at 11 35 54 am

Same thing doesn't happen for runtime: custom.

Ideally should be filled with the value from the facet.

joaoandremartins:
Fixes #1285 

When the user presses the app.yaml override button for the first time, it contains a suggestion.


patflynn:
I probably won't get a chance to look at PRs until later in the week.  If
you can get approval from Etan don't block on me.

PR #1289, opened 2017-03-06
Adds Flexible framework detection

### LINKED ISSUE #1277: Suggest adding Flexible framework support is app.yaml is found
Similarly to what happens with Standard and appengine-web.xml.

joaoandremartins:
Fixes #1277

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.intellij.appengine.facet.flexible;
+
+import com.intellij.facet.FacetType;
+import com.intellij.framework.detection.FacetBasedFrameworkDetector;
+import com.intellij.framework.detection.FileContentPattern;
+import com.intellij.openapi.fileTypes.FileType;
+import com.intellij.openapi.roots.ModifiableRootModel;
+import com.intellij.patterns.ElementPattern;
+import com.intellij.patterns.StandardPatterns;
+import com.intellij.util.indexing.FileContent;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.yaml.YAMLFileType;
+
+/** Detects App Engine Flexible framework in a project. */
+public class AppEngineFlexibleFrameworkDetector
+    extends FacetBasedFrameworkDetector<
+        AppEngineFlexibleFacet, AppEngineFlexibleFacetConfiguration> {
+
+  public AppEngineFlexibleFrameworkDetector() {
+    super("appengine-java-flexible");
+  }
+
+  @Override
+  public void setupFacet(@NotNull AppEngineFlexibleFacet facet, ModifiableRootModel model) {
+    new AppEngineFlexibleSupportProvider().setupFacet(facet, model);
+  }
+
+  @NotNull
+  @Override
+  public FacetType<AppEngineFlexibleFacet, AppEngineFlexibleFacetConfiguration> getFacetType() {
+    return FacetType.findInstance(AppEngineFlexibleFacetType.class);
+  }
+
+  @NotNull
+  @Override
+  public FileType getFileType() {
+    return YAMLFileType.YML;
+  }
+
+  @NotNull
+  @Override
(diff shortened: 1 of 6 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
```diff
@@ -84,7 +84,50 @@ public boolean isSupportAlreadyAdded(@NotNull Module module,
         || !facetsProvider.getFacetsByType(module, AppEngineStandardFacet.ID).isEmpty();
   }
 
-  static class AppEngineFlexibleSupportConfigurable extends FrameworkSupportInModuleConfigurable {
+  /** Initializes the Flexible facet by settings the default paths for app.yaml and Dockerfile
+   * and generating the necessary run configurations.
+   */
+  public void setupFacet(@NotNull AppEngineFlexibleFacet facet,
```
joaoandremartins: I just factored this into a new method. Code should be unchanged.

What do you think of turning this into a static method? So we don't have to instantiate AppEngineFlexibleSupportProvider if we want to call this from the outside.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
```diff
@@ -0,0 +1,64 @@
...
+  public FileType getFileType() {
+    return YAMLFileType.YML;
+  }
+
+  @NotNull
+  @Override
+  public ElementPattern<FileContent> createSuitableFilePattern() {
+    return StandardPatterns.or(
+        FileContentPattern.fileContent().withName("app.yaml"),
+        FileContentPattern.fileContent().withName("app.yml"));
```
etanshaul: does AE support ".yml" extension

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/AppEngineStandardFrameworkDetector.java
```diff
@@ -39,7 +39,7 @@
     FacetBasedFrameworkDetector<AppEngineStandardFacet, AppEngineStandardFacetConfiguration> {
 
   public AppEngineStandardFrameworkDetector() {
-    super("appengine-java");
+    super("appengine-java-standard");
```
etanshaul: Since you changed this here, do you also need to change it in `AppEngineStandardFrameworkType`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
```diff
@@ -0,0 +1,64 @@
...
+    extends FacetBasedFrameworkDetector<
+        AppEngineFlexibleFacet, AppEngineFlexibleFacetConfiguration> {
+
+  public AppEngineFlexibleFrameworkDetector() {
+    super("appengine-java-flexible");
+  }
+
+  @Override
+  public void setupFacet(@NotNull AppEngineFlexibleFacet facet, ModifiableRootModel model) {
+    new AppEngineFlexibleSupportProvider().setupFacet(facet, model);
```
etanshaul: will this also result in the run configurations being created? I noticed that in the equivalent method for Standard support we doing things differently.

joaoandremartins: Yes, it creates the run configurations.

What happens in Standard?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
joaoandremartins: Yes, I just tried. It doesn't support ".xpto", so it probably only accepts yaml and yml, so we should be fine.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/AppEngineStandardFrameworkDetector.java
joaoandremartins: This is the detectorId, it's independent from the framework type ID in AppEngineStandardFrameworkType.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
joaoandremartins: If you meant the usage of AppEngineWebIntegration, I failed to see the relevance of using a similar mechanism here and tried to simplify things a bit.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
etanshaul: was asking because I figured if it doesn't support .yml then we can leave it out of this code. not a big deal

joaoandremartins: I'm glad you asked because, the same way ".xpto" isn't supported, it could also be that ".yml" wouldn't. And an error would be thrown.

PR #1296, opened 2017-03-13
IDEA 2017.1 compatibility changes

joaoandremartins:
Fixes the received user report for

java.lang.NoSuchMethodError: com.intellij.ide.ui.laf.darcula.DarculaUIUtil.paintFocusRing(Ljava/awt/Graphics;IIII)V
...

and a couple other issues with javax.annotation.Nullable.


patflynn:
I guess these changes break 2016?  I don't think we can merge this until 2017 is released and we're committing our next release to it. 

Alternatively it may make sense to create an EAP branch and merge this in there. We could even push a release targeting the EAP only so that those users can use our plugin.

joaoandremartins:
Yeah, I won't merge til we move to IJ 2017 (i.e., while tests don't pass).

EAP branch sounds good, but we'd still need to release the plugin normally right? e.g., Alpha and default channels, etc.

patflynn:
You can apparently release multiple binaries with the same 'version' but
targeting different idea platform versions. I haven't tested it though.

PR #1299, opened 2017-03-16
Handle malformed app.yamls.

joaoandremartins:
Previously, whenever a malformed app.yaml was found, anything that required parsing it (deployment, facet, etc.) broke.


joaoandremartins:
@etanshaul can you please take a look?

etanshaul:
This is a stylistic thing, but I think the code will be more readable if instead of allowing the `ScannerException` to propagate out of app.yaml parsing class, you caught it there and instead rethrew something like an `AppYamlParseException`. This will make the code clearer imo.

etanshaul:
also, are there any unit tests you could add to the Editor test classes to cover this?

joaoandremartins:
Changed to a checked exception and added unit tests.
@etanshaul PTAL

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -186,9 +187,9 @@ public Project getProject() {
         }
         return Optional.of(createFlexRunner(loggingHandler, Paths.get(source.getFilePath()),
             deploymentConfiguration, deploy));
-      } catch (ScannerException se) {
+      } catch (MalformedYamlFile myf) {
```
etanshaul: You should add the suffix 'Exception' to the name. I'm pretty sure this is a universal convention.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
```diff
@@ -264,6 +265,8 @@ public AppEngineStandardRuntime getAppEngineStandardDeclaredRuntime(
       Map<String, String> yamlMap = (Map<String, String>) parseResult;
 
       return yamlMap.containsKey(key) ? Optional.of(yamlMap.get(key)) : Optional.empty();
+    } catch (ScannerException se) {
+      throw new MalformedYamlFile();
```
etanshaul: pass the exception object to this new exception you are throwing

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -167,23 +168,29 @@ public Project getProject() {
...
+          callback.errorOccurred(
+              GctBundle.getString("appengine.deployment.error.staging.dockerfile"));
+          return Optional.empty();
+        }
+        return Optional.of(createFlexRunner(loggingHandler, Paths.get(source.getFilePath()),
+            deploymentConfiguration, deploy));
+      } catch (MalformedYamlFileException myf) {
         callback.errorOccurred(
-            GctBundle.getString("appengine.deployment.error.staging.dockerfile"));
+            GctBundle.message("appengine.appyaml.malformed") + "\n" + myf.getMessage());
```
etanshaul: Where does this get shown? Just confirming because I see you added the exception message here too.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
```diff
@@ -213,14 +214,21 @@ public AppEngineStandardRuntime getAppEngineStandardDeclaredRuntime(
     }
   }
 
+  /**
+   * @throws ScannerException when an app.yaml isn't syntactically well formed
```
etanshaul: update javadoc to mention correct exception

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
```diff
@@ -213,14 +214,21 @@ public AppEngineStandardRuntime getAppEngineStandardDeclaredRuntime(
...
+   */
   @Override
-  public Optional<String> getServiceNameFromAppYaml(@NotNull String appYamlPathString) {
+  public Optional<String> getServiceNameFromAppYaml(@NotNull String appYamlPathString)
+      throws MalformedYamlFileException {
     return getValueFromAppYaml(appYamlPathString, SERVICE_TAG_NAME);
   }
 
+  /**
+   * @throws ScannerException when an app.yaml isn't syntactically well formed
```
etanshaul: update javadoc

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
```diff
@@ -229,8 +237,15 @@ public AppEngineStandardRuntime getAppEngineStandardDeclaredRuntime(
     }
   }
 
+  /**
+   * Returns the value of a key-value pair for a given {@code key}, on the file located at
+   * {@code appYamlPathString}.
+   * @return a String with the value, or an empty Optional if app.yaml isn't a regular file, or
+   * if there is any error getting the value
+   * @throws ScannerException when an app.yaml isn't syntactically well formed
```
etanshaul: update javadoc

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
joaoandremartins: This gets shown, for example, when a valid run configuration is saved, the app.yaml file is made invalid and, immediately after, there is an attempt to run that run configuration. It gets shown on the log window.

I thought it would be nice to be more helpful, even though this case should be pretty unusual. It would be better to show it, e.g., in the deployment or facet UI, but I thought the text got too big.

The exception message basically gives more detail around where the yaml file is invalid, which can be helpful in large files. But the same can be accomplished with a YAML file editor.

google-cloud-tools-plugin/ultimate/build.gradle
```diff
@@ -39,6 +39,7 @@ dependencies {
     compile ('com.google.cloud.tools:appengine-plugins-core:' + toolsLibVersion) {
         exclude group: 'com.google.guava', module: 'guava'
     }
+    compile 'org.yaml:snakeyaml:1.18'
```
etanshaul: is it correct to have this on the ultimate module's build file? what from ultimate needs this? I thought that since this is all flex related, and flex can be run on community, that you could add this to the "regular" build file only.

joaoandremartins: Take a look at AppEngineStandardSupportProviderTest. It requires javaee stuff, because it is testing bits of AppEngineStandardUltimateWebIntegration, and it extends JavaeeFrameworkSupportProviderTestCase too.

etanshaul: two things then: can you make it testCompile instead of compile?

or better yet, split up the test so that you separate the community from the ultimate parts? that way you don't have to add the dependency just for testing.

joaoandremartins: Made it testCompile since all the tests make use of some javaee component.

PR #1300, opened 2017-03-17
Add _config.yml to select theme to render Markdown files correctly

joaoandremartins:
Also fixes line spacing issues with list, and adds bold face to bullet topics so they're more easily readable.


joaoandremartins:
This is how it renders locally:

screenshot from 2017-03-17 12 36 30

PR #1303, opened 2017-03-21
Fixes IDEA 2017 EAP breakages

### LINKED ISSUE #1302: Need a way to support EAP IntelliJ releases
our plugin is broken on 2017.1 EAP right now due to a breaking API change: com.intellij.ide.ui.laf.darcula.DarculaUIUtil.paintFocusRing(Ljava/awt/Graphics;IIII)V

Ideally we'd be able to release versions of our plugin which support the EAP.  One potential way is to have an EAP branch with CI and do releases from it that target only the EAP version and up.

etanshaul:
sort of fixes #1302 ("sort of" because it doesn't provide an actual mechanism for supporting EAP releases, but rather makes the new 2017 EAP work with the same codebase).

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/ui/CustomizableComboBox.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/ui/CustomizableComboBox.java
@@ -219,7 +219,8 @@ public abstract class CustomizableComboBox extends JPanel {
     super.paint(graphics);
     if ((textField.isFocusOwner() || (getPopup() != null && getPopup().isPopupVisible()))) {
       if (isUsingDarculaUiFlavor()) {
-        DarculaUIUtil.paintFocusRing(graphics, 3, 3, getWidth() - 4, getHeight() - 4);
+        DarculaUIUtil.paintSearchFocusRing((Graphics2D) graphics,
+            new Rectangle(3, 3, getWidth() - 4, getHeight() - 4));
       }
     }
   }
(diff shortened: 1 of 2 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/ui/CustomizableComboBox.java
```diff
@@ -219,7 +219,8 @@ public void paint(Graphics graphics) {
     super.paint(graphics);
     if ((textField.isFocusOwner() || (getPopup() != null && getPopup().isPopupVisible()))) {
       if (isUsingDarculaUiFlavor()) {
-        DarculaUIUtil.paintFocusRing(graphics, 3, 3, getWidth() - 4, getHeight() - 4);
+        DarculaUIUtil.paintSearchFocusRing((Graphics2D) graphics,
```
etanshaul: from my testing, paintSearchFocusRing has the same UI outcome

etanshaul: casting Graphics to Graphics2d is apparently always safe: http://stackoverflow.com/questions/179415/java2d-is-it-always-safe-to-cast-graphics-into-graphics2d

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/SetupCloudRepositoryAction.java
```diff
@@ -503,12 +503,8 @@ public void run() {
       @NotNull Collection<VirtualFile> files) {
     final ChangeListManager changeListManager = ChangeListManager.getInstance(project);
     final FileIndexFacade fileIndex = FileIndexFacade.getInstance(project);
-    return Collections2.filter(files, new Predicate<VirtualFile>() {
-      @Override
-      public boolean apply(@javax.annotation.Nullable VirtualFile file) {
```
etanshaul: I updated this file too because the Nullable annotation was removed

joaoandremartins:
#1296

etanshaul:
oops. I missed that one. The difference is that this fix will work on both builds. Although I am doing some more testing to be sure.

etanshaul:
nevermind. paintSearchFocusRing creates an oval shape instead of a rectangle which doesn't look right. I'll close this or repurpose it to create a mechanism to support a separate EAP release

etanshaul:
closing. see #1296

PR #1309, opened 2017-03-22
Release v17.2.1 2017

etanshaul:
This is the release to support 2017 EAP and up. Is it based off of our last release from master with the additional commits needed to get things working for 2017.

Note the target of this PR is the 2017 branch, not master.

I had to set the version to 17.2.1_2017 (note the _2017) to disambiguate the tag names for the release process.


etanshaul:
looks like I missed some of the trickery in the Jenkins release job that was setting the version parameter based off of the tag name. So I can't set the version this way. Looking at alternatives.

loosebazooka:
I think this might be all Patrick, I'm not familiar with what's going on here. Willing to look over any gradle related stuff though.

etanshaul:
Yeah, this is all jenkins related. Not sure yet what is going on. my original comment about the version thing may have been wrong. I'm still looking

etanshaul:
Looks like this may be due to flakiness in `RestSignatureInspectionTest`. If so, I'll fix in a separate PR and re-release.

etanshaul:
closing to attempt a new release

patflynn:
Need to think about a good versioning scheme for this, but could we just label it v17.2.2 for now?  Since we only target 2017 and up it wouldn't trigger an update for people who havent upgraded to 2017.1 (which BTW is now out of EAP now)

PR #1314, opened 2017-03-23
Bumps appengine-plugins-core version to 0.2.9.

joaoandremartins:
0.2.9 provides devappserver1 integration, which will be required by #1315

### DIFF
--- a/google-cloud-tools-plugin/ultimate/build.gradle
+++ b/google-cloud-tools-plugin/ultimate/build.gradle
@@ -37,7 +37,7 @@ dependencies {
     compile files(project(':google-cloud-tools-plugin').sourceSets.main.output)
     compile files("${System.properties['java.home']}/../lib/tools.jar")
     compile ('com.google.cloud.tools:appengine-plugins-core:' + toolsLibVersion) {
-        exclude group: 'com.google.guava', module: 'guava'
+        exclude group: 'org.yaml', module: 'snakeyaml'
     }
     testCompile 'org.yaml:snakeyaml:1.18'
     testCompile project(path: ':google-account-plugin')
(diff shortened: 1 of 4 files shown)


patflynn:
could you add a PR description? I think I'm missing context.

joaoandremartins:
@patflynn updated

joaoandremartins:
Btw, still need review here. @etanshaul @patflynn

google-cloud-tools-plugin/ultimate/build.gradle
```diff
@@ -37,7 +37,7 @@ dependencies {
     compile files(project(':google-cloud-tools-plugin').sourceSets.main.output)
     compile files("${System.properties['java.home']}/../lib/tools.jar")
     compile ('com.google.cloud.tools:appengine-plugins-core:' + toolsLibVersion) {
-        exclude group: 'com.google.guava', module: 'guava'
```
etanshaul: why is the exclude needed in cloud-tools build.gradle, but not this one?

joaoandremartins: I think that guava is being transitively passed to the main module from another dependency.
Since it's not being passed to the ultimate module, and build completes successfully without it, I thought of getting rid of it.
I added the common library dependency to the ultimate module, and at the time just copied the common library dependency bit without checking if this exclusion was necessary.

patflynn: I don't think you want to remove this. Guava is provided by the IJ platform, and incompatibility between guava versions can cause errors at runtime.

joaoandremartins: If my interpretation of this build file is correct, then ultimate only depends on the main module source code, not its dependencies.

But I'll add Guava back to be safe.

joaoandremartins: I'm referring to this line,
compile files(project(':google-cloud-tools-plugin').sourceSets.main.output)
but I'm with you, for safety.

joaoandremartins:
@patflynn need your approval

PR #1315, opened 2017-03-23
Integrates Devappserver1

joaoandremartins:


etanshaul:
can we please get a screenshot of the updated UI?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineServerModel.java
```diff
@@ -233,6 +243,11 @@ public void setArtifact(@Nullable Artifact artifact) {
   }
 
   @Override
+  public List<File> getServices() {
+    return ImmutableList.of(Paths.get(artifactPointer.getArtifact().getOutputPath()).toFile());
```
etanshaul: This doesn't seem null-safe? for instance getArtifact() is nullable.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/executor/AppEngineStandardRunTask.java
```diff
@@ -44,9 +47,11 @@
    * @param runnerId typically "Run" or "Debug", to indicate type of local run. To be used in
    *     metrics
    */
-  public AppEngineStandardRunTask(@NotNull RunConfiguration runConfig, @Nullable String runnerId) {
+  public AppEngineStandardRunTask(@NotNull RunConfiguration runConfig, @Nullable String runnerId,
+      boolean useDevappserver2) {
     this.runConfig = runConfig;
     this.runnerId = runnerId;
+    this.useDevappserver2 = useDevappserver2;
```
etanshaul: nitpicking: 

camel case DevAppServer2

Also perhaps make it devAppServer2 instead of useDevAppServer2 so its more inline with pojo boolean naming conventions: e.g. the getter/setter will be isDevAppServer2/setDevAppServer2

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/executor/AppEngineStandardRunTask.java
```diff
@@ -62,7 +67,9 @@ public void execute(ProcessStartListener startListener) {
         .startListener(startListener)
         .build();
 
-    CloudSdkAppEngineDevServer devServer = new CloudSdkAppEngineDevServer(sdk);
+    AppEngineDevServer devServer = useDevappserver2
```
etanshaul: just curious: if the common lib uses the default name for devAppServer2, and explicitly calles out devAppServer1 (e.g. CloudSdkAppEngineDevServer1), why did you do it the other way around? does it have something to do with the default value we set?

joaoandremartins: Because devappserver2 is what we consider to be experimental. i.e., the default is devappserver1, 2 if the user overrides the option.

loosebazooka:
FYI, all dev app server1 changes aren't in 0.2.9 plugins core.

joaoandremartins:
@loosebazooka I'm waiting for your a-ok to merge this PR, after it's gone through code review.

loosebazooka:
I'm confused, I mean that https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/346 isn't included yet, and it's what we want to end up on before pushing dev appserver 1 changes. I presume this wont affect this PR too much, but it's possible that it could.

joaoandremartins:
@loosebazooka I gave it a test spin with devappserver1 and it seemed to work without a hitch. I kind of assumed you were referring to internal changes that weren't related to the work on our side, but we'll obviously adjust if we need to.

Actually, from https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/346, I don't see any changes in the API, but we might need to provide more parameters than what we otherwise were?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/executor/AppEngineStandardRunTask.java
joaoandremartins: I think isUseDevAppServer2 is appropriate here. devAppServer2 might be too generic for a member name.

etanshaul:
as discussed offline, we decided to update to devappserver1 without a toggleable option to switch to v2. There will also need to be some params that are currently exposed for v2 that will need to be removed for v1. I'll do this in a different PR as it has sufficiently diverged from this one.

joaoandremartins:
This PR becomes obsolete, then.
I'll go ahead and close it.

PR #1320, opened 2017-03-27
Allows specifying a JDK to CloudSdk objects

### LINKED ISSUE #1316: Intellij 2017 uses IDE bundled JRE for deployment
For some reason, its not using the project configured JDK. This causes staging errors:

"Unable to stage app: Cannot get the System Java Compiler. Please use a JDK, not a JRE."

joaoandremartins:
This change depends on an appengine-plugins-core 0.2.10 version, where specifying a JDK at CloudSdk construction time will be possible.

This fixes #1316

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
@@ -43,6 +43,8 @@ import com.google.gson.Gson;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.projectRoots.Sdk;
+import com.intellij.openapi.roots.ui.configuration.projectRoot.ProjectSdksModel;
 import com.intellij.openapi.ui.Messages;
 import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.vcs.impl.CancellableRunnable;
@@ -62,6 +64,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Map;
+import java.util.Optional;
 
 /**
  * A Cloud SDK (gcloud) based implementation of the {@link AppEngineHelper} interface.
@@ -217,7 +220,7 @@ public class CloudSdkAppEngineHelper implements AppEngineHelper {
     CloudToolsPluginInfoService pluginInfoService =
         ServiceManager.getService(CloudToolsPluginInfoService.class);
 
-    return new CloudSdk.Builder()
+    CloudSdk.Builder sdkBuilder = new CloudSdk.Builder()
         .sdkPath(CloudSdkService.getInstance().getSdkHomePath())
         .async(true)
         .addStdErrLineListener(logListener)
@@ -227,8 +230,11 @@ public class CloudSdkAppEngineHelper implements AppEngineHelper {
         .appCommandCredentialFile(credentialsPath.toFile())
         .appCommandMetricsEnvironment(pluginInfoService.getExternalPluginName())
         .appCommandMetricsEnvironmentVersion(pluginInfoService.getPluginVersion())
-        .appCommandOutputFormat("json")
-        .build();
+        .appCommandOutputFormat("json");
+
+    getProjectJavaSdk(project).ifPresent(sdkBuilder::javaSdkPath);
+
+    return sdkBuilder.build();
   }
 
   private static final String CLIENT_ID_LABEL = "client_id";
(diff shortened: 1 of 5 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -358,6 +364,18 @@ private Path getFileFromResourcePath(String resourcePath) {
     return appYaml;
   }
 
+  private Optional<Path> getProjectJavaSdk(Project project) {
+    ProjectSdksModel projectSdksModel = new ProjectSdksModel();
```
patflynn: is ProjectRootManager.getInstance(project).getProjectSdk() the better API to use here?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -227,8 +230,11 @@ public CloudSdk createSdk(
         .appCommandCredentialFile(credentialsPath.toFile())
         .appCommandMetricsEnvironment(pluginInfoService.getExternalPluginName())
         .appCommandMetricsEnvironmentVersion(pluginInfoService.getPluginVersion())
-        .appCommandOutputFormat("json")
-        .build();
+        .appCommandOutputFormat("json");
+
+    getProjectJavaSdk(project).ifPresent(sdkBuilder::javaSdkPath);
```
patflynn: if project SDK is null we should probably just fail.

joaoandremartins: Because the Cloud SDK might be created for running processes that don't use Java, I think it's better to allow passing here without a JDK.
We're failing in other places before creating the Cloud SDK if there is no JDK set. Like before running the local server.

patflynn: That argument is making me question whether the java sdk should be a cloud sdk configuration vs a part of the individual command arguments.

joaoandremartins: I don't have very strong feelings either way. I kind of think this solution is more elegant because there is only one place to specify the JDK, but let me know which one you prefer.

patflynn: @joaoandremartins Yeah this one is a bit marginal.  I think that the jdk location ideally would be surfaced as part of the API for calling the individual commands that require it, so that the project sdk is only read at the time the action is performed.  

However since the scope of a Cloud SDK is usually confined to running a single action, it's something we can fix later if you want.

joaoandremartins: From the point of view of dissociating the API from its internals, which do you think makes more sense -- specifying the JDK at Cloud SDK creation time or at command config time?

Thinking about a day we would bypass the Cloud SDK and have a pure Java API, having this dependency at the command level would stop making sense. From this point of view, the JDK location would be a dependency of the Cloud SDK and not of the command.
Also, the Cloud SDK path is another such example -- it's not used for devappserver commands and yet we still define it at the Cloud SDK level.

patflynn: I think command config is the right answer. 

If staging or the local devserver are commands that require a JDK and allow it to be configured by the user, that remains true regardless of whether it's packaged as a maven jar or bundled with the Cloud SDK.

joaoandremartins: In that case, I'll need to go ahead and re-write https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/352 to get the JDK from the command config.

joaoandremartins: This also doesn't fit well with the current CloudSdk architecture, where e.g. [CloudSdkDevAppEngineDevServer](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/master/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer.java#L105) translates a configuration into flags and then the [CloudSdk](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/master/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java#L251) object executes it.

I think the change from specifying jdkHome from CloudSdk config to command config needs a lot more thinking, and I'm not comfortable doing that in the time frame required to fix the Standard staging issue. I could go ahead and hack something to make the command config way fit it with the current CloudSdk model, but I don't think that would be the right approach.

joaoandremartins:
@patflynn PTAL

PR #1323, opened 2017-03-28
Reset module config

### LINKED ISSUE #1283: Removing the Flexible facet from the Module Settings dialog should hide module selector and Module Settings button
As is, if the facet is removed, the run config can't be persisted without the user specifying an app.yaml.

However, if the button is clicked, an exception is thrown.

But it'd be cooler if we hid the Module Settings dialog and the module selector, kinda like what happens when a run config is created without Flex facet in any module.

screen shot 2017-03-03 at 10 13 50 am

etanshaul:
fixes #1283

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
@@ -82,6 +82,8 @@ import javax.swing.event.TreeModelEvent;
 public class AppEngineFlexibleDeploymentEditor extends
     SettingsEditor<AppEngineDeploymentConfiguration> {
   private static final String DEFAULT_SERVICE = "default";
+  private Project project;
+  private DeploymentSource deploymentSource;
   private final AppEngineProjectService appEngineProjectService =
       AppEngineProjectService.getInstance();
 
@@ -106,12 +108,12 @@ public class AppEngineFlexibleDeploymentEditor extends
   private JButton moduleSettingsButton;
   private JCheckBox hiddenValidationTrigger;
   private JLabel noSupportedModulesWarning;
-  private DeploymentSource deploymentSource;
 
   private static final boolean PROMOTE_DEFAULT = false;
   private static final boolean STOP_PREVIOUS_VERSION_DEFAULT = false;
 
   public AppEngineFlexibleDeploymentEditor(Project project, AppEngineDeployable deploymentSource) {
+    this.project = project;
     this.deploymentSource = deploymentSource;
     version.getEmptyText().setText(GctBundle.getString("appengine.flex.version.placeholder.text"));
     appYamlTextField.addBrowseFolderListener(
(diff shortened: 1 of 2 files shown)


google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditorTest.java
```diff
@@ -389,13 +389,12 @@ public void testDockerfileOverride() {
     assertEquals("an override", editor.getDockerfileTextField().getText());
   }
 
-  public void testDockerfileStartsDisabledAndWithModuleSetting() {
+  public void testDockerfileStartsDisabled() {
```
etanshaul: I need to verify this. Specifically checking to see if this unit test was valid and why my changes changed the behavior. @joaoandremartins if anything jumps out at you let me know since you wrote this originally.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditorTest.java
```diff
@@ -389,13 +389,12 @@ public void testDockerfileOverride() {
...
     ApplicationManager.getApplication().runWriteAction(
         () -> ModuleManager.getInstance(getProject()).disposeModule(javaModule));
     editor = new AppEngineFlexibleDeploymentEditor(getProject(), deploymentSource);
     assertEquals(1, editor.getModulesWithFlexFacetComboBox().getItemCount());
     assertFalse(editor.getDockerfileTextField().isEnabled());
-    assertEquals(dockerfile.getPath(), editor.getDockerfileTextField().getText());
```
etanshaul: @joaoandremartins I removed this because with my changes, the text field was empty string so this assertion was failing.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -106,12 +108,12 @@
   private JButton moduleSettingsButton;
   private JCheckBox hiddenValidationTrigger;
   private JLabel noSupportedModulesWarning;
-  private DeploymentSource deploymentSource;
 
   private static final boolean PROMOTE_DEFAULT = false;
   private static final boolean STOP_PREVIOUS_VERSION_DEFAULT = false;
 
   public AppEngineFlexibleDeploymentEditor(Project project, AppEngineDeployable deploymentSource) {
+    this.project = project;
```
joaoandremartins: I'd rather we don't set project as a member of the class, if not strictly necessary. Pass it to the one method that requires it from the constructor, instead.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditorTest.java
joaoandremartins: Maybe the order counts.
I'm wondering if when the combo box model is changed, toggleDockerfileSection() is triggered by the item listener?

etanshaul:
ptal

PR #1324, opened 2017-03-29
Removes unnecessary references to AppEngineFlexibleDeploymentConfiguration

joaoandremartins:
I don't remember exactly why I went for this design, but I can clearly see now it wasn't a great decision.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
@@ -16,7 +16,6 @@
 
 package com.google.cloud.tools.intellij.appengine.facet.flexible;
 
-import com.google.cloud.tools.intellij.appengine.cloud.AppEngineDeploymentConfiguration;
 import com.google.cloud.tools.intellij.appengine.cloud.AppEngineHelper;
 import com.google.cloud.tools.intellij.appengine.cloud.CloudSdkAppEngineHelper;
 import com.google.cloud.tools.intellij.appengine.cloud.flexible.AppEngineFlexibleDeploymentArtifactType;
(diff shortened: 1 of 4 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -215,7 +211,7 @@ private boolean isValidConfigurationFile(String path) {
   /**
    * Validates the configuration and turns on/off any necessary warnings.
    */
-  private void validateConfiguration() {
+  private void toggleWarnings() {
```
patflynn: displayWarnings or showWarnings?  

'toggle' means to switch state, which I guess we don't always do here.

patflynn:
@joaoandremartins cool, but maybe include a bit more context as to why it was a bad decision in the PR description. :)

joaoandremartins:
Sorry about the lack of context.
I was passing a deployment configuration to the facet editor, which doesn't make a lot of sense.
I suspect I did this to enable the "Add Framework Support -> Runnable deployment run configuration out of the box" flow, but I don't think this even achieved that.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -78,13 +77,13 @@
   private JLabel noDockerfileLabel;
   private JLabel errorIcon;
   private JLabel errorMessage;
-  private AppEngineDeploymentConfiguration deploymentConfiguration;
+  private AppEngineFlexibleFacetConfiguration facetConfiguration;
   private AppEngineHelper appEngineHelper;
 
-  FlexibleFacetEditor(@Nullable AppEngineDeploymentConfiguration deploymentConfiguration,
+  FlexibleFacetEditor(AppEngineFlexibleFacetConfiguration facetConfiguration,
```
etanshaul: facetConfiguration @NotNull still?

joaoandremartins: Oh I'm never sure which one is (not) required.
Aren't fields non-null by default? Or is it the other way around?

etanshaul: I don't think either is default. Having these just aids in static analysis. I.e. the IDE will give useful warnings if you pass null to an @NotNull argument.

joaoandremartins:
@etanshaul @patflynn ptal

Issue #1298, opened 2017-03-16
Use PsiFiles on the facet generate button, if file already exists


joaoandremartins:
To get rid of the "file change on the file system" warning.

@etanshaul is this what you were working on in the Generate buttons?

joaoandremartins:
Besides the generate buttons, the app.yaml generation on Add Framework Support should also be changed to stop using FileUtil.

etanshaul:
closed as won't fix

Issue #1151, opened 2016-12-07
Generated Dockerfile doesn't work with Maven plugin


meltsufin:
The IntelliJ plugin generates a Dockerfile that contains the line referencing `target.war`.
```
...
ADD target.war $JETTY_BASE/webapps/root.war
```
This works for deploying to App Engine through IntelliJ because it is configured to build `target.war`.
However, it breaks `mvn appengine:deploy` command for users who then want to deploy the project using the appengine-maven-plugin becasue Maven build the application archive based on `{maven module name}.war` pattern.

etanshaul:
as discussed, this probably breaks Gradle too when used like this

patflynn:
@meltsufin good catch.  This is nasty and should be fixed as part of our flex re-work in January.

patflynn:
What about normalizing the name of the archive at staging?  conceivably we
could also let the user specify the final staged archive name/path.

etanshaul:
We do currently normalize the name of the archive for staging (we rename it to "target.war/jar"), if I understood you correctly. Doing this, however, lead to the bug described here.

Wouldn't specifying the name also lead to this same issue?

etanshaul:
Document a couple of offline discussion points with @patflynn:

- However we decide to support this, we need a migration strategy for users with exiting Dockerfiles with the normalized "target.jar/war" names.
- A potentially first step way to handle build system naming differences is, in the facet config, to give the user the option to change the target artifact name in the Dockerfile, defaulting it to a best guess.

chanseokoh:
> A potential first step to handle build system naming differences is, in the facet config, to give the user the option to change the target artifact name in the Dockerfile, defaulting it to a best guess.

Will you allow changing the target name later in the facet config or in the deploy config? If so, I think you need to detect if the Dockerfile is the one you actually generated and not a user-supplied one.

etanshaul:
no, only at generation time. at least at first.

nkibler:
I'm having trouble understanding the steps to repro this. Is it one of the following?
 
1. You start out with a project that you've added the App Engine facet to through IntelliJ, then you add the App Engine Maven plugin and try running `mvn appengine:deploy`
2. You start out with a Maven project that has the App Engine Maven plugin added, then you try adding the facet through IntelliJ (and generate a Dockerfile), then you try running `mvn appengine:deploy`

etanshaul:
[...]

PR #1327, opened 2017-03-30
Generate app.yaml for framework support from template

etanshaul:
In support of #1151 

Updates the generation of app.yaml when adding framework support to come from the IJ template mechanism.

2 future PRs are coming in order to close the bug:

1. Update all template generation (from facet too) for app.yaml and Dockerfile to use the same mechanism
2. Update the Dockerfile template to be dynamic so that the artifact target name gets set appropriately


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
```diff
@@ -165,6 +128,48 @@ public static void addSupport(@NotNull AppEngineFlexibleFacet facet,
...
+   * already exists it will not overwrite the file.
+   */
+  private static void generateAppYaml(@NotNull AppEngineFlexibleFacet facet,
+      @NotNull ModifiableRootModel rootModel) {
+    Project project = facet.getModule().getProject();
+    FileTemplate appYamlTemplate = FileTemplateManager.getInstance(project)
+        .getInternalTemplate(AppEngineTemplateGroupDescriptorFactory.APP_YAML_TEMPLATE);
+
+    try {
+      VirtualFile virtualFile = rootModel.getContentRoots()[0].findFileByRelativePath("src/main");
```
joaoandremartins: Maybe worth checking if content roots size > 0?

joaoandremartins: Maybe worth checking if content roots size > 0?

joaoandremartins: This check used to be made before when this method was inlined. But after extraction, might make sense to check it here, in case it's called by another method who doesn't perform that check.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
```diff
@@ -165,6 +128,48 @@ public static void addSupport(@NotNull AppEngineFlexibleFacet facet,
...
+
+      if (virtualFile != null) {
+        PsiDirectory directory = PsiManager.getInstance(project)
+            .findDirectory(virtualFile);
+
+        if (directory != null) {
+          PsiDirectory appEngineDirectory;
+          try {
+            directory.checkCreateSubdirectory(CONFIG_DIR_NAME);
```
joaoandremartins: Do you really need to create the directory like this? Isn't it automatically created by createFromTemplate?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
```diff
@@ -165,6 +128,48 @@ public static void addSupport(@NotNull AppEngineFlexibleFacet facet,
...
+          if (appEngineDirectory != null
+              && FileTemplateUtil.canCreateFromTemplate(
+                  new PsiDirectory[]{appEngineDirectory}, appYamlTemplate)) {
+            FileTemplateUtil.createFromTemplate(appYamlTemplate, "app.yaml",
+                FileTemplateManager.getInstance(project).getDefaultProperties(),
+                appEngineDirectory);
+          }
+        }
+      }
+    } catch (Exception e) {
```
joaoandremartins: Do we want to catch a more specific exception?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/AppEngineTemplateGroupDescriptorFactory.java
```diff
@@ -32,12 +33,19 @@
 
   @NonNls
   public static final String APP_ENGINE_WEB_XML_TEMPLATE = "GctAppEngineWeb.xml";
+  @NonNls
+  public static final String APP_YAML_TEMPLATE = "app.yaml";
```
joaoandremartins: Maybe not for this PR, but it would be nice to find a suitable central place to store this. I noticed we're referring to it in string literals in several different places.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
etanshaul: nope, createFromTemplate will fail if the directory doesn't exist

etanshaul: well to be more specific. createFromTemplate requires a directory. and to pass it a directory, that directory needs to exist

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
etanshaul: unfortunately this is what createFromTemplate throws

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/AppEngineTemplateGroupDescriptorFactory.java
etanshaul: yea, my next planned PR for this (templatizing the remaining instances of app.yaml / Dockerfile) may address this

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
etanshaul: I don't think contentRoots can be empty. Even if the project has no files, the root will be the project folder itself. Unless you can think of another way?

etanshaul: oh, I guess I can just pass contentRoots directly to generateAppYaml instead of rootModel. This way the size check already will have happened

joaoandremartins: I don't know. I think it's possible to delete every content root from a project/module. Can't hurt to guard?

joaoandremartins: Yeah, passing the module's content root with @NonNull could be good

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
```diff
@@ -165,6 +128,48 @@ public static void addSupport(@NotNull AppEngineFlexibleFacet facet,
     setupDeploymentRunConfiguration(facet.getModule());
   }
 
+  /**
+   * Generates an app.yaml configuration file in the src/main/appengine directory. If an app.yaml
+   * already exists it will not overwrite the file.
+   */
+  private static void generateAppYaml(@NotNull AppEngineFlexibleFacet facet,
+      @NotNull VirtualFile[] contentRoots) {
```
joaoandremartins: I think there's the same problem as before -- contentRoots can be empty.

Since you only need one, you could just pass a @NotNull VirtualFile moduleContentRoot

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
etanshaul: see 0c694dee0af4d0b4afbfbb41f8a2b1aa4a9f7fa7

I know its not guarded in the method itself. but its a private method, so its ok I think.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
etanshaul: ah, good point

etanshaul: done

google-cloud-tools-plugin/resources/fileTemplates/internal/app.yaml.ft
```diff
@@ -0,0 +1,4 @@
+runtime: java
+env: flex
+env_variables:
+  'DBG_ENABLE': 'true'
```
patflynn: this is on by default so you can remove it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
patflynn: do we expect this to happen? Should we just rethrow as a runtime exception?

etanshaul: on second, probably fail so we can know about it. If the directory was located / created and creating the file from the template fails then this really is unexpected. i'll refactor.

joaoandremartins: We can know about it without failing. Maybe a ping?

etanshaul: Sure, but I think the point is that is shouldn't fail. If it does then its really an exception.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
```diff
@@ -165,6 +128,50 @@ public static void addSupport(@NotNull AppEngineFlexibleFacet facet,
...
+        if (appEngineDirectory != null
+            && FileTemplateUtil.canCreateFromTemplate(
+            new PsiDirectory[]{appEngineDirectory}, appYamlTemplate)) {
+          try {
+            FileTemplateUtil.createFromTemplate(appYamlTemplate, "app.yaml",
+                FileTemplateManager.getInstance(project).getDefaultProperties(),
+                appEngineDirectory);
+          } catch (Exception e) {
+            // If the file already exists, this exception will be thrown by createFromTemplate
+            logger.warn("Failed to create app yaml from template. " + e.getMessage());
```
patflynn: should this be a dialog or something?  Is there anything we can use on the exception instance it to verify that it's caused by the file already existing?

etanshaul:
@patflynn @joaoandremartins aside from our discussion about the behavior of log.error, I ended up keeping it at log.debug:

That exception is thrown and caught for cases where the user tries to generate the app.yaml when one already exists so its not at all a case where we want things to blow up.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
etanshaul: I could add an explicit check to if the file exists prior to calling createFromTemplate.

But we then still have to problem of having to handle the exception thrown by createFromTemplate (and it does throw Exception, not a subclass) which could happen for various reasons.

etanshaul: I'm not sure what you mean by if it could be a dialog?

patflynn:
nm, I get it. But definitely add a comment explaining why you're catching
and swalling Exception.

joaoandremartins:
I guess this also closes #1298?

PR #1330, opened 2017-03-30
Fixes module combo box and button state

joaoandremartins:
Correctly disables the modules combo box and module settings button if both app.yaml and Dockerfile overrides are turned on, and enables them if otherwise.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
@@ -158,6 +158,9 @@ public class AppEngineFlexibleDeploymentEditor extends
     appYamlOverrideCheckBox.addActionListener(event -> {
       boolean isAppYamlOverrideSelected = ((JCheckBox) event.getSource()).isSelected();
       appYamlTextField.setVisible(isAppYamlOverrideSelected);
+
+      setModuleComboBoxAndButtonEnabled(
+          !(isAppYamlOverrideSelected && dockerfileOverrideCheckBox.isSelected()));
       toggleDockerfileSection();
     });
 
@@ -174,6 +177,9 @@ public class AppEngineFlexibleDeploymentEditor extends
             dockerfileOverride = dockerfileTextField.getText();
             dockerfileTextField.setText(getDockerfilePath());
           }
+
+          setModuleComboBoxAndButtonEnabled(
+              !(isDockerfileOverrideSelected && appYamlOverrideCheckBox.isSelected()));
         }
     );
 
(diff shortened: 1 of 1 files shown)


joaoandremartins:
Right now, only the module combo box gets disabled on resetForm and then it isn't possible to change its state without reopening the window.

etanshaul:
does this address a certain issue #?

joaoandremartins:
I never got to creating an issue for this bug.

etanshaul:
ok, no worries

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -158,6 +158,9 @@ protected void textChanged(DocumentEvent event) {
     appYamlOverrideCheckBox.addActionListener(event -> {
       boolean isAppYamlOverrideSelected = ((JCheckBox) event.getSource()).isSelected();
       appYamlTextField.setVisible(isAppYamlOverrideSelected);
+
+      setModuleComboBoxAndButtonEnabled(
```
etanshaul: setModuleControls would be good

joaoandremartins: Why do you suggest?
The method only sets the module bits to enabled/disabled, and I was basing on other methods like setEnabled().
I can replace Enabled by State?

etanshaul:
Its a more general issue, but its a little confusing that we don't disable the module controls when only checking "use a different app.yaml", but not Dockerfile. I say this because with the UI its not clear that the Dockerfile is also coming from the module. Probably not the right place to fix this though.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
etanshaul: yea I guess you're right. It felt a little weird with it being long. But I have no problem with setModuleControlsEnabled/State

joaoandremartins:
Not sure if I know what you mean.
We also change the module elements state when "use a different dockerfile" is checked.

etanshaul:
it appears you only change the state if _both_ are selected. I was referrring to the case where only "use another yaml" is selected.

etanshaul:
my point being is that from the UI it only clear that module settings applies to app.yaml, not to Dockerfile. but like I said, not the place to fix this.

joaoandremartins:
Aah yes. That's on purpose, because the Dockerfile is still being gotten from the module.
Though you're right, that relationship isn't downright visible.

PR #1332, opened 2017-03-31
17.2.2 release notes

joaoandremartins:


CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # Change Log
 All notable changes to this project will be documented in this file.
-## [unreleased]
```
joaoandremartins: Is this required to stay?

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
...
+## [17.2.2]
+
+This is another maintenance release to address the invalid Java Runtime Environment (JRE) on 
+standard staging ([#1316](https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1316))
```
etanshaul: Maybe just be a little more explicit: on staging an App Engine standard app

CHANGELOG.md
etanshaul: I'm not sure what its for. I don't see it show up in the rendered notes. Maybe @patflynn knows?

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # Change Log
 All notable changes to this project will be documented in this file.
-## [unreleased]
+## [17.2.2]
```
patflynn: shouldn't this be 17.2.2-2017?

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
...
+
+This is another maintenance release to address the invalid Java Runtime Environment (JRE) on 
+staging an App Engine standard app ([#1316](https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1316))
+issue:
+
+> Unable to stage app: Cannot get the System Java Compiler. Please use a JDK, not a JRE.
```
patflynn: This doesn't seem to follow our standard format for changelog.

CHANGELOG.md
etanshaul: we discussed offline that for the release notes we can just call it 17.2.2. you think we should be explicit that its 17.2.2_2017?

joaoandremartins: The "-2017" is just an artifact of us having two separate releases for each IJ product.
I wouldn't like to make it too public..

CHANGELOG.md
joaoandremartins: I wasn't aware we had a changelog format.
The reason I added this bit here was so the issue becomes more discoverable. I can remove it if you want.

etanshaul: If you look at our other sections, we have a "Fixed" header where we list fixes.It has the issue referenced at the end.

patflynn:
that's how it's displayed to users by jetbrains, so I'd prefer we kept the
full version string in sync.

patflynn:
It's also fine to call out this particular issue in the summary at the top,
but also list it in the list of fixed issues.

joaoandremartins:
Ah, hadn't understood what you meant.

google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -37,11 +37,24 @@
...
 
+      <h2>[17.2.2-2017]</h2>
+
+      <h3>Fixed</h3>
+
+      <ul>
+        <li>Invalid Java Runtime Environment (JRE) on staging an App Engine standard app
+        (<a href="https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1316">#1316</a>):
+
+        <blockquote>Unable to stage app: Cannot get the System Java Compiler. Please use a JDK, not a JRE.</blockquote>
```
etanshaul: will this render properly? not sure if you can put a blockquote in an li like this (block element, inside of an inline element)?

CHANGELOG.md
etanshaul: nm. li isn't inline. so its probably ok.

google-cloud-tools-plugin/resources/META-INF/plugin.xml
etanshaul: nm. li isn't inline. so its probably ok.

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # Change Log
 All notable changes to this project will be documented in this file.
-## [unreleased]
+## [17.2.2-2017]
```
etanshaul: technically its 17.2.2_2017

google-cloud-tools-plugin/resources/META-INF/plugin.xml
joaoandremartins: I was actually waiting to do an alpha release to check that.
This bit will only render if the plugin is in a repository.

etanshaul:
lgtm pending version display fix. I'm not sure about the 'unreleased' section - if its a convention or not. So I'm ok removing it since it doesn't break anything.

PR #1336, opened 2017-04-03
Release v17.2.2

joaoandremartins:


gradle.properties
```diff
@@ -18,5 +18,5 @@ ideaEdition = IC
 ideaVersion = 2017.1
 javaVersion = 1.8
 ijPluginRepoChannel = alpha
-version = 17.2.2_2017-SNAPSHOT
+version = 17.2.3
```
etanshaul: see https://github.com/GoogleCloudPlatform/google-cloud-intellij/commit/e2696fac201422162ffd32b4b76d33de566271ae

the version should probably be 17.2.3_2017-SNAPSHOT

etanshaul: looks like the tag is good, so you can just push a commit to this PR without bothering with another commit

joaoandremartins: Didn't we decide that for the next version, master would start targeting 2017?

If so, 17.2.3 (even 17.3) should be the next version.

etanshaul: yes but technically this is the 2017 branch so the next release from this branch would actually be _2017 if we happened to do it for whatever reason. If we target 2017 from master next, like you suggested, we'd be releasing from master.

joaoandremartins: Cool, I can change it after this PR if needed

patflynn:
I'm a bit confused what's going on.

We seem to have released two versions:
https://github.com/GoogleCloudPlatform/google-cloud-intellij/releases/

17.2.2 and 17.2.2_2017

17.2.2 is showing up as an update in my 2016.3 IJ release which I would not
expect since these are cut off the 2017 branch.

joaoandremartins:
@patflynn I released 17.2.2 off of master, by mistake. Is there a way to remove it from the alpha channel?

etanshaul:
what is 17.2.2 based off of? If its identical to the previous release in alpha e.g. 17.2.1 then its ok if its there.

etanshaul:
also, you can delete the release from our github page. also you probably want to remove the tag too.

patflynn:
Aha. No biggie. In that case like Etan mentioned we can just delete the
release.

etanshaul:
I went ahead and deleted the accidental alpha release from JB (they provide a ui for it) since it was based of the latest tip in master which we don't want out there.

PR #1340, opened 2017-04-03
Templatize app.yaml and Dockerfile generation

etanshaul:
More work in support of #1151 

Still a few TODOs in the code I need to address.

Dockerfile generation is not yet working:  I may need a new PR to go in before this to create new Dockerfile file types as it seems the generation from template may not work without it.

This PR updates all remaining code that creates app.yaml and Dockerfiles to generate them from templates. For #1151 there will be one more PR to dynamically generate the artifact name.


etanshaul:
please hold off on review until I can address the Dockerfile generation issue

etanshaul:
I went ahead and added the Docker file type extensions into this PR. Dockerfile generation from template now works. It still has a downside that the template name has to have a .docker (or any) extension for it to work. I get around this by just removing the extension (renaming the file) after its generated. I'm checking again to see if I can get around this.

etanshaul:
Looks like we have to have a default extension for the template system to work. So here's the way its working:

- From the Dockerfile templates, a file named `Dockerfile.docker` is generated. 
  - This extension is associated with the new `DockerFileType` created in this PR. 
- The file is renamed (using the Psi infrastructure) to `Dockerfile`.

patflynn:
That's a shame. Could you file a bug with JB? Bazel BUILD files will have
the same issue.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/file/DockerFileType.java
```diff
@@ -0,0 +1,63 @@
...
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import javax.swing.Icon;
+
+/**
+ * Representation of the Docker configuration file type.
+ */
+public class DockerFileType extends LanguageFileType {
```
joaoandremartins: I wonder if this is already available from something like https://plugins.jetbrains.com/plugin/7724-docker-integration

etanshaul: That plugin isn't open, so we can't depend on it in our plugin.

etanshaul:
https://youtrack.jetbrains.com/issue/IDEA-170960

etanshaul:
@patflynn this can be revisited.

I'm looking at testing

etanshaul:
With regards to the existing Docker plugin:

1) We can't use this because, as far as I can tell, IJ cannot generate templates for extensionless file types
   (see https://youtrack.jetbrains.com/issue/IDEA-170960)
2) This PR will not cause any interference with the Docker plugin (if it happens to be installed)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
```diff
@@ -375,37 +377,34 @@ private PsiElement generateFromTemplate(String templateName, String templateDire
...
-                  configTemplate,
-                  fileName,
-                  templateProperties,
-                  configDirectory);
-            } catch (Exception e) {
-              // If the file already exists, this exception will be thrown by createFromTemplate
-              // We want to silently skip the generation in this case.
-              logger.debug("Failed to create app yaml from template. " + e.getMessage());
-            }
+        if (configDirectory != null
```
patflynn: what if this if condition is false? Should that happen?

etanshaul: that shouldn't happen. I'll add some logging for that case.

etanshaul: to elaborate, that condition is checking 2 things:

1) that the directory exists: the code above ensures that the directory is created ore retrieved if it already existed -> this would only fail for some unexpected IO error
2) that the template can be created -> this is static and not based on any user action so should really never happen

It probably makes sense to raise an ide level error in this case

etanshaul: @patflynn ptal. Added some error logging.

etanshaul:
@patflynn 982fd20

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/file/DockerFileLanguage.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package com.google.cloud.tools.intellij.appengine.file;
+
+import com.intellij.lang.Language;
+
+/**
+ * Barebones implementation for the Docker language required by {@link DockerFileType}.
+ */
+public class DockerFileLanguage extends Language {
```
patflynn: I think it'd be nice to also explicitly call these classes DockerFileTemplateLanguage and DockerFileTemplateType

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/file/DockerFileTypeFactory.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.intellij.openapi.fileTypes.FileTypeConsumer;
+import com.intellij.openapi.fileTypes.FileTypeFactory;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Factory for producing the Docker file type, registered in plugin.xml.
+ */
+public class DockerFileTypeFactory extends FileTypeFactory {
```
patflynn: same here.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectService.java
```diff
@@ -38,7 +39,11 @@
 
   public enum FlexibleRuntime {
     custom,
```
patflynn: I think this violates java style. Enum elements should be all caps.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectService.java
```diff
@@ -38,7 +39,11 @@
...
-    java
+    java;
+
+    public String configLabel() {
```
patflynn: should this be toString as well?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectService.java
```diff
@@ -38,7 +39,11 @@
 
   public enum FlexibleRuntime {
```
patflynn: javadoc

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
```diff
@@ -302,6 +320,99 @@ String getServiceNameFromAppEngineWebXml(XmlFile appengineWebXml) {
...
+      }
+    });
+  }
+
+  @Nullable
+  private PsiElement generateFromTemplate(String templateName, String templateDirectoryName,
+      String fileName, Properties templateProperties, Module module) {
+    FileTemplate configTemplate = FileTemplateManager.getDefaultInstance().getInternalTemplate(
+        templateName);
+    VirtualFile virtualFile = ModuleRootManager.getInstance(module).getContentRoots()[0]
```
patflynn: this makes me think of that dialog IJ sometimes shows where there are multiple paths in different content roots that match when you try to move a class.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
```diff
@@ -302,6 +320,99 @@ String getServiceNameFromAppEngineWebXml(XmlFile appengineWebXml) {
...
+            // If the file already exists, this exception will be thrown by createFromTemplate
+            // We want to silently skip the generation in this case.
+            logger.debug("Failed to create app.yaml from template. " + e.getMessage());
+          }
+        } else {
+          logger.error("Failed to create app.yaml from template in directory: " + configDirectory);
+        }
+      }
+    } else {
+      logger.error("Failed to located source root for app.yaml creation.");
```
patflynn: this triggers the whole error reporting flow right?  We don't want to do that just because the user doesn't have a src/main.

etanshaul: yea. thats a valid point

etanshaul: I changed it to warn

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/AppEngineProjectService.java
etanshaul: ah right, I meant to fix that in this PR, but forgot. thanks

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
etanshaul: hmm, can you explain more? i.e. what the ramifications of that are? better multi-module support?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -147,7 +147,7 @@ public Project getProject() {
           callback.errorOccurred(GctBundle.getString("appengine.deployment.error.staging.yaml"));
           return Optional.empty();
         }
-        if (runtimeOptional.filter(runtime -> runtime == FlexibleRuntime.custom).isPresent()
```
patflynn: I believe errorprone should be flagging these. IIRC Google Java style is to use .equals even for string and enums.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
patflynn: I think it's even weirder than that as you could have multiple content roots that contained the same path in a given module.  Let's not try to fix it here.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
etanshaul: oddly its not flagging it and I do have the errorProne plugin installed

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/project/DefaultAppEngineProjectService.java
etanshaul: sure. but this reminded me that I want to do a couple quick tests for multimodule support before I merge.

etanshaul: the generation works with regards to multi module projects - this is because the generation itself is scoped to a specific module (that has the facet).

I did uncover another related bug - the inspection of runtime custom vs java isn't correctly scoped to a module and breaks for multimodule. Will file a separate bug and fix separately

patflynn:
:man_shrugging:

PR #1353, opened 2017-04-06
Adds a checkbox to deploy all configs for standard

### LINKED ISSUE #1346: Add option to deploy all configs with App Engine Standard deployment
When a user is deploying a standard app, they should be able to have us automatically deploy all the cron.xml, dos.xml, etc files during deployment.

patflynn:
Fixes #1346 

@etanshaul let me know if the approach makes sense and I'll add some tests.


google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -119,6 +119,7 @@ appengine.deployment.source.not.found.error=Deployment source not found: {0}.
 appengine.deployment.invalid.source.error=Invalid deployment source.
 appengine.deployment.status.deploying=Deploying to App Engine
 appengine.deployment.error.invalid.cloudsdk=Invalid Cloud SDK directory path.
+appengine.deployment.deployallconfigs.checkbox.label=Deploy all App Engine service configuration files (indexes, dos, dispatch, cron, queues)
```
etanshaul: How about something like:

"Include all App Engine service configuration files (..) with the deployment"

Just to make it totally clear that were're still deploying the artifact.

patflynn: @etanshaul that's a good point. My wording there is awful.

What about ''Update dos, dispatch, cron, task queues, and datastore indexes" ?

patflynn: screenshot from 2017-04-06 17 09 04

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploy.java
```diff
@@ -37,17 +38,28 @@
...
 
 /**
  * Deploys an application to App Engine.
  */
 public class AppEngineDeploy {
 
   private static final Logger logger = Logger.getInstance(AppEngineDeploy.class);
+  private static final String APPENGINE_GENERATED_DIR = "WEB-INF/appengine-generated/";
+  private static final ImmutableList<String> APPENGINE_CONFIG_FILE_PATHS = ImmutableList.of(
+      APPENGINE_GENERATED_DIR + "index.yaml", // Deploy indexes first since the app needs them.
```
etanshaul: since you mentioned index.yaml needs to be deployed first, are you sure the ordering here is honored downstream (e.g. in the core lib etc.)?

patflynn:
I can remove this. It appears as though gcloud picks its own ordering.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: yeah, a lot better this way

etanshaul:
yeah. this approach looks good.

patflynn:
added editor test.

etanshaul:
I'll cherry pick that commit from master into 17.2 and 2017 and see if that fixes it.

PR #1368, opened 2017-04-17
Don't recreate / overwrite flex deployment configs when migrating to new flex support

### LINKED ISSUE #1366: Flex migration - run config is lost
Start with a flex project on a build before the flex updates. Create a deployment run config with a bunch of settings.

Swap over to master, reopen the same project.

1) The run config is empty and no deployment sources are available yet -> this is OK because framework support has not been added yet. Add framework support (either manually, or via the detection if you had an app.yaml). 
_Update: on second thought, some guidance here would be nice - the user wouldn't necessarily know to add framework support, especially if there is no automatic detection. Perhaps some wording on the deployment config, or through documentation._

2) Re-open the run config. Its empty. 

Ideally, the run config that was created earlier with the old build should be restored.

etanshaul:
fixes #1366 

The problem was that the existing flex run configuration was getting wiped out. This update changes it so that a new flex deploy run config is create only if one does not already exist.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
@@ -204,7 +204,8 @@ public class AppEngineFlexibleSupportProvider extends FrameworkSupportInModulePr
   }
 
   private static boolean hasFlexibleDeploymentConfiguration(List<RunConfiguration> runConfigs) {
-    return runConfigs.stream()
+    return runConfigs
+        .stream()
         .anyMatch(runConfig -> {
           if (runConfig instanceof DeployToServerRunConfiguration) {
             DeploymentConfiguration deployConfig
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleSupportProvider.java
```diff
@@ -175,25 +179,48 @@ private static void generateAppYaml(@NotNull AppEngineFlexibleFacet facet,
...
+      }
+
+      runManager.addConfiguration(settings, false /* shared */);
     }
+  }
+
+  private static boolean hasFlexibleDeploymentConfiguration(List<RunConfiguration> runConfigs) {
+    return runConfigs
+        .stream()
+        .anyMatch(runConfig -> {
```
patflynn: I kind of like the flatMap(select()) style in the last answer here: http://stackoverflow.com/questions/35740543/java-8-stream-check-if-instanceof

It's a shame there isn't a utility method for this in the JDK.

etanshaul: I can actually clean this up a bit though using filters

etanshaul:
@patflynn I may have went overboard, but see 200eeb5

I could have gone even further with another map call before the anyMatch, but held myself back. All this may be less readable though.

PR #1374, opened 2017-04-20
Don't exclude flex from deploying assets (cron, dos, indexes etc.)

### LINKED ISSUE #1373: App Engine asset deployment (dos, cron etc.) should be available for flex as well
we enabled it for standard, but we should allow it for flex as well

etanshaul:
Fixes #1373

Testing this out still


patflynn:
it's unfortunately more complicated than this.  

App Engine Flex apps don't generate yaml files into WEB-INF/appengine-generated.   Typically the files will be in the source as yamls.

etanshaul:
on hold - see the issue for details

etanshaul:
obsolete - see #1497

PR #1375, opened 2017-04-20
Don't do whitelist inspections for java8 runtime standard apps

etanshaul:
fixes #1368


etanshaul:
Without `<runtime>java8</runtime>`:

image

With `<runtime>java8</runtime>`:

image

image

Issue #1334, opened 2017-03-31
AE Flex Facet UI: Should we hide Dockerfile config when runtime  != custom


patflynn:
I may have forgotten or missed a conversation about this, but I initially thought we wanted to hide the 'Dockerfile' configuration in the Flex facet unless the runtime was set to 'custom'.  Flex tries to hard to hide docker from the user for non-custom scenarios, so ideally the user wouldn't see this field unless they had explicitly decided to go down that path.

Another thing that looks like a bug is that the Dockerfile shows up as configured except greyed out when the app.yaml is set to custom and the project has a Dockefile under ../src/main/docker.  This is confusing because it could give the impression that the Dockerfile is being used, but just can't be changed.  Hiding the configuration completely would also fix this bug.

Doing something similar to what's proposed for the deployment configuration here: #1333  could be nice. So that the runtime is displayed, and if it's custom, the dockerfile selection UI is displayed as a child of that UI element.

joaoandremartins:
The decision was to hide Dockerfile on runtime != custom from the deployment UI, but not from the facet, so as to decrease the amount of "magic" to the users. i.e., why was Dockerfile there before and why is not there now? What do I need to do to get it back there?

Why do you say that Flex "tries hard to hide docker" for non-custom scenarios?

Dockerfile will always be set to something, even if grayed out, unless the user sets it to an empty value. The caption underneath the Dockerfile textbox, when it's disabled, attempts to explain the state of things:

![screenshot from 2017-03-31 14 27 05](https://cloud.githubusercontent.com/assets/3290856/24563927/2d9689c2-161e-11e7-90c4-20b8653c9346.png)

There might be something I'm missing, but #1333 seems to only focus on the deployment UI.

All in all, I still think Dockerfile being always there in facet for the purposes of reducing mysticism is still a valid decision.

patflynn:
On Fri, Mar 31, 2017 at 2:28 PM, João Martins <notifications@github.com>
wrote:

> The decision was to hide Dockerfile on runtime != custom from the
> deployment UI, but not from the facet, so as to decrease the amount of
> "magic" to the users. i.e., why was Dockerfile there before and why is not
> there now? What do I need to do to get it back there?
>
> Why do you say that Flex "tries hard to hide docker" for non-custom
> scenarios?
>
App Engine Flex does not require docker or Dockerfiles and emphasizes their
standard runtimes.  It's weird and can be confusing to have a configuration
[...]

PR #1376, opened 2017-04-21
Hide docker file section when runtime != custom in flex facet project structure form

nbashirbello:
Fixes part of #1334 that references hiding the docker file section when runtime != custom.

Not does address location to the docker file section as discussed in #1334.


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -234,15 +234,20 @@ private void showWarnings() {
 
   private void toggleDockerfileSection() {
     boolean enabled = false;
+    boolean visible = true;
```
etanshaul: do we even need this variable anymore? it looks like it just gets assigned to enabled

nbashirbello: We need it when a MalformedYamlFileException is thrown.

etanshaul: not sure I follow; looks like when the exception is thrown its just ignored. I think its valid to say that if the parsing fails, then we should hide it entirely (not have it visible, but disabled)

patflynn:
@nbashirbello thanks nabila!

When you say 'fixes part of..' could you see which part? :)

Also for changes that affect the UI before and after screenshots are super helpful.

nbashirbello:
@patflynn 

I have updated the comments of the PR to reflect your suggestions. I will include screenshots when I  adjust the UI.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -234,16 +234,15 @@ private void showWarnings() {
...
-      enabled = isRuntimeCustom();
-      // Shows no Dockerfile label if runtime is Java and appYaml is valid.
-      noDockerfileLabel.setVisible(!enabled && isValidConfigurationFile(appYaml.getText()));
+      visible = isRuntimeCustom();
     } catch (MalformedYamlFileException myf) {
-      noDockerfileLabel.setVisible(false);
+      // do nothing
     }
-    dockerfile.setEnabled(enabled);
-    genDockerfileButton.setEnabled(enabled);
```
etanshaul: are these enabled by default always? or should we explicitly setEnabled still?

nbashirbello: Yes, when you add a element to a form, it is enabled by default

PR #1395, opened 2017-05-02
Add kokoro to 17.2 branch

patflynn:


kokoro/windows/presubmit.cfg
```diff
@@ -0,0 +1,2 @@
+# Location of the presubmit build bash script in repository.
```
akerekes: `bash` is for ubuntu builds

PR #1396, opened 2017-05-02
Don't overwrite existing app engine run configurations

### LINKED ISSUE #1272: Adding GAE support to new module resets existing GAE run configurations
This was tested with deployment run configs, but it should also happen with local run and debugger for Standard.

To reproduce, just add Standard support to a module, fill in a run config and add another one. The existing run config will be reset.

When the Flex facet lands, this will also happen when Flex support is added, then removed, and added again.

etanshaul:
fixes #1272


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/AppEngineStandardWebIntegration.java
```diff
@@ -111,34 +111,51 @@ public void registerFrameworkInModel(FrameworkSupportModel model,
...
-        configurationType.getDisplayName(), factory);
-    DeployToServerRunConfiguration<?, ?> runConfiguration
-        = (DeployToServerRunConfiguration<?, ?>)settings.getConfiguration();
-
-    if (server != null) {
-      runConfiguration.setServerName(server.getName());
-    }
 
-    runManager.addConfiguration(settings, false /*isShared*/);
+    boolean hasExistingDeployConfiguration = runManager.getAllConfigurationsList()
```
patflynn: could we use getClonfigurationList(ConfigurationType) here?

etanshaul: Yeah probably. I missed that one. I'll update.

etanshaul: I think its cleaner to leave it as is. I _could_ do something like this:

```
    RunManagerImpl runManager = RunManagerImpl.getInstanceImpl(module.getProject());
    ConfigurationType configType = runManager.getConfigurationType(AppEngineCloudType.ID);

    boolean hasExistingDeployConfiguration = configType == null ||
        runManager.getConfigurationsList(configType)
            .stream()
            .map(config -> ((DeployToServerRunConfiguration) config).getServerType())
            .anyMatch(serverType -> serverType instanceof AppEngineCloudType);
```
But since configType is nullable it makes the code a little uglier.

PR #1398, opened 2017-05-02
Add kokoro release config and script

patflynn:


kokoro/ubuntu/release.sh
```diff
@@ -0,0 +1,80 @@
...
+     "$(pwd)" && exit 1;
+export GIT_TAG_NAME=$(git describe)
+[ -z "$GIT_TAG_NAME" ] && \
+    echo "ERROR: 'git describe' failed. We're probably not in a git workspace."\
+     && exit 1;
+export TAG_VERSION=${GIT_TAG_NAME:1}
+
+if [ "$SOURCE_VERSION" != "$TAG_VERSION" ]
+    then
+        echo "ERROR: The version ($SOURCE_VERSION) configured in gradle.properties does not match"\
```
etanshaul: is this correct? won't the version in gradle.properties be a SNAPSHOT version (whereas the release tag won't)?

patflynn:
The version in the gradle.properties at the tag should be a non snapshot version

PR #1401, opened 2017-05-03
Add GCS based build dependency cache to make builds speed and reliable.

patflynn:
It will show up in the kokoro build workspace under src/gfile

### DIFF
--- a/kokoro/common.cfg
+++ b/kokoro/common.cfg
@@ -1,0 +1,3 @@
+
+# Include GCS cache of build dependencies
+gfile_resources: "/bigstore/cloud-tools-for-java-team-kokoro-build-cache"


akerekes:
How do you want to populate it?
Could you add the changes to the build file that configures the cache in this PR?

kokoro/common.cfg
```diff
@@ -0,0 +1,3 @@
+
+# Include GCS cache of build dependencies
+gfile_resources: "/bigstore/cloud-tools-for-java-team-kokoro-build-cache"
```
etanshaul: probably should add a line break

patflynn:
The cache is manually populated.

PR #1403, opened 2017-05-03
Fix race condition in unit test causing flakiness

etanshaul:
This specific test was flaky because the `CompletableFuture#supplyAsync` lambda returned by the mock might not complete before the rest of the test runs (causing the expected exception not to occur). 

This fix refactors the logic to force the CompletableFuture to be done and finish with the desired exception via `completeExceptionally`.


etanshaul:
for some unknown reason this test is still failing on kokoro. once #1402 is merged in we'll be able to see whats happening

etanshaul:
I think I fixed it

PR #1410, opened 2017-05-04
Migrate legacy run deploy configurations to work with this plugin (#1280)

etanshaul:
(cherry picked from master commit f157692)


etanshaul:
will cherry pick to 2017 as well once this is merged

PR #1418, opened 2017-05-04
Refactors logging around possible flaky test

etanshaul:
This test was relying on @BeforeClass to properly inject the TestLoggerFactory so that the log.error statement (in the constructor called by the test) does not throw an assertion error - depending on the type of factory, it may or may not throw an error - TestLoggerFactory does not. 

Regardless, we are occasionally seeing test failures in jenkins still, so this mechanism may be flaky.

This refactors it by not doing a log.error in the constructor called by the test, and getting rid of the factory injection altogether.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugProcessHandler.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugProcessHandler.java
@@ -23,7 +23,8 @@ import com.google.common.annotations.VisibleForTesting;
 
 import com.intellij.execution.process.ProcessHandler;
 import com.intellij.openapi.diagnostic.Logger;
-import com.sun.istack.NotNull;
+
+import org.jetbrains.annotations.NotNull;
 
 import java.io.OutputStream;
 
@@ -35,7 +36,7 @@ public class CloudDebugProcessHandler extends ProcessHandler {
 
   private static final Logger LOG = Logger.getInstance(CloudDebugProcessHandler.class);
 
-  private final CloudDebugProcess process;
+  private CloudDebugProcess process;
 
   /**
    * Initialize the cloud debug process handler.
@@ -63,7 +64,6 @@ public class CloudDebugProcessHandler extends ProcessHandler {
 
   @VisibleForTesting
   CloudDebugProcessHandler() {
-    this.process = null;
   }
 
   @Override
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugProcessHandler.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugProcessHandler.java
@@ -19,7 +19,6 @@ package com.google.cloud.tools.intellij.debugger;
 import com.google.cloud.tools.intellij.debugger.ui.LogoutDebugProcessDetacher;
 import com.google.cloud.tools.intellij.login.CredentialedUser;
 import com.google.cloud.tools.intellij.login.Services;
-import com.google.common.annotations.VisibleForTesting;
 
 import com.intellij.execution.process.ProcessHandler;
 import com.intellij.openapi.diagnostic.Logger;
@@ -36,7 +35,7 @@ public class CloudDebugProcessHandler extends ProcessHandler {
 
   private static final Logger LOG = Logger.getInstance(CloudDebugProcessHandler.class);
 
-  private CloudDebugProcess process;
+  private final CloudDebugProcess process;
 
   /**
    * Initialize the cloud debug process handler.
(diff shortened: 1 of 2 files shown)


etanshaul:
my theory is that the flakiness is due to some classloader ordering variability: if the test class was loaded before the framework code that injects the factory, then the old workaround (of manually setting the TestLogger) would not do anything and the test would fail.

patflynn:
@etanshaul can you include the test cases and failures that you're trying to fix either in the PR description or as an issue linked to it?

thx

etanshaul:
Sure. it's actually all the tests in CloudDebugProcessHandler that seem to fail nondeterministically. This PR updates this test exclusively.

patflynn:
Yeah I'd just like to have the stacktraces and test failure output associated to the PR.

etanshaul:
Sure. I'll add them when

etanshaul:
Sounds good I'll add that

etanshaul:
I can't actually locate the stacktrace for this at the moment (since its flaky only on Jenkins and I can't find the PR that reproduces this). However, it happens on the following line:

CloudDebugProcessHandlerTest:42

This line initializes the CloudDebugProcessHander which then throws a RuntimeException and ultimately an AssertionError for the Unit test. Looking at the constructor (prior to my changes), you can see the log.error that gets called all the time from the unit tests. This is what led me to believe that setting the proper logging factory is flaky.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugProcessHandler.java
```diff
@@ -55,10 +57,15 @@ public CloudDebugProcessHandler(CloudDebugProcess process) {
...
       }
     } else {
-      LOG.error("process or process state is null. This should only happen in tests");
+      LOG.error("process state is null.");
     }
   }
 
+  @VisibleForTesting
+  CloudDebugProcessHandler() {
+    this.process = null;
```
patflynn: Why this redundant constructor?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugProcessHandler.java
```diff
@@ -38,9 +40,9 @@
   /**
    * Initialize the cloud debug process handler.
    */
-  public CloudDebugProcessHandler(CloudDebugProcess process) {
+  public CloudDebugProcessHandler(@NotNull CloudDebugProcess process) {
     this.process = process;
```
patflynn: what about doing something like Preconditions.checkNotNull(process.getProcessState()) right here?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugProcessHandler.java
etanshaul: the way the test is currently constructed it requires the initialization of this class without setting the process

patflynn: ok.  What about leaving out the redundant initializer. We don't typically explicitly set fields to null.

etanshaul: sure, it was because the field is final, but I can remove that if you think its less ugly

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugProcessHandler.java
```diff
@@ -19,9 +19,11 @@
 import com.google.cloud.tools.intellij.debugger.ui.LogoutDebugProcessDetacher;
 import com.google.cloud.tools.intellij.login.CredentialedUser;
 import com.google.cloud.tools.intellij.login.Services;
+import com.google.common.annotations.VisibleForTesting;
 
 import com.intellij.execution.process.ProcessHandler;
 import com.intellij.openapi.diagnostic.Logger;
+import com.sun.istack.NotNull;
```
patflynn: wrong annotation :)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugProcessHandler.java
patflynn: oh that's a good point. this seems weird. What about simply not supporting this style of test?  Can we just require that the test provide a mock process?

etanshaul:
fixed the wrong annotation (good catch), and removed the redundant setting of the field to null (making the field non-final). But I think your suggestion is best - updating the test to pass in a mock (or something similar). I'll update

PR #1429, opened 2017-05-08
Don't throw a runtime exception for missing artifact in deploy config

### LINKED ISSUE #1428: Deployment run config migrated from old plugin throws exception
When trying to run the migrated deployment configuration out of the box (i.e. just clicking play before editing it), throws an IDE exception:

image

Editing the run config and setting all the fields fixes the problem (without having to delete and recreate the run config), however, it would be nice to have it working out of the box (either working right away, or showing a proper error message - the latter is probably more likely since the user will have to enter configuration that was missing in the old plugin).

etanshaul:
fixes #1428 

We shouldn't be throwing an exception here as this occurs when migrating from the old plugin. Instead the user is now shown this in the console output for the deployment:

image

Upon editing the deploy config and filling in all the details, the deploy works as expected.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
@@ -124,7 +124,6 @@ public class CloudSdkAppEngineHelper implements AppEngineHelper {
 
     if (!(source instanceof AppEngineDeployable)) {
       callback.errorOccurred(GctBundle.message("appengine.deployment.invalid.source.error"));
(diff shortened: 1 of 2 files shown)


patflynn:
weird. I thought I cmmented on this PR but I don't see it.

In case I'm going crazy:
would it be possible to make the run configuration's deployment source a valid type during the migration process?

etanshaul:
hmm. interesting question. we may be able to convert it to an AppEngineDeployable. I'll see if this would work.

etanshaul:
@patflynn It doesn't look like we update the deployment source at conversion time (via IJ's migration utilitities) because we don't have access to the project (no project is open yet, and we need the project to get the deployment sources). All we have is are the dom elements from which the run configurations are deserialized.

There are some other things we might be able to do, like attempt convert the deployment source at deploy time. But I think this is out of scope here.

PR #1433, opened 2017-05-09
Convert legacy local dev run configuration names

### LINKED ISSUE #1427: Rename local dev run config migrated from the old plugin
The run configuration gets migrated correctly, and is runnable out of the box, however it would be nice to rename it to use the new naming scheme. Here's what it currently will look like after migration:

image

etanshaul:
fixes #1427


etanshaul:
@patflynn, @nbashirbello  ready for review:

This PR updates the converter to also update the name of the Local Dev Run Config (appends "(migrated)" to it):

image

Question: should we also update the name of the deployment run config to have that wording? Currently I am not - the out of the box naming from the old run config is fine (as show in the screenshot above). But I think I will to them both to keep it consistent and to indicate that it was, in fact, migrated.

etanshaul:
done - also updating the name of the converted deploy configuration to have the "(migrated)" suffix.

PR #1437, opened 2017-05-09
Release v17.2.4

etanshaul:


nbashirbello:
A little confused as to what it happening here :( Help!!

etanshaul:
just re-releasing what @patflynn  already released + the extra bug fix commits. there are no files changed because the version I'm releasing matches exactly what he releaed

patflynn:
yeah it's a bit weird. I think we'll need to come up with a better strategy so that we don't have to retag things.

PR #1441, opened 2017-05-11
Merge 2017 back into master

etanshaul:
A couple known issues still need to be resolved (work in progress):

1) IJ test framework's more agressive leak detection in 2017 is getting triggered again by something introduced in master. I think its isolated to the new tests in the flex/standard deployment editor test classes.

2) The flex / standard deployment UI is slightly misaligned: the service and version fields are pushed far to the right

I've done some initial sanity check testing, but this will require more extensive testing to validate.


etanshaul:
What would be very helpful for validation:

@patflynn - validate that the Kokoro configs are correct for master

@nbashirbello & @patflynn - give the plugin a test drive, checking critical paths introduced in both 2017(/17.2) and master

kokoro/common.cfg
```diff
@@ -0,0 +1,3 @@
+
```
patflynn: these kokoro config files stay out. They've been moved.

gradle.properties
```diff
@@ -15,8 +15,8 @@
 #
 
 ideaEdition = IC
```
patflynn: what do you think about default to IU?

etanshaul:
also, I just realized that this loses all the commits from 2017 and we want to maintain that history. Let me see what I can do ...

kokoro/ubuntu/release.sh
```diff
@@ -59,27 +59,21 @@ if [ "$SOURCE_VERSION" != "$TAG_VERSION" ]
...
 
-echo "Installing aktau/github-release.."
+echo "Installing github-release"
 go get github.com/itchio/gothub
 cd github/google-cloud-intellij
 echo "Building plugins"
 ./gradlew buildPlugin
 
 echo "Creating Github release for tag: $GIT_TAG_NAME"
-
```
patflynn: we want to keep release.sh as is. This is actually a newer version than the branch.

etanshaul:
I may have shot myself in the foot here by not thinking about the need to preserve the commit history from 2017 until after doing this merge. Best case I can back out and force push to this branch; worst case I have to do this merge over again. I'll experiment with a few different approaches.

etanshaul:
Ok, I finally fixed it I think. The history from 2017 is back. Ill continue validating this merge and making fixes from code review.

etanshaul:
Reminder to self:

When merging this in after approval, make sure to disable squashing and enable merge commits so that master gets all the commit history hear.

Also, I can squash individual commits related to review into 1 "merge 2017 to master" commit.

kokoro/common.cfg
etanshaul: which configs specifically need to be removed? just common.cfg?

kokoro/ubuntu/release.sh
etanshaul: I don't think I follow this comment - are you saying the version in 2017 is newer than this? What needs to change exactly?

gradle.properties
etanshaul: I'll do this in a different PR

kokoro/ubuntu/release.sh
etanshaul: I think maybe the version in master is incorrect:
the echo statement says its pull aktau/github-releases, but then the command pulls from itchio/gothub. 

Also, should I have kepts all those exports and --name lines below?

kokoro/common.cfg
patflynn: all the kokoro cfg and script file changes should be removed. Master has the latest config.

etanshaul:
@patflynn - done, I've reverted all the Kokoro changes in the merge. Its now the same as master.

Issue #1304, opened 2017-03-21
App Engine Flex custom runtime should allow the user to specify the Docker directory instead of Dockerfile path


patflynn:
@meltsufin 
@loosebazooka 
@nbashirbello 
@joaoandremartins 

I believe the build tools now use a directory as the input to the docker part of staging.

loosebazooka:
Yeah for pure flex deployments. This is not true for compat style deployments.

joaoandremartins:
What to do in case the provided directory doesn't contain a Dockerfile? Allow its creation like we've been doing til now and error if the provided directory doesn't contain one? Or if it doesn't, use gcloud gen-config to generate one?

joaoandremartins:
It would also be good in the future to provide the rationale for this decision, just so I'm not running in the dark. What's the motivation? Get more files required by Docker from that same directory?

patflynn:
yep. People may want to include more than just their archive in the final
image.

gen-config runtime detection won't work on source, so that's not a good
option unless we can tell the command exactly what kind of runtime we want
it to generate. (which would be nice)

nbashirbello:
Wanted to confirm the behavior of the Flex facet UI when we only save the dockerfile directory as opposed to the dockerfile path. I am leaning towards leaving the Flex facet UI the same, i.e having the Dockerfile textbox and the generate button that generates a new dockerfile. But on save, we only save the directory of the selected or newly generated Dockerfile and when we open the Flex facet UI, we populate the dockerfile textbox with the dockerfile directory and append “/Dockerfile” to it. Thoughts?

chanseokoh:
@nbashirbello that approach needs the assumption that the filename should always be `Dockerfile`. Do we have such a guarantee? (I'm thinking maybe yes in our App Engine story.) If yes, the current `Dockerfile` path field should be preventing manually entering or selecting a file whose name is not `Dockerfile`. Is it the case?

patflynn:
I think that sounds a little weird.  I wouldn't understand that all the
files in the directory would be copied in to the staging folder if I just
saw the full Dockerfile path.

nbashirbello:
@patflynn Are you talking about the deployment dialog?
<img width="1082" alt="screen shot 2017-05-05 at 2 30 53 pm" src="https://cloud.githubusercontent.com/assets/6805942/25759230/8a6deed2-319f-11e7-9356-6467fd7a6b08.png">
[...]

Issue #1408, opened 2017-05-04
Cannot apply changes in Settings dialog if there is a problem with the Cloud SDK configuration


chashnikov:
The same thing as with [run configurations](https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1407): I cannot apply changes and close the Settings dialog if there is a problem with the Cloud SDK in Build, Execution, Deployment | Clouds. It's very inconvenient because I may have changes in other parts of Settings dialog and I'll lose them if I press 'Cancel'.

chashnikov:
In order to fix this you need to avoid throwing exceptions from com.google.cloud.tools.intellij.appengine.cloud.AppEngineCloudConfigurable#apply. The error will be shown in UI anyway, and you may also show it if a user tries to deploy to a misconfigured cloud, but don't forbid to apply changes in the Settings dialog.

patflynn:
@nkibler can you update this bug?  I thought you had some PRs that addressed this.

nkibler:
This is slightly different. It's the same problem reported in #1407 but it's for the Google App Engine Cloud settings as shown below:

<img width="1168" alt="screen shot 2017-08-01 at 11 10 18 am" src="https://user-images.githubusercontent.com/14352762/28832339-277609d6-76aa-11e7-8aa9-b665c2aa1608.png">

nkibler:
So we don't have the same opportunity to follow the pattern we applied to the run configuration. Specifically, these configurations are fundamentally different, right down to the SDK level.

The run configuration we changed in #1407 provided a construct to check the configuration for validity and IDEA took care of the rest (calling it, showing the red X, opening the config when it's still invalid and pressing the run button, etc.).

This, however, extends from `UnnamedConfigurable`, which does not provide a similar construct: https://upsource.jetbrains.com/idea-ce/file/idea-ce-d00d8b4ae3ed33097972b8a4286b336bf4ffcfab/platform/platform-api/src/com/intellij/openapi/options/UnnamedConfigurable.java

We have two options:

1. Leave it as-is and the user cannot save the cloud configurable until they satisfy the validation requirements (non-empty Cloud SDK directory).
2. Remove the validation entirely and let it fail on runtime.

In this instance, I'm favoring (1) because it's nice to see where the problem is (2 would not show this) and this form is much shorter than the run configuration. @patflynn @etanshaul WDYT? Maybe we go for (1) now and could re-evaluate if the form becomes larger.

[...]

PR #1443, opened 2017-05-12
Specify docker directory instead of docker file for Flex custom runtime

nbashirbello:
Fix for issue #1304


.idea/compiler.xml
```diff
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <option name="DEFAULT_COMPILER" value="error-prone" />
```
etanshaul: don't commit this; to prevent this from happening, install the error-prone compile plugin in IntelliJ and then make sure you set the java compiler (in IJ settings) to the error prone compiler

nbashirbello: Thanks

nbashirbello:
PR ready for review

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -148,7 +149,7 @@ public Project getProject() {
           return Optional.empty();
         }
         if (runtimeOptional.filter(runtime -> runtime == FlexibleRuntime.CUSTOM).isPresent()
-            && !Files.exists(Paths.get(deploymentConfiguration.getDockerFilePath()))) {
+            && !Files.exists(Paths.get(deploymentConfiguration.getDockerDirectoryPath() + DOCKERFILE))) {
```
etanshaul: There should be a better way to do this: like using one of the file api's to check if a file exists within the directory. This avoids the concatenation and potential OS issues (may not work in Windows).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -298,14 +299,14 @@ protected void resetEditorFrom(@NotNull AppEngineDeploymentConfiguration configu
...
-    dockerfileTextField.setText(configuration.getDockerFilePath());
+    dockerDirectoryTextField.setText(configuration.getDockerDirectoryPath());
     gcpProjectSelector.setText(configuration.getCloudProjectName());
     appYamlTextField.setVisible(configuration.isOverrideAppYaml()
         || modulesWithFlexFacetComboBox.getItemCount() == 0);
     archiveSelector.setText(configuration.getUserSpecifiedArtifactPath());
     appYamlOverrideCheckBox.setSelected(configuration.isOverrideAppYaml()
         || modulesWithFlexFacetComboBox.getItemCount() == 0);
-    dockerfileOverrideCheckBox.setSelected(configuration.isOverrideDockerfile()
+    dockerDirectoryOverrideCheckBox.setSelected(configuration.isOverrideDockerfile()
```
etanshaul: should the method name `isOverrideDockerfile` in the configuration class also be updated?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -374,18 +375,19 @@ private void validateConfiguration() throws ConfigurationException {
...
       if (isCustomRuntime()) {
-        if (StringUtils.isBlank(getDockerfilePath())) {
+        String dockerDirectoryPath = getDockerDirectoryPath();
+        if (StringUtils.isBlank(dockerDirectoryPath)) {
           throw new ConfigurationException(
-              GctBundle.message("appengine.flex.config.browse.dockerfile"));
+              GctBundle.message("appengine.flex.config.browse.docker.directory"));
         }
-        if (!isValidConfigurationFile(getDockerfilePath())) {
+        if (!isValidConfigurationFile(dockerDirectoryPath + DOCKERFILE)) {
```
etanshaul: same comment here as above about concatenating to the path

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleStage.java
```diff
@@ -72,7 +76,8 @@ public void stage(@NotNull Path stagingDirectory) {
             GctBundle.getString("appengine.deployment.error.staging.yaml"));
       }
       if (isCustomRuntime
-          && !Files.exists(Paths.get(deploymentConfiguration.getDockerFilePath()))) {
+          && !Files.exists(Paths.get(
+          deploymentConfiguration.getDockerDirectoryPath() + DOCKERFILE))) {
```
etanshaul: same thing

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -186,9 +187,17 @@ public void apply() throws ConfigurationException {
...
-        throw new ConfigurationException(
-            GctBundle.getString("appengine.deployment.error.staging.dockerfile"));
+      if (isRuntimeCustom()) {
+        String dockerfileDir = dockerDirectory.getText();
+        if (dockerfileDir.isEmpty() || !Files.isDirectory(Paths.get(dockerfileDir))) {
+          throw new ConfigurationException(
+              GctBundle.getString("appengine.deployment.error.staging.docker.directory"));
+        }
+
+        if (!isValidConfigurationFile(dockerfileDir + DOCKERFILE)) {
```
etanshaul: same

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -237,16 +246,25 @@ private void showWarnings() {
...
+    } else {
+      try {
+        if (isRuntimeCustom()) {
+          String dockerDir = dockerDirectory.getText();
+          if (dockerDir.isEmpty() || !Files.isDirectory(Paths.get(dockerDir))) {
+            errorMessage
+                .setText(
+                    GctBundle.getString("appengine.deployment.error.staging.docker.directory"));
+            showError = true;
+          } else if (!isValidConfigurationFile(dockerDir + DOCKERFILE)) {
```
etanshaul: same

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -237,16 +246,25 @@ private void showWarnings() {
...
+            errorMessage
+                .setText(GctBundle.getString("appengine.deployment.error.staging.dockerfile"));
```
etanshaul: is this the right name for the error message? why "error staging"?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -287,19 +306,35 @@ public void onFacetInitialized(@NotNull Facet facet) {
...
       this.fileName = fileName;
       this.configFileGenerator = configFileGenerator;
-      this.filePicker = filePicker;
+      this.directoryPicker = directoryPicker;
       this.configurationValidator = configurationValidator;
     }
 
     @Override
     public void actionPerformed(ActionEvent event) {
+      boolean genAppYaml = "app.yaml".equals(fileName);
```
etanshaul: this feels a bit off. this is supposed to be a generic generator yet you add app.yaml specific stuff here. I'm thinking this can be refactored to avoid this.

etanshaul:
General observation:

for this change to actually be useful, don't we also have to update the Dockerfile template to pull in the contents of the directory? As it currently stands, only the war / jar artifact is added.

@patflynn @nbashirbello

patflynn:
For custom use-cases we should expect the user to modify the Dockerfile so I think it's OK as is, but you're right that we could add the contents of the Docker dir.  The problem is that we don't want to add the full content of the staging directory, and that's where the docker build will happen.

I wonder what the spotify plugin does.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -90,7 +90,7 @@ appengine.flex.config.user.specified.artifact.title=Select a JAR or WAR file
 appengine.flex.config.user.specified.artifact.error=Browse to a JAR or WAR file.
 appengine.flex.config.custom.deploy.title=App Engine flexible environment Configuration
 appengine.flex.config.browse.app.yaml=Browse to an app.yaml file.
-appengine.flex.config.browse.dockerfile=Browse to a Dockerfile.
+appengine.flex.config.browse.docker.directory=Browse to a docker directory.
```
patflynn: I think you should capitalize docker

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -122,8 +122,10 @@ appengine.stop.modules.version.canceled.message=Undeploy Canceled
 appengine.deployment.error.during.staging=Deployment failed due to staging failing with exit code: {0}.
 appengine.deployment.exception.during.staging=Deployment failed due to an exception while staging the project.
 appengine.deployment.error.staging.yaml=The specified app.yaml configuration file does not exist or is not a valid file.
-appengine.deployment.error.staging.dockerfile=The specified Dockerfile configuration file does not exist or is not a valid file.
-appengine.deployment.error.staging.gotosettings=Set a valid file in Module Settings or use another one.
+appengine.deployment.error.staging.dockerfile=There is no Dockerfile in specified directory or it is not a valid file.
```
patflynn: what is a 'valid file'?

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -122,8 +122,10 @@ appengine.stop.modules.version.canceled.message=Undeploy Canceled
...
+appengine.deployment.error.staging.docker.directory=The specified docker directory does not exist or it is not a valid directory.
```
patflynn: same here.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -122,8 +122,10 @@ appengine.stop.modules.version.canceled.message=Undeploy Canceled
...
+appengine.deployment.error.staging.directory.gotosettings=Set a valid directory in Module Settings or use another one.
```
patflynn: I'd be tempted to use a key label that matched what it's used for instead of what it says.  Maybe appengine.deployment.error.staging.directory.invalid?

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -122,8 +122,10 @@ appengine.stop.modules.version.canceled.message=Undeploy Canceled
...
+appengine.deployment.error.staging.file.gotosettings=Set a valid file in Module Settings or use another one.
```
patflynn: same here.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -78,6 +78,7 @@
   private static final String REFRESH_TOKEN_LABEL = "refresh_token";
   private static final String GCLOUD_USER_TYPE_LABEL = "type";
   private static final String GCLOUD_USER_TYPE = "authorized_user";
+  private static final String DOCKERFILE = "/Dockerfile";
```
patflynn: I'd be tempted to set this to 'Dockerfile', since '/Dockerfile' alone is a valid path to the root of the file system, which will never be what we want.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
patflynn: Yes, Path supports a way to build paths.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.form
```diff
@@ -180,12 +180,12 @@
...
             </properties>
           </component>
-          <component id="9e96a" class="javax.swing.JLabel" binding="dockerfileLabel">
+          <component id="9e96a" class="javax.swing.JLabel" binding="dockerDirectoryLabel">
             <constraints>
               <grid row="2" column="0" row-span="1" col-span="1" vsize-policy="0" hsize-policy="0" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
             </constraints>
             <properties>
-              <text value="Dockerfile:"/>
+              <text value="Docker Directory:"/>
```
patflynn: this should come from the resource bundle.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -186,9 +187,17 @@ public void apply() throws ConfigurationException {
...
 
     try {
-      if (isRuntimeCustom() && !isValidConfigurationFile(dockerfile.getText())) {
-        throw new ConfigurationException(
-            GctBundle.getString("appengine.deployment.error.staging.dockerfile"));
+      if (isRuntimeCustom()) {
+        String dockerfileDir = dockerDirectory.getText();
+        if (dockerfileDir.isEmpty() || !Files.isDirectory(Paths.get(dockerfileDir))) {
+          throw new ConfigurationException(
+              GctBundle.getString("appengine.deployment.error.staging.docker.directory"));
```
patflynn: this error and the one below should include the actually configured path.

patflynn: @etanshaul 
@nbashirbello 
I think this the kind of thing Nikolay suggested we not do in apply methods #1408

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -237,16 +246,25 @@ private void showWarnings() {
...
       errorMessage.setText(GctBundle.getString("appengine.deployment.error.staging.yaml"));
       showError = true;
-    }
-
-    try {
-      if (isRuntimeCustom() && !isValidConfigurationFile(dockerfile.getText())) {
-        errorMessage.setText(GctBundle.getString("appengine.deployment.error.staging.dockerfile"));
+    } else {
+      try {
+        if (isRuntimeCustom()) {
```
patflynn: this looks exactly like the code in apply.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
patflynn: +1.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: see my UX doc for proposed updates. perhaps those changes can be anticipated in this PR.

https://docs.google.com/document/d/1rE4Ir9UnRXoLnRxxxqF63KmBL8xkl_W1KOwqUwdtM6Y

I specifically do not have the "valid file" part because we don't do anything intelligent with the Dockerfile yet.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
etanshaul: yea, in my UX update docs I change this so that the facet dialog can be saved even with errors.

I'd suggest to make the minimal set of changes here in this PR to accommodate updating to a Docker directory since once we agree on the UX updates there will be a separate PR to revamp this.

patflynn:
sgtm

etanshaul:
@patflynn regarding adding the other assets to the Dockerfile:

I agree that as a first step we can expect users with runtime: custom to add what is needed to the Dockerfile.  One way around way you said about not adding everything in the staging directory, would be to instead add the files individually from the specified docker directory (instead of just everything in there).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
nbashirbello: This is shared with the deployment ui

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -148,7 +150,7 @@ public Project getProject() {
           return Optional.empty();
         }
         if (runtimeOptional.filter(runtime -> runtime == FlexibleRuntime.CUSTOM).isPresent()
-            && !Files.exists(Paths.get(deploymentConfiguration.getDockerFilePath()))) {
+            && (!new File(deploymentConfiguration.getDockerDirectoryPath(), DOCKERFILE).exists())) {
```
etanshaul: instead of exists() i'd suggest to use isFile() to make sure its not a directory.

But better yet, I think you can use the Paths api:

```
Path path = Paths.get(deploymentConfiguration.getDockerDirectoryPath(), DOCKERFILE));
..
.. && (!Files.isRegularFile(path))...
```

Same comment for all other cases doing this logic

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -78,6 +79,7 @@
   private static final String REFRESH_TOKEN_LABEL = "refresh_token";
   private static final String GCLOUD_USER_TYPE_LABEL = "type";
   private static final String GCLOUD_USER_TYPE = "authorized_user";
+  private static final String DOCKERFILE = "Dockerfile";
```
etanshaul: DOCKERFILE_NAME ?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -278,28 +300,46 @@ public void onFacetInitialized(@NotNull Facet facet) {
...
     // Used to refresh the warnings.
     private final Runnable configurationValidator;
 
     GenerateConfigActionListener(
         Project project,
         String fileName,
         Consumer<Path> configFileGenerator,
-        TextFieldWithBrowseButton filePicker,
+        TextFieldWithBrowseButton directoryPicker,
+        boolean isDirectory,
```
etanshaul: I'm not really a fan of passing in a boolean like this to switch between the logic.

Is it possible to normalize the directory outside of this class? e.g. if its not a directory, to grab the parent directory outside of here and pass it into this generator?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -237,16 +248,26 @@ private void showWarnings() {
...
       errorMessage.setText(GctBundle.getString("appengine.deployment.error.staging.yaml"));
       showError = true;
-    }
-
-    try {
-      if (isRuntimeCustom() && !isValidConfigurationFile(dockerfile.getText())) {
-        errorMessage.setText(GctBundle.getString("appengine.deployment.error.staging.dockerfile"));
+    } else {
+      try {
+        if (isRuntimeCustom()) {
```
etanshaul: may change depending on if/how you handle @patflynn's comment here, but as is its a little hard to follow this code. Could it be refactored for clarity (extracting named methods where appropriate etc.)?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -237,16 +248,26 @@ private void showWarnings() {
...
+          String dockerDirectoryTest = dockerDirectory.getText();
+          if (dockerDirectoryTest.isEmpty() || !Files.isDirectory(Paths.get(dockerDirectoryTest))) {
```
etanshaul: dockerDirectoryTe**x**t? or dockerDirectoryName? also you could do something like  StringUtils.isEmpty for a null safe check too.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -374,18 +375,19 @@ private void validateConfiguration() throws ConfigurationException {
...
           throw new ConfigurationException(
-              GctBundle.message("appengine.flex.config.browse.dockerfile"));
+              GctBundle.message("appengine.flex.config.browse.docker.directory"));
         }
-        if (!isValidConfigurationFile(getDockerfilePath())) {
+        if (!isValidConfigurationFile(Paths.get(dockerDirectoryPath, DOCKERFILE).toString())) {
           throw new ConfigurationException(
               GctBundle.getString("appengine.deployment.error.staging.dockerfile") + " "
-                  + GctBundle.getString("appengine.deployment.error.staging.gotosettings"));
+                  + GctBundle.getString("appengine.deployment.error.staging.directory.gotosettings"));
```
etanshaul: This message doesn't mention a Docker direction, just a "directory". Shouldn't it indicate that its the Docker directory we are talking about here?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -374,18 +375,19 @@ private void validateConfiguration() throws ConfigurationException {
     if (!isValidConfigurationFile(getAppYamlPath())) {
       throw new ConfigurationException(
           GctBundle.getString("appengine.deployment.error.staging.yaml") + " "
-              + GctBundle.getString("appengine.deployment.error.staging.gotosettings"));
+              + GctBundle.getString("appengine.deployment.error.staging.file.gotosettings"));
```
etanshaul: Similar comment here as for the Docker directory messaging: this just mentions an "invalid file", not a bad app.yaml which is why the error is shown.

Aren't these messages shown at the bottom of the deployment config? If so they should mention what they are for.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.form
```diff
@@ -65,7 +65,7 @@
               <grid row="2" column="0" row-span="1" col-span="1" vsize-policy="0" hsize-policy="1" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
             </constraints>
             <properties>
-              <text value="Generate..."/>
+              <text value="Generate Dockerfile..."/>
```
etanshaul: I know it was hardcoded here before..but this should also be in the message bundle.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
nbashirbello: Updated

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
nbashirbello: We also need to know, after we generate the dockerfile of directory, if we are setting the directoryPicker to have the location of the file it just generated or its parent

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
nbashirbello: Should we use regular file instead, because we are checking for that: There is no Dockerfile in specified directory or it is not a regular file.

etanshaul: I'm not sure what you mean by regular file? Not a directory (if so, I'd say that's implied and not worth the potentially confusing wording)?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
etanshaul: ok. Ideally, this would be refactored a somehow, but if it ends up not being cleaner, then at a minimum I'd suggest to document the behavior of this class then since it takes a little effort to decipher as is - what the outcome of setting `isDirectory` one way or the other etc.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -321,20 +349,40 @@ public void actionPerformed(ActionEvent event) {
...
+            destinationFilePath.toString());
         configurationValidator.run();
       }
     }
   }
 
+  /**
+   * An object representing the outcome of a configuration validation check.
+   */
+  static class Result {
```
etanshaul: ValidationResult ?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -166,33 +171,23 @@ public JComponent createComponent() {
...
-
-    try {
-      if (isRuntimeCustom() && !isValidConfigurationFile(dockerfile.getText())) {
-        throw new ConfigurationException(
-            GctBundle.getString("appengine.deployment.error.staging.dockerfile"));
-      }
-    } catch (MalformedYamlFileException myf) {
-      throw new ConfigurationException(
-          GctBundle.getString("appengine.appyaml.malformed"));
+    Result result = validateConfiguration();
```
etanshaul: why call validateConfiguration twice? showWarnings already calls it. I know that showWarnings is called from other places than apply(). You could do something like return the validationResult from showWarnings (and rename it to validateAndShowWarnings, or something).

etanshaul:
@nbashirbello Aside from the last few comments I just made, things look good.

I noticed that there are some failing tests- its possible that if you haven't merged from master in a while then you are missing some of my fixes to the flaky tests. If thats the case, updating should fix it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
nbashirbello: Done

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/FlexibleFacetEditor.java
```diff
@@ -278,28 +289,59 @@ public void onFacetInitialized(@NotNull Facet facet) {
...
     private final String fileName;
-    private final TextFieldWithBrowseButton filePicker;
+    private final TextFieldWithBrowseButton directoryPicker;
     private final Consumer<Path> configFileGenerator;
+    private final boolean isDirectory;
     // Used to refresh the warnings.
     private final Runnable configurationValidator;
 
+    /**
+     * Constructor
```
etanshaul: I don't think this is proper javadoc /  google format. I'd add a sentence (and I think it has to end in a period).

PR #1449, opened 2017-05-15
Clear declared fields in unit test to prevent triggering 'too many projects leaked' detection in 2017

etanshaul:
Still testing things out. To see if this gets rid of flakiness.


google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/debugger/ServerToIDEFileResolverTest.java
```diff
@@ -72,24 +75,27 @@ public void ignore_testGetFileFromPath_fullPath() {
     assertEquals(
         fileResolver.getFileFromPath(project, "path/to/prj/src/main/com/java/package/Class.java"),
         file1.getVirtualFile());
+    UsefulTestCase.clearFields(fileResolver);
   }
 
   // When searching for the package and class name.
-  public void testGetFileFromPath_packageClass() {
+  public void testGetFileFromPath_packageClass() throws IllegalAccessException {
```
patflynn: why aren't we doing this in the teardown like the other tests?

etanshaul:
this wasn't leading anywhere. will try a different approach in a different PR

PR #1452, opened 2017-05-16
Fixes for IJ 2017.2

etanshaul:
Updates for Intellij 2017.2

These changes _are_ backwards compatible with 2017.1.

We need to decide if we release this as:

1) A new branch / release targeting 2017.2 EAP and up, keeping the 2017.1 release as is. This release would then target a range:   `<idea-version since-build="172.2103" until-build="172.*"/>` which is 2017.2 EAP and up: this is how its currently configured in this PR. If we do this, then we'd need to upload the IJ binary to the GCS so I can comment the `intellijRepo` line back in.

2) Merge this into the 2017 branch and create a new 2017 release (this will work because it backwards compatible). Caveat: we'd have to update the version to range from 2017.1 to 2017.2*; there is no way with the gradle intellij plugin to have a since/until versioning scheme that spans major releases, so we'd have to disable the auto versioning and hardcode this range into the plugin.xml's.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -57,10 +57,10 @@ subprojects {
     intellij {
         type = ideaEdition
         version = ideaVersion
-        updateSinceUntilBuild = true
+        updateSinceUntilBuild = false
         downloadSources = true
         sandboxDirectory = "${rootProject.buildDir}/idea-sandbox"
-//        intellijRepo = "https://storage.googleapis.com/cloud-tools-for-java-team-kokoro-build-cache/idea-distributions"
+        intellijRepo = "https://storage.googleapis.com/cloud-tools-for-java-team-kokoro-build-cache/idea-distributions"
         publish {
             username System.getenv('IJ_REPO_USERNAME')
             password System.getenv('IJ_REPO_PASSWORD')
--- a/build.gradle
+++ b/build.gradle
@@ -60,7 +60,7 @@ subprojects {
         updateSinceUntilBuild = true
         downloadSources = true
         sandboxDirectory = "${rootProject.buildDir}/idea-sandbox"
-        intellijRepo = "https://storage.googleapis.com/cloud-tools-for-java-team-kokoro-build-cache/idea-distributions"
+//        intellijRepo = "https://storage.googleapis.com/cloud-tools-for-java-team-kokoro-build-cache/idea-distributions"
         publish {
             username System.getenv('IJ_REPO_USERNAME')
             password System.getenv('IJ_REPO_PASSWORD')
--- a/google-account-plugin/resources/META-INF/plugin.xml
+++ b/google-account-plugin/resources/META-INF/plugin.xml
@@ -18,8 +18,9 @@
   <id>com.google.gct.login</id>
   <name>Google Account</name>
   <vendor>Google</vendor>
-  <!-- "version" set by gradle-intellij-plugin -->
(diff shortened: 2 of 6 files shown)


etanshaul:
@patflynn - thoughts for option 1 vs 2? 

Its a shame that we'd need to hardcode the versions with option 2, but I think I prefer this option. This way, we could potentially avoid having to maintain multiple releases (2017.1 / 2017.2) for a while, at least until we can't avoid non-backwards compatible changes.

etanshaul:
also will fix #1453

patflynn:
I'm fine with option 2.

patflynn:
can you also merge this change into master?

etanshaul:
sure, I'll update the versions to be hardcoded in plugin.xml, and also cherrypick this into master

etanshaul:
@patflynn ptal.

This is intended to be merged into 2017 and released from there.

I have the versions in both plugin.xml's set as follows:

`<idea-version since-build="171.3780" until-build="172.*"/>`

The "since" version matches our current released minimum: https://plugins.jetbrains.com/plugin/8079-google-cloud-tools

build.gradle
```diff
@@ -57,7 +57,7 @@ subprojects {
     intellij {
         type = ideaEdition
         version = ideaVersion
-        updateSinceUntilBuild = true
+        updateSinceUntilBuild = false
```
etanshaul: This will prevent gradle-intellij-plugin from overriding our hardcoded versions in plugin.xml

google-account-plugin/resources/META-INF/plugin.xml
```diff
@@ -18,8 +18,9 @@
   <id>com.google.gct.login</id>
   <name>Google Account</name>
   <vendor>Google</vendor>
-  <!-- "version" set by gradle-intellij-plugin -->
-  <!-- "idea-version since-build" set by gradle-intellij-plugin -->
+  <!-- "idea-version since-build" set to cover 2017.1 - 2017.2 -->
+  <!-- Set manually because the gradle-intellij-plugin cannot span across major release versions -->
+  <idea-version since-build="171.3780" until-build="172.*"/>
```
patflynn: thank you for remembering the account plugin

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AppEngineStandardMavenLibrary.java
```diff
@@ -39,29 +38,28 @@
       DependencyScope.PROVIDED),
   JSTL(
       GctBundle.message("appengine.library.jstl.api.name"),
-      new RepositoryLibraryProperties("javax.servlet", "jstl", RepositoryUtils.ReleaseVersionId),
-      DependencyScope.PROVIDED),
+      new RepositoryLibraryProperties("javax.servlet", "jstl",
+          AppEngineStandardMavenLibrary.RELEASE_VERSION_ID), DependencyScope.PROVIDED),
```
patflynn: it's a bit weird to full qualify it no?

etanshaul: its needed. otherwise I get "forward reference.." compilation errors.

PR #1457, opened 2017-05-16
Updates the avatar icon when the GoogleLoginAction updates

### LINKED ISSUE #1399: Toolbar Login button avatar icon sometimes remains 'anonymous' despite the user's profile photo displaying correctly in the details view
The below icon state of the login toolbar button is for when the user has not logged in, or when the profile pic has failed to load.   However we are seeing it in cases where clicking on the button shows a logged in user with a profile photo.

screenshot from 2017-05-03 09 39 10

potentially related: #1377

Any fix needs to be patched into Android Studio.

nkibler:
Fixes #1399.

**Cause:** Not entirely known, but the call to `Presentation.setIcon()` was called once on a background thread, scheduled by `ApplicationManager.getApplication().invokeLater()`. I debugged this call and ensured it was actually called, but there may have been a problem with repainting from this thread.

**Solution:** Remove the custom action button that holds the icon and use the already-provided icon field on the action itself. To update, override the action's method `AnAction.update()` to sync the icon with the current account status periodically. This is safe to do as often as `.update()` is called because it has no long-running operations; it simply updates references and repaints the action's icon if needed.


nkibler:
Hmm so this fixes the problem, but I'm not exactly sure why the property change event that's fired by the Presentation and received by the ActionButton doesn't work. I'm still digging into why updating the icon during the event notification doesn't seem to work as it should, but it's internal to the IntelliJ platform SDK.

nkibler:
PTAL, I've removed the GoogleLoginActionButton entirely and relied on the action's icon itself. This removes a lot of unnecessary complexity and leverages the update() method in the action for switching the icon reference based on state.

I also included a little fix to close the sign-in popup when transitioning to the browser so it doesn't float over top of the browser endlessly until you go back to IntelliJ and remove focus from it.

etanshaul:
hey @nkibler. thanks for the first pr!

Can you elaborate on what complexity is removed by getting rid of the GoogleLoginActionButton? On quick glance it looks like the button just encapsulated some of the behavior that you now inlined instead.

Also, could you very briefly explain the behavior updates that we should expect to see with the changes here? This will help when giving it a test drive ourselves.

nkibler:
Sure! That's correct, removing the custom button allows the icon switching behavior to be inlined with the action. It's especially useful because it doesn't require manual calls to notifyStatusIndicator() every time the account is changed, like it does now. Instead, it leverages AnAction's update() function, which is frequently called by the framework: https://upsource.jetbrains.com/idea-ce/file/idea-ce-d00d8b4ae3ed33097972b8a4286b336bf4ffcfab/platform/editor-ui-api/src/com/intellij/openapi/actionSystem/AnAction.java?nav=8552:8558:focused&line=341

With this PR, you'll notice the following:
- The icon in the toolbar for the account plugin correctly displays the icon for the selected account, or the default icon if there is none
- The sign-in/add account popup is closed when pressing the "Sign in" or "Add Account" buttons before opening the browser. This prevents the popup from globally floating on your screen until you re-activate IntelliJ and remove focus from the popup

google-account-plugin/src/com/google/cloud/tools/intellij/login/CredentialedUser.java
```diff
@@ -49,19 +51,12 @@ protected CredentialedUser(String email) {
    * Creates a credentialed user.
    */
   public CredentialedUser(GoogleLoginState state,
-      final IGoogleLoginCompletedCallback updateUserCallback) {
+      @Nullable final IGoogleLoginCompletedCallback callback) {
```
patflynn: the callback name should include what event it handles.  I guess in this case it should be something like `onLoginComplete`.

google-account-plugin/src/com/google/cloud/tools/intellij/login/CredentialedUser.java
```diff
@@ -111,20 +106,18 @@ protected void setActive(boolean isActive) {
...
       name = userInfo.getName();
-      IUserPropertyCallback pictureCallback = new IUserPropertyCallback<Image>() {
-        @Override
-        public void setProperty(Image newImage) {
-          image = newImage;
-          updateUserCallback.onLoginCompleted();
+      GoogleLoginUtils.provideUserPicture(userInfo, newImage -> {
+        image = newImage;
+        if (callback != null) {
+          callback.onLoginCompleted();
```
patflynn: ugh. my suggested name will make this line ugly, but I think that's the fault of the interface definition. I'm fine with `loginCompletedCallback` or something like that to make this less ugly.

patflynn:
@nkibler sorry for the slow latency on this PR, and thanks for fixing this so fast. 

A few meta things (which we all need to do):

- For bug fixes in the PR description try to describe the cause as best you can and how your change fixes it.
- It's not completely clear here what's the refactor and what's the bugfix, or whether the refactor fixes the bug.  If you can, try and separate out changes into the smallest 'whole' change, and send multiple PRs. That'll keep your descriptions and the PR review short and focused.   Basically try to do a single thing per PR. 
- Testing is very difficult in this codebase but we need to improve coverage. If writing a test wasn't practical I'd definitely believe you, but mention in the description why there's no test.

None of us are doing the above points consistently, but I'd like us to start.

google-account-plugin/src/com/google/cloud/tools/intellij/login/IntellijGoogleLoginService.java
```diff
@@ -621,14 +588,7 @@ public void initializeUsers() {
...
 
-        IGoogleLoginCompletedCallback callback = new IGoogleLoginCompletedCallback() {
-          @Override
-          public void onLoginCompleted() {
-            uiFacade.notifyStatusIndicator();
-          }
-        };
-
-        users.addUser(new CredentialedUser(delegate, callback));
+        users.addUser(new CredentialedUser(delegate, null));
```
etanshaul: sometimes we add a comment next to unamed parameters to make things clearer

e.g. `new CredentialedUser(delegate, null /*variableNameFromSignature*/)`

This is a bit controversial, especially if you have IJ's new named parameter syntactic sugar thing on (I turned it off because I found it was buggy). But I think it helps readability.

nkibler:
@patflynn Sure thing! I added my notes on the potential cause and solution to the description. Sorry about the bugfix/refactor merge; I thought the refactor would explicitly fix the bug but the original commit was all I needed to fix the bug. As for tests, I didn't see any existing UI tests for this plugin so I was going to ask about adding those after this PR went through.

patflynn:
That's right. No UI tests as of now :(. This will change soon, now that IJ is packaging a UI testing framework with the latest SDK release.  In the mean time just unit test what you can, and use the platform test case for any headless components you can integration test.

google-account-plugin/src/com/google/cloud/tools/intellij/login/CredentialedUser.java
nkibler: Done.

google-account-plugin/src/com/google/cloud/tools/intellij/login/CredentialedUser.java
nkibler: Done.

google-account-plugin/src/com/google/cloud/tools/intellij/login/IntellijGoogleLoginService.java
nkibler: Done.

google-account-plugin/src/com/google/cloud/tools/intellij/login/ui/GoogleLoginUsersPanel.java
```diff
@@ -186,11 +187,9 @@ public void mouseDragged(MouseEvent event) {
     boolean noUsersAvailable =
         (listModel.getSize() == 1) && (listModel.get(0) instanceof NoUsersListItem);
     JButton addAccountButton = new JButton(noUsersAvailable ? SIGN_IN : ADD_ACCOUNT);
-    addAccountButton.addActionListener(new ActionListener() {
-      @Override
-      public void actionPerformed(ActionEvent event) {
-        Services.getLoginService().logIn();
-      }
+    addAccountButton.addActionListener(event -> {
+      SwingUtilities.getWindowAncestor(this).dispose();
```
etanshaul: > I also included a little fix to close the sign-in popup when transitioning to the browser so it doesn't float over top of the browser endlessly until you go back to IntelliJ and remove focus from it.

Is this the fix? If not could you point us to it? (sorry if you did already and I missed it)

nkibler: Yep, this is it.

google-account-plugin/src/com/google/cloud/tools/intellij/login/IntellijGoogleLoginService.java
```diff
@@ -243,17 +241,11 @@ public void onSuccess() {
...
-
-            @Override
-            public void onLoginCompleted() {
-              uiFacade.notifyStatusIndicator();
-              if (callback != null) {
-                callback.onLoginCompleted();
-              }
+          users.addUser(new CredentialedUser(state, () -> {
+            if (callback != null) {
+              callback.onLoginCompleted();
```
etanshaul: oh, I missed this one - should this name be updated too per @patflynn comments?

nkibler: Done.

google-account-plugin/src/com/google/cloud/tools/intellij/login/ui/GoogleLoginUsersPanel.java
patflynn: @nkibler what are the steps to reproduce the bug?

nkibler: 1. Open IDEA.
2. Click the profile image button to open the Google Account popup.
3. Click the "Sign In" button or the "Add Account" button, whichever is appropriate for your account state.
4. Observe the Google Account popup remains visible on top of the opened browser login window. See the picture below.

screen shot 2017-05-17 at 12 07 44 pm

etanshaul: interesting. I'm not able to reproduce that. on OSX at least

etanshaul: not on linux either

nkibler: Hmm, well this is unrelated anyway so I can file a separate bug and remove this from this PR if you'd like.

nkibler: Filed #1461 and removed this change from the PR. PTAL

PR #1471, opened 2017-05-18
Add app.yaml file in postStartupActivity

### LINKED ISSUE #1442: The App Engine Flex create new wizard fails when creating a project with app.yaml
In the create new project wizard, selecting GAE flex and then “Create app.yaml” causes an error in DefaultAppEngineProjectService#generateFromTemplate that the app.yaml already exists and fails the new project creation

etanshaul:
fixes #1442


etanshaul:
oops. this fails if the project already started. will update in a sec

patflynn:
would it be hard to add a platform test case to reproduce this bug?

etanshaul:
I'm not immediately seeing a way to run a platform test case pre startup which is what would be require to cover this entirely.

I did add a test to cover the post-startup case though: b972c7d.

PR #1473, opened 2017-05-22
Don't set field name color to indicate error

### LINKED ISSUE #1237: Label color BLACK contrasts with Darkula theme
When a field gets in an error state, then on a good state, the label foreground colour is set to BLACK, which doesn't play nice with non-default themes.

screenshot from 2017-02-09 13 39 32

Same thing happens for the Directory Name label. And we're possibly doing it elsewhere in the plugin.

etanshaul:
fixes #1237


etanshaul:
still one other case of this I found that I need to fix

etanshaul:
The 2 cases that I found that were problematic and now fixed here are:

1) The Clone from GCR dialog (see the original issue)
2) The Create App Engine Region dialog. It was setting the error text red (and then back to black). Now it looks like this (the text is black, and I added an icon instead):

image

PR #1480, opened 2017-05-30
Convert the local run configuration type so that migration happens only once

### LINKED ISSUE #1479: Local dev server run config migration triggered on already converted configs

etanshaul:
fixes #1479 

The problem before was that the check that determines if migration of the local run config is needed was based on the "type" attribute of the run config which previously was the same between the old plugin on this plugin. The migration wasn't converting the type causing this to get triggered over an over again.

The fix is to not just update the name, but also migrate the type to a new key.

### DIFF
--- a/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/AppEngineServerConfigurationType.java
+++ b/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/AppEngineServerConfigurationType.java
@@ -114,6 +114,6 @@ public class AppEngineServerConfigurationType extends J2EEConfigurationType {
   @NotNull
   @Override
   public String getId() {
-    return "GoogleAppEngineDevServer";
+    return "gcp-app-engine-local-run";
   }
 }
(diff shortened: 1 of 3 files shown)


etanshaul:
This was effect of the bug:

image

Everything still worked properly, except that conversion was triggered over and over again leading to the name being rewritten each time (with "(migrated)" being appended on each conversion) instead of just once.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/run/AppEngineServerConfigurationType.java
```diff
@@ -114,6 +114,6 @@ public AppServerIntegration getIntegration() {
   @NotNull
   @Override
   public String getId() {
-    return "GoogleAppEngineDevServer";
+    return "gcp-app-engine-local-run";
```
etanshaul: Ideally this would be a static final member to be re-used by `AppEngineRunConfigurationConverter`, but since this class is in the Ultimate module and the converter is not (it probably should be - but I'm leaving this for a separate PR) I can't do this.

PR #1481, opened 2017-06-01
Fixes user icon size at higher DPI scales

### LINKED ISSUE #1391: Google Account Login on HIDPI screen to small
When on a HiDPI screen on fedora (25) the google avatar is shown too small (should be 2x), see image:

image

nkibler:
Fixes #1391.

Can anyone verify this on a HiDPI monitor? I can't locally verify but judging by the code in the `JBUI.pixScale()` method, it should fix the issue.

**Problem:** The user's image is scaled manually to 16x16, regardless of the calculated pixel scale (which takes into account user settings and system settings).

**Solution:** Use the provided `JBUI.pixScale()` method to determine the appropriate size of this icon before scaling it down.

### DIFF
--- a/google-account-plugin/src/com/google/cloud/tools/intellij/login/ui/GoogleLoginAction.java
+++ b/google-account-plugin/src/com/google/cloud/tools/intellij/login/ui/GoogleLoginAction.java
@@ -20,31 +20,27 @@ import com.google.cloud.tools.intellij.login.CredentialedUser;
 import com.google.cloud.tools.intellij.login.Services;
 import com.google.cloud.tools.intellij.login.util.AccountMessageBundle;
 
-import com.intellij.openapi.actionSystem.ActionToolbar;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.actionSystem.Presentation;
 import com.intellij.openapi.actionSystem.RightAlignedToolbarAction;
-import com.intellij.openapi.actionSystem.ex.CustomComponentAction;
 import com.intellij.openapi.project.DumbAware;
 import com.intellij.openapi.ui.popup.ComponentPopupBuilder;
 import com.intellij.openapi.ui.popup.JBPopup;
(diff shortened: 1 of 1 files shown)


patflynn:
I can do this on my laptop tomorrow :)

etanshaul:
interestingly im not able to reproduce the original problem on my macbook (posted this accidentally on the issue before).

Still playing around with dpi settings to see if I can.

google-account-plugin/src/com/google/cloud/tools/intellij/login/ui/GoogleLoginAction.java
```diff
@@ -69,15 +65,14 @@ public void update(AnActionEvent e) {
       if (image == null) {
         presentation.setIcon(GoogleLoginIcons.DEFAULT_USER_AVATAR);
```
patflynn: does Google provide how and low res avatar images we can use?

nkibler: Oh you mean loading over HTTP instead of local assets? We could do that if you want.

patflynn: I guess I'm just asking of we can get high dpi versions of the resources instead of scaling.

nkibler: For the default icons, we have 16x16 and 32x32 already, the latter is for retina displays. The scaling we're doing is on the icon we pull from their profile, which I believe is coming in at 96x96 to support the account popup size. We just resize for the smaller icon.

patflynn: ah ok that makes sense.

patflynn:
@etanshaul I can actually reproduce this on my work machine. Unfortunately can't get the local run version of intellij to respect my scaling settings.

patflynn:
@nkibler built this branch locally and installed from disk into IJ on High DPI screen that was exhibiting this issue.  Verified the fix works.

Issue #1477, opened 2017-05-24
Refresh flex deployment UX


etanshaul:
Will be based on UX changes to facet:
https://docs.google.com/document/d/1rE4Ir9UnRXoLnRxxxqF63KmBL8xkl_W1KOwqUwdtM6Y/edit

etanshaul:
also see (mostly outdated): #1275 and #1333

PR #1484, opened 2017-06-02
Refactors common deployment ux into shared class

etanshaul:
In support of deployment UX refresh ( #1477 )

doing this in parts to keep the PRs small.

This extracts the common UX configuration elements from the standard and the flexible deploy UX's into a common form (previously they were duplicated).

Also slightly updates the look (notice the top/bottom margin, and the order of the elements). Look at the environment / project / region section:
image

Also snuck in a very minor update to the flex/standard environment label (capitalization was wrong).


patflynn:
Will take a look this afternoon

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -51,6 +51,9 @@ appengine.tools.menu.run.server.error.title=Project is not an App Engine standar
 appengine.tools.menu.run.server.error.message=To use the App Engine standard local development server, add App Engine standard framework support to one or more modules in your project.
 appengine.tools.menu.deploy.error.title=This project is not an App Engine project
 appengine.tools.menu.deploy.error.message=To deploy to App Engine, add App Engine standard or flexible framework support to one or more modules in your project.
+appengine.deployment.config.environment.label=Environment:
+appengine.deployment.config.region.label=Region:
+appengine.deployment.config.project.label=Project:
 appengine.version.label=Version\:
```
patflynn: I'm noticing the '\:' for version but not for the new labels. does it matter?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
```diff
@@ -0,0 +1,45 @@
...
+
+import com.google.cloud.tools.intellij.resources.ProjectSelector;
+
+import javax.swing.JLabel;
+import javax.swing.JPanel;
+
+/**
+ * Common App Engine deployment configuration UI shared by flexible and standard.
+ */
+public class AppEngineDeploymentConfigurationPanel {
```
patflynn: So I guess this panel is instantiated automagically by the jetbrains UI framework?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -108,12 +111,20 @@
...
   private JLabel noSupportedModulesWarning;
+  private JLabel environmentLabel;
 
   private static final boolean PROMOTE_DEFAULT = false;
   private static final boolean STOP_PREVIOUS_VERSION_DEFAULT = false;
 
   public AppEngineFlexibleDeploymentEditor(Project project, AppEngineDeployable deploymentSource) {
     this.deploymentSource = deploymentSource;
+
+    gcpProjectSelector = appEngineConfig.getProjectSelector();
```
patflynn: I'm kinda surprised we need to do this if the appEngienConfig is composing the other pieces.

patflynn: I may need a primer on how this works.

etanshaul: yes this is hideous; i'm already working on the next PR where I plan to refactor this. Likely I will not reassign to local members and just get them via the getters.

If you're ok keeping it here, it will shortly be fixed in the next PR

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: no its not needed; removed

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
etanshaul: yes, this is the common panel which is included by both the flex and the standard forms.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
etanshaul: basically whats happening is that the common panel is housing these common UX components. Then the flex and the standard editors may need access to some of these common elements so that they can handle events (and labels etc.) differently. 

In this PR I was just reassigning the already existing members (since everything was duplicated) as you see in this line you referred to. I did this partly to minimize the diff in the PR.

But I think it would be better to not do this reassignment and get the common elements directly from the common panel. If you have a better pattern I'm very open to suggestions.

PR #1485, opened 2017-06-05
Convention is to include '...' after tools menu action labels.

patflynn:


etanshaul:
I need to check this, but I think I explicitly included or did not include these for a reason: for instance if the local run config is already configured, then it runs it immediately (no subsequent UI panel pops up).

etanshaul:
Yea - create a new AE standard project, configured the local run server, then choose this option from the menu.

Notice that the server runs immediately.

I guess having ellipses here is still up for debate; but I used them to indicate that the action will or will not execute immediately.

patflynn:
I'm closing this based off of @etanshaul distinction between the action types.

etanshaul:
I think you forgot to close it; closing

PR #1488, opened 2017-06-06
Update deploy parameter UX and move more components to shared form

etanshaul:
More progress in support of deploy UX refresh ( #1477 )

Moves remainder of common (shared between flex and standard) elements and logic to shared class.

Updates the UI of these common elements.

Flex:
image
Notice the "Deployment parameters" section. The "configuration files" (app.yaml / Dockerfile) has not been updated - this will be in the next PR

Standard:
image


etanshaul:
Here is the old look of the Flex deployment config just for quick comparison:
(remember only everything above the "Configuration files" section has been updated in this PR)

image

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -137,6 +138,7 @@ appengine.stop.modules.version.execution.error.with.code=Stop application failed
 appengine.stop.modules.version.success.dialog.message=Application has been successfully stopped.
 appengine.stop.modules.version.success.dialog.title=Application Stopped
 appengine.stop.modules.version.canceled.message=Undeploy Canceled
+appengine.deployment.parameters.panel.title=Deployment parameters
```
patflynn: I think we titles should be title-caps. e.g 'Deployment Parameters'

patflynn:
I'm loving these small PRs. Thanks!

etanshaul:
@patflynn see https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1488/commits/2846847fdc6eab8b882c32f17f3d41f89cec2b34 for enabling stop previous version for standard, as discussed

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
```diff
@@ -31,6 +44,99 @@
...
+    configuration.setPromote(promoteCheckbox.isSelected());
+    configuration.setCloudProjectName(projectSelector.getText());
+    configuration.setStopPreviousVersion(stopPreviousVersionCheckbox.isSelected());
+  }
+
+  /**
+   * Sets the project / version to allow the deployment line items to be decorated with additional
+   * identifying data. See {@link AppEngineRuntimeInstance#getDeploymentName}.
+   */
+  public void setDeploymentProjectAndVersion(DeploymentSource deploymentSource) {
```
nkibler: Again, not sure about team style but a @param doc is needed here.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
```diff
@@ -17,12 +17,25 @@
...
+import javax.swing.JCheckBox;
 import javax.swing.JLabel;
 import javax.swing.JPanel;
+import javax.swing.event.TreeModelEvent;
 
 /**
- * Common App Engine deployment configuration UI shared by flexible and standard.
+ * Common App Engine deployment configuration UI shared by flexible and standard deployments.
  */
 public class AppEngineDeploymentConfigurationPanel {
```
nkibler: Can this be final? Or does part of the magic involve subclassing it?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
```diff
@@ -42,4 +148,35 @@ public JLabel getEnvironmentLabel() {
...
+    return promoteCheckbox;
+  }
+
+  @VisibleForTesting
+  public void setProjectSelector(ProjectSelector projectSelector) {
+    this.projectSelector = projectSelector;
+  }
+
+  @VisibleForTesting
+  public void setApplicationInfoPanel(
```
nkibler: nit: You can use all 100 chars here and put this signature on one line

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -76,7 +69,6 @@
 import javax.swing.JList;
 import javax.swing.JPanel;
 import javax.swing.event.DocumentEvent;
-import javax.swing.event.TreeModelEvent;
 
 /**
  * Flexible deployment run configuration user interface.
```
nkibler: Ditto about the final comment from above

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -294,24 +241,14 @@ private void resetModuleConfigSelection(Project project) {
...
-  private void refreshApplicationInfoPanel() {
-    if (gcpProjectSelector.getProject() != null && gcpProjectSelector.getSelectedUser() != null) {
-      appInfoPanel.refresh(gcpProjectSelector.getProject().getProjectId(),
-          gcpProjectSelector.getSelectedUser().getCredential());
-    } else {
-      appInfoPanel.setMessage(GctBundle.getString("appengine.infopanel.noproject"),
-          true /* isError*/);
-    }
-  }
+
```
nkibler: nit: Remove extra unnecessary newline here and above, there seems to be 3 here between the constructor and the method below

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/standard/AppEngineStandardDeploymentEditor.java
```diff
@@ -236,27 +143,27 @@ protected JComponent createEditor() {
...
-  JCheckBox getPromoteCheckbox() {
-    return promoteCheckbox;
+  JCheckBox getStopPreviousVersionCheckbox() {
+    return commonConfig.getStopPreviousVersionCheckbox();
   }
 
   @VisibleForTesting
-  JCheckBox getStopPreviousVersionCheckbox() {
-    return stopPreviousVersionCheckbox;
+  public JCheckBox getDeployAllConfigsCheckbox() {
```
nkibler: Do these need to be public? I don't see where they're used in a test outside this package.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
etanshaul: we haven't really been enforcing this if its clear enough from the signature, but I'm open..

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
etanshaul: there is no subclassing of it. is the google style to make everything final that can be final? what about fields, method parameters, etc.

etanshaul:
I went ahead and made the classes final; we can discuss offline what we want our style to be going forward

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
etanshaul: leaving this out unless you / @patflynn feel strongly about it. imo this can get noisy

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
nkibler: Correct, the recommendation is to follow Effective Java's advice: "design and document for inheritance or else prohibit it." So for classes, the easiest rule to follow is always make them final unless it should be sub-classed, in which case its usage should be documented.

For instance variables, final should be used if the field should not be reassigned. This is the preferred behavior for most fields too to prevent mutability issues for the field itself.

For method params, it's mostly discouraged because it causes more harm than good (burden of enforcement, extra noise, etc.). Also as a quick note, as of Java 8, it's no longer necessary to use final on local variables to make them accessible from an inner class, which is often why local vars are final.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -592,13 +508,23 @@ TextFieldWithBrowseButton getArchiveSelector() {
...
+    return commonConfig.getDeployAllConfigsCheckbox();
+  }
+
+  @VisibleForTesting
+  ProjectSelector getProjectSelector() {
+    return commonConfig.getProjectSelector();
+  }
+
+  @VisibleForTesting
+  public AppEngineDeploymentConfigurationPanel getCommonConfig() {
```
nkibler: Ditto about removing public visibility here too, if possible.

PR #1489, opened 2017-06-06
Fixes the testRemoveConsolePane failure on EAP

### LINKED ISSUE #1487: flaky tests
```
com.google.cloud.tools.intellij.debugger.CloudExecutionStackTest > testGetTopFrame_nullFramesReturnsNull SKIPPED

com.google.cloud.tools.intellij.debugger.CloudExecutionStackTest > testGetTopFrame_nonNull SKIPPED

com.google.cloud.tools.intellij.debugger.CloudDebugProcessTest > testRemoveConsolePane FAILED
    java.lang.AssertionError: Already disposed: Project (Disposed) testRemoveConsolePane

        Caused by:
        java.lang.AssertionError: Already disposed: Project (Disposed) testRemoveConsolePane

            Caused by:
            java.lang.AssertionError: Already disposed: Project (Disposed) testRemoveConsolePane

363 tests completed, 1 failed, 2 skipped
:google-cloud-tools-plugin:test FAILED
```

nkibler:
Fixes #1487.

**Problem:** This is specific to the `CloudDebugProcessTest.testRemoveConsolePane` method and only on the EAP latest snapshot of IntelliJ. The async call to create IDE representations of the breakpoints is happening after the project is disposed; I believe it's actually happening in the `tearDown()` process of the test. When this block is executed, it attempts to access the first breakpoint handler; since this doesn't exist yet, the CloudDebugProcess creates one. But by this point, the project is disposed so attempting to access its components throws an `AssertionError`.

This seems like a threading issue; maybe the IntelliJ versions before the EAP latest snapshot immediately evaluate blocks scheduled with `SwingUtilities.invokeLater()` in tests, but the latest snapshot delays until the test is over. I'm going to dig further into this test specifically to see if there's a better long-term solution here.

**Solution:** Bail fast from the scheduled block if the project has been disposed and there are no breakpoint handlers created.


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugProcess.java
```diff
@@ -261,9 +261,12 @@ public void run(@NotNull ProgressIndicator indicator) {
...
     getStateController().startBackgroundListening();
-    SwingUtilities.invokeLater(new Runnable() {
-      @Override
-      public void run() {
+    SwingUtilities.invokeLater(() -> {
+      // If a breakpoint handler has not been created before, getBreakpointHandler() will attempt to
+      // create one. But if the project is disposed of, this creation will fail because it will try
+      // to access the project's components when creating the PsiManager. This guards against a
+      // failure in this scenario.
+      if (breakpointHandlers != null || !getXDebugSession().getProject().isDisposed()) {
```
patflynn: There's nothing currently relying on the fact getBreakpointHandler initializes a null breakpointHandlers here?

nkibler: It likely is, but the `|| !getXDebugSession().getProject().isDisposed()` clause will ensure it will still run, as long as the project hasn't been disposed. I can't imagine anything is depending on this executing even when the project is disposed because they would be seeing the `AssertionError` thrown just like this test.

patflynn: something feels kind of messed up about this. i.e. adding code to guard against a case that should I guess never happen during normal execution of the program..

nkibler: Yeah that's true, let me dig into the cause further and figure out what's going on.

nkibler: Actually this seems to have been the trigger: https://github.com/JetBrains/intellij-community/commit/95a2625ec31ae0327ccf962c31e18ace2301e3e0

Now the events stacked up in the EDT aren't processed until after the project is disposed. We could manually call this somewhere in the test instead, what do you think?

patflynn: rings a bell. @etanshaul may have an opinion on this.

etanshaul: This is the first I'm seeing this change by JB.

It seems strange to me that they would wait until the project is disposed to process these anyc events.

> We could manually call this somewhere in the test instead, what do you think?
 
Manually call what? 

I agree that it feels wrong to modify anything but the test in this case unless we can be sure there's problem with the non-test logic.

nkibler: Sorry, manually call `UIUtil.dispatchAllInvocationEvents()`, which should process anything waiting on the event dispatch thread, including the problematic block in this PR.

etanshaul: ah cool. that sounds reasonable to me. does it fix the test?

etanshaul: I do see many instances of that being invoked manually in tests in the community source. If it fixes the problem then I'm good with that solution.

patflynn:
SGTM

Issue #1476, opened 2017-05-24
Refresh flex facet UX


etanshaul:
https://docs.google.com/document/d/1rE4Ir9UnRXoLnRxxxqF63KmBL8xkl_W1KOwqUwdtM6Y/edit

PR #1493, opened 2017-06-07
Update app engine flexible deployment UX

### LINKED ISSUE #1477: Refresh flex deployment UX

etanshaul:
Fixes #1477 

Updates the remainder of the deployment UX.

Screenshots coming..

Still missing test coverage refresh

This PR updates the flex config portion of the deployment UI (the portion with the app.yaml and dockerfile config) to match the refresh that was done in #1476.

Also, as discussed offline, I got rid of the previously available app.yaml override and Dockerfile override features in favor of clarity and simplicity.


google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -78,7 +78,7 @@ appengine.flex.configtype.custom.label=Custom
 appengine.flex.user.specified.artifact.label=Select a deployment archive:
 appengine.flex.dockerfile.location.label=Dockerfile path\:
 appengine.flex.version.placeholder.text=auto generated
-appengine.flex.facet.config.title=App Engine flexible configuration
+appengine.flex.facet.config.title=App Engine Flexible Configuration
```
etanshaul: This appears as a panel title, so I think its ok to have "Flexible" capitalized. Feel free to disagree.

etanshaul:
No modules with flex support:
image

app.yaml with runtime: java (the edit button links to the module settings):
image

app.yaml with runtime: custom (the Dockerfile path links to the module settings):
image

Showing multiple app.yamls (from different modules) - paths are displayed relative to the project root:
image

etanshaul:
See the above screenshots for an overview of what things look like.

Things **not** fixed here (will file separate issues):
- Move error messages closer to issue. Notice how now they still show up at the bottom:
image

- Don't block saving the dialog on errors. Currently its still preventing the user from saving until errors are resolved.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -147,116 +129,92 @@ protected void textChanged(DocumentEvent event) {
...
-      }
-    });
-
-    resetModuleConfigSelection(project);
-    modulesWithFlexFacetComboBox.addItemListener(event -> toggleDockerfileSection());
-    modulesWithFlexFacetComboBox.setRenderer(new ListCellRendererWrapper<Module>() {
+    appYamlCombobox.setRenderer(new ListCellRendererWrapper<AppEngineFlexibleFacet>() {
       @Override
-      public void customize(JList list, Module value, int index, boolean selected,
+      public void customize(JList list, AppEngineFlexibleFacet facet, int index, boolean selected,
```
nkibler: nit: Use the Google Java formatter to fix this, but the params should either be altogether on the next line (indented by 4 spaces), or if they don't fit like that either, all on separate lines indented by 4 spaces.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -147,116 +129,92 @@ protected void textChanged(DocumentEvent event) {
...
+  private void resetRuntimeDisplay() {
+    if (isValidConfigurationFile(getAppYamlPath())) {
+      try {
+        Optional<FlexibleRuntime> flexibleRuntime =
+            appEngineProjectService.getFlexibleRuntimeFromAppYaml(getAppYamlPath());
+
+        runtimePanel.setLabelText(flexibleRuntime.map(FlexibleRuntime::toString).orElse(""));
+        runtimePanel.setVisible(true);
+        runtimePanel.setExplanationLabelVisibility(
+            !flexibleRuntime.map(FlexibleRuntime::isCustom).orElse(false));
```
nkibler: Why not use `isCustomRuntime()` here since this logic is already separated into its own method?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleRuntimePanel.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.awt.Font;
+
+import javax.swing.JLabel;
+import javax.swing.JPanel;
+
+/**
+ * UI panel containing App Engine flexible runtime information.
+ */
+public class AppEngineFlexibleRuntimePanel {
```
nkibler: final?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleRuntimePanel.java
```diff
@@ -0,0 +1,62 @@
...
+
+  private JPanel runtimePanel;
+  private JLabel runtimeLabel;
+  private JLabel runtimeExplanationLabel;
+
+  public AppEngineFlexibleRuntimePanel() {
+    runtimeExplanationLabel.setFont(
```
nkibler: How is this guaranteed to not be null here?

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditorTest.java
```diff
@@ -242,117 +214,6 @@ public void testValidateConfiguration_invalidCloudSdk() {
     }
   }
 
-  public void testValidateConfiguration_blankYaml() {
```
nkibler: Why remove these tests? Are they not valid anymore?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleRuntimePanel.java
etanshaul: its part of the magic of IJ UX form binding. This class is bound to a GUI form of the same name (with a .form) extension. The platform is responsible for initializing the class properly.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditorTest.java
etanshaul: They are not - all these tests were removed as part of the removal off the app.yaml / Dockerfile override feature mentioned in the first comment of the PR.

I still have a todo to fix some tests and add more coverage for the changes.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
etanshaul: I had a mental todo to look into refactoring this if possible.

But the reason is that if you look a few lines up, I already extract the flexRuntime. I didn't want to call isCustomRuntime which would just duplicate this extraction (which requires IO - parsing the app.yaml file).

etanshaul:
I just noticed that with these changes, the flex cost label location didn't make sense anymore. I moved it to be underneath the environment label (and added an icon to it). Of course this works as expected for flex compat too.

image

_Note: I think prior to merging this we can give it a quick look over to see if we can optimize the spacing / margins in the layout everywhere_

patflynn:
@etanshaul This is looking pretty slick! Starting review noww

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -139,20 +140,20 @@ appengine.stop.modules.version.success.dialog.message=Application has been succe
...
-appengine.deployment.error.staging.file.gotosettings=Set a valid file for app.yaml in Module Settings or use another one.
 appengine.deployment.error.not.logged.in=You must be logged in to perform this action.
 appengine.deployment.error.suggestflexfacet=If you can't find your artifacts in the deployment run configuration, consider adding Google App Engine Flexible support to your module.
 appengine.deployment.source.not.found.error=Deployment source not found: {0}.
 appengine.deployment.invalid.source.error=Invalid deployment source.
 appengine.deployment.status.deploying=Deploying to App Engine
 appengine.deployment.error.invalid.cloudsdk=Invalid Cloud SDK directory path.
 appengine.deployment.deployallconfigs.checkbox.label=Update dos, dispatch, cron, queues, and datastore indexes
-appengine.deployment.flexible.nogaesupport=No modules with Google App Engine support found. Please provide your configuration files
+appengine.deployment.flexible.nogaesupport=No modules with Google App Engine flexible support found.
```
patflynn: Is this caused by the lack of the facet? Can we be specific about that? e.g. No modules with the GAE Flex facet found..?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFacet.java
```diff
@@ -39,4 +41,16 @@ public AppEngineFlexibleFacet(@NotNull FacetType facetType,
       AppEngineFlexibleFacetConfiguration> getFacetType() {
     return FacetTypeRegistry.getInstance().findFacetType(AppEngineFlexibleFacetType.ID);
   }
+
+  /**
+   * Returns the {@link AppEngineFlexibleFacet} present in the supplied module, or null if there
+   * isn't one.
+   */
+  @Nullable
+  public static AppEngineFlexibleFacet getAppEngineFacetByModule(@Nullable Module module) {
```
patflynn: Why accept a null module?

patflynn: why accept null module params?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFacet.java
```diff
@@ -39,4 +41,16 @@ public AppEngineFlexibleFacet(@NotNull FacetType facetType,
       AppEngineFlexibleFacetConfiguration> getFacetType() {
     return FacetTypeRegistry.getInstance().findFacetType(AppEngineFlexibleFacetType.ID);
   }
+
+  /**
+   * Returns the {@link AppEngineFlexibleFacet} present in the supplied module, or null if there
```
patflynn: I think literals like null should be {@code null}

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleRuntimePanel.java
patflynn: It's pretty weird, especially when you don't see a marker interface or anything to indicate magic.

patflynn:
a couple of UI thoughts:
- the italics message about setting a custom runtime looks kind of weird as a child of runtime.  Maybe we could pull it up to a footnote of the panel?  
- Do all run config validation errors always have 'invalid data' prefixes/

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFacet.java
etanshaul: good point. updated

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: yea its caused by no modules / no modules with flex facet.

I hesitated to mention the facet since the way users are typically expected to add facets is via framework support.

Maybe I should mention framework support more explicitly:
No modules with Google App Engine **framework** support found.

etanshaul:
`the italics message about setting a custom runtime looks kind of weird as a child of runtime. Maybe we could pull it up to a footnote of the panel?`
My thought was that this message is directly applicable to runtimes so I had it under the runtime section. I don't feel too strongly about it though.

`Do all run config validation errors always have 'invalid data' prefixes`
Yea this is part of what the framework does when throwing ConfigurationExceptions for validation. This will be removed in a future PR where we no longer block on form errors.

etanshaul:
@patflynn tests have been fixed and updated, and comments addressed. only potential todo is moving the location of the custom runtime / Dockerfile message.

etanshaul:
@patflynn 
here's what the message looks like un-nested:
image

wdyt? maybe remove the italics too..

patflynn:
hmm, what do you think about removing the message completely?

etanshaul:
done e1ece90

image

etanshaul:
found a bug: saving the configuration doesn't save the selected app.yaml in the dropdown. The path is saved, but need to restore the correct selection in the combobox. Fix coming.

patflynn:
the path to the app.yaml is saved?  shouldn't it be read from the module
facet at execution time?

etanshaul:
hmm currently what's persisted is the app.yaml path itself. this is sort of a legacy thing brought over from when we allowed app.yaml overrides.

Its not technically wrong, but instead we could persist the selected facet instead (the facet corresponding to the selected app.yaml in the dropdown)

etanshaul:
on second thought, changing what's persisted would affect user migration. its safer to keep storing the app.yaml path as is.

etanshaul:
ahh. I see your point. We have the potential now for the app.yaml (and Dockerfile for that matter) to be stale. Ill look into this now.

etanshaul:
some things to note about the above:

Once fixed, and released, users upgrading that have a prior custom flex deployment will have to re-enter the app.yaml (and Dockerfile if applicable). This is a consequence of moving these to the facet.

Attempting to do some auto migration like checking to see if there is a stored app.yaml / Dockerfile in the configuration and then moving it to the facet wouldn't be robust as there isn't a one-to-one mapping between deployment config and facet config.

PR #1496, opened 2017-06-09
Update serialized data for flex deployments

### LINKED ISSUE #1495: Flex deployment configuration can be stale
See the end of #1493 for some discussion on this

Currently, app.yaml and Dockerfile strings are directly serialized in the flex deployment config. However, we have since moved the source of these to the flex facet. 

This means that deployments can be stale: user creates a valid deploy config, saves it. Then updates the app.yaml / Dockerfile in the flex facet. Now the user runs the deploy config -> the app.yaml / Dockerfile is stale.

The solution is to instead serialize some key into the facet so that these config files can be retrieved at runtime (deployment / stage time).

etanshaul:
fixes #1495

Instead of serializing the app.yaml and Dockerfile strings in the deploy config, we serialize the module name. The module name is guaranteed to be unique and can be used as a key into the facet configuration to retrieve the config files at deploy time.


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
```diff
@@ -143,18 +144,26 @@ public Project getProject() {
...
-        Optional<FlexibleRuntime> runtimeOptional =
-            AppEngineProjectService.getInstance().getFlexibleRuntimeFromAppYaml(
-                deploymentConfiguration.getAppYamlPath());
+        AppEngineFlexibleFacet flexFacet =
+            AppEngineFlexibleFacet.getFacetByModuleName(
+                deploymentConfiguration.getModuleName(), project);
 
-        if (!Files.exists(Paths.get(deploymentConfiguration.getAppYamlPath()))) {
+        if (flexFacet == null
+            || !Files.exists(Paths.get(flexFacet.getConfiguration().getAppYamlPath()))) {
```
patflynn: I guess there are two different potential problems here:
1) the module name that was saved doesn't exist anymore (it was removed or renamed)
2) the module flex facet doesn't have an actual app.yaml file specified

for 1 can we check this when the user brings up the runtime config editor and unset the module if it doesn't exist anymore?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleStage.java
```diff
@@ -62,23 +68,29 @@ public AppEngineFlexibleStage(
    */
   public void stage(@NotNull Path stagingDirectory) {
     try {
+      AppEngineFlexibleFacet flexibleFacet =
+          AppEngineFlexibleFacet.getFacetByModuleName(
+              deploymentConfiguration.getModuleName(), project);
+
+      if (flexibleFacet == null
+          || !Files.exists(Paths.get(flexibleFacet.getConfiguration().getAppYamlPath()))) {
+        throw new RuntimeException(
```
patflynn: here too it would be nice to differentiate between the error conditions and have more helpful messages.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -215,8 +215,12 @@ private void reloadAppYamls(Project project) {
   protected void resetEditorFrom(@NotNull AppEngineDeploymentConfiguration configuration) {
     commonConfig.resetEditorFrom(configuration);
 
-    dockerfileDirectoryPathLink.setHyperlinkText(
-        tryTruncateConfigPathForDisplay(configuration.getDockerDirectoryPath()));
+    appYamlCombobox.setSelectedItem(
+        AppEngineFlexibleFacet.getFacetByModuleName(configuration.getModuleName(), project));
+    // todo confirm that the below is not needed
+//    dockerfileDirectoryPathLink.setHyperlinkText(
```
nkibler: Looks like this was debug code and should be removed.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/CloudSdkAppEngineHelper.java
etanshaul: for the second case, we get the following error printed in the deploy console:

`Failed to deploy '[2017-06-09 13:31:47] demo:war. Project: test-default-1344. Version: asdf': The specified app.yaml configuration file does not exist or is not a valid file.`

I think that is reasonable. I don't know of a way (or even if its desireable) to dynamically invalidate a run config after updating something in the facet.

For 1 - you are totally right..I feel silly for not thinking of that. I wish there was something more resilient that we could key off of. If not, then your suggestion is probably a good one.

etanshaul: Interesting..I just testing the behavior for point 1:

After deleting the module:
- Attempting to just rerun the deploy config gives the same error as printed above "..specified app.yaml ..does not exist ...". This is just an error message in the deploy console - no IDE exception.
- Editing the run config that previously was set to the now deleted module, has the module automatically unset:

image

I think, again, that these are reasonable behaviors. But we may be able to improve error messaging as you suggest in your other comment. wdyt?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
etanshaul: thanks. done

etanshaul:
@patflynn update error messages and granularity, let me know if this addresses your comment.

Also, regarding your 2 comments:

`the module name that was saved doesn't exist anymore (it was removed or renamed)`
`the module flex facet doesn't have an actual app.yaml file specified`

I checked the behavior for both of these (see details above in my response), so I think its good to go here. 

@patflynn @nkibler, ptal again

PR #1504, opened 2017-06-13
Replace process termination logic with Java8 Process#isAlive

etanshaul:
Came across this class when working on other things. Now that we are on Java8 this can be cleaned up (and a todo in the code can be checked off).


etanshaul:
note if you look under the hood at `Process#isAlive`, this update is equivalent to what was there before.

PR #1497, opened 2017-06-12
Adds an app engine config directory chooser to the flex facet ui

etanshaul:
In support of #1373 

Breaking up #1373 into two PRs for reviewability:

1) (this pr) - adds a new field to the flex facet for setting an additional config directory - see screenshots below
2) (next pr) - consumes this field for deploying additional assets during flex deployment

UI changes:

valid state
image

invalid directory state:
image

PR #1506, opened 2017-06-14
Enable deployment of additional configs for flex deployment

etanshaul:
finishes up #1373

@patflynn @nkibler.  Still needs some tests, but please let me know what you think of this approach.

This enables the deploy additional config checkbox for flex deployments:
screen shot 2017-06-14 at 4 58 04 pm

Open question:
- As implemented, the additional configs are pulled from the "additional app engine config" directory as configured in the flex facet (#1497). If this directory is misconfigured (not a real directory, unspecified etc.), then deployment will proceed as is and silently skip deploying all configs - even if the checkbox is checked. Only a warning will be logged. I'm still trying to determine if/how we want to inform the the user about this; any suggestions welcome.

Additional notes:
- Got rid of some extra noise at the beginning of AppEngineDeploy#deploy that was determining app.yaml name for flex because, for now, we only allow it to be called app.yaml


patflynn:
I'm curious what your feeling on this feature is. I'm personally starting
to have second thoughts. I get an icky feeling about the appengine 'config'
directory in the module facet and the additional services deployment
checkbox for flex.

Maybe just remove completely for now?  Sorry I know this is a bit late in
the process.

Eventually I think a reasonable experience could simply be letting users
right click deploy a file that matches the right name... also menu actions..

etanshaul:
I hear you, but without this, or something similar like you suggest for
down the road, isn't the experience limited (without the ability to deploy
these configs)? I agree that the config directory widget in the facet feels
a bit convoluted, but I just want to make sure we make the right trade off.

patflynn:
How about just adding right-click deploy plus documentation for now and
seeing how users react?

etanshaul:
Just to make sure we're on the same page. Currently, without this feature,
a flex user won't have any ability within the IDE to deploy these assets
which would cause projects requiring them not to work (without getting them
deployed outside the IDE).

Assuming this is correct, then even with the right-click feature, a user
would have to do this separately from the deploy of their app. I.e first
deploy the app, then deploy the desired config - a two step process,
whereas with gcloud they can just "gcloud app depoy [all deployables]".

If we can't find a good UX for this then your suggestion for just
right-click is probably the best. But throwing another suggestion: what
about having in the deploy run config some sort of widget where the user
can manually browse to a set of files to deploy alongide the artifact?

patflynn:
I'm not enormously concerned by that simply because it's true about the
flex deployment feature today.  I do want to get to something richer, but
I'm a little worried about storing module state about the appengine config
file locations.  I also just kind of don't like the way that it clutters
the UI and makes the user think about it. so I'm tempted to spend a little
more time figuring out how to better integrate it in a way that's true to
the scope of cron, indexes, etc in App Engine projects.

The concern that apps will be broken if we don't deploy indexes at app
deploy time is actually kind of mitigated by the fact that it's broken even
if you deploy them at the same time because the indexes usually take longer
to build than for the app to be deployed.  That's one of the reasons why
the cloud sdk team decoupled the deployment of those files.

I am a bit concerned that I'm focusing on the multi-service scenario when
the common case is probably single service (maybe?)

I'm happy to brainstorm ways we could have a cleaner model.

etanshaul:
I agree. I can table these changes for now (reverting the config directory UI change) until we come up with a nicer solution.

etanshaul:
closing this PR

PR #1519, opened 2017-06-27
- 'set --o nounset' so that use of uninitialized variable causes failure

patflynn:
- fail if any other command returns an error.

I considered using 'set -o errexit' but it seemed like losing the
 benefit of specific error messages for which command failed wasn't
  worth it.  Open to argument about that.


etanshaul:
is the `nounset` doing anything here (in the context of the other changes you made)? Or is it just to guard against general uses of unitialized vars?

patflynn:
general guard.

PR #1520, opened 2017-06-27
- sudo install gothub.

patflynn:


etanshaul:
I wonder though if this will require a password entry (and will fail since its not interactive)

patflynn:
Yeah I don't know, the kokoro team suggested it so hopefully it works.

PR #1522, opened 2017-06-27
Fix display path truncation to be system independent

### LINKED ISSUE #1521: app.yaml path in deployment config isn't visually truncated in windows
Doesn't affect functionality, but instead of displaying a more shortened project-relative path, it shows the entire absolute path.

etanshaul:
fixes #1521


etanshaul:
Path is now properly relativized for display in windows (and the os delimiters are preserved):

image

PR #1536, opened 2017-07-05
Rename Docker language to prevent conflicts with other plugins

### LINKED ISSUE #1535: Plugin conflicts with IDEA .ignore plugin
When trying to run alongside the ignore plugin (https://plugins.jetbrains.com/plugin/7495--ignore) we get a plugin conflict warning:

image

From the logs:
image

This is caused by both plugins registering a language with the "Docker" id.

etanshaul:
fixes #1535


etanshaul:
Only visible change is in the 'language template' setting:

image

Also, this ID change is a more appropriate name for what it represents.

Splaktar:
Thank you for fixing this! Looking forward to the plugin update so that I can use the .ignore plugin again!

Issue #1244, opened 2017-02-14
Setup a nightly release to 'nightly' or 'alpha' channel.


patflynn:
We should setup a nightly release to an appropriate intellij repo channel for people who want to live on the bleading edge.

etanshaul:
fixed via #1543

PR #1542, opened 2017-07-06
Publish plugin to the nightly JB repo channel

etanshaul:
#1244 

Not exactly nightly, but a simple solution that uses the kokoro continuous integration for deployment to the nightly channel

### DIFF
--- a/kokoro/ubuntu/continuous.sh
+++ b/kokoro/ubuntu/continuous.sh
@@ -28,5 +28,17 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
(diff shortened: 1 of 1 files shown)


nkibler:
I've been thinking.. do we want to tie the continuous build to a nightly release like this? Would it make more sense to automate release cutting according to a schedule (i.e. nightly) and push it to different channels depending on the type of release we're making? I'm also not sure what's possible with Kokoro.

etanshaul:
@nkibler, that was my original goal, but as far I could tell it isn't possible with Kokoro.

In terms of the types of channels, JB provides 3: the main release, "nightly", and "alpha". I was basically just re-purposing the nightly channel as a continuous deployment channel (makes it continuously up to date with master). Feel free to disagree with this approach though.

etanshaul:
It looks like there are issues with this approach though. The continuous kokoro script is being run on PRs against master too which is not what we want.

etanshaul:
closing for now since this doesn't work as intended.

PR #1547, opened 2017-07-10
Cleanup artifact logic

etanshaul:
follow up to #1546

remove unnecessary boolean flags and require the standard facet for target artifacts in the local dev server run config

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
@@ -160,7 +160,7 @@ public class AppEngineUtil {
   }
 
   public static void setupAppEngineArtifactCombobox(@NotNull Project project,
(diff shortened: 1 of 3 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/util/AppEngineUtil.java
```diff
@@ -159,8 +159,8 @@ private static boolean doesArtifactMatchEnvironment(Artifact artifact,
     return moduleDeploymentSources;
   }
 
-  public static void setupAppEngineArtifactCombobox(@NotNull Project project,
-      final @NotNull JComboBox comboBox, final boolean withAppEngineFacetOnly) {
+  public static void setupAppEngineStandardArtifactCombobox(@NotNull Project project,
```
nkibler: nit: Can you run the format tool on this? It should put both parameters on the following line

etanshaul: done. the rest of the method also needed some formatting it seems

Issue #1407, opened 2017-05-04
Cannot apply changes in deployment run configuration if there is a validation error


chashnikov:
If I create a 'Google App Engine Deployment' run configuration in 'Run/Debug Configuration' dialog I cannot apply the changes and close the dialog if some validation error is reported (e.g. if project isn't specified or there is some problem in the Cloud SDK). So I need to either fix the problem (but it may be not trivial, e.g. in order to specify a project I may need to firstly create it) or press 'Cancel' in the dialog losing all other changes I did in it. It's very inconvenient. 

It's better to use a different approach, look like it's done for Application run configurations for example: if you don't specify Main class for an Application run configuration IDEA will show an error message, but it'll allow you to apply the changes and close the dialog. However you try to run such a configuration, an error will be shown.

chashnikov:
The problem is that you throw an exception from com.google.cloud.tools.intellij.appengine.cloud.AppEngineDeploymentRunConfigurationEditor#applyEditorTo method. You should do this only in very rare cases when it is indeed not possible to **save** the entered data. In order to validate configuration you need to implement checkConfiguration method in com.google.cloud.tools.intellij.appengine.cloud.AppEngineDeploymentConfiguration and throw exception from it.

patflynn:
Let's make this bug about making sure we do this consistently across all editors.

PR #1550, opened 2017-07-14
Adds a utility class for working with the PicoContainer in tests.

nkibler:
Progress towards #1407.

This is a test utility that replaces registered components in the MutablePicoContainer with mocks and keeps track of the original instances, which have to be replaced when the test is tearing down.


google-cloud-tools-plugin/testSrc/com/intellij/appengine/PicoContainerTestUtil.java
```diff
@@ -0,0 +1,98 @@
...
+        (MutablePicoContainer) ApplicationManager.getApplication().getPicoContainer();
+    Object originalInstance = applicationContainer.getComponentInstanceOfType(clazz);
+    applicationContainer.unregisterComponent(clazz.getName());
+    applicationContainer.registerComponentInstance(clazz.getName(), newInstance);
+    return originalInstance;
+  }
+
+  /** Represents a bound component in the {@link MutablePicoContainer}. */
+  @AutoValue
+  abstract static class Component {
```
nkibler: I thought the same thing, but figured we could benefit from it elsewhere potentially in the future. It's also only marked as a test dep so it won't affect the plugin's built size.

google-cloud-tools-plugin/testSrc/com/intellij/appengine/PicoContainerTestUtil.java
```diff
@@ -0,0 +1,98 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.intellij.appengine;
+
+import com.google.auto.value.AutoValue;
+
+import com.intellij.openapi.application.ApplicationManager;
+
```
patflynn: these imports seem oddly spaced.

nkibler: This is how the IntelliJ import optimizer is formatting it, but I agree, it shouldn't be like this. In fact, Google recently revised their style rules to say all imports should be in the same block: https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing

Should I use this approach (static in one block, others in another block) on all new files?

patflynn:
are you running google java format?

nkibler:
Yep, but I think the project is configured to split up the imports:
screenshot from 2017-07-19 12 58 46

PR #1551, opened 2017-07-14
Fixes all error-prone errors and some compiler warnings.

nkibler:
This cleans up the output of Gradle builds so it's not cluttered with ErrorProne errors or compiler warnings. Note that there are still many compiler warnings to take care of (mostly unchecked and deprecated warnings), but they aren't specifically called out in the build output because there are too many of them for the `google-account-plugin` and the `google-cloud-tools-plugin` modules.


patflynn:
How come I don't see any errorprone errors when I do gradlew buildPlugin on master?

nkibler:
Looks like error-prone isn't running for `./gradlew buildPlugin`, but it does run for `./gradlew build`. Let me dig into why that's happening.

nkibler:
Ah these error-prone issues are in the tests (under `testSrc/`), which aren't compiled for the `buildPlugin` task.

PR #1553, opened 2017-07-17
Use new syntax for applying plugins on subprojects

loosebazooka:
https://docs.gradle.org/3.4/userguide/plugins.html#sec:subprojects_plugins_dsl


build.gradle
```diff
@@ -17,17 +17,16 @@
 buildscript {
     repositories {
         maven { url "https://maven-central.storage.googleapis.com"}
-        maven { url "https://plugins.gradle.org/m2/" }
     }
     dependencies {
-        classpath 'gradle.plugin.org.jetbrains:gradle-intellij-plugin:0.1.10'
         classpath "net.ltgt.gradle:gradle-errorprone-plugin:0.0.8"
```
nkibler: Can we do this for errorprone too?

loosebazooka: I don't think errorprone is coming from the gradle plugin repository (https://plugins.gradle.org/) in this case

loosebazooka: actually I just looked it up and it can come from there

loosebazooka: lemme update the build and check.

loosebazooka:
So there's something in there which I'm not super clear on, the inclusion of the google maven repository mirror. I wonder if kokoro is slower or unaffected without that in there.

nkibler:
It was introduced in #1406 under the assumption that it does speed up builds; maybe @patflynn knows more?

patflynn:
When I tested it out I did see a decent speedup in the build times.
 alexsloan saw the opposite.

patflynn:
I'm still inclined to keep it though because gcs is less flaky than maven central

patflynn:
@loosebazooka we merging this?

loosebazooka:
I guess not, I forgot I wrote this. It's up to you guys really, if you want to.

build.gradle
```diff
@@ -19,16 +19,13 @@ buildscript {
         maven { url "https://maven-central.storage.googleapis.com"}
```
loosebazooka: This CL needs some changes if you still plan on merging, these lines are probably unnecessary, gradle pulls in all plugin dependencies when using the new format from jCenter. I would probably just rewrite it if you guys wanted to actually do this.

PR #1558, opened 2017-07-19
Adds the CloudToolsRule and @MockComponent annotation

nkibler:
Progress towards #1407.

This creates a `CloudToolsRule` that makes writing JUnit4 tests a breeze by performing a lot of userful set-up/tear-down tasks, including the creation of a separate `TestFixture` for setting up the IntelliJ's SDK state.

Here's the new functionality added by `CloudToolsRule`:

```
@Rule public final CloudToolsRule cloudToolsRule = new CloudToolsRule(this);
```

**Always:**
1. Initializes the fields annotated with `@Mock` by calling `MockitoAnnotations.initMocks(testInstance)`
2. Creates an `IdeaProjectTestFixture`, which sets up static application state for the IntelliJ SDK and creates a `Project`

**Additional Annotations:**

`@TestService`:
- Annotates fields of any type, but usually accompanies a `@Mock` annotation
- Before the test runs, it swaps the actual implementation of the service that is registered in the application's `PicoContainer` with the field's value
- After the test runs, it swaps the real implementation back
- Sample:
    ```
    @Mock @TestService private CloudSdkService mockCloudSdkService;

    @Test
    public void myTest() {
      when(mockCloudSdkService.validateCloudSdk()).thenReturn(ImmutableSet.of());
      // Do something that depends on this mocked call...
    }
    ```

`@TestFixture`:
- Annotates fields of type `com.intellij.testFramework.fixtures.IdeaProjectTestFixture`
- Injects the created `IdeaProjectTestFixture` into the annotated field, giving access to the underlying `Project`
- Sample:
    ```
    @TestFixture private IdeaProjectTestFixture testFixture;

    @Test
    public void myTest() {
      Project project = testFixture.getProject();
      // Do something with the project...
    }
    ```

`@TestModule`:
- Annotates fields of type `com.intellij.openapi.module.Module`
- Creates a new `Module` and adds it to the test fixture's `Project`, then injects the value of the created `Module` into the annotated field
- Optionally allows for the addition of a `Facet` to the `Module` by specifying the ID of the associated `com.intellij.facet.FacetType`
- Samples:
    ```
    @TestModule private Module myModule;
    @TestModule(facetTypeId = MyFacetType.ID) private Module myModuleWithFacet;
    ```

`@TestFile`:
- Annotates fields of type `java.io.File`
- Creates a new `File` with the given name and optionally the given contents
- Manages the deletion of the file after the test is torn down
- Samples:
    ```
    @TestFile(name = "my.file")
    private File myFile;

    @TestFile(name = "my-file-with.contents", contents = "Some contents")
    private File myFileWithContents;
    ```

**Improvement Example:**
Before this rule, to mock some components registered with the `PicoContainer`, you would need code like:

```
@Mock private CloudSdkService mockCloudSdkService;
@Mock private AppEngineProjectService mockAppEngineProjectService;

@Override
public void setUp() throws Exception {
  super.setUp();
  MockitoAnnotations.initMocks(this);
  PicoContainerTestUtil.getInstance()
      .replaceComponentWithMock(CloudSdkService.class, mockCloudSdkService);
  PicoContainerTestUtil.getInstance()
      .replaceComponentWithMock(AppEngineProjectService.class, mockAppEngineProjectService);
}

@Override
public void tearDown() throws Exception {
  PicoContainerTestUtil.getInstance().tearDown();
  super.tearDown();
}
```

Now, this code looks like:

```
@Rule public final CloudToolsRule cloudToolsRule = new CloudToolsRule(this);

@Mock @MockComponent private CloudSdkService mockCloudSdkService;
@Mock @MockComponent private AppEngineProjectService mockAppEngineProjectService;
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,14 +126,14 @@ A summary of its features can be found below.
 
 **Additional Annotations:**
 
-`@MockService`:
-- Annotates fields of any type, but should also have the `@Mock` annotation
+`@TestService`:
+- Annotates fields of any type, but usually accompanies a `@Mock` annotation
 - Before the test runs, it swaps the actual implementation of the service that is registered in the
-  application's `PicoContainer`
+  application's `PicoContainer` with the field's value
 - After the test runs, it swaps the real implementation back
 - Sample:
     ```
-    @Mock @MockService private CloudSdkService mockCloudSdkService;
+    @Mock @TestService private CloudSdkService mockCloudSdkService;
 
     @Test
     public void myTest() {
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,9 +101,87 @@ the Apache license found in the LICENSE file.
 A number of issues in the 
 [issue tracker](https://github.com/GoogleCloudPlatform/gcloud-intellij/issues/new)
 have been tagged as "[Help Wanted](https://github.com/GoogleCloudPlatform/gcloud-intellij/labels/help%20wanted)." 
-These are relatively small, self-contained changes that are good places to start. 
+These are relatively small, self-contained changes that are good places to start.
+
+## Project Utilities
+
+### `CloudToolsRule`
+We have a custom JUnit `TestRule` that reduces the amount of boilerplate code we have in our unit
+tests. Notably, this rule handles a lot of the set-up and tear-down required for a proper IntelliJ
+project context.
+
+For example:
+
+```
+@Rule public final CloudToolsRule cloudToolsRule = new CloudToolsRule(this);
+```
+
+A summary of its features can be found below.
+
+**Always:**
+1. Initializes the fields annotated with `@Mock` by calling
+   `MockitoAnnotations.initMocks(testInstance)`
+2. Creates an `IdeaProjectTestFixture`, which sets up static application state for the IntelliJ SDK
+   and creates a `Project`
+
+**Additional Annotations:**
+
+`@MockService`:
+- Annotates fields of any type, but should also have the `@Mock` annotation
+- Before the test runs, it swaps the actual implementation of the service that is registered in the
+  application's `PicoContainer`
+- After the test runs, it swaps the real implementation back
+- Sample:
+    ```
+    @Mock @MockService private CloudSdkService mockCloudSdkService;
+
+    @Test
+    public void myTest() {
+      when(mockCloudSdkService.validateCloudSdk()).thenReturn(ImmutableSet.of());
+      // Do something that depends on this mocked call...
+    }
+    ```
+
+`@TestFixture`:
+- Annotates fields of type `com.intellij.testFramework.fixtures.IdeaProjectTestFixture`
+- Injects the created `IdeaProjectTestFixture` into the annotated field, giving access to the
+  underlying `Project`
+- Sample:
+    ```
+    @TestFixture private IdeaProjectTestFixture testFixture;
+
+    @Test
+    public void myTest() {
+      Project project = testFixture.getProject();
+      // Do something with the project...
+    }
+    ```
+
+`@TestModule`:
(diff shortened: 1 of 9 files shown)


nkibler:
PTAL, I added a few more things related to setting up the test fixture that's really helpful for tests like the *DeploymentEditorTests.

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/MockComponent.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * <p>Now this mock can be used like any other Mockito variable:
+ *
+ * <pre>
+ *   when(mockCloudSdkService.validateCloudSdk()).thenReturn(ImmutableSet.of());
+ * </pre>
+ */
+@Target(ElementType.FIELD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface MockComponent {}
```
etanshaul: nit: but maybe MockService to better comply with the IJ lingo?

etanshaul:
do you mind adding to the description some sample code showing the consumption of the various components previously provided by the PlatformTestCase? having it documented here might help

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/MockComponent.java
nkibler: Done, also replaced "component" with "service" in the `PicoContainerTestUtil` class.

nkibler:
PTAL, added a ton of details of this change in the description.

patflynn:
The PR description is fantastic.  What about putting it in a .md file somewhere that people will find it.

nkibler:
Done, added to the `CONTRIBUTING.md` docs.

CONTRIBUTING.md
```diff
@@ -101,9 +101,87 @@ the Apache license found in the LICENSE file.
...
+**Always:**
+1. Initializes the fields annotated with `@Mock` by calling
+   `MockitoAnnotations.initMocks(testInstance)`
+2. Creates an `IdeaProjectTestFixture`, which sets up static application state for the IntelliJ SDK
+   and creates a `Project`
+
+**Additional Annotations:**
+
+`@MockService`:
+- Annotates fields of any type, but should also have the `@Mock` annotation
```
etanshaul: Not a big deal, but wondering if its possible to have only the @MockService annotation?

nkibler: Yeah you can. We can prevent it and throw an exception if the `@Mock` annotation is missing, but it's not _that_ important because it technically doesn't matter how it's mocked, or even if it isn't mocked and it's a real object. It just does a simple swap and doesn't care what the value really is.

CONTRIBUTING.md
```diff
@@ -101,9 +101,87 @@ the Apache license found in the LICENSE file.
...
+    @TestFixture private IdeaProjectTestFixture testFixture;
+
+    @Test
+    public void myTest() {
+      Project project = testFixture.getProject();
+      // Do something with the project...
+    }
+    ```
+
+`@TestModule`:
```
etanshaul: What is the connection between this Module's project (module#getProject) and the project supplied by the test fixture? Are they the same?

CONTRIBUTING.md
etanshaul: I see. so maybe the name MockService is a bit off? maybe something like @TestService?

That way, @Mock @TestService is a bit clearer?

CONTRIBUTING.md
nkibler: Yeah they should be, the modules are created via:

```
ModuleManager.getInstance(testFixture.getProject()).newModule(...);
```

CONTRIBUTING.md
nkibler: Done, renamed `@MockService` to `@TestService`.

PR #1559, opened 2017-07-21
Migrates App Engine deployment config validation to the config class

### LINKED ISSUE #1407: Cannot apply changes in deployment run configuration if there is a validation error

nkibler:
**Preface:** Sorry for the huge PR, I couldn't manage to split it up nicely after I implemented it because all of the validation changes need to be made together. Plus it's mostly tests. :)

Fixes #1407.

Instead of preventing the ability to save the deployment config settings when there's a validation issue, this change allows invalid configuration to be saved. It moves the validation to the config class itself instead of throwing exceptions in the editors. Here's an example of what the window looks like now when there's a config validation error:

screenshot from 2017-07-21 15 30 54

By pressing the "Apply" button, the configuration is saved and an error icon is shown on the run configuration dropdown:

screenshot from 2017-07-21 15 31 36

The user can either edit the configuration again by choosing "Edit Configurations..." from the dropdown, or if they attempt to run with the invalid config, the edit window will be brought up:

screenshot from 2017-07-21 15 34 01

If they attempt to press "Run" with invalid configuration, a final dialog is presented that explains the config is invalid and gives the user a chance to go back and edit it, or continue anyway:

screenshot from 2017-07-21 15 34 33

The deployment process will throw an appropriate error if the configuration is invalid for deployment.


common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/TestUtils.java
```diff
@@ -172,4 +170,54 @@ public static void assertIsSerializable(@NotNull Serializable object) {
...
+   * expectedThrowable} when executed.
+   *
+   * @param expectedThrowable the class of the exception that should be thrown
+   * @param runnable the {@link ThrowingRunnable} that should throw
+   * @param <T> the type of exception that should be thrown
+   * @return the thrown exception of type {@code expectedThrowable}
+   * @throws AssertionError if the given {@link ThrowingRunnable} throws a different type of
+   *     exception or does not throw at all
+   */
+  public static <T extends Throwable> T expectThrows(
```
etanshaul: Is this a direct copy of what we have internally?

can't we use `assertThrows` directly provided by junit? (also I think that expectThrows was deprecated in favor of assertThrows).

nkibler: AFAIK `assertThrows` is only provided in JUnit 5. The recommended way in JUnit 4 is to use either try/catch or `ExpectedException`. The try/catch idiom is lengthy and makes tests harder to read (hence why JUnit 5 introduces `assertThrows`) and `ExpectedException` will not execute statements beyond the one that immediately follows the expectation setup: https://github.com/junit-team/junit4/wiki/Exception-testing

Bringing an `expectThrows` method in now prevents clashing with the future `assertThrows`, but follows the same pattern.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
```diff
@@ -60,15 +84,10 @@ public String getGoogleUsername() {
   }
 
   @Attribute(ENVIRONMENT_ATTRIBUTE)
-  public String getEnvironment() {
+  public AppEngineEnvironment getEnvironment() {
```
etanshaul: Have you tested how this affects plugin migration? I.e users that have run confine set up with older versions

nkibler: Yep, no migration is needed. Luckily, the way IJ stores enum attributes like this is by name so this transition is safe.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
```diff
@@ -99,6 +118,16 @@ public String getModuleName() {
     return moduleName;
   }
 
+  @Attribute("appYamlPath")
+  public String getAppYamlPath() {
```
etanshaul: appYaml, and Dockerfile weren't previously part of the persisted configuration, correct?

I remember instead using moduleName as a the persisted key which can be used as a pointer into the facet configuration. How do these work together now?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
```diff
@@ -123,6 +128,12 @@ public void applyEditorTo(@NotNull AppEngineDeploymentConfiguration configuratio
     configuration.setPromote(promoteCheckbox.isSelected());
     configuration.setCloudProjectName(projectSelector.getText());
     configuration.setStopPreviousVersion(stopPreviousVersionCheckbox.isSelected());
+    configuration.setDeployAllConfigs(deployAllConfigsCheckbox.isSelected());
```
etanshaul: I need to test what actually happens in the UI but I believe deploy all configs should apply only to standard (and be hidden from flex) which is why this was not previously in the common editor class.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -232,79 +224,19 @@ protected void resetEditorFrom(@NotNull AppEngineDeploymentConfiguration configu
...
-    if (deploymentSource instanceof AppEngineDeployable) {
-      environment = ((AppEngineDeployable) deploymentSource).getEnvironment().name();
-    }
-    configuration.setEnvironment(environment);
-    configuration.setUserSpecifiedArtifact(
-        deploymentSource instanceof UserSpecifiedPathDeploymentSource);
+
+    configuration.setEnvironment(AppEngineEnvironment.APP_ENGINE_FLEX);
     configuration.setUserSpecifiedArtifactPath(archiveSelector.getText());
+    configuration.setAppYamlPath(getAppYamlPath());
```
etanshaul: similar to above comment - I suspect this (and Dockerfile) isn't needed. That is what the moduleName is for.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
etanshaul: oh I see - the flex editor is explicitly hiding it; but since its standard only config (for now) shouldn't this live in the standard editor?

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/TestUtils.java
```diff
@@ -172,4 +170,54 @@ public static void assertIsSerializable(@NotNull Serializable object) {
...
+    String message =
+        String.format(
+            "Expected '%s' to be thrown, but nothing was thrown.",
+            expectedThrowable.getSimpleName());
+    throw new AssertionError(message);
+  }
+
+  /**
+   * Returns true if the given {@code instance} can be assigned to the given {@code clazz},
+   * otherwise returns false.
```
patflynn: don't you usually {@code false} because it's a literal?

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/TestUtils.java
```diff
@@ -172,4 +170,54 @@ public static void assertIsSerializable(@NotNull Serializable object) {
...
+   */
+  private static boolean isInstanceOfType(Object instance, Class<?> clazz) {
+    String className = clazz.getName();
+    for (Class<?> type = instance.getClass(); type != null; type = type.getSuperclass()) {
```
patflynn: instanceof

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/TestUtils.java
nkibler: Yep, good catch. I must have gotten lazy. Removed this method though so it's now gone.

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/TestUtils.java
nkibler: Done, though the dynamic equivalent is Class.isInstance()

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
nkibler: Discussed offline; agreed to leave these out of the config and derive them from the facet config, which can be determined by the `moduleName`. If, for some reason, the user has issues with their facet config (e.g. missing `app.yaml`), they will be notified on deployment.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
nkibler: I wanted to separate the responsibility of setting config options to the panel that owns the UI elements; in this case, the field is in the common panel. If we want to restrict it to standard editors only, IMO the UI should also move out of the common panel and into the standard editor. WDYT?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
nkibler: Done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
```diff
@@ -138,4 +146,110 @@ public void setVersion(String version) {
...
+    }
+  }
+
+  /**
+   * Returns true if the runtime specified in the {@link #getAppYamlPath() appYamlPath} is custom,
+   * otherwise returns false.
+   *
+   * @throws MalformedYamlFileException if the {@link #getAppYamlPath() appYamlPath} is malformed
+   */
+//  private boolean isCustomRuntime() throws MalformedYamlFileException {
```
etanshaul: remove this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
etanshaul: yea good point. I forgot that the ui element lived in the common panel (this was in anticipation of implementing it for flex as well - but this was delayed). I'm fine with it as you have it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
nkibler: Whoops, thanks. Done.

nkibler:
@patflynn I'm going to merge this in, but feel free to add more comments and I'll address them in a follow-up PR.

PR #1564, opened 2017-07-25
Adds a feedback link to the cloud tools menu

### LINKED ISSUE #1560: add send feedback action to tool menu
it should open a window to the new issue page.

etanshaul:
fixes #1560

we should update the new issue template prior to releasing


etanshaul:
image

Couldn't find a great icon for this yet

chanseokoh:
It is easy to pre-fill non-sensitive information when accessing the URL (OS, Java version, Cloud SDK version, etc), and it could be helpful:

https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.bugreport/src/com/google/cloud/tools/eclipse/bugreport/ui/BugReportCommandHandler.java#L67

etanshaul:
ah, nice to know that we can dynamically prefill the form

etanshaul:
@chanseokoh out of curiosity, why do you dynamically supply the form contents in the eclipse action if you already have a static template for new issues in github? is it legacy?

chanseokoh:
The dynamic pre-fill code in CT4E is to extract and auto-fill info about users' environment, like the plugin/OS/Java/Cloud SDK versions. Users don't have to manually enter them when they open a new issue with the built-in menu in CT4E.

The static template is for users directly opening a new issue on GitHub.

etanshaul:
ah missed that. thanks

etanshaul:
yeah ill add a test.

I updated it to auto populate the issue template

etanshaul:
Updated wording:
image

Pre-filled template:
image

etanshaul:
@nkibler test added. also note the updates since your last comment.

etanshaul:
I did this so that I could test the only non-trivial logic of this class.
The public method just calls a static utility to open a browser. I suppose
I could refactor the class entirely.

nkibler:
I don't think it needs refactoring, just physically call `actionPerformed`, mock and mark a `BrowserLauncher` with `@TestService`, then expect `.browse(String, WebBrowser)` to be called with the URL you expect as the first param and `null` as the second. Does that make sense?

nkibler:
FYI I bring this up because of http://go/unit-test-practices#public-apis

etanshaul:
Ah good call. I hadn't dug into the utility to see that it calls an application service underneath.

patflynn:
thanks for doing this!

for icons what do you think about something like this guy?
screen shot 2017-07-27 at 3 21 07 pm

etanshaul:
that kind of icon is exactly what I was looking for. Where did you pull that from?

patflynn:
cloud console :).   May be worth pinging whoever is our UX resource now to
see if they have an svg of this we can use.

etanshaul:
just remembered to update the tests here, ptal.

etanshaul:
(I'm still looking into getting an icon here)

patflynn:
feel free to merge and create a new issue to update the icon.

etanshaul:
sure - can you approve @patflynn? I'll add in the icon separately.

PR #1566, opened 2017-07-25
Moves deployment list wipeout to the project-specific runtime instance

### LINKED ISSUE #1432: Deployment list gets wiped out if an IDE window is opened or closed
This is especially frustrating with long-running deployments like Flex, that disappear from the list and there is no indication of when they're finished, or the exit status. There is also no exception thrown or error logged anywhere. The Application Servers window only goes green for a bit. This is happening on Linux.

To reproduce, open or close an IDE window in another project with a populated deployment list. I only tried to open projects in the same window, although it might also be reproducible by opening in a new window.

nkibler:
Fixes #1432.

This works by moving the deployment list wipeout logic to the `AppEngineRuntimeInstance` and registering the project listener hooks to be project-specific.


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineRuntimeInstance.java
```diff
@@ -132,11 +139,18 @@ public void computeDeployments(@NotNull ComputeDeploymentsCallback callback) {
...
         runnable.cancel();
       }
       createdDeployments.clear();
     }
+
+    createdDeployments
+        .keys()
+        .forEach(
+            project ->
+                ProjectManager.getInstance()
```
etanshaul: Is this not handled by the platform as part of cleanup on project close?

chanseokoh: Even before that, `createdDeployments` is cleared just above. It's empty.

nkibler: Haha whoops, good call @chanseokoh. And @etanshaul, I'm not sure. Is there any way to check that? During deployment, is there something I can monitor (e.g. underlying process) that would close correctly if the platform was doing this?

etanshaul: look at `ProjectManagerImpl#addProjectManagerListener`

looks like its registering the removal callback with the disposer on each add

nkibler: Hmm I only see that when it's explicitly passed a `Disposable`:

```
@Override
  public void addProjectManagerListener(@NotNull final ProjectManagerListener listener, @NotNull Disposable parentDisposable) {
    addProjectManagerListener(listener);
    Disposer.register(parentDisposable, () -> removeProjectManagerListener(listener));
  }
```

PR #1575, opened 2017-07-28
Fix auto value dependency

etanshaul:
@nkibler I think I missed this during review of https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1550

But it looks like this apt scope is something copied over from android studio? In any case, its causing test compilation issues and removing it seems to fix things.

### DIFF
--- a/common-test-lib/build.gradle
+++ b/common-test-lib/build.gradle
@@ -19,5 +19,4 @@ dependencies {
     compile 'org.mockito:mockito-all:1.9.5'
 
     compileOnly 'com.google.auto.value:auto-value:1.4.1'
-    apt         'com.google.auto.value:auto-value:1.4.1'
 }


etanshaul:
On second look it seems like its coming from the gradle-apt-plugin. I need to look more to see whats going on here.

nkibler:
Right, since AutoValue is really just an annotation processor that generates code, the `apt` command should configure the `JavaCompile` task to run this processor. It should also configure IntelliJ to put the generated code in its classpath since the `idea` plugin is also applied.

I haven't seen any compilation issues, can you post repro steps and a stacktrace?

etanshaul:
@nkibler to reproduce:

1) run ./gradlew clean
2) run `CloudToolsFeedbackActionTest` from within the IDE (e.g. click "play" on the test class)

Notice the compilation error on `PicoContainerTestUtil`:

image

Oddly, I can't get it to work again by commenting out the `apt` scope like I thought I did before. I'll update if I can figure out whats going on.

etanshaul:
@nkibler 

Did we decide here to just update ide builder to delegate to gradle? If we are still running into that issue we discussed offline can you briefly describe whats going here?

nkibler:
Yeah, we should delegate the build to gradle. Unfortunately, the change doesn't seem to be version controlled; I know you were looking into that and seeing if we should track the file that changes were made in.

As far as the issue goes, even when we delegate the build to gradle, the following issue stands:
> When you try to run a unit test from the IDE after a gradle clean, it will fail. A second attempt to run the test will work.

Note that this does not affect external gradle commands (a `gradle test` after a `gradle clean` will work). With the build delegated to gradle, it has something to do with a classpath issue that I spent some time digging into but didn't get far. I suggest we track it, but keep it low priority since it's a small edge case and isn't blocking.

etanshaul:
closing as obsolete

PR #1578, opened 2017-08-02
Null check for flex environment in deployment source

### LINKED ISSUE #1576: NullPointer when opening a flex maven project
All I did was open a maven flex project and I got this:

image

It is reproducible for me.

etanshaul:
fixes #1576 

When a maven deployment source is freshly deserialized, it doesn't have an environment - this is to ensure that the environment is fresh. 
We needed a null check here.

### DIFF
--- a/google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationTest.java
+++ b/google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationTest.java
@@ -236,11 +236,26 @@ public final class AppEngineDeploymentConfigurationTest {
     assertThat(error).hasMessage("Browse to an app.yaml file.");
   }
 
+  @Test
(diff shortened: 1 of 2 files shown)


google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationTest.java
```diff
@@ -246,6 +252,15 @@ private void setUpValidFlexConfiguration() {
     configuration.setModuleName("some-module-name");
   }
 
+  private void setUpConfigurationWithNullEnvironment() {
```
nkibler: nit: Since this will only be used in this one test, you could just call `setUpValidFlexConfiguration()` then call the `when()` in the test to return `null` for the environment.

etanshaul: done

PR #1579, opened 2017-08-02
Logs errors instead of throwing RuntimeExceptions on deploy

### LINKED ISSUE #1563: IDE errors are thrown on AppEngine deploy when missing config
**Current Behavior:** If the run configuration is missing some required value to deploy to App Engine, an IDE exception is thrown. For example, if the user clears their `app.yaml` field in the facet config and attempts to deploy (after #1559), an IDE-level error is thrown.

**Expected Behavior:** A warning should be logged in the console with a descriptive message, but an IDE error should not be thrown.

**Steps to reproduce:**

1. Create a valid AppEngine deploy configuration.
2. Open the facet configuration, either through the module settings or the run configuration edit window.
3. Clear the `app.yaml` field.
4. Run the deploy task.
5. Observe the thrown IDE error:
```
Error occurred during App Engine flexible staging: no app.yaml configuration file was specified.
java.lang.RuntimeException: Error occurred during App Engine flexible staging: no app.yaml configuration file was specified.
        at com.google.cloud.tools.intellij.appengine.cloud.flexible.AppEngineFlexibleStage.stage(AppEngineFlexibleStage.java:80)
        ...
```

nkibler:
Fixes #1563.


etanshaul:
can you point me to the change (lines) that fixes the issue?

I see the refactoring: returning a boolean rather than throwing and catching a RuntimeException, but other than that I'm not seeing behavioral differences; I may have missed it though.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleStage.java
```diff
@@ -120,9 +119,20 @@ public void stage(@NotNull Path stagingDirectory) {
       if (isCustomRuntime) {
         FileUtils.copyDirectory(Paths.get(dockerDirectory).toFile(), stagingDirectory.toFile());
       }
-    } catch (IOException | InvalidPathException | MalformedYamlFileException ex) {
-      loggingHandler.print(ex.getMessage() + "\n");
-      throw new RuntimeException(ex);
+    } catch (Exception e) {
```
etanshaul: stupid style thing: at one point we had a static analysis tool failing the build for one letter variables (including exception variables). I thought this was coming from the google guide, but I could be wrong. I personally didn't like that rule anyway.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/executor/AppEngineFlexibleDeployTask.java
```diff
@@ -67,12 +65,10 @@ public void execute(ProcessStartListener startListener) {
...
-    try {
-      flexibleStage.stage(stagingDirectory);
-    } catch (RuntimeException re) {
-      deploy.getCallback()
-          .errorOccurred(GctBundle.message("appengine.deployment.exception.during.staging"));
-      logger.error(re);
+    if (!flexibleStage.stage(stagingDirectory)) {
+      String message = GctBundle.message("appengine.deployment.exception.during.staging");
+      deploy.getCallback().errorOccurred(message);
+      logger.error(message);
```
etanshaul: this logger.error will still cause an IDE level error to occur

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleStage.java
nkibler: Oh interesting, I've never heard of that. In fact, this line from the Java Practices guide calls this out and says it's ok for a few certain applications: http://go/java-practices/naming#single-letter-variables-e-i-etc

etanshaul: good. I thought it was dumb to disallow it for everything.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/executor/AppEngineFlexibleDeployTask.java
nkibler: Removed, PTAL.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/executor/AppEngineFlexibleDeployTask.java
```diff
@@ -67,12 +65,9 @@ public void execute(ProcessStartListener startListener) {
       return;
     }
 
-    try {
-      flexibleStage.stage(stagingDirectory);
-    } catch (RuntimeException re) {
-      deploy.getCallback()
-          .errorOccurred(GctBundle.message("appengine.deployment.exception.during.staging"));
-      logger.error(re);
```
etanshaul: instead of removing, you can replace with a logger.warn

nkibler: Done.

etanshaul:
looks good. did you confirm that the detailed messages are shown in the console to the user?

nkibler:
Yep, here's the full console view for a sample error (missing `app.yaml`):

```
Deploying '[2017-08-03 10:48:41] artifact.jar. Project: my-project. Version: auto'...
Created temporary staging directory: /tmp/gae-staging-my-project
Error occurred during App Engine flexible staging: no app.yaml configuration file was specified.
Failed to deploy '[2017-08-03 10:48:41] artifact.jar. Project: my-project. Version: auto': Deployment failed due to an exception while staging the project.
```

etanshaul:
is it possible to add a test that covering this change? (that failed prior to removing the log.error)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleStage.java
etanshaul: why not catch the specific exceptions anymore?

etanshaul:
One thing I just noticed that concerns me slightly - since you are no longer reporting IDE errors at all from the stage method, what about errors that happen do to file io? Previously these would have been reported.

I know I previously said I thought its ok not to report anything from that method, but I overlooked this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleStage.java
nkibler: I took them out because they weren't handled individually in any way, but this would also catch runtime exceptions we probably don't want to be caught so I'll change it back.

nkibler:
> is it possible to add a test that covering this change? (that failed prior to removing the log.error)

The modified `stage_exception` test in `AppEngineFlexibleDeployTaskTest` would fail if an IDE error is thrown because an exception would be thrown by `task.execute()`.

> since you are no longer reporting IDE errors at all from the stage method, what about errors that happen do to file io?

I can let the `stage()` method throw `IOExceptions` and let the task throw an IDE error instead of warning if it does throw, WDYT?

etanshaul:
> I can let the stage() method throw IOExceptions and let the task throw an IDE error instead of warning if it does throw, WDYT?

sgtm

nkibler:
PTAL, I've propagated `IOExceptions` to the deploy task and thrown them as IDE errors, adding tests to verify this works as expected.

etanshaul:
lgtm, but whats going on with the windows ci?

nkibler:
As discussed offline, apparently the message of the `IOExceptions` thrown when copying files is OS-dependent; since Windows does not include the text "Not a directory" as expected, the test was failing. I've removed the message validation entirely for this test and just ensure an `IOException` is propagated.

PR #1590, opened 2017-08-07
Fix nightly release by setting a timestamp as the version

etanshaul:
The nightly release is currently broken because the artifact in the JB repo will only get updated if the version of the plugin changes.

Previously, each nightly release would attempt to upload a new artifact of the same version (unless we just performed an actual production release) which wouldn't work.

This overrides the version in gradle.properties to a simple date stamp to ensure its unique for each nightly run.


etanshaul:
will test this out from this branch

etanshaul:
It works: 

https://plugins.jetbrains.com/plugin/8079-google-cloud-tools

Click on the nightly channel. See the last upload: `20170807`

PR #1601, opened 2017-08-10
Catch invalid java sdk validation exception from core lib.

etanshaul:
I'm not 100% on if this is the best approach. WDYT?

Also need some tests.


etanshaul:
fixes #1602

nkibler:
I like it, it's better than throwing an IDE error IMO. Think you can get tests in this PR too?

etanshaul:
https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/441

etanshaul:
On second thought - I think perhaps the better solution here is, on SDK validation, to pass in the project JDK to the CloudSDK builder (like we do for "normal" non-validation-only scenarios). 

This would be in addition to catching the core-lib exception.

nkibler:
Oh interesting, I didn't know that was an option. Yeah I agree, especially if we're doing that normally for other commands.

nkibler:
Ah I see, yeah then let's wait to hear from them about decoupling these two. If they don't want to decouple them, would you still be in support of the solution in this PR? It basically moves a user error (messed up system JDK) from being an IDE error to being a validation error; we still show it, it's just about how we show it and if it gets reported as a crash to us.

etanshaul:
"them" is actually us :)

I'll think about it. Talking to @patflynn offline another thing we could do is just not validate the jdk in the core lib.

patflynn:
@etanshaul close or merge?

etanshaul:
this requires changes to the core library prior to merging, and could affect other tooling too. I need to revisit that before making the decision here

PR #1605, opened 2017-08-10
Fixes EAP test failure by using the non-null Mockito contains() method

### LINKED ISSUE #1584: EAP broken due to mockito - kotlin problems
One current example: `CloudToolsFeedbackActionTest`

Mockito matchers/captors return null causing runtime exceptions for kotlin methods expecting not-null params. A workaround we discussed:

https://github.com/nhaarman/mockito-kotlin

nkibler:
Fixes #1584.

This _could_ happen to us again if Jetbrains swaps out another Java source for a Kotlin one and makes a parameter `@NotNull` that we are using a Mockito matcher/captor/etc. on. For now, this fixes the EAP test failure though.

### DIFF
--- a/google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/CloudToolsFeedbackActionTest.java
+++ b/google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/CloudToolsFeedbackActionTest.java
@@ -16,7 +16,7 @@
 
 package com.google.cloud.tools.intellij;
 
-import static org.junit.Assert.assertTrue;
+import static org.mockito.ArgumentMatchers.contains;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
@@ -31,7 +31,6 @@ import java.io.File;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
-import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
 
 /** Tests for {@link CloudToolsFeedbackAction}. */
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/CloudToolsFeedbackActionTest.java
```diff
@@ -59,23 +56,22 @@ public void testDisplayableSdkVersion_isEmpty_whenSdkIsInvalid() {
     when(sdkService.getSdkHomePath()).thenReturn(missingCloudSdk.toPath());
 
     feedbackAction.actionPerformed(null /*event*/);
-    verify(browserLauncher).browse(urlArg.capture(), eq(null));
 
-    String expectedEmptySdkVersionMessage = "- Google Cloud SDK version: \n";
-    assertTrue(urlContainsMessage(expectedEmptySdkVersionMessage));
+    String expected = "- Google Cloud SDK version: \n";
+    verify(browserLauncher).browse(urlContains(expected), eq(null), eq(null));
```
etanshaul: why is it using this overloaded version suddenly (before it was the one with 2 params)

nkibler: Discussed offline; there's something strange going on here before/after the EAP update and with/without Mockito 2. I'll take a look at this again once the Mockito 2 PR is merged to eliminate variables when debugging.

nkibler:
Abandoning for now while I figure out a solution that works for both EAP and non-EAP.

PR #1606, opened 2017-08-10
Upgrades our test code to use Mockito 2.+

nkibler:


google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/application/GoogleApiClientAppEngineAdminServiceTest.java
```diff
@@ -90,8 +87,8 @@ public void testGetApplicationForProjectId_success() throws IOException, GoogleA
 
     String projectId = "my-project";
     assertEquals(result, service.getApplicationForProjectId(projectId, mock(Credential.class)));
-    verify(appengineClientMock.apps(), times(1)).get(eq(projectId));
```
etanshaul: aren't we loosing verification by not testing the # of invocations anymore?

nkibler: Nope, the single-param `verify()` method ensures it's only called once so this parameter is unnecessary.

common-lib/src/test/java/com/google/cloud/tools/intellij/BasePluginInfoServiceTest.java
```diff
@@ -31,10 +31,10 @@
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
-import org.mockito.runners.MockitoJUnitRunner;
+import org.mockito.junit.MockitoJUnitRunner;
 
 /** Tests for {@link BasePluginInfoService}. */
-@RunWith(MockitoJUnitRunner.class)
+@RunWith(MockitoJUnitRunner.Silent.class)
```
etanshaul: can you explain the "silent" part and why some tests aren't replaced with this?

nkibler: In Mockito 2, the MockitoJUnitRunner enforces strict mocks; if you stub something on a mock that isn't called, it throws an error. The `Silent` runner will not fail the test in this scenario. I figured it might be easier to mark this common base class with `Silent` for now until we begin replacing usages of it with `CloudToolsRule`.

etanshaul: so this is just happening in base test classes where the mock is intended to be used by sub classes? I wonder if there's a way to mark it as "silent" at a more granular level rather than replacing the runner for the whole class.

nkibler: Whoops I thought this was on a different class, nevermind. I've removed the `Silent` runner here and removed the unnecessary stubs. PTAL.

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/BasePluginTestCase.java
```diff
@@ -26,18 +26,17 @@
...
 import org.junit.runner.RunWith;
-import org.mockito.runners.MockitoJUnitRunner;
+import org.mockito.junit.MockitoJUnitRunner;
 import org.picocontainer.MutablePicoContainer;
 
 /**
  * Test base class that provides a mock Intellij application and project.
  */
-@RunWith(MockitoJUnitRunner.class)
+@RunWith(MockitoJUnitRunner.Silent.class)
```
etanshaul: what about this silent?

nkibler: Ah yes, this is the one I was thinking of. I removed it and only 1 test failed (`DefaultCloudSdkVersionNotifierTest.testNotifyIfCloudSdkNotSupported_nullSdkPath`) due to an `UnnecessaryStubbingException`, so I fixed that test and removed the `Silent` runner. PTAL

PR #1608, opened 2017-08-16
Fixes EAP test failure

### LINKED ISSUE #1584: EAP broken due to mockito - kotlin problems

nkibler:
Fixes #1584.

The crux of the issue is that the EAP switched the `BrowserLauncher` abstract class to a Kotlin source. The first parameter in `BrowserLauncher.browse(String, WebBrowser)` is marked as non-null, which in Kotlin, generates a runtime check that throws an NPE if the passed value is `null` ([see here](https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#null-safety)). Most Mockito `ArgumentMatchers` return `null` when stubbing/verifying; it's an easy way for them to return a value that ultimately doesn't matter and is class-agnostic.

Couple that with the fact that they changed `BrowserLauncher.browse(String, WebBrowser)` from an abstract method to a concrete one that delegates to the 3-parameter method: `BrowserLauncher.browse(String, WebBrowser, Project)`. This method is declared final (the default for Kotlin functions) so Mockito can't stub it and the real implementation will run (with the runtime null-check).

The simple solution in this PR is to use the `BrowserLauncher` directly (instead of indirectly through the `BrowserUtil` class) and directly invoke the 3-parameter `browse(String, WebBrowser, Project)` method, which is still abstract in the Kotlin source. This way Mockito can still stub/verify the method without a hacky workaround that would be required if `BrowserUtil` was still used.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/CloudToolsFeedbackAction.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/CloudToolsFeedbackAction.java
@@ -23,7 +23,7 @@ import com.google.cloud.tools.intellij.flags.PropertiesFileFlagReader;
 import com.google.cloud.tools.intellij.ui.GoogleCloudToolsIcons;
 import com.google.cloud.tools.intellij.util.GctBundle;
 import com.google.common.net.UrlEscapers;
-import com.intellij.ide.BrowserUtil;
+import com.intellij.ide.browsers.BrowserLauncher;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.project.DumbAwareAction;
@@ -60,7 +60,7 @@ public class CloudToolsFeedbackAction extends DumbAwareAction {
 
   @Override
   public void actionPerformed(AnActionEvent event) {
-    BrowserUtil.browse(formatUrl());
+    BrowserLauncher.getInstance().browse(formatUrl(), /* browser= */ null, /* project= */ null);
   }
 
   private static String formatUrl() {
--- a/google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/CloudToolsFeedbackActionTest.java
+++ b/google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/CloudToolsFeedbackActionTest.java
@@ -16,8 +16,11 @@
 
 package com.google.cloud.tools.intellij;
 
-import static org.junit.Assert.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.contains;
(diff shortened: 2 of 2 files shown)


nkibler:
It may also help to see [the new Kotlin source](https://upsource.jetbrains.com/idea-ce/file/idea-ce-8193059effa92ee7f818d6257e6ecde193df6dc9/platform/platform-api/src/com/intellij/ide/browsers/BrowserLauncher.kt) for `BrowserLauncher`. Notice how the `browse(String, WebBrowser)` method has changed from:

```
public abstract void browse(@NotNull String url, @Nullable WebBrowser browser);
```

... to:

```
fun browse(url: String, browser: WebBrowser?) {
  browse(url, browser, null)
}
```

This is part of the problem since Kotlin functions are compiled to final methods and Mockito cannot stub this method.

etanshaul:
nit: but wrapping the utility in a service and mocking that service would also work I think. 

Also, haven't thought about it much but what if instead of using BrowserUtil, CloudToolsFeedbackAction used `BrowserLauncher#open` directly? Would this have the same Kotlin problem?

I'm ok with this though.

nkibler:
> nit: but wrapping the utility in a service and mocking that service would also work I think.

This would work by introducing untestable code; that is to say, if we do this and wanted to maintain test coverage for the service we create, we would run into the same issue. IMO having full coverage of our own code is more important.

> Also, haven't thought about it much but what if instead of using BrowserUtil, CloudToolsFeedbackAction used BrowserLauncher#open directly? Would this have the same Kotlin problem?

The NPE would still be an issue but can be solved in this instance with the `contains()` method. The method stubbing issue could be solved by invoking the `BrowserLauncher#browse(String, WebBrowser, Project)` method directly, but if you just used the `browse(String, WebBrowser)`, you would have the same issue.

Also I should have pointed out that once the EAP becomes stable, we can remove the stub hack altogether.

etanshaul:
Regarding the first point: I see what you're saying but a service would at least allow us to test that the browse utility is invoked with the proper parameter which is all we really cared about (for this test at least), without having to do this hack. It does introduce an untestable black box, but as long as its documented that its just a wrapper for the utility its fine imo.

For the second point: maybe I misunderstood, but if using `BrowserLauncher#browse(String, WebBrowser, Project)` gets around the problem then can't we just use that?

Approving since I'm OK with the current solution and will leave it up to you

etanshaul:
And yeah, since we can remove all this when the EAP goes ga then much of the discussion is moot anyway. lgtm

nkibler:
I agree that using `BrowserLauncher#browse(String, WebBrowser, Project)` is a simple solution and doesn't require the hack. It also prevents us from having to remember to clean up the hack in the future and guards us to future changes made to the `BrowserUtil` class that might break this again.

PTAL, I've changed it to use the `BrowserLauncher` directly.

Issue #1610, opened 2017-08-18
Fix custom runtime for flex deployments support


nkibler:
As we discussed and designed, we will:

- Add a field to specify the filename that the artifact will be staged as in the deployment config
  - This defaults to `target.jar` or `target.war`, depending on the type of artifact that is being deployed
- Remove Dockerfile generation support and rely on the user to create or bring their own

This affords us the following benefits:

- Our plugin can work from the same Dockerfile as other plugins used to deploy (e.g. `appengine-maven-plugin`) since there are no hidden assumptions about the name of the staged artifact (e.g. `target.war`)
- We can stop generating simple Dockerfiles that don't fit the needs of most users who are using a custom runtime, removing lots of complicated logic from our plugin
- This change is backwards compatible since the built artifact will default to `target.jar` or `target.war` if no name is provided in the run configuration; users who had previously generated Dockerfiles from our plugin that reference a `target.jar` or `target.war` artifact will continue to work

For more discussion on some of the problems and caveats, read #1151.

nkibler:
This has been completed and will be released in 2017.9.1, so I'm closing this. Note that we opted to leave the `Dockerfile` generation as-is since it can still work with this new staged artifact name, the user will just set it to be `target.jar/war`.

PR #1611, opened 2017-08-18
Adds new field for staged artifact name

nkibler:
Progress towards #1610.

When creating a deployment configuration to App Engine flexible with a custom runtime, the following field is shown:

untitled

If the user selects a deployment archive that is a WAR file and this staged artifact name field is empty, the placeholder text shows that the artifact will be staged as `target.war`:

war

Similarly, if the user selects a JAR artifact, the placeholder text shows that the artifact will be staged as `target.jar`:

jar

Finally, if the user hovers over the label for this field, the following tooltip is shown to help users understand what this new field means:

tooltip


google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -126,6 +126,8 @@ appengine.flex.deployment.cost.warning.afterLink=for pricing information.
 appengine.flex.deployment.appyaml.edit.button.text=edit
 appengine.flex.deployment.docker.directory.label=Docker directory:
 appengine.flex.deployment.use.different.docker.directory=Use a different Docker directory
+appengine.flex.deployment.staged.artifact.name.label=Staged artifact name:
+appengine.flex.deployment.staged.artifact.name.tooltip=The filename to use when staging your artifact for deployment. Directories are allowed; all paths are relative to the temporary staging directory.
```
etanshaul: "Directories are allowed"? Can you clarify?

etanshaul: Also, should probably have some wording to indicate that the Dockerfile artifact name should match up with this.

nkibler: > "Directories are allowed"? Can you clarify?

You can have something like "my/dir/with/my/artifact.war" if you want.

> Also, should probably have some wording to indicate that the Dockerfile artifact name should match up with this.

Yeah was trying to keep it short, but I'll add something in about that and reword it a bit.

etanshaul: Ah. The directory wording had me thinking that you were saying that I could reference a directory e.g. "/some/path/my-dir" instead of a file (which isn't correct). Could just be me, or we can reword.

nkibler: Ah I see your point, would using "subdirectories" help?

etanshaul: hmm, how about something like

> The filename to use when staging your artifact for deployment in the staging directory. For subdirectories, include path relative to the staging directory.

Still not perfect, but maybe an improvement.

nkibler: What about mentioning the parallel with the Dockerfile? How about:

> The filename to use when staging your artifact for deployment. This should match the reference in your Dockerfile. For subdirectories, include the path relative to the staging directory.

patflynn:
Staging won't mean much to the end-user using custom runtime, but Docker
will be front of their mind, so I think it would make sense to lead with
that.

e.g. 'This will be the path to your Java artifact in the Docker context
directory.  You'll be able to use this path in your Dockerfile to include
your artifact in your Docker image.'

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentArtifactType.java
```diff
@@ -16,38 +16,45 @@
...
    */
   @NotNull
   public static AppEngineFlexibleDeploymentArtifactType typeForPath(@Nullable Path deployPackage) {
-    if (deployPackage != null) {
-      if (deployPackage.toString().endsWith(".jar")) {
-        return JAR;
-      } else if (deployPackage.toString().endsWith(".war")) {
-        return WAR;
-      }
+    if (deployPackage == null) {
```
patflynn: I'm probably being pedantic but this is pretty much a tightened up valueOf. Why not just get rid of the extension name (which is inherent in the enum constant name), and implement as something like:
https://gist.github.com/patflynn/61d3d455ad46cf1f316596489b8b3223

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentArtifactType.java
```diff
@@ -16,38 +16,45 @@
...
+  }
 
   /**
-   * Returns the right {@code AppEngineFlexibleDeploymentArtifactType} for the given
-   * {@code deployPackage}.
+   * Returns the right {@code AppEngineFlexibleDeploymentArtifactType} for the given {@code
+   * deployPackage}.
    */
   @NotNull
   public static AppEngineFlexibleDeploymentArtifactType typeForPath(@Nullable Path deployPackage) {
```
patflynn: why do we accept null Path here?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentArtifactType.java
```diff
@@ -16,38 +16,45 @@
...
+        .orElse(UNKNOWN);
   }
 
-
-  @Override
-  public String toString() {
-    return "." + name().toLowerCase(Locale.US);
+  @NotNull
+  public String getDefaultArtifactName() {
+    return extension.isEmpty() ? "" : "target" + extension;
```
patflynn: this doesn't seem like the right place for this.  the target.jar/war is to support legacy run configs.  I'd expect it in the run configuration code manager code somewhere.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -186,14 +194,22 @@ private void resetRuntimeDisplay() {
     }
   }
 
+  private void updateStagedArtifactNameEmptyText() {
+    Path artifactPath = Paths.get(Strings.nullToEmpty(deploymentSource.getFilePath()));
+    AppEngineFlexibleDeploymentArtifactType deploymentArtifactType =
+        AppEngineFlexibleDeploymentArtifactType.typeForPath(artifactPath);
+    stagedArtifactNameTextField
+        .getEmptyText()
+        .setText(deploymentArtifactType.getDefaultArtifactName());
```
patflynn: I'm being real lazy here, but are we setting the default name to the name of the deployment artifact if this is a new run config?

nkibler: No, but I'm not sure how tell that a run configuration is "new". IDEA manages the lifecycle of these config objects and from a quick debug session, it looks like `AppEngineDeploymentConfigurator.createDefaultConfiguration(DeploymentSource)` is called repeatedly while the edit configuration window is open, even for existing configurations. I don't know when we'd have the chance to default this only for new configurations and not existing ones.

The user will have to review this field anyways and considering they are using a custom runtime with their own Dockerfile, they should know that they need to fill this out. WDYT?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentArtifactType.java
nkibler: Not sure, both invocations of it should pass in `@NotNull` paths so I've changed this, PTAL.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentArtifactType.java
nkibler: Done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentArtifactType.java
nkibler: This needs to be used by both the editor and the stage task; it doesn't make sense to make them depend on each other, so I was using the type as a common dep for this to exist. I could create a separate utility just for this purpose but it seems like overkill. WDYT?

patflynn:
Could we use a run configuration migration to change all the existing
configs to use target.war/jar and then default the editor to the artifact
name?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -186,14 +194,22 @@ private void resetRuntimeDisplay() {
...
+  }
+
   private void reloadAppYamls(Project project) {
-    appYamlCombobox.setModel(new DefaultComboBoxModel<>(
```
etanshaul: This is also a note to self, but in the future for PRs that have a lot of reformatting it could be helpful to first send a PR with just the file reformatting, followed by the PR with the actual logic changes. Makes it easier to tell what has or hasn't non-trivially changed.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentArtifactType.java
```diff
@@ -16,38 +16,38 @@
...
     }
-    return UNKNOWN;
   }
 
-
-  @Override
-  public String toString() {
-    return "." + name().toLowerCase(Locale.US);
+  @NotNull
+  public String getDefaultArtifactName() {
```
etanshaul: I think after the updates that we discussed offline to have a migration to cover "old" run configs, then we can stop thinking of artifacts of type `target.xxx` as the default name. This may just be a naming thing.

etanshaul:
Looks good to me so far. I think after the run config migration as @patflynn suggested, this will be good to go.

nkibler:
Added the `StagedArtifactNameConverterProvider`, PTAL while I clean up the `target.jar/war` logic elsewhere.

patflynn:
So what does this look like for new users who create a run config? 

Does the staged artifact name (which should match the deployment source artifact name) show up light grey and they can override it?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentArtifactType.java
patflynn: +1 I guess we don't need this anymore?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleStage.java
```diff
@@ -111,10 +111,16 @@ public boolean stage(@NotNull Path stagingDirectory) throws IOException {
         }
       }
 
-      Path stagedArtifactPath =
-          stagingDirectory.resolve(
-              "target"
-                  + AppEngineFlexibleDeploymentArtifactType.typeForPath(deploymentArtifactPath));
+      String stagedArtifactName = deploymentConfiguration.getStagedArtifactName();
+      if (Strings.isNullOrEmpty(stagedArtifactName)) {
+        stagedArtifactName =
```
patflynn: is this right? shouldn't the stagedArtifactName be the deployment source artifact name if this value is empty?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -186,14 +194,22 @@ private void resetRuntimeDisplay() {
     }
   }
 
+  private void updateStagedArtifactNameEmptyText() {
+    Path artifactPath = Paths.get(Strings.nullToEmpty(deploymentSource.getFilePath()));
```
patflynn: this logic needs to be updated now doesn't it?

empty text should always be the source artifact name.

nkibler:
Yeah sorry, I pushed the new migration logic first before finishing the clean-up since it adds a ton of code; was trying to break up the review that way. I'm almost finished with the clean-up and will push it shortly.

patflynn:
cool!

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentArtifactType.java
nkibler: Done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleStage.java
nkibler: Done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
nkibler: Done.

nkibler:
PTAL, I'm taking new screenshots now and will update the PR description with them.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/migration/StagedArtifactNameConverterProvider.java
```diff
@@ -0,0 +1,125 @@
...
+    private static final Predicate<Element> HAS_ARTIFACT_PATH =
+        element -> element.getAttribute(USER_SPECIFIED_ARTIFACT_PATH_ATTRIBUTE) != null;
+
+    @Override
+    public boolean isConversionNeeded(RunManagerSettings runManagerSettings) {
+      return runManagerSettings
+          .getRunConfigurations()
+          .stream()
+          .flatMap(TO_DEPLOYMENT_SETTINGS)
+          .anyMatch(IS_STAGED_ARTIFACT_NAME_NULL.and(HAS_ARTIFACT_PATH));
```
etanshaul: A little confused why you are AND'ing with HAS_ARTIFACT_PATH.

The user specified path stuff is for the case when the user browses to an arbitrary jar/war on the filesystem. So why would having that path be a requirement for conversion?

etanshaul: isn't the goal of the migrator to determine if its an "old" run config (e.g. one that was created prior to these change) and then, if so, default the artifact field to "target.jar/war"?

If its determined _not_ to be an old configuration then the field can be null/empty which indicates that we should use the real artifact in the deployment (which can be populated later, not here in the migration).

Am I missing something?

nkibler: If we only supported one type of artifact (i.e. only WARs), you're right; we wouldn't need to know anything about the artifact. But since we support WARs and JARs and we had separate Dockerfile templates for each, we need to determine whether the artifact is a WAR or a JAR and default the name accordingly.

For example, the deployment source could be a Maven artifact that produces a JAR. If we need to migrate the deployment config, we need to set the staged artifact name to `target.jar`. But to know this, we have to find out what type of deployment source it is and find the right field in the XML that holds the name of the artifact.

Does this make sense?

etanshaul: aha I forgot about the jar/war distinction.

Perhaps we can do something clever (hacky?): set the artifact name to just "target" (no extension), and have the deployment config check this. Or perhaps even better, we could have a boolean field that marks this run config as legacy, then the deployment config sets the artifact name to "target.jar/war" if this field is true.

nkibler: When you say "the deployment config sets the artifact name", do you mean the staging task? I don't think we should change the getters in `AppEngineDeploymentConfiguration` to do anything other than return the value deserialized from the XML.

Also, I feel like using a converter to set a config as "legacy" is strange and it means we will always have to support this legacy field going forward. Part of the attraction of using the converter is that we don't have to support any legacy logic in our actual plugin's code (everything is in the converter) and we can delete the converter if we really want to at any point.

IMO, keeping this complexity in the converter is the right move. I wouldn't mind taking the approach I mentioned in my first comment in this chain, which is parsing a different field for non-user specified artifacts to find the artifact name. It's not as bad as it sounds (hopefully). :)

WDYT?

etanshaul: both good points. SGTM.

nkibler:
Jumping back to this; a proper solution to migrate the configs is non-trivial because the project is not-yet-loaded at the time of migration. This makes determining the deployment source's artifact name extremely difficult because you would have to manually parse XML files to find it and each source type has its own convention of where these files are and what their contents are.

Could we reconsider not doing any migration for users with existing run configurations using a custom runtime on flex? This would drastically simplify the solution and the impact to the user is a one-time error on deployment.

etanshaul:
@nkibler just to refresh my memory, and to make it clear on this thread, could you please describe what the user experience will be after making this change without migration support?

nkibler:
Without migration, the artifact staged will only be renamed if there is a value in the run config for `stagedArtifactName`. Otherwise, it's copied over without renaming.

For users who are deploying to App Engine flex with a custom runtime and have a `Dockerfile` that references a name other than the artifact's name (e.g. `target.war`), they will see an error when attempting to deploy that says:

> Step 2/2 : ADD target.war /webapps/root.war
lstat target.war: no such file or directory

We could include a log during staging that mentions what filename the artifact is being staged as so users who see this error might understand what to do more easily.

etanshaul:
> We could include a log during staging that mentions what filename the artifact is being staged as so users who see this error might understand what to do more easily.

Sounds good.

Let's go ahead and separate migration from the rest of this and merge in the PR without migration (get rid of all the migration stuff from here). We can then discuss offline with @patflynn since he had a stake in this to determine if we release without it. Sound good?

PR #1615, opened 2017-08-23
Google Cloud Storage integration setup

etanshaul:
As discussed, I'll be splitting this feature up into small reviewable PRs. This PR includes:

- plugin.xml entry point for the GCS panel (commented out so that the feature is hidden until ready)
- GCS API added to the build
- empty GCS tools panel:

screen shot 2017-08-22 at 9 06 30 pm

screen shot 2017-08-22 at 9 06 42 pm

todo: @nkibler I tried using the new `@CloudToolsRule` instead of extending from PlatformTestCase for the simple test in this PR, but the tests failed. Need to look into it more to see why.

### DIFF
--- a/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
+++ b/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
@@ -336,6 +336,9 @@ clouddebug.runconfig.formoredetails=For more details,
 clouddebug.runconfig.documentation.url=https://cloud.google.com/debugger/
 clouddebug.runconfig.documentation.url.text=click here
 
+gcs.panel.toolbar.refresh.hover.text=Refresh
+gcs.panel.toolbar.refresh.hover.description=Refresh Google Cloud Storage buckets
+
 settings.error.closing.file=Error closing settings file {0}.\nError details: {1}
 #{0} and {1} in the below message represent open and closing tags of a hyperlink.
 appengine.more.info=For more information about App Engine, please visit {0}this{1} page.
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsToolWindowPanel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsToolWindowPanel.java
@@ -27,7 +27,7 @@ import com.intellij.openapi.ui.SimpleToolWindowPanel;
 import com.intellij.util.ui.JBUI;
 import javax.swing.JPanel;
 
-/** Creates a Google Cloud Storage side tool panel. */
+/** Creates a Google Cloud Storage tool panel. */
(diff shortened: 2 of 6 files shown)


google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -336,6 +336,9 @@ clouddebug.runconfig.formoredetails=For more details,
 clouddebug.runconfig.documentation.url=https://cloud.google.com/debugger/
 clouddebug.runconfig.documentation.url.text=click here
 
+gcs.panel.toolbar.refresh.hover.text=Refresh
```
patflynn: instead of having a refresh button could we listen for changes?
https://cloud.google.com/storage/docs/pubsub-notifications

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsToolWindowPanel.java
```diff
@@ -0,0 +1,65 @@
...
+import com.intellij.openapi.actionSystem.ActionManager;
+import com.intellij.openapi.actionSystem.ActionToolbar;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.actionSystem.DefaultActionGroup;
+import com.intellij.openapi.project.DumbAwareAction;
+import com.intellij.openapi.ui.SimpleToolWindowPanel;
+import com.intellij.util.ui.JBUI;
+import javax.swing.JPanel;
+
+/** Creates a Google Cloud Storage tool panel. */
```
patflynn: it looks like it defines the panel rather than creating it.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: I like this. But how to do feel about adding this as an enhancement after the initial implementation? I wont block the PR on this unless you feel strongly about doing this now.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsToolWindowPanel.java
etanshaul: Good point

patflynn:
Nope. I approved.

patflynn:
being too terse. 'nope' as in do not block.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: note to self for future implementation: probably already implied but to be explicit - when implement a listener like this, we'd wouldn't automatically refresh the list. Instead we'd have some text at the top, for example:

`Bucket changes detected. Click here to refresh the list.`

PR #1619, opened 2017-08-25
GCS bucket content explorer setup

etanshaul:
This is a partial functionality PR to keep things small(ish). It sets up the GCS bucket content explorer editor. 

The editor itself is unfinished. It just lists all the bucket contents in a flattened view (and just the blob name). **Everything else in this editor is unfinished (all UI polish, icons, toolbars, breadcrumbs, directory structure etc.).**

### Some info on creating custom editor windows:
Custom editors, created by using the `FileEditorProvider` extension point, are backed by `VirtualFile`s. This makes sense since most editor use-cases involve opening up some sort of real file. In our case, we want to open up an editor, but there is no "file" - instead we have a `Bucket` datastructure (from the GCS API). To achieve this, I created a GCS VirtualFile implementation which basically wraps a GCS bucket. In order to get the full functionality of the editor UI, I created several of these file related components: `GcsBucketVirtualFile`, `GcsBucketFileType` etc.

Bucket explorer editor window after opening a bucket in the left panel:
screen shot 2017-08-24 at 9 15 16 pm

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.intellij.gcs;
+
+import com.intellij.codeHighlighting.BackgroundEditorHighlighter;
+import com.intellij.openapi.fileEditor.FileEditor;
+import com.intellij.openapi.fileEditor.FileEditorLocation;
+import com.intellij.openapi.fileEditor.FileEditorState;
+import com.intellij.openapi.util.Key;
+import java.beans.PropertyChangeListener;
+import javax.swing.JComponent;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Google Cloud Storage {@link FileEditor} implementation. Instantiates the UI panel that allows
+ * browsing of GCS bucket contents.
+ */
+final class GcsBucketContentEditor implements FileEditor {
+
+  private final GcsBucketContentEditorPanel bucketContentPanel;
+  private final GcsBucketVirtualFile bucketVirtualFile;
+
+  GcsBucketContentEditor(@NotNull GcsBucketVirtualFile bucketVirtualFile) {
+    this.bucketVirtualFile = bucketVirtualFile;
+    bucketContentPanel = new GcsBucketContentEditorPanel(bucketVirtualFile);
+  }
+
+  @NotNull
+  @Override
+  public JComponent getComponent() {
+    return bucketContentPanel.getComponent();
+  }
+
+  @Nullable
+  @Override
+  public JComponent getPreferredFocusedComponent() {
+    return bucketContentPanel.getComponent();
+  }
+
+  @NotNull
+  @Override
+  public String getName() {
+    return "gcs-bucket-content";
+  }
+
+  @Override
+  public void setState(@NotNull FileEditorState state) {}
+
+  @Override
+  public boolean isModified() {
+    return false;
+  }
+
+  @Override
+  public boolean isValid() {
+    return bucketVirtualFile.isValid();
+  }
+
+  @Override
+  public void selectNotify() {}
+
+  @Override
+  public void deselectNotify() {}
+
+  @Override
+  public void addPropertyChangeListener(@NotNull PropertyChangeListener listener) {}
+
+  @Override
+  public void removePropertyChangeListener(@NotNull PropertyChangeListener listener) {}
+
+  @Nullable
+  @Override
+  public BackgroundEditorHighlighter getBackgroundHighlighter() {
+    return null;
+  }
+
+  @Nullable
+  @Override
+  public FileEditorLocation getCurrentLocation() {
+    return null;
+  }
+
+  @Override
+  public void dispose() {}
+
+  @Nullable
+  @Override
+  public <T> T getUserData(@NotNull Key<T> key) {
+    return null;
+  }
+
+  @Override
+  public <T> void putUserData(@NotNull Key<T> key, @Nullable T value) {}
+}
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
@@ -49,7 +49,7 @@ final class GcsBucketContentEditor implements FileEditor {
   @Nullable
   @Override
   public JComponent getPreferredFocusedComponent() {
-    return bucketContentPanel.getComponent();
+    return getComponent();
   }
 
   @NotNull
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
@@ -39,7 +39,8 @@ final class GcsBucketContentEditor implements FileEditor {
 
   GcsBucketContentEditor(@NotNull GcsBucketVirtualFile bucketVirtualFile) {
     this.bucketVirtualFile = bucketVirtualFile;
-    bucketContentPanel = new GcsBucketContentEditorPanel(bucketVirtualFile);
+    bucketContentPanel = new GcsBucketContentEditorPanel();
+    bucketContentPanel.setTableModel(bucketVirtualFile);
   }
 
   @NotNull
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
@@ -32,6 +32,8 @@ import org.jetbrains.annotations.Nullable;
  */
 final class GcsBucketContentEditor implements FileEditor {
 
+  private static final String GCS_FILE_EDITOR_NAME = "gcs-bucket-content";
+
   private final GcsBucketContentEditorPanel bucketContentPanel;
   private final GcsBucketVirtualFile bucketVirtualFile;
 
@@ -55,7 +57,7 @@ final class GcsBucketContentEditor implements FileEditor {
   @NotNull
   @Override
   public String getName() {
-    return "gcs-bucket-content";
+    return GCS_FILE_EDITOR_NAME;
   }
 
   @Override
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
@@ -40,7 +40,7 @@ final class GcsBucketContentEditor implements FileEditor {
   GcsBucketContentEditor(@NotNull GcsBucketVirtualFile bucketVirtualFile) {
     this.bucketVirtualFile = bucketVirtualFile;
     bucketContentPanel = new GcsBucketContentEditorPanel();
-    bucketContentPanel.setTableModel(bucketVirtualFile);
+    bucketContentPanel.setTableModel(bucketVirtualFile.getBucket().list().iterateAll());
   }
 
   @NotNull
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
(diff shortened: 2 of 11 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
```diff
@@ -0,0 +1,109 @@
...
+  @NotNull
+  @Override
+  public JComponent getComponent() {
+    return bucketContentPanel.getComponent();
+  }
+
+  @Nullable
+  @Override
+  public JComponent getPreferredFocusedComponent() {
+    return bucketContentPanel.getComponent();
```
nkibler: Maybe `return getComponent()` instead? It's functionally the same but lets the reader know that it should be the same component as `getComponent()`.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @NotNull
+  @Override
+  public String getName() {
+    return "gcs-bucket-content";
```
nkibler: What is this name used for? Does it make sense to pull into a constant somewhere else?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @Override
+  public void setState(@NotNull FileEditorState state) {}
+
+  @Override
+  public boolean isModified() {
+    return false;
```
nkibler: I'm guessing this (and the unimplemented methods below) will be filled out later?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
```diff
@@ -0,0 +1,88 @@
...
+  private JPanel bucketContentEditorPanel;
+  private JPanel bucketContentToolbarPanel;
+  private JPanel breadCrumbsPanel;
+  private JTable bucketContentTable;
+
+  private static final List<String> TABLE_COL_HEADER =
+      Arrays.asList("Name", "Size", "Type", "Last Modified");
+
+  GcsBucketContentEditorPanel(@NotNull GcsBucketVirtualFile bucketVirtualFile) {
+    Iterable<Blob> blobs = bucketVirtualFile.getBucket().list().iterateAll();
```
nkibler: Generally, it's discouraged to have constructors do any real work because it "removes seams needed for testing, forcing subclasses/mocks to inherit unwanted behavior. Too much work in the constructor prevents instantiation or altering collaborators in the test." - go/cl-constructors

What about having an instance field for this `GcsBucketVirtualFile` and expose a method to fill the table with data?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketEditorProvider.java
```diff
@@ -0,0 +1,53 @@
...
+  @NotNull
+  @Override
+  public FileEditor createEditor(@NotNull Project project, @NotNull VirtualFile file) {
+    return new GcsBucketContentEditor((GcsBucketVirtualFile) file);
+  }
+
+  @NotNull
+  @Override
+  public String getEditorTypeId() {
+    return "gcs-bucket-content-editor";
```
nkibler: Ditto about maybe needing a constant if this actually means something and needs to be unique or easily findable.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
```diff
@@ -44,6 +51,8 @@
  */
 final class GcsBucketPanel {
 
+  private Project project;
```
nkibler: final?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
```diff
@@ -69,6 +79,21 @@ protected void textChanged(DocumentEvent event) {
                 refresh();
               }
             });
+
+    bucketList.addMouseListener(
+        new MouseAdapter() {
+          @Override
+          public void mouseClicked(MouseEvent event) {
+            if (event.getClickCount() == 2) {
```
nkibler: There's really no double-click listener? Poor Swing :(

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketVirtualFile.java
```diff
@@ -0,0 +1,180 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A {@link VirtualFile} representation of a Google Cloud Storage bucket.
+ *
+ * <p>This is essentially just a wrapper for a {@link Bucket} needed in order to create a custom
+ * editor window for exploring GCS bucket contents. Custom editors are created by providing {@link
+ * FileEditorProvider} implementations which are backed by VirtualFiles.
+ */
+class GcsBucketVirtualFile extends VirtualFile {
```
nkibler: Can this be final?

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanelTest.java
```diff
@@ -0,0 +1,76 @@
...
+  public void testBucketContentTableInitialization() {
+    List<Blob> blobs = Lists.newArrayList(blob);
+    when(blobPage.iterateAll()).thenReturn(blobs);
+    when(blob.getName()).thenReturn("blobName");
+
+    editorPanel = new GcsBucketContentEditorPanel(bucketVirtualFile);
+
+    JTable bucketTable = editorPanel.getBucketContentTable();
+    assertThat(bucketTable.getColumnCount()).isEqualTo(4);
+    assertThat(bucketTable.getRowCount()).isEqualTo(1);
```
nkibler: Could you assert that the info in the row is what you would expect?

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/gcs/GcsBucketEditorProviderTest.java
```diff
@@ -0,0 +1,51 @@
...
+import com.google.cloud.tools.intellij.testing.TestFixture;
+import com.intellij.openapi.vfs.newvfs.impl.StubVirtualFile;
+import com.intellij.testFramework.fixtures.IdeaProjectTestFixture;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.mockito.Mock;
+
+/** Tests for {@link GcsBucketEditorProvider}. */
+public class GcsBucketEditorProviderTest {
```
nkibler: What about testing the other methods, like `createEditor()`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
etanshaul: not necessarily. Some may be implemented some will be left as is. isModified for example is used to determine if the underlying virtual file was modifed externally so that things like updating the icon of the editor (like with a `*` can be done. If we have a simple refresh button its not needed. But if we do something like @patflynn suggested and listen to changes over pubsub then I could see hooking into this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
etanshaul: yeah, I thought the same thing ..

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
etanshaul: the table model needs to be set on initialization of the panel. I could perhaps move the setting of the model to the `.form` itself but that's not really helping your concern. Maybe I didn't understand your suggestion?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketVirtualFile.java
etanshaul: its being mocked in a test, so no :( (at least not without moving things around). Apparently there is an incubating feature in Mockito 2 to allow this:
https://github.com/mockito/mockito/wiki/What%27s-new-in-Mockito-2#unmockable

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanelTest.java
etanshaul: I could assert that the column headers match - I'll add that. Row data is dynamic so obviously that wouldn't make sense.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/gcs/GcsBucketEditorProviderTest.java
etanshaul: done

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanelTest.java
etanshaul: done

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
etanshaul: done

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
etanshaul: done

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketEditorProvider.java
etanshaul: from the docs:

> @return id of type of the editors that are created with this FileEditorProvider. Each FileEditorProvider should have unique non null id. The id is used for saving/loading of EditorStates.

I'll move it to a class-level constant

etanshaul: done - i just made a private field. There isn't any sort of registry for these. Also the platform implementations of FileEditorProvider that I saw do the same thing.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditor.java
etanshaul: This is the name that identifies an open editor instance. I'm not 100% sure what its used for internally. Its not used for the name of the editor (the name you see in tab in the gui). its also not used to prevent multiple editors of the same "kind" from being opened (I had to add some logic myself to do this: see GcsBucketPanel#findOpenGcsVirtualFile

I'll just move it to a private constant to make it more visible

etanshaul: done

etanshaul:
@nkibler, I think I addressed all your comments.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
nkibler: This class is instantiated by `GcsBucketContentEditor`, so you control the lifecycle of this object. I'm suggesting to move out the API call with the model initialization to an instance method and call the method from `GcsBucketContentEditor` when you want (which in this case, might be right after you initialize it).

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanelTest.java
nkibler: > Row data is dynamic so obviously that wouldn't make sense.

Isn't the data in the row pulled from the Blob itself, which you have mocked here in this test? It looks like the code under test only sets the name of the blob for now and puts dashes in the other columns. Could you assert that the name in the first row is set to "blobName" and the rest are dashes for now?

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/gcs/GcsTestUtils.java
```diff
@@ -0,0 +1,44 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import com.google.api.gax.paging.Page;
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.Bucket;
+import java.util.Iterator;
+
+/** Utilities for GCS related tests. */
+class GcsTestUtils {
```
nkibler: nit: final?

etanshaul: about the row data - yeah, I'll update. Not sure what I was thinking.

etanshaul:
@nkibler updated from latest comments. ptal

etanshaul:
`This class is instantiated by GcsBucketContentEditor, so you control the lifecycle of this object. I'm suggesting to move out the API call with the model initialization to an instance method and call the method from GcsBucketContentEditor when you want (which in this case, might be right after you initialize it).`

Missed this one. I'll address it.

etanshaul:
is this (https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1619/commits/fa2bc2491e5976cba15618a4fd9cf5f820e82eca) what you had in mind?

PR #1622, opened 2017-08-30
Adds editor listener to update staged artifact name placeholder

### LINKED ISSUE #1621: Staged artifact name field missing placeholder text
screen shot 2017-08-29 at 11 21 38 am

This was for an existing run configuration. I expected to see some sort of placeholder text in the field indicating what the name of the artifact will be. Without it, its unclear if the user needs to set it or not.

nkibler:
Fixes #1621.

Previously there was just a listener on the artifact path field to update the staged artifact name placeholder text. This would not be triggered if the user selects (or previously had selected) a non-user specified source, like a Maven target. This adds an additional listener to the state of the settings editor itself, which receives updates any time the state has changed (e.g. when a deployment source has been selected).

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
@@ -126,7 +126,6 @@ public final class AppEngineFlexibleDeploymentEditor
                           AppEngineFlexibleDeploymentEditor.this.deploymentSource)
                       .setFilePath(archiveSelector.getText());
                 }
-                updateStagedArtifactNameEmptyText();
               }
             });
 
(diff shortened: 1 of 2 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -97,6 +97,8 @@ public AppEngineFlexibleDeploymentEditor(Project project, AppEngineDeployable de
     commonConfig.getDeployAllConfigsCheckbox().setSelected(false);
     commonConfig.getDeployAllConfigsCheckbox().setVisible(false);
 
+    addSettingsEditorListener(editor -> updateStagedArtifactNameEmptyText());
```
etanshaul: since this listens to all changes in the editor, will we run into double calls to `updateStagedArtifactNameEmptyText` when its called on archive selector's doc listener?

nkibler: Ah good point, I just tested it and it looks like you're right, this handler is called even for changes to the artifact path field. I removed the line below to update as part of the document listener, PTAL.

PR #1624, opened 2017-08-31
Downgrade storage library to compatible guava version

etanshaul:
Explicitly depend on Guava 20.0

Why am I doing this?

When working on GCS integration I ran into incompatibilities between the version of Guava brought in by GCS (20) and the one brought in by the plugin (19).

This also removes the provided dep. check for guava.


etanshaul:
@loosebazooka can you ptal to make sure this isn't a dumb idea?

google-cloud-tools-plugin/build.gradle
```diff
@@ -62,6 +62,8 @@ dependencies {
         exclude group: 'org.yaml', module: 'snakeyaml'
     }
     compile files('lib/google-api-services-source.jar')
+
+    compile 'com.google.guava:guava:20.0'
```
nkibler: GCS is on 20 still? *sigh*

etanshaul: it is: https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/pom.xml#L139

etanshaul:
@loosebazooka any chance you could take a quick look to see if this makes sense?

loosebazooka:
will do

build.gradle
```diff
@@ -76,7 +76,6 @@ subprojects {
             failOnVersionConflict()
 
             def provided = [["org.apache.httpcomponents", "httpclient"],
-                            ["com.google.guava", "guava"],
                             ["com.google.api-client", "google-api-client"]]
```
loosebazooka: isn't this still provided by the intellij libs?

loosebazooka: I believe the reason this was added in was to prevent a classpath loading problem when intellij loads from https://github.com/JetBrains/intellij-community/tree/master/lib but they also seem to be on guava 21.

loosebazooka: Also I think this check was to prevent **any library** from pulling in the transitive `guava`. So this might lead to other dependency clash issues that we might be unaware of.

etanshaul: Also I think that link to IJ you put is ahead of whats released. I'm seeing guava 19 being pulled in by the IJ 2017.1 distribution.

I'm open to other approaches to solve this problem. The problem being that the GCS client lib requires guava 20.

loosebazooka:
The one weird thing about force is that it gives us another thing to track. I wonder if there's a way to ensure someone doesn't bring in something newer that what we force?

etanshaul:
Maybe we can progammatically grab the latest version of that library that is used and then force that version?

I wish there was a way to force the latest version of a single dependency (while still keeping the failOnVersionConflict strategy).

etanshaul:
something like this:

```
dependencies.all { dep ->
                if (dep.group == "com.google.api" && dep.name == "api-common") {
                    force "${dep.group}:${dep.name}:${dep.version}"
                }
            }
```

etanshaul:
nevermind. that won't force the latest.

etanshaul:
will merge as is and consider ways to improve this

patflynn:
What does 'com.google.api:api-common:1.1.0' conflict with?

etanshaul:
it conflicts with an previous RC version of the same library also brought in by the cloud storage library.

loosebazooka:
could you exclude that version specifically in the cloud storage library pull in?

etanshaul:
@loosebazooka let me try that

patflynn:
I'm confused. :)

How can the conflict be between an older version of the GCS library?
Presumably we're only using one version of GCS API?

etanshaul:
We are using a single version of GCS. But that version of GCS that I downgraded to (in order to avoid Guava conflicts) was itself bringing in 2 conflicting versions of api-common. But since we do a failOnVersionConflict, this was failing by us (they were not forcing the version and just letting both through).

patflynn:
Should we open a bug against them to have a fail on version conflict build
setting?

etanshaul:
That would be reasonable, but they _may_ do that now (I need to check). This problem is encountered in an older version of GCS that I downgraded to in order to get Guava to play nice with the provided IJ version of Guava.

etanshaul:
@loosebazooka I tried a couple of the things we discussed:

- It seems you cannot exclude a specific version of a library; the recommended approach is to use "force" which is what this PR is already doing
- Moving the resolutionStrategy directly into the compile dependency (to make things more readable) doesn't work since this is not a valid method in this context

PR #1629, opened 2017-09-05
Adds the ThreadUtil utility class

nkibler:
This utility class replaces usages of calls to `ApplicationManager.getApplication().executeOnPooledThread()`. The problem with the previous approach is that even in unit tests, it will use a background `ExecutorService` that is controlled by the IDEA SDK and is inaccessible from our tests. This makes testing code that submits runnables to this service hard to test because it's unknown when the task will be executed.

This PR adds the `ThreadUtil` class, where invocations to execute tasks in the background look like: `ThreadUtil.getInstance().executeInBackground()`. The `CloudToolsRule` also binds a direct executor to the implementation, causing tasks to be executed synchronously on the current thread for unit tests that use the `CloudToolsRule`.

This is prep work for the project selector search feature.

### DIFF
--- a/common-lib/src/main/java/com/google/cloud/tools/intellij/util/ThreadUtil.java
+++ b/common-lib/src/main/java/com/google/cloud/tools/intellij/util/ThreadUtil.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.intellij.util;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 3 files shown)


patflynn:
@brendandouglas

brendandouglas:
LGTM. You might also want to consider overriding MockApplication (e.g. as done [here](https://github.com/bazelbuild/intellij/blob/e6b919a16092b5463872d26f3d532ead5ddb64e5/base/tests/utils/unit/com/google/idea/blaze/base/TestUtils.java)).

Then there's no need to redirect all usages of executeOnPooledThread, and it's less prone to future breakages.

nkibler:
Thanks for the pointer! As of now, we are using a real `Application`, as built by the [`HeavyTestFixtureBuilderImpl`](https://upsource.jetbrains.com/idea-ce/file/idea-ce-d00d8b4ae3ed33097972b8a4286b336bf4ffcfab/platform/testFramework/src/com/intellij/testFramework/fixtures/impl/HeavyTestFixtureBuilderImpl.java). Do you know what impact replacing the real `Application` with a `MockApplication` would have in this scenario? We build our test fixture here: https://github.com/GoogleCloudPlatform/google-cloud-intellij/blob/ab7441cca97183cd33d78ca84e7e62510241130d/common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/CloudToolsRule.java#L95

brendandouglas:
Ah, MockApplication isn't suitable for integration tests. In that case it's probably fine as is.

Quite a few IntelliJ callers seem to handle this only calling 'executeOnPooledThread' when not in test mode (e.g. AbstractLayoutCodeProcessor#runLayoutCodeProcess).

nkibler:
IMO adding logic in code that is aware of whether it's being run in a test or not is often a code smell. For now, we'll keep it as is. We may explore alternatives in the future if we want to reduce the test fixture to a lightweight one.

@patflynn PTAL?

PR #1645, opened 2017-09-18
Adds the staged artifact name converter logic

### LINKED ISSUE #1151: Generated Dockerfile doesn't work with Maven plugin

nkibler:
Fixes #1151.

This adds a new legacy bit to `AppEngineDeploymentConfigurations` as a one-time conversion to maintain the legacy behavior of staging artifacts as `target.jar` or `target.war`. The staging process for flexible run configurations will stage the artifact as `target.jar/war` if this legacy bit is set, preserving past functionality. When the user opens the editor, the `target.jar/war` staged artifact name is explicitly set in their run configuration and the legacy bit is cleared; their configuration is fully "migrated" after this.

This PR also logs the name that the artifact is being staged as during deployment so the user has more information about what's going on.


google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -169,7 +169,9 @@ appengine.deployment.error.invalid.cloudsdk=Invalid Cloud SDK directory path.
 appengine.deployment.deployallconfigs.checkbox.label=Update dos, dispatch, cron, queues, and datastore indexes
 appengine.deployment.flexible.nogaesupport=No modules with Google App Engine flexible framework support found.
 appengine.deployment.error.cancelled=Deployment process was cancelled.
+appengine.deployment.staging.artifact.as=Staging artifact as: {0}
 appengine.converter.description=Convert projects created by the legacy App Engine plugin to be compatible with the new plugin
+appengine.staged.artifact.name.converter.description=Converts run configurations to set a legacy bit for preserving the old staged artifact name.
```
etanshaul: Is this message visible to users? If so maybe update the wording - its clear to us but may be too cryptic for users.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
```diff
@@ -49,6 +49,7 @@
 
   public static final String USER_SPECIFIED_ARTIFACT_PATH_ATTRIBUTE = "userSpecifiedArtifactPath";
   public static final String STAGED_ARTIFACT_NAME = "stagedArtifactName";
+  public static final String STAGED_ARTIFACT_NAME_LEGACY = "stagedArtifactNameLegacy";
```
etanshaul: Didn't we say we wouldn't call this "legacy"? I think this was @patflynn's suggestion

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -246,6 +246,16 @@ protected void resetEditorFrom(@NotNull AppEngineDeploymentConfiguration configu
     }
     archiveSelector.setText(configuration.getUserSpecifiedArtifactPath());
 
+    if (configuration.isStagedArtifactNameLegacy()) {
```
etanshaul: if it is "legacy" don't we also want to populate the UI field with placeholder text showing the name we are going to use (e.g. _.jar/war)?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
nkibler: IIRC, @patflynn's comment was to not simply call it "legacy" because it's too obscure and we may have other "legacy" things to migrate in the future.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
nkibler: I just realized that this is visible if the user selects the `Details...` link in the conversion dialog. What do you suggest?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
nkibler: By setting the configuration's staged artifact name appropriately, the field's text shows this legacy value `target.jar/war` (see line 260 below):

```
stagedArtifactNameTextField.setText(configuration.getStagedArtifactName());
```

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: Something like this maybe?

`Converter for deployment run configs that marks the deployment as using target.jar/war for the staged artifact name in App Engine flexible deployments`

Its not very elegant, but something along those lines.

etanshaul:
This is the PR I mentioned that fixed the problem I had of the converters running multiple times. We should just double check that we don't have to do anything here to prevent this (if we don't, then I'm not sure why I had to to fix this problem originally):

https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1480

etanshaul:
@nkibler This could just be an artifact of the way I'm running it (in debug / sandbox mode), but if I set a breakpoint on the `isConversionNeeded` method, it gets called every time I re-run the IDE (built from the sources).

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
nkibler: Considering this statement is prefixed with `The following conversions will be performed:`, what about:

> App Engine flexible deployment run configurations will have their staged artifact name explicitly set to target.jar/war

nkibler:
It shouldn't be a problem that `.isConversionNeeded()` is called multiple times since it has no side effects. It may call that every time but only call `process()` if it hasn't done it before.

etanshaul:
But if isConversionNeeded returns true (which from your implementation it looks like it would be after resetting the legacy bit to false) wouldn't it re-run process? Otherwise, why test isConversionNeeded in the first place?

nkibler:
Theoretically, yes, IntelliJ shouldn't need to call `.isConversionNeeded()` again if it's already been processed. We can't change their behavior, though, and I've tested it by running it multiple times and it does not call `.process()` again. I'm not sure we should do anything else.

etanshaul:
cool. if your testing confirmed that it doesn't get called multiple times then lgtm regarding this issue.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: I'd add "existing" to the front of the sentence.

Also consider the following added to the end (feel free to ignore if you think its too verbose):

`Existing App Engine flexible deployment run configurations will have their staged artifact name explicitly set to target.jar/war for interoperability with existing Dockerfiles.`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/migration/StagedArtifactNameConverterProvider.java
```diff
@@ -0,0 +1,126 @@
...
+              return Stream.of(settings);
+            }
+          }
+          return Stream.empty();
+        };
+
+    static final Predicate<Element> IS_GCP_APP_ENGINE_DEPLOY =
+        element -> "gcp-app-engine-deploy".equals(element.getAttributeValue("type"));
+
+    static final Predicate<Element> IS_LEGACY_BIT_NOT_SET =
```
etanshaul: nit: but you are checking that the bit doesn't exist, not that its "not set", right?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/migration/StagedArtifactNameConverterProvider.java
```diff
@@ -0,0 +1,126 @@
...
+      runManagerSettings
+          .getRunConfigurations()
+          .stream()
+          .filter(IS_GCP_APP_ENGINE_DEPLOY)
+          .flatMap(TO_DEPLOYMENT_SETTINGS)
+          .filter(IS_LEGACY_BIT_NOT_SET)
+          .forEach(
+              element ->
+                  element.setAttribute(STAGED_ARTIFACT_NAME_LEGACY, Boolean.toString(true)));
+      //conversionContext.getRunManagerSettings().
```
etanshaul: remove comment

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
nkibler: Done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/migration/StagedArtifactNameConverterProvider.java
nkibler: Yeah, sorry, English is weird. I've changed this to say IS_LEGACY_BIT_NONEXISTENT

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/migration/StagedArtifactNameConverterProvider.java
nkibler: Whoops, thanks.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/migration/StagedArtifactNameConverterProvider.java
```diff
@@ -0,0 +1,124 @@
...
+    StagedArtifactNameConversionProcessor(ConversionContext conversionContext) {
+      this.conversionContext = conversionContext;
+    }
+
+    @Override
+    public boolean isConversionNeeded(RunManagerSettings runManagerSettings) {
+      return runManagerSettings
+          .getRunConfigurations()
+          .stream()
+          .filter(IS_GCP_APP_ENGINE_DEPLOY)
```
etanshaul: super nit: but some of this stream logic is duplicated with #process and perhaps could be extracted

etanshaul: actually nm - I didn't notice the difference on IS_LEGACY_BIT_NONEXISTENT.

etanshaul:
Can you show a screenshot of what shows up in the new artifact field when its _not_ a legacy name (not target.jar/war)? I'm wondering if its shown as normal text, or placeholder text?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/migration/StagedArtifactNameConverterProvider.java
```diff
@@ -0,0 +1,124 @@
...
+          if (deployment != null) {
+            Element settings = deployment.getChild("settings");
+            if (settings != null) {
+              return Stream.of(settings);
+            }
+          }
+          return Stream.empty();
+        };
+
+    static final Predicate<Element> IS_GCP_APP_ENGINE_DEPLOY =
```
etanshaul: wasn't there a suggestion to also filter for flex apps since isn't it that case that this will cause conversion on standard apps as is?

nkibler: Ah yes, let me add that restriction.

nkibler: Done, PTAL

PR #1646, opened 2017-09-18
Defaults new App Engine standard projects to Java8

### LINKED ISSUE #1641: Make Java 8 runtime the default for app engine standard

nkibler:
Fixes #1641.

I believe this is all that is required, but correct me if I'm wrong.


google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineJavaeeSupportContributor.java
```diff
@@ -44,7 +44,7 @@
   @Override
   public void setupFrameworkSupport(JavaeeFrameworkSupportContributionModel model) {
     AppEngineStandardFacet appEngineStandardFacet = model.getFacet(AppEngineStandardFacetType.ID);
-    if (appEngineStandardFacet == null) {
+    if (appEngineStandardFacet == null || appEngineStandardFacet.isJava8Runtime()) {
```
etanshaul: Won't this cause all the other setup to be skipped too? We still want the run configs created for java 8 runtimes.

nkibler: Ah I see, fixed. PTAL

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineJavaeeSupportContributor.java
```diff
@@ -48,7 +48,9 @@ public void setupFrameworkSupport(JavaeeFrameworkSupportContributionModel model)
       return;
     }
 
-    setWebXmlServletVersion(model.getModule());
+    if (!appEngineStandardFacet.isJava8Runtime()) {
```
etanshaul: technically it would be nice to check if its a java 7 runtime (think about java 9 in the future).

etanshaul: lgtm otherwise

nkibler: Yeah I thought about this, but I found a ton of hidden assumptions elsewhere regarding the exclusive Java7/Java8 setup that won't exist in the future, and I figured it would be best to tackle Java9 separately. See usages of `AppEngineStandardFacet.isJava8Runtime()`, for example.

etanshaul: what do you mean? like that for Java 7 the runtime is just "java"?

nkibler:
PTAL, I added the `.getRuntimeLanguageLevel()` as discussed offline.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/inspections/AppEngineForbiddenCodeInspection.java
```diff
@@ -66,7 +68,13 @@
     Module module = ModuleUtilCore.findModuleForPsiElement(file);
     final AppEngineStandardFacet appEngineStandardFacet
         = AppEngineStandardFacet.getAppEngineFacetByModule(module);
-    if (appEngineStandardFacet == null || appEngineStandardFacet.isJava8Runtime()) {
+    if (appEngineStandardFacet == null) {
+      return null;
+    }
+
+    Optional<LanguageLevel> runtimeLanguageLevel = appEngineStandardFacet.getRuntimeLanguageLevel();
+    if (!runtimeLanguageLevel.isPresent()
```
etanshaul: nit: but I usually like using functional idioms when dealing with Optionals if possible (one of the reasons being to avoid having to do a get())

nkibler: I think this one is unavoidable because we want to short-circuit if the `runtimeLanguageLevel` does not exist or if it's at least 1.8. I'm not sure how to represent that easily via functional methods. (I don't want to wrap the following code in a `.filter(...).ifPresent(() -> { ... })` chain because it would indent a large block of code.

But I used a functional approach in `AppEngineStandardUnsupportedJavaVersionCheck` and `AppEngineJavaeeSupportContributor`, PTAL.

patflynn:
just to confirm:
runtime: java 8 means the user can specify language level 8 and below.
runtime: java 7 means the user can specify language level 7 and below.

more explicitly:
java 8 runtime can be used with java 7 language level.

etanshaul:
Thats a good point. While I think the logic is correct (since language level is just a proxy for runtime in this context) it is a bit confusing for readability.

nkibler:
Fair enough, what about using `JavaVersion` instead of `LanguageLevel` to avoid this confusion? See my latest commit for details.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/AppEngineStandardFacet.java
```diff
@@ -85,10 +94,22 @@ public boolean isNonStandardCompatEnvironment() {
...
+  public Optional<JavaVersion> getRuntimeJavaVersion() {
     AppEngineStandardWebApp appEngineStandardWebApp = getAppEngineStandardWebXml();
-    return appEngineStandardWebApp != null
-        && "java8".equalsIgnoreCase(appEngineStandardWebApp.getRuntime().getStringValue());
+    if (appEngineStandardWebApp == null) {
+      return Optional.empty();
+    }
+
+    String runtime = appEngineStandardWebApp.getRuntime().getStringValue();
+    return Optional.ofNullable(RUNTIMES_MAP.get(runtime));
```
patflynn: shouldn't null here resolve to java7?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/AppEngineStandardUnsupportedJavaVersionCheck.java
```diff
@@ -76,29 +74,26 @@ public void runActivity(@NotNull Project project) {
...
-              }
+              Arrays.stream(projectModules)
+                  .filter(this::usesJava8OrGreater)
+                  .filter(
+                      module -> {
+                        AppEngineStandardFacet facet =
+                            AppEngineStandardFacet.getAppEngineFacetByModule(module);
+                        return !(facet == null || facet.isNonStandardCompatEnvironment())
+                            && facet
+                                .getRuntimeJavaVersion()
```
patflynn: this bit confuses me a bit.  Why should getRuntimeJavaVersion return a stream to be filtered?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/AppEngineStandardFacet.java
nkibler: @etanshaul discussed this yesterday and decided that we would return an `Optional.empty()` for not specified runtimes and unrecognized runtimes here; the caller can decide to interpret this as Java7 if they want. But I'm open to suggestions, I have mixed feelings about both approaches.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/AppEngineStandardUnsupportedJavaVersionCheck.java
nkibler: It's not a stream, it's an `Optional`, which provides the `.filter()` method to further filter the `Optional`, potentially turning it to `.empty()` if the filter doesn't match the value.

patflynn:
I'd be tempted to have the facet own default to java 7 so that the client's don't have to think about it and cause this code handling the optional is kinda ugly.

nkibler:
Done, PTAL.

nkibler:
Leaving this open until we're ready to merge.

PR #1649, opened 2017-09-19
Update generated Dockerfile artifact names

### LINKED ISSUE #1648: Change name of artifact in generated Dockerfile
Now that we are no longer renaming the staged artifact name by default for flex deploys, we should also update the name of the artifact we use in the generated dockerfile to make it clearer that it needs to be updated by the user.

Some appropriate name, plus comments above it in the Dockerfile should suffice.

etanshaul:
fixes #1648 

I updated the names of the artifact in the Docker templates to `${project.build.finalName}`. I went with this because presumably the user could configure the maven resource filtering to filter in this name into the Dockerfile. I am aware this won't work for native / gradle. But there are comments in the template now instructing the user to replace these.

I need to test that these new variables don't overly obfuscate the gcloud output when deployed as is.

@patflynn @nkibler wdyt?

### DIFF
--- a/google-cloud-tools-plugin/resources/fileTemplates/internal/JarDockerfile.gaedocker.ft
+++ b/google-cloud-tools-plugin/resources/fileTemplates/internal/JarDockerfile.gaedocker.ft
@@ -1,4 +1,5 @@
 FROM gcr.io/google_appengine/openjdk
-ADD target.jar /app/
+# Replace occurrences of ${project.build.finalName} with the name of the deployed jar
+ADD ${project.build.finalName}.jar /app/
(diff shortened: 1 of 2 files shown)


nkibler:
Can you make sure that the Maven variable can be filtered in too?

etanshaul:
This may be deal breaker for this approach.

Error message with previous template (target.war/jar) when the staged artifact name doesn't match:
image

Error message from this PR with the maven property:
image
Looks like it causing a parsing error in the Dockerfile.

imo this kills this approach. I think we should just go with a simple name + comment.

Something like: `artifact-name.jar/war`

etanshaul:
updated from discussion

PR #1652, opened 2017-09-20
Fixes setting deploy all asset checkbox to true for AE standard apps

### LINKED ISSUE #1651: Update dos, dispatch, cron etc. is not set by default
Should have been fixed by https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1639 but when I generate a new app engine standard project (using the wizard) then open the deploy config, the checkbox is not set.

etanshaul:
fixes #1651


etanshaul:
reverts the previous approach and just sets the UI field to true by default

etanshaul:
no problem :)

PR #1655, opened 2017-09-21
GCS bucket content breadcrumb navigation and refresh

etanshaul:
Root of bucket:
image

Nested folder:
image

Wrapping breadcrumbs:
image


etanshaul:
@nkibler @patflynn ptal

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
```diff
@@ -64,6 +65,20 @@ public void mouseClicked(MouseEvent event) {
             }
           }
         });
+
+    refreshButton.addActionListener(
```
nkibler: Can you add tests for these two new listeners?

etanshaul: looking at this, but to have the PR ready to go are there any other comments?

PR #1657, opened 2017-09-25
Handle empty GCS buckets and empty blob directories

etanshaul:
Does 2 main things:

1) Handles initial display of empty GCS buckets and empty directories
2) Handles update of buckets and directories when moving from empty to non-empty and vice versa by refreshing. For example, a bucket is initially empty, the user then updates the contents of the bucket outside of the IDE, then refreshes the bucket in the IDE.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
@@ -98,7 +98,7 @@ final class GcsBucketContentEditorPanel {
   void initTableModel() {
     List<Blob> blobs = getBlobsStartingWith("");
     if (blobs.isEmpty()) {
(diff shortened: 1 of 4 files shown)


etanshaul:
Empty bucket:
image

Empty directory:
image

etanshaul:
@nkibler @patflynn ptal

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
```diff
@@ -125,6 +161,14 @@ void updateTableModel(String prefix) {
         bucket.list(BlobListOption.currentDirectory(), BlobListOption.prefix(prefix)).iterateAll());
   }
 
+  /**
+   * Tests if a given directory prefix is empty. A directory is empty if there are no blobs or if
+   * the only blob matches the current directory prefix.
+   */
+  private boolean isEmptyDirectory(String prefix, List<Blob> blobs) {
```
nkibler: Make this static

PR #1658, opened 2017-09-25
Updates GCS panel and bucket icons

etanshaul:
image

Notice the icon on the Google Cloud Storage panel, and the icons next to each bucket.


etanshaul:
also updated is the icon on the editor panel:
image

PR #1661, opened 2017-09-26
Makes the project selector searchable

### LINKED ISSUE #1660: Implement type-ahead search for the project selector
Currently, there is no way to search through the list of projects in the project selector. This is especially problematic for users with many projects, causing them to scroll through a very long list to try and find their target project.

The project selector should support type-ahead search. As the user types, the projects should be filtered by matching project names or project IDs.

nkibler:
Fixes #1660.


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/GoogleUserModelItem.java
```diff
@@ -195,4 +214,36 @@ private void loadUserProjects() {
       loadErrorState(ex.getMessage());
     }
   }
+
+  @Override
+  public TreeNode getChildAt(int index) {
+    return (TreeNode) getFilteredChildren().toArray()[index];
```
etanshaul: can getFilteredChildren be null?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
```diff
@@ -109,25 +104,31 @@ public ProjectSelector(final boolean queryOnExpand) {
...
-            && event.getTreePath() != null
-            && event.getTreePath().getLastPathComponent() instanceof GoogleUserModelItem) {
-          GoogleUserModelItem userItem = (GoogleUserModelItem) event.getTreePath()
-              .getLastPathComponent();
-          for (int index = 0; index < userItem.getChildCount(); index++) {
-            DefaultMutableTreeNode loadedItem = (DefaultMutableTreeNode) userItem.getChildAt(index);
-            if (loadedItem instanceof ResourceProjectModelItem
-                && getText().equals(((ResourceProjectModelItem) loadedItem).getProjectId())) {
-              popupPanel.tree.setSelectionPath(new TreePath(loadedItem.getPath()));
+    treeModel.addTreeModelListener(
```
etanshaul: doesn't look like it, but just to make sure I didn't miss anything: nothing was changed in this block other than formatting right?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
```diff
@@ -400,18 +415,37 @@ private TreePath find(DefaultMutableTreeNode root, String str) {
...
 
-    public void initializeContent(String selectedProjectId) {
+    void setFilter(String filter) {
+      for (int i = 0; i < tree.getRowCount(); i++) {
+        TreePath path = tree.getPathForRow(i);
+        if (path.getLastPathComponent() instanceof GoogleUserModelItem) {
+          GoogleUserModelItem userModelItem = (GoogleUserModelItem) path.getLastPathComponent();
+          userModelItem.setFilter(filter);
+
+          boolean isExpanded = tree.isExpanded(path);
```
etanshaul: nit: move this above where its used, or just inline tree#isExpanded

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
```diff
@@ -400,18 +415,37 @@ private TreePath find(DefaultMutableTreeNode root, String str) {
...
+          DefaultTreeModel model = (DefaultTreeModel) tree.getModel();
+          model.reload(userModelItem);
+
+          if (isExpanded) {
+            tree.expandPath(path);
```
etanshaul: what is this for?

etanshaul:
I think a demo, like we discussed, would be great for helping with review too

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/GoogleUserModelItem.java
nkibler: Nope, worst case it's an empty stream.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
nkibler: Whoops, yeah you're right. Sorry about that, auto formatter must have kicked in and I didn't notice it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
nkibler: Obsolete, PTAL.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
nkibler: I thought this was needed to re-expand the tree if the model was updated and the tree was collapsed, but I just removed it and it works fine still. PTAL

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
```diff
@@ -167,6 +168,18 @@ public void componentHidden(ComponentEvent event) {
...
         });
 
+    getTextField()
+        .getDocument()
+        .addDocumentListener(
+            new DocumentAdapter() {
+              @Override
+              protected void textChanged(DocumentEvent e) {
+                if (popupPanel != null) {
+                  popupPanel.setFilter(getText());
```
etanshaul: In the latest iteration of this, hows the delay on various systems? I think its important that we validate that it responds quickly enough (<10ms or whatever so as not to be visible / annoying to the user).

etanshaul: Just to add my initial impression after giving it a test drive

On my linux box it feels pretty snappy overall. However, typing in the first letter has a noticeable delay. Subsequent letters though have a negligible delay. Could that initial delay be optimized?

etanshaul: I was able to get it into a state where the delay on the first letter keeps increasing (exponentially it seems):

Each time I type a letter into the selector field, backspace, then type it again the delay gets longer. Eventually it gets to the point even of crashing the IDE. Is there maybe some state that is being held onto mistakenly?

nkibler: So there's no explicitly set delay; any delay you're experiencing is it applying the filter to every project, telling the tree there was a model update, and redrawing all of the matching projects.

Out of curiosity, how many projects do you have? Do you have multiple accounts setup too? I tried typing the first letter over and over again for a few minutes with my 300+ projects and didn't notice any sort of build-up of delay.

etanshaul: I don't have more than 300+ projects and have it set up with only one account.

I noticed that I was able to trigger this delay by playing first typing in the selector, clicking off, then returning to it. I don't really have exact reproduction steps.

etanshaul:
Could we add key binding for the 'enter' key here so that the current match gets selected? I'm fine with it if you think it should be in a different PR, I just think this feature would be really nice.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
nkibler: Fixed it; there was a memory leak on the `popupPanel`. PTAL

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
```diff
@@ -434,7 +437,7 @@ void setFilter(String filter) {
           GoogleUserModelItem userModelItem = (GoogleUserModelItem) path.getLastPathComponent();
           userModelItem.setFilter(filter);
           DefaultTreeModel model = (DefaultTreeModel) tree.getModel();
-          model.reload(userModelItem);
+          model.nodeStructureChanged(userModelItem);
```
etanshaul: does `nodeStructureChanged` have a different visible outcome from `reload`?

etanshaul:
sorry for the delay on this. lgtm with respect to the code. but I'd just like to make sure the response time is adequate so would you mind reporting back what you find on various machines (ideally testing on a slower setup)?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
nkibler: Not visible but from what I could tell, it performed better than `reload()`, probably because it lets one of the listeners do the reloading instead of forcing it here.

nkibler:
I'll check it now on my laptop, but it's still not exactly a super slow machine.

nkibler:
It LGTM on my laptop, super snappy with the latest memory leak fixes.

etanshaul:
I finally tested on my slow windows vm (sorry for the delay) and its as snappy as can be expected. LGTM - lets get this in!

PR #1663, opened 2017-09-26
Execute GCS bucket operations asynchronously

etanshaul:
will update target branch once https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1657 is merged

Loads bucket contents asynchronously. Shows a loader when operation is pending:
image

### DIFF
--- a/common-lib/src/main/java/com/google/cloud/tools/intellij/util/ThreadUtil.java
+++ b/common-lib/src/main/java/com/google/cloud/tools/intellij/util/ThreadUtil.java
@@ -51,7 +51,7 @@ public final class ThreadUtil {
   }
 
   /**
-   * Executes the given {@link Runnable} on even dispatching thread.
+   * Executes the given {@link Runnable} on the event dispatching thread.
    *
    * <p>If {@link #invokeLaterOnEDT} is {@code true} the runnable will be invoked asynchronously,
    * otherwise it will be invoked synchronously.
@@ -87,7 +87,7 @@ public final class ThreadUtil {
   }
 
   /**
-   * Indicates that runnables to be invoked on the EDT should be invoked later in the background.
+   * Indicates that runnables to be invoked on the EDT should be invoked asynchronously.
    */
   @VisibleForTesting
   public void setBackgroundEDTInvocator() {
--- a/common-lib/src/main/java/com/google/cloud/tools/intellij/util/ThreadUtil.java
+++ b/common-lib/src/main/java/com/google/cloud/tools/intellij/util/ThreadUtil.java
@@ -17,6 +17,7 @@
 package com.google.cloud.tools.intellij.util;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.intellij.openapi.application.ApplicationManager;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
@@ -27,6 +28,7 @@ public final class ThreadUtil {
   private static final ThreadUtil INSTANCE = new ThreadUtil();
 
   private ExecutorService backgroundExecutorService;
+  private boolean invokeLaterOnEDT = true;
 
   /** Visibility is restricted internally; use {@link #getInstance()} instead. */
   private ThreadUtil() {
@@ -49,6 +51,22 @@ public final class ThreadUtil {
   }
 
   /**
+   * Executes the given {@link Runnable} on even dispatching thread.
+   *
+   * <p>If {@link #invokeLaterOnEDT} is {@code true} the runnable will be invoked asynchronously,
+   * otherwise it will be invoked synchronously.
+   *
+   * @param runnable the {@link Runnable} to run on the EDT.
+   */
+  public void invokeLaterOnEDT(Runnable runnable) {
+    if (invokeLaterOnEDT) {
+      ApplicationManager.getApplication().invokeLater(runnable);
+    } else {
+      ApplicationManager.getApplication().invokeAndWait(runnable);
+    }
+  }
+
+  /**
    * Sets the {@link ExecutorService} used for executing background tasks.
    *
    * <p>This should only be used by unit tests to replace the default {@link ExecutorService} with
@@ -66,5 +84,25 @@ public final class ThreadUtil {
   @VisibleForTesting
   public void setBackgroundExecutorService(ExecutorService backgroundExecutorService) {
     this.backgroundExecutorService = backgroundExecutorService;
+  }
+
+  /**
+   * Indicates that runnables to be invoked on the EDT should be invoked later in the background.
+   */
+  @VisibleForTesting
+  public void setBackgroundEDTInvocator() {
+    this.invokeLaterOnEDT = true;
+  }
+
+  /**
+   * Indicates that runnables to be invoked on the EDT should be invoked synchronously.
+   *
+   * <p>This should only be used by unit tests to replace behavior of {@link
+   * #executeInBackground(Runnable)} from running Swing EDT tasks in the background to having them
+   * block.
+   */
+  @VisibleForTesting
+  public void setSynchronousEDTInvocator() {
+    this.invokeLaterOnEDT = false;
   }
 }
--- a/common-lib/src/main/java/com/google/cloud/tools/intellij/util/ThreadUtil.java
+++ b/common-lib/src/main/java/com/google/cloud/tools/intellij/util/ThreadUtil.java
@@ -17,7 +17,6 @@
 package com.google.cloud.tools.intellij.util;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.intellij.openapi.application.ApplicationManager;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
@@ -28,7 +27,6 @@ public final class ThreadUtil {
   private static final ThreadUtil INSTANCE = new ThreadUtil();
 
   private ExecutorService backgroundExecutorService;
-  private boolean invokeLaterOnEDT = true;
 
   /** Visibility is restricted internally; use {@link #getInstance()} instead. */
   private ThreadUtil() {
@@ -51,22 +49,6 @@ public final class ThreadUtil {
   }
 
   /**
-   * Executes the given {@link Runnable} on the event dispatching thread.
-   *
-   * <p>If {@link #invokeLaterOnEDT} is {@code true} the runnable will be invoked asynchronously,
-   * otherwise it will be invoked synchronously.
-   *
-   * @param runnable the {@link Runnable} to run on the EDT.
-   */
-  public void invokeLaterOnEDT(Runnable runnable) {
-    if (invokeLaterOnEDT) {
-      ApplicationManager.getApplication().invokeLater(runnable);
-    } else {
-      ApplicationManager.getApplication().invokeAndWait(runnable);
-    }
-  }
-
-  /**
    * Sets the {@link ExecutorService} used for executing background tasks.
    *
    * <p>This should only be used by unit tests to replace the default {@link ExecutorService} with
@@ -84,25 +66,5 @@ public final class ThreadUtil {
   @VisibleForTesting
   public void setBackgroundExecutorService(ExecutorService backgroundExecutorService) {
     this.backgroundExecutorService = backgroundExecutorService;
-  }
-
-  /**
-   * Indicates that runnables to be invoked on the EDT should be invoked asynchronously.
-   */
-  @VisibleForTesting
-  public void setBackgroundEDTInvocator() {
-    this.invokeLaterOnEDT = true;
-  }
-
-  /**
-   * Indicates that runnables to be invoked on the EDT should be invoked synchronously.
-   *
-   * <p>This should only be used by unit tests to replace behavior of {@link
-   * #executeInBackground(Runnable)} from running Swing EDT tasks in the background to having them
-   * block.
-   */
-  @VisibleForTesting
-  public void setSynchronousEDTInvocator() {
-    this.invokeLaterOnEDT = false;
   }
 }
--- a/common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/DelayedSubmitExecutorServiceProxy.java
+++ b/common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/DelayedSubmitExecutorServiceProxy.java
@@ -29,7 +29,7 @@ import org.jetbrains.annotations.NotNull;
  * ExecutorService#submit(Runnable)}. This proxy instead stores the supplied {@link Runnable} and
  * then executes it only after {@link #doSubmit()} is invoked.
  */
-public class DelayedSubmitExecutorServiceProxy extends AbstractExecutorService {
+public final class DelayedSubmitExecutorServiceProxy extends AbstractExecutorService {
 
   private final ExecutorService executor;
   private Runnable task;
@@ -81,7 +81,7 @@ public class DelayedSubmitExecutorServiceProxy extends AbstractExecutorService {
   /** Submits the stored task. */
   public Future<?> doSubmit() {
     if (task == null) {
-      throw new IllegalStateException("There is no task to submit.");
+      throw new AssertionError("There is no task to submit.");
     }
     return executor.submit(task);
   }
--- a/common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/DelayedSubmitExecutorServiceProxy.java
+++ b/common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/DelayedSubmitExecutorServiceProxy.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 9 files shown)


etanshaul:
@nkibler @patflynn ptal

common-lib/src/main/java/com/google/cloud/tools/intellij/util/ThreadUtil.java
```diff
@@ -49,6 +51,22 @@ public static ThreadUtil getInstance() {
...
 
   /**
+   * Executes the given {@link Runnable} on the event dispatching thread.
+   *
+   * <p>If {@link #invokeLaterOnEDT} is {@code true} the runnable will be invoked asynchronously,
+   * otherwise it will be invoked synchronously.
+   *
+   * @param runnable the {@link Runnable} to run on the EDT.
+   */
+  public void invokeLaterOnEDT(Runnable runnable) {
```
patflynn: Very confused by this method.  It sounds like it's always invoking later, but then it checks a flag in the implementation.  Might be helpful to include use-cases in the javadoc

common-lib/src/main/java/com/google/cloud/tools/intellij/util/ThreadUtil.java
```diff
@@ -67,4 +85,24 @@ public static ThreadUtil getInstance() {
   public void setBackgroundExecutorService(ExecutorService backgroundExecutorService) {
     this.backgroundExecutorService = backgroundExecutorService;
   }
+
+  /**
+   * Indicates that runnables to be invoked on the EDT should be invoked asynchronously.
+   */
+  @VisibleForTesting
+  public void setBackgroundEDTInvocator() {
```
patflynn: this doesn't look very threadsafe.

common-lib/src/main/java/com/google/cloud/tools/intellij/util/ThreadUtil.java
etanshaul: Yeah, I'm not overly crazy about this pattern either. Its used strictly for testing: ONLY unit tests should be replacing the default asynchronous strategy with a synchronous one to aid in testing async code. I was using the same approach as was already used in the ThreadUtil class.

etanshaul: I previously had suggested to use to DI to handle these situations: e.g. abstract out the async code into an application service (injected via plugin.xml); then replace it with a mock for unit testing. wdyt?

nkibler: FWIW, this isn't exactly as I suggested. To follow the pattern that already exists in this class (without creating another settable value), you could change this method to simply call:

```
backgroundExecutorService.submit(() -> ApplicationManager.getApplication().invokeAndWait(runnable));
```

Everything else would "take care of itself" as the `backgroundExecutorService` is already swapped for a direct executor in `CloudToolsRule`.

patflynn: ah I see.  Your strategy of letting the runner be swapped for testing sounds good to me.  
@nkibler any thoughts?

nkibler: I think the application service registered via the `plugin.xml` is just extra overhead for relatively the same solution as I listed above. It still modifies a supposedly static object (the `PicoContainer`) in a test utility. The only difference is that there is no `@VisibleForTesting` method exposed, which should fail compilation anyways if production code were to call it.

It's up to you, either way works for me.

etanshaul: ah yes. @nkibler - my implementation was a bit silly. I'll refactor to use the already existing executor.

etanshaul: I'm OK with this approach since it isolates the smallest component needed for testing, instead of creating a new service which is more overhead (like you said).

patflynn: 👍

etanshaul: on second look, is:

`backgroundExecutorService.submit(() -> ApplicationManager.getApplication().invokeAndWait(runnable));`

Actually technically equivalent to:

`ApplicationManager.getApplication().invoke(runnable)` ?

That will still block the EDT, no? So in other words, I'm not 100% sure if we can refactor what I have.

nkibler: I'm assuming you meant to type `.invokeAndWait()` in the second statement since there is no `.invoke()`? No they aren't equivalent. The point of using the `backgroundExecutorService` is that it will execute a given `Runnable` using a cached thread pool that is separate from the main thread. `.invokeLater()` technically does this as well, but we don't control the `ExecutorService` that the `Application` uses, so we can't swap it out in unit tests to force it to execute the tasks on the current thread.

etanshaul: No I mean invoke() on the second case.

My point was with your suggested refactoring. You said to just replace what I have with:

```
backgroundExecutorService.submit(() -> ApplicationManager.getApplication().invokeAndWait(runnable));
```

What about non-test cases? In this case the background executor will not be a direct one and asynchronously call invokeAndWait. However, invokeAndWait will block the EDT thus not producing the correct results (the UI will block).

Perhaps I'm being stupid and missing something here.

etanshaul: oops replace any time I mention `invoke` with `invokeLater`

nkibler: Ah I see. "Invoke and wait" in this sense means to schedule the `Runnable` to run on the EDT and block the current thread until it returns. It won't stop the EDT from running, then force this task to be executed on it before continuing, as that would cause major application issues.

More info from the `.invokeAndWait()` Javadoc:
> Causes `runnable.run()` to be executed synchronously on the AWT event dispatching thread, when the IDE is in the specified modality state (or a state with less modal dialogs open). This call blocks until all pending AWT events have been processed and (then) `runnable.run()` returns.

etanshaul: So are we agreeing that we have to keep these boolean setters? :)

etanshaul:
@patflynn @nkibler cleaned things up - PTAL. The EDT wrapper wasn't needed (thanks @nkibler for the discussion).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
```diff
@@ -155,10 +167,19 @@ private void showBlobTable() {
...
+    showLoader();
+    ThreadUtil.getInstance()
+        .executeInBackground(
+            () -> {
+              List<Blob> blobs =
+                  Lists.newArrayList(
+                      bucket
+                          .list(BlobListOption.currentDirectory(), BlobListOption.prefix(prefix))
+                          .iterateAll());
+              hideLoader();
```
nkibler: Shouldn't this be done on the EDT as well? Not sure how thread-safe JPanel methods are.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanelTest.java
```diff
@@ -88,6 +88,7 @@ public void testEmptyBucket_noBucketContentTable() {
     JTable bucketTable = editorPanel.getBucketContentTable();
     assertThat(bucketTable.getColumnCount()).isEqualTo(0);
     assertThat(bucketTable.getRowCount()).isEqualTo(0);
+    assertFalse(editorPanel.getLoadingPanel().isVisible());
```
nkibler: Is there another assertion you can add elsewhere to verify the panel is shown before the call is done? And what happens when the call fails?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
etanshaul: fixed

etanshaul:
Added coverage for testing that loader is shown.

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/DelayedSubmitExecutorServiceProxy.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.TimeUnit;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A proxy {@link ExecutorService} that delays the execution of {@link
+ * ExecutorService#submit(Runnable)}. This proxy instead stores the supplied {@link Runnable} and
+ * then executes it only after {@link #doSubmit()} is invoked.
+ */
+public class DelayedSubmitExecutorServiceProxy extends AbstractExecutorService {
```
nkibler: nit: final?

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/DelayedSubmitExecutorServiceProxy.java
```diff
@@ -0,0 +1,88 @@
...
+
+  @Override
+  public void execute(@NotNull Runnable command) {
+    executor.execute(command);
+  }
+
+  /** Stores the submitted {@code task} to be submitted later on {@link #doSubmit()}. */
+  @Override
+  public Future<?> submit(Runnable task) {
+    this.task = task;
```
nkibler: I know you only submit one task currently, but could you add in support for a list of tasks? If you add an additional call in your code under test without realizing its impact here, there could be some very weird behavior (the last task submitted will be the only one executed).

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/DelayedSubmitExecutorServiceProxy.java
```diff
@@ -0,0 +1,88 @@
...
+ * then executes it only after {@link #doSubmit()} is invoked.
+ */
+public class DelayedSubmitExecutorServiceProxy extends AbstractExecutorService {
+
+  private final ExecutorService executor;
+  private Runnable task;
+
+  /** Creates a new instance that is a proxy for the given {@code executor}. */
+  public DelayedSubmitExecutorServiceProxy(ExecutorService executor) {
+    this.executor = executor;
```
nkibler: Could you just initialize this within this class instead of passing it in? Since it's a test util, I would imagine every caller is going to want a direct executor service anyway.

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/DelayedSubmitExecutorServiceProxy.java
```diff
@@ -0,0 +1,88 @@
...
+  @Override
+  public Future<?> submit(Runnable task) {
+    this.task = task;
+    return new FutureTask<>(task, null);
+  }
+
+  /** Submits the stored task. */
+  public Future<?> doSubmit() {
+    if (task == null) {
+      throw new IllegalStateException("There is no task to submit.");
```
nkibler: IMO an `AssertionError` is more appropriate, especially since this will be used by tests.

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/DelayedSubmitExecutorServiceProxy.java
etanshaul: done

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/DelayedSubmitExecutorServiceProxy.java
etanshaul: done

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/DelayedSubmitExecutorServiceProxy.java
etanshaul: I kind of feel like that is overextending the indented behavior of this executor service. How do you feel instead about throwing an exception if trying to submit a task when one has already be submitted?

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/DelayedSubmitExecutorServiceProxy.java
etanshaul: do feel strongly about this? Its not impossible to conceive of test usages of other types of executors.

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/DelayedSubmitExecutorServiceProxy.java
nkibler: Executor services are designed to accept multiple tasks and schedule each runnable on its queue. There are several scenarios where a single method could submit multiple tasks to the executor service.

But since this is not needed now, I'll leave it optional.

etanshaul: fair enough. I can update it

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/DelayedSubmitExecutorServiceProxy.java
nkibler: If a test wants to use anything other than a direct executor, then this wrapper becomes almost useless; there's not much value in determining when a task would execute on a different thread because you would still have to continuously check for completion from your test's thread, even when using this wrapper.

I'll leave this up to you.

nkibler:
Why didn't the presubmit tests run?

etanshaul:
because of the target branch of this PR. It depends on https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1657 which needs to be merged first. I'll then update the target

etanshaul:
btw, I'm still updating this PR to deal with failure scenarios (from the GCS api).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
```diff
@@ -169,18 +171,34 @@ private void showBlobTable() {
...
+                ApplicationManager.getApplication()
+                    .invokeAndWait(
+                        () -> {
+                          hideLoader();
+                          showError();
+                        });
+                return;
+              }
+
+              final List<Blob> loadedBlobs = blobs;
```
nkibler: I'm guessing you have to do this because the compiler can't tell that `blobs` is "effectively final" since its instantiation is separated from its definition?

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanelTest.java
```diff
@@ -291,6 +292,38 @@ public void testBlobSizeDisplay_GB() {
...
+
+    editorPanel = new GcsBucketContentEditorPanel(bucketVirtualFile.getBucket());
+    editorPanel.initTableModel();
+
+    assertTrue(editorPanel.getErrorPanel().isVisible());
+
+    // Re-initialize mocks so the exception is not thrown
+    GcsTestUtils.setupVirtualFileWithBucketMocks(bucketVirtualFile);
+
+    assertFalse(editorPanel.getErrorPanel().isVisible());
```
nkibler: Don't you have to make a call to `updateTableModel()` or something to retry the list call?

etanshaul: good eye. I'm in the process of fixing this test now

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
etanshaul: yes. and I can't make blobs final for obvious reasons

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanelTest.java
etanshaul: fixed

etanshaul:
something is still up with these tests. looking into it.

Issue #1607, opened 2017-08-11
region indicator in deployment UI shows error when project is not selected


patflynn:
The region should just be empty or NA when there's no project specified

![screenshot from 2017-08-11 10 29 38](https://user-images.githubusercontent.com/7862557/29217543-1cae64ac-7e80-11e7-89a6-bc9b47bf7577.png)

of note is that there's some weird position problem as well with the google cloud console link

etanshaul:
fixed via #1682

PR #1682, opened 2017-10-05
Fix for the GAE deploy config's region indicator showing an error when a project is not selected

nbashirbello:
Fix for #1607


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ModelItem.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package com.google.cloud.tools.intellij.resources;
+
+import javax.swing.tree.DefaultMutableTreeNode;
+
+/**
+ * Represents an existing or non-existing GCP project
+ */
+public class ModelItem extends DefaultMutableTreeNode {
```
etanshaul: Why the need to add this in between?

Also the name of it doesn't really indicate that its a GCP related object.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
```diff
@@ -148,14 +148,22 @@ public void setDeploymentProjectAndVersion(DeploymentSource deploymentSource) {
...
+  /**
+   * Updates the text of the panel as follows:
+   *   if the project text box is empty, no message is displayed,
+   *   if the project text represents a valid project, the project details are displayed,
+   *   if the project text represents an invalid project, an error message is displayed.
+   */
   private void refreshApplicationInfoPanel() {
-    if (projectSelector.getProject() != null && projectSelector.getSelectedUser() != null) {
+    if (projectSelector.getText().isEmpty()) {
+      applicationInfoPanel.setMessage("", false /* isError*/);
```
etanshaul: why do you need to set an empty message to indicate no message?

nbashirbello: I am not actually sure if this method is only called the first time the panel is loaded. It is also called from a public method, so I covered all cases.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ModelItem.java
nbashirbello: I added the ModelItem, so that we can tell when the project field has no value (null), an invalid project(InvalidProjectModelItem) and a valid project (ResourceProjectModelItem).

I agree with the name comment. Any suggestions? I am not a good name-generator :)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineApplicationInfoPanel.java
```diff
@@ -67,10 +67,22 @@ public AppEngineApplicationInfoPanel() {
...
+   * Updates the panel as follows:
+   *   if the project textbox specifies a valid project, it displays the project's information,
+   *   if the project textbox specifies an invalid project, it displays an error message,
+   *   if the project textbox is empty, no message is displayed.
+   */
   public void refresh(final ProjectSelectionChangedEvent event) {
     if (event == null) {
       ApplicationManager.getApplication().executeOnPooledThread(
-          () -> setMessage(GctBundle.getString("appengine.infopanel.noproject"),
+          () -> setMessage("", false /* isError*/));
```
etanshaul: If we need to clear out the message explicitly like this, can you add a `clearMessage` (or something like that) method for better readability?

etanshaul:
looks like there no tests that broke as a result of this change; is it possible to add some to cover this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
```diff
@@ -267,12 +269,13 @@ public Long getProjectNumber() {
...
+   ModelItem modelItem = getCurrentModelItem();
+   return modelItem instanceof ResourceProjectModelItem ?
+       ((ResourceProjectModelItem)modelItem).getProject() : null;
   }
 
   @Nullable
-  private ResourceProjectModelItem getCurrentModelItem() {
+  private ModelItem getCurrentModelItem() {
     if (Strings.isNullOrEmpty(getText())) {
       return null;
```
etanshaul: Its a bit confusing that you are using both null and `InvalidProjectModelItem`. It looks like if the selector is empty (no text) you return null, but if the selector has text but is not an existing project you return InvalidProjectModelItem. Why the need for both?

etanshaul: Oh I think I get it. Its because you specifically need to handle this null case (where there is no text) differently than an "invalid" project.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
```diff
@@ -550,14 +553,20 @@ protected void doLogin() {
...
       if (newSelection.getParent() instanceof GoogleUserModelItem) {
         user = ((GoogleUserModelItem) newSelection.getParent()).getCredentialedUser();
       }
-      event = new ProjectSelectionChangedEvent(newSelection.getProject(), user);
+
+      if (newSelection instanceof InvalidProjectModelItem) {
+        event = new ProjectSelectionChangedEvent(null, user);
+      } else if (newSelection instanceof ResourceProjectModelItem) {
+        event = new ProjectSelectionChangedEvent((
+            (ResourceProjectModelItem)newSelection).getProject(), user);
```
etanshaul: is the formatting off here? (space after the cast)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
```diff
@@ -550,14 +553,20 @@ protected void doLogin() {
...
     CredentialedUser user = null;
     ProjectSelectionChangedEvent event = null;
     if (newSelection != null) {
       if (newSelection.getParent() instanceof GoogleUserModelItem) {
         user = ((GoogleUserModelItem) newSelection.getParent()).getCredentialedUser();
       }
-      event = new ProjectSelectionChangedEvent(newSelection.getProject(), user);
+
+      if (newSelection instanceof InvalidResourceProjectModelItem) {
+        event = new ProjectSelectionChangedEvent(null, user);
```
etanshaul: This doesn't look safe. For example, `AppEngineApplicationPanel#refresh` listens on this event and doesn't account for null `Project`s.

nbashirbello: There is a check in AppEngineApplicationInfoPanel#refresh for null project

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineApplicationInfoPanel.java
nbashirbello: Done

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
etanshaul: Ah ok I missed that you updated it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/InvalidResourceProjectModelItem.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.intellij.resources;
+
+/**
+ * A non-existing project specified in the Project field of the App Engine Deployment configuration
+ * dialog.
+ */
+public class InvalidResourceProjectModelItem extends GcpModelItem {
```
etanshaul: looks like you can make this package-private and final

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/GcpModelItem.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package com.google.cloud.tools.intellij.resources;
+
+import javax.swing.tree.DefaultMutableTreeNode;
+
+/**
+ * Represents an existing or non-existing GCP project
+ */
+public class GcpModelItem extends DefaultMutableTreeNode {
```
etanshaul: same comment about about package-private / final.

nit: GcpModelItem -> ProjectModelItem (maybe fits better with other naming in this package)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineApplicationInfoPanel.java
```diff
@@ -126,6 +139,17 @@ public void setMessage(String text, boolean isError) {
     }, isError);
   }
 
+  public void clearMessage() {
```
etanshaul: package-private?

etanshaul:
lgtm after latest round of comments.

Also would be nice to add some tests if possible. If you run into issues around testing asynchronous operations, we've recently added some utilities to help with that - @nkibler or I could point you to it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java
etanshaul: On second look, could you just pass in the `InvalidResourceProjectModelItem` here instead of null? That way you could make the paramenter not nullable and then do the check for the type inside of the event itself. Also allowing you to get rid of the if/else here.

nbashirbello: I won't be able to pass the  InvalidResourceProjectModelItem here because the signature is ProjectSelectionChangedEvent(Project, CredentialedUser)

etanshaul: ah right - sgtm

PR #1687, opened 2017-10-10
Adds a tool menu action for GCS

etanshaul:
image

When clicked, the action will open the GCS tool window panel and focus on the project selector. 

The platform is making it really difficult for me to add a test to cover this, but will look a little bit more.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsToolWindowAction.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsToolWindowAction.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.intellij.gcs;
+
+import com.google.cloud.tools.intellij.ui.GoogleCloudToolsIcons;
+import com.google.cloud.tools.intellij.util.GctBundle;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.project.DumbAwareAction;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.wm.ToolWindow;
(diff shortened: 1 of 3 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsToolWindowAction.java
```diff
@@ -0,0 +1,49 @@
...
+        GoogleCloudToolsIcons.CLOUD_STORAGE);
+  }
+
+  @Override
+  public void actionPerformed(AnActionEvent event) {
+    Project project = event.getProject();
+    if (project != null) {
+      ToolWindow window =
+          ToolWindowManager.getInstance(project).getToolWindow("Google Cloud Storage");
+      window.activate(null /*runnable*/);
```
nkibler: What happens if the window is already active?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsToolWindowAction.java
```diff
@@ -0,0 +1,49 @@
...
+        GctBundle.message("gcs.tools.menu.action.description"),
+        GoogleCloudToolsIcons.CLOUD_STORAGE);
+  }
+
+  @Override
+  public void actionPerformed(AnActionEvent event) {
+    Project project = event.getProject();
+    if (project != null) {
+      ToolWindow window =
+          ToolWindowManager.getInstance(project).getToolWindow("Google Cloud Storage");
```
nkibler: Could `.getToolWindow()` ever reasonably return `null`? The Javadoc says it can but I'm wondering in what real-world scenario this could happen.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsToolWindowAction.java
etanshaul: It will remain open and put focus on the project selector field

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsToolWindowAction.java
etanshaul: It won't return null since we register a tool window of this ID in plugin.xml. But its conceivable that another plugin (or some bad code in ours) could unregister this tool window, but I don't think we need to handle this realistically.

nkibler: Ok cool, that's what I figured.

etanshaul:
I see no good way of testing this without doing something awkward refactoring in the action to use injected services.

The `ToolWindowManager` when run from the test returns a headless manager that creates a stub `ToolWindow`; this toolWindow is a real object and has empty implementation for all its method - crucially for the `isActive` and `activate` methods. So in other words `ToolWindow#activate` does nothing at all when run from the test and therefore there is nothing I can assert.

Issue #1693, opened 2017-10-11
Project selector contents not reloaded after login


etanshaul:
1) log out of all cloud projects
2) open a ui that has the project selector; e.g. tools -> google cloud tools -> setup on cloud source repos
3) open the project selector and note that you are logged out
4) sign in from the button on the bottom of the project selector and complete the oauth flow

Expected: user is signed in and cloud projects are listed

Actual: despite being signed in, the panel is not properly refreshed:
![image](https://user-images.githubusercontent.com/1735744/31457279-51d8dc72-ae8a-11e7-88cb-4ea8e04d8e7c.png)

etanshaul:
@nkibler I've confirmed that this was introduced as part of da734da35987b9c8496f67f38ad14f281b3966e4

etanshaul:
@parkjam4 just a heads up that this bug will likely cause you to see some weird behavior (like in the screenshot above) when you are giving the GCS alpha build a spin for documentation purposes. 

I'll update the build once @nkibler fixes this.

PR #1694, opened 2017-10-12
Syncs project selector on login events

### LINKED ISSUE #1688: Cloud Storage window needs to be re-opened after login
1) Log out of your cloud account
2) Open the GCS bucket panel and note that buckets will not be displayed
3) Log in to the cloud account
4) Refresh the GCS buckets

-> The buckets won't refresh. Only upon minimizing and maximizing the panel will they be loaded.

etanshaul:
fixes #1688

Note for the full experience to be correct, #1693 needs to be fixed. I tested this with da734da stashed to make sure things work as expected.

This does 2 things:
1) Adds a message in the UI in the bucket listing panel when the user is logged out:
image
2) Fixes all the project selector synchronization issues by adding a login/logout listener and re-syncing the project selector on these events.

The listener mechanism for login events that was already in place wasn't robust enough. There was no way to to register the listener extension properly: 
- If I regestered it in `plugin.xml` it would fail because the listener needs to be an inner class of `ProjectSelector` to have access to it.
- If I programmatically register the extension in the `ProjectSelector`'s constructor, this would not work because the account plugin is loaded before the cloud tools plugin.

So instead I moved it to use the built in IntelliJ message bus pub/sub mechanism.


etanshaul:
chasing a "Too many projects leaked" error in the test framework again

etanshaul:
I _think_ the tests are fixed now.

@nkibler | @patflynn | @nbashirbello PTAL

google-account-plugin/testSrc/com/google/cloud/tools/intellij/login/CredentialedUserRosterTest.java
```diff
@@ -47,8 +43,8 @@ public void setUp() {
   }
 
   /**
-   * Tests that {@link com.google.intellij.login.Users#addUser(CredentialedUser)}
-   * stores the proper users and currently manages the active user.
+   * Tests that {@link CredentialedUserRoster#addUser(CredentialedUser)} stores the proper users and
```
nkibler: Are these comments necessary? Typically there are no javadoc comments on tests; it should be easily readable so the reader knows what it's testing.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -346,6 +346,7 @@ gcs.panel.bucket.listing.no.project.selected=To view your buckets select a Cloud
 gcs.panel.bucket.listing.error.loading.buckets=Could not load buckets for selected project.
 gcs.panel.bucket.listing.no.buckets.found=No buckets found for selected project.
 gcs.panel.bucket.listing.loading.text=loading ...
+gcs.panel.bucket.listing.not.logged.in=To view your buckets login to a Cloud Project.
```
nkibler: Technically you're logging into your Google account to view your Cloud projects. Maybe reword this?

google-account-plugin/testSrc/com/google/cloud/tools/intellij/login/CredentialedUserRosterTest.java
etanshaul: true. I wasn't planning on refactoring the existing code except for the broken reference. but I agree these are redundant.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: ah good point. how about: "... login to your Google Cloud Platform account" ?

etanshaul: ill check how we word this elsewhere too

nkibler: SGTM, but maybe split "log in" since "login" is a noun for the act of logging in. :)

PR #1695, opened 2017-10-12
 Prevent browsing GCS bucket blobs when logged out

### LINKED ISSUE #1692: Cloud storage credentials remain valid even after logging out
1) sign in
2) open the gcs panel and choose a bucket. note that the bucket contents are visible
3) sign out using the top right widget
4) refresh the bucket contents

-> Contents load properly and the credentials remain valid. After logging out this should not happen. The access token should be revoked on sign out.

etanshaul:
Fixes #1692

Attempting any action after logging results in:
image

Logging back in and then refreshing allows resuming where the user left off.


etanshaul:
I was able to reuse internal classes to display a module picker:
image

```
List<Module> modules = Lists.newArrayList(ModuleManager.getInstance(project).getModules());
ChooseModulesDialog dialog = new ChooseModulesDialog(event.getProject(), modules, "blah", "blah");
dialog.show();
```

Its worth looking into if we can use this to then add framework support instead of adding the facet directly. Ideally the framework support window would popup so that the configuration is available to the user (adding libraries, generating files etc.).

Issue #1685, opened 2017-10-09
Provide a way to add appengine framework support outside of the 'add framework support' flow.


patflynn:
As a workaround to the bug mentioned in #1666, and to improve discoverability of app engine framework file generation, we should see how else we can help users convert their projects to app engine.  Either through a contextual action or something in the toolbar.

etanshaul:
A toolbar action is not a bad idea. We could programmatically add the facet when clicked.

etanshaul:
@nbashirbello this just came to me when looking at the update on the JetBrains bug:

If, with these new actions, we just add the facet then we may be missing out on all of the other things that go along with adding framework support: like automatically creating the appropriate run configs.

Could you possibly check if:
1) There is a way programmitically trigger the addition of framework support (e.g. Flexible support, and Standard support).
2) and if so, does this do all of the other things typically done when this support is added like auto-creating the run configs.

If possible, I think this is better than what we discussed about adding the facet directly. If its not possible, then adding the facet is ok. WDYT? Sorry for throwing this in after agreeing to a different approach.

etanshaul:
Actually maybe I'm wrong here - like we discussed be there probably isn't a way to do this without a new UI since framework support is scoped to a specific module. If this is the case, then just adding the facet is probably better.

etanshaul:
Posted this in the wrong place before..

But I was able to resuse internal classes to open up a module picker. Its worth seeing if we can use this to add framework support instead of adding the facet directly (ideally showing the framework support dialog so that the user can generate files etc.). This should then take care of adding the run configurations automatically.

![image](https://user-images.githubusercontent.com/1735744/31520304-6a1116b2-af73-11e7-877e-11dd18941421.png)

```
List<Module> modules = Lists.newArrayList(ModuleManager.getInstance(project).getModules());
 ChooseModulesDialog dialog = new ChooseModulesDialog(event.getProject(), modules, "blah", "blah");
 dialog.show();
```

PR #1697, opened 2017-10-13
Adds a 'Add App Engine framework support' menu item - UI only

nbashirbello:
Partial fix for #1685


etanshaul:
could you post some screenshots?

nbashirbello:
addfacet1

nbashirbello:
Based on what direction we decide to go, I'll rename the class to reflect facets or framework

etanshaul:
How about keeping it higher level: "Enable App Engine support". Basically hide "framework" / "facet" from the user.

patflynn:
Do we need the logos in the second submenu?

@etan are you referring to the menu item label? If so I'm a little unsure
about using 'enable' since that doesn't necessary feel like it would be
adding to or modifying the project.  For instance it could be read as
enabling app engine support in the plugin. (Unlikely I know, but users with
no context could be confused)

etanshaul:
Fair enough. How about just “Add App Engine support” then?

Also I agree that the sub menus could be icon less.

patflynn:
works for me.

nbashirbello:
Updated image
screen shot 2017-10-13 at 2 03 25 pm

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -202,6 +202,8 @@ appengine.run.settings.loglevel.label=Application log level:
 appengine.run.settings.adminhost.label=Admin host:
 appengine.run.settings.adminport.label=Admin port:
 appengine.infopanel.noproject=Can't retrieve region for selected project.
+appengine.add.standard.facet.tools.menu.description=Adds the Google App Engine Standard framework support to a module.
```
etanshaul: nit: remove the "the" after "Adds"

google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -463,6 +463,10 @@
     </action>
 
     <group id="GoogleCloudTools" text="Google Cloud Tools" class="com.google.cloud.tools.intellij.CloudToolsMenuGroupAction" popup="true">
+      <group id="AddAppEngineFacet" text="Add App Engine support" class="com.google.cloud.tools.intellij.appengine.facet.AddAppEngineFacetMenuGroupAction" popup="true">
```
etanshaul: I guess this text can't be moved to the message bundle?

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: Standard should not be capitalized. Same for Flexible in the next message.

etanshaul:
lgtm, we should probably comment out the extension points in plugin.xml until the action is implemented

PR #1699, opened 2017-10-13
Add GCS blob name and bucket name copy actions

### LINKED ISSUE #1671: Add ability to get GCS link to a blob
At first it can just be by right clicking a blob > get link.

etanshaul:
fixes #1671 

POC - still needs a little cleanup. 

See the issues for rationale behind this.

image

blob name: an absolute path to the blob; e.g. `folder1/folder2/myblob.zip`

What do you think of this approach? I was also considering having a little bottom notification panel popup on clicking on one of the copy actions to acknowledge that the text was copied (and show its contents).

### DIFF
--- a/common-lib/src/main/java/com/google/cloud/tools/intellij/ui/CopyToClipboardActionListener.java
+++ b/common-lib/src/main/java/com/google/cloud/tools/intellij/ui/CopyToClipboardActionListener.java
@@ -23,7 +23,7 @@ import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 
 /** {@link ActionListener} that copies text to the clipboard on action performed. */
-public class CopyToClipboardActionListener implements ActionListener {
+public final class CopyToClipboardActionListener implements ActionListener {
   private final String text;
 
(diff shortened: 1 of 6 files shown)


common-lib/src/main/java/com/google/cloud/tools/intellij/ui/CopyToClipboardActionListener.java
```diff
@@ -0,0 +1,38 @@
...
+package com.google.cloud.tools.intellij.ui;
+
+import java.awt.Toolkit;
+import java.awt.datatransfer.Clipboard;
+import java.awt.datatransfer.StringSelection;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+
+/** {@link ActionListener} that copies text to the clipboard on action performed. */
+public class CopyToClipboardActionListener implements ActionListener {
```
nkibler: nit: final?

etanshaul:
Updated: now in addition to the right-click menu shown in the PR description, there is a right click menu on the bucket listing panel for copying the bucket name (since this is a perhaps more natural place to find this):

image

PR #1702, opened 2017-10-16
Adds metrics for GCS events

etanshaul:
Adds metrics for the following events:

- bucket listing (for a certain cloud project) - will ping on each listing
- blob browsing - will ping on each browse operation (e.g. each folder traversal)
- usage of each right-click menu action
- StorageExceptions from bucket listing, and blob browsing

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
@@ -21,8 +21,10 @@ import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.Storage.BlobListOption;
 import com.google.cloud.storage.StorageException;
 import com.google.cloud.tools.intellij.login.Services;
+import com.google.cloud.tools.intellij.stats.UsageTrackerProvider;
 import com.google.cloud.tools.intellij.ui.CopyToClipboardActionListener;
 import com.google.cloud.tools.intellij.util.GctBundle;
+import com.google.cloud.tools.intellij.util.GctTracking;
 import com.google.cloud.tools.intellij.util.ThreadUtil;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
@@ -127,6 +129,8 @@ final class GcsBucketContentEditorPanel {
       return;
     }
 
+    pingMetrics(GctTracking.GCS_BLOB_BROWSE);
+
     Consumer<List<Blob>> afterLoad =
         blobs -> {
           if (blobs.isEmpty()) {
@@ -153,6 +157,8 @@ final class GcsBucketContentEditorPanel {
       initTableModel();
       return;
     }
+
+    pingMetrics(GctTracking.GCS_BLOB_BROWSE);
 
     tableModel.setRowCount(0);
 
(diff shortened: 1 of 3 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
```diff
@@ -188,8 +194,19 @@ private void showRightClickMenu(MouseEvent event) {
     Blob selectedBlob = tableModel.getBlobAt(bucketContentTable.rowAtPoint(event.getPoint()));
 
     if (selectedBlob != null) {
+      copyBucketNameMenuItem.addActionListener(
+          e ->
+              UsageTrackerProvider.getInstance()
```
nkibler: Why sometimes use `pingMetrics` and other times use this syntax?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
```diff
@@ -188,8 +194,19 @@ private void showRightClickMenu(MouseEvent event) {
...
+                  .trackEvent(GctTracking.GCS_BUCKET_LIST_COPY_BUCKET_NAME)
+                  .ping());
+
+      copyBlobNameMenuItem.addActionListener(
+          e -> pingMetrics(GctTracking.GCS_BLOB_BROWSE_COPY_BLOB_NAME));
       copyBlobNameMenuItem.addActionListener(
           new CopyToClipboardActionListener(selectedBlob.getName()));
+
+      copyBucketNameMenuItem.addActionListener(
```
nkibler: It looks like this won't be differentiated from the `GCS_BUCKET_LIST_COPY_BUCKET_NAME` metric because the `copyBucketNameMenuItem` is the same instance, right?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
```diff
@@ -240,6 +257,10 @@ private void loadBlobsStartingWith(String prefix, Consumer<List<Blob>> afterLoad
                         () -> {
                           hideLoader();
                           showError();
+
+                          UsageTrackerProvider.getInstance()
```
nkibler: Ditto about using `pingMetrics`.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanel.java
```diff
@@ -283,6 +304,10 @@ private void hideError() {
     errorPanel.setVisible(false);
   }
 
+  private void pingMetrics(String metricsEvent) {
```
nkibler: Can this be static? Also, wouldn't it be more appropriate to put this in the `UsageTrackerProvider` class and static import it here for usage? That way you could re-use it elsewhere to reduce boilerplate.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/util/GctTracking.java
```diff
@@ -64,6 +64,17 @@
 
   public static final String CLOUD_SDK_MALFORMED_PATH = "cloudsdk.malformedpath";
 
+  public static final String GCS_BUCKET_LIST = "cloud.storage.bucket.list";
+  public static final String GCS_BUCKET_LIST_EXCEPTION = "cloud.storage.bucket.list.exception";
+  public static final String GCS_BUCKET_LIST_COPY_BUCKET_NAME =
+      "cloud.storage.bucket.list.copy.blob.name";
```
nkibler: s/blob/bucket/?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/util/GctTracking.java
```diff
@@ -64,6 +64,17 @@
...
+  public static final String GCS_BLOB_BROWSE = "cloud.storage.blob.browse";
+  public static final String GCS_BLOB_BROWSE_COPY_BUCKET_NAME =
+      "cloud.storage.blob.browse.copy.bucket.name";
+  public static final String GCS_BLOB_BROWSE_COPY_BLOB_NAME =
+      "cloud.storage.blob.browse.copy.blob.name";
+  public static final String GCS_BLOB_BROWSE_EXCEPTION = "cloud.storage.blow.browse.exception";
```
nkibler: s/blow/blob/?

etanshaul:
PTAL.

Cleaned it up a bit. Decided to get rid of that pingMetrics method altogether since it was 1 line of code it was replacing.

PR #1704, opened 2017-10-18
Forces the popup panel initialization on show every time

### LINKED ISSUE #1693: Project selector contents not reloaded after login

nkibler:
Fixes #1693.

I spent way too much time trying to find a proper solution, which is to essentially rewrite the `ProjectSelector`. This is a quick fix that forces the panel to be re-initialized every time the pop-up is shown. It also applies the filter every time the pop-up is shown, not just when the text changes.


etanshaul:
nice - this doesn't reintroduce the leak from before though I assume?

nkibler:
No because the `JTree` is mostly what's being recreated on re-initialization. The `PopupPanel` is still retained, but its rendered `JTree` is continually recreated.

PR #1706, opened 2017-10-20
Implementation for "Add App Engine Support" menu action

### LINKED ISSUE #1685: Provide a way to add appengine framework support outside of the 'add framework support' flow.

nbashirbello:
Fixes #1685

This automatically creates the appropriate run configs and gives the user the same options to configure the project.

On selecting the Add Google App Engine support -> Standard/Flex, this opens the Choose Module dialog to allow the user to select a module (only modules that do not have the GAE facet and modules that support the selected facet type are shown), and then opens the appropriate Add Framework Support panel in a dialog

screen shot 2017-10-13 at 2 03 25 pm

screen shot 2017-10-20 at 12 01 27 pm

screen shot 2017-10-20 at 12 01 36 pm


nbashirbello:
Some minor clean up is left

etanshaul:
nice! is it ready for review now?

What happens when you select a module that already has app engine support?

etanshaul:
oh I just reread your description - I guess it will exclude modules that already have this support from the list.

Do you think this is the best approach? What if instead we show all the modules, but then notify the user if they choose one that already as the facet? Perhaps this is less magical and makes things more explicit. Not saying this is the right way just wondering what you think.

nbashirbello:
@etanshaul You can start a reviewing :)

I wanted the experience to be the same as when you select "Add Framework Support..." in the context menu. The facets that cannot be enabled for your module are hidden. 

If not modules can 'support' the selected facet, an error message is displayed.

etanshaul:
ah yea. good point.

nbashirbello:
@etanshaul 
This is one of the small changes I haven't been able to figure out.

The mainPanel consist of 2 panels, cloudSdkPanel and appEngineStandardLibraryPanel
screen shot 2017-10-20 at 2 07 48 pm
. I would like to change the layout so the inner panels location are (top, left). Any ideas on how to do that?

etanshaul:
Have you tried playing around with the layout manager of the mainPanel?

nbashirbello:
Updated GAE standard dialog:

screen shot 2017-10-20 at 2 57 01 pm

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AddAppEngineFrameworkSupportAction.java
```diff
@@ -0,0 +1,100 @@
...
+    Project project = event.getProject();
+    if (project == null) {
+      return;
+    }
+
+    List<Module> suitableModules = new ArrayList<>(
+        Arrays.asList(ModuleManager.getInstance(project).getModules()));
+    DefaultFacetsProvider facetsProvider = new DefaultFacetsProvider();
+    FrameworkSupportInModuleProvider provider = getModuleProvider();
+    suitableModules.removeIf(module -> !provider.isEnabledForModuleType(ModuleType.get(module)) ||
```
etanshaul: nit: do you have the google code formatter setup? formatting on this line looks a bit off; maybe run it on the whole file since its new code

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AddAppEngineFrameworkSupportAction.java
```diff
@@ -0,0 +1,100 @@
...
+        provider.isSupportAlreadyAdded(module, facetsProvider));
```
etanshaul: I'm probably missing something, but how is this removeIf clause working?

My understanding is that the goal here is to remove all modules from the list for which the framework support (flex or standard, depending on which we are dealing with) is already added. I don't see anything specific to flex / standard facet here.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/AddAppEngineStandardFrameworkSupportToolsMenuAction.java
```diff
@@ -0,0 +1,64 @@
...
+    Project project = module.getProject();
+    String contentRootPath = "";
+    VirtualFile[] roots = ModuleRootManager.getInstance(module).getContentRoots();
+    if (roots.length > 0) {
+      contentRootPath = roots[0].getPath();
+    }
+
+    FrameworkSupportModelImpl model = new FrameworkSupportModelImpl(project, contentRootPath,
+        LibrariesContainerFactory.createContainer(project));
+    return provider.createConfigurable(model);
```
etanshaul: Why the is the approach for this method different here than it is in `AddAppEngineFlexibleFrameworkSupportToolsMenuAction`?

I see that we have `AppEngineStandardSupportProvider.AppEngineSupportConfigurable` (which should probably be renamed to `AppEngineStandardSupportConfigurable`). Can we reuse that?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AddAppEngineFlexibleFrameworkSupportToolsMenuAction.java
```diff
@@ -0,0 +1,49 @@
...
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Creates a shortcut in the tools menu to add the App Engine Flexible framework support to a
+ * module.
+ */
+public class AddAppEngineFlexibleFrameworkSupportToolsMenuAction extends AddAppEngineFrameworkSupportAction {
+  public AddAppEngineFlexibleFrameworkSupportToolsMenuAction() {
+    super(GctBundle.message("appengine.flexible.facet.name"),
+        GctBundle.message("appengine.flexible.name.in.message"));
```
etanshaul: could you just collapse this into each description? I.e. just have 2 description messages; 1 for flex and 1 for standard. I think that would make things clearer.

etanshaul: (and you could get rid of this "name.in.message" message)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AddAppEngineFrameworkSupportAction.java
```diff
@@ -0,0 +1,105 @@
...
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Action to add either App Engine Flexible or Standard framework support to a module which
+ * includes creating appropriate facet and run configurations.
+ */
+public abstract class AddAppEngineFrameworkSupportAction extends AnAction {
+  private final String FRAMEWORK_NAME_IN_MESSAGE;
+
+  public AddAppEngineFrameworkSupportAction(@NotNull String nameInTitle,
```
etanshaul: package-private constructor?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/AddAppEngineStandardFrameworkSupportToolsMenuAction.java
nbashirbello: The constructor for AppEngineStandardSupportProvider.AppEngineSupportConfigurable is private so we use AppEngineStandardSupportProvider #createConfigurable

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AddAppEngineFlexibleFrameworkSupportToolsMenuAction.java
nbashirbello: The name.in.message is used for 2 other dialogs, should I still collapse the description?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AddAppEngineFrameworkSupportAction.java
nbashirbello: The classes that inherit are in different packages: com.google.cloud.tools.intellij.appengine.facet.standard and com.google.cloud.tools.intellij.appengine.facet.flexible

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AddAppEngineFrameworkSupportAction.java
etanshaul: aha. I see now.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/AddAppEngineStandardFrameworkSupportToolsMenuAction.java
etanshaul: it doesn't have to be private though, no?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/AddAppEngineFrameworkSupportAction.java
etanshaul: sorry, missed that

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AddAppEngineFlexibleFrameworkSupportToolsMenuAction.java
etanshaul: I ran a 'find usages' and see the "name.in.message" being used only for these new Actions. 1 usage for the flex and 1 usage for the standard one.

nbashirbello: In AddAppEngineFrameworkSupportAction, the name.in.message is assigned to a FRAMEWORK_NAME_IN_MESSAGE and utilized in the ChooseModule and error dialog

etanshaul: Ok. Maybe its the naming thats bothering me. How about:

appengine.flexible.facet.name > appengine.flexible.facet.name.title
appengine.flexible.name.in.message > appengine.flexible.facet.name

Also updating the variable names accordingly.

Otherwise, just running the formatter as I mentioned in another comment and LGTM.

PR #1711, opened 2017-10-24
Fixes project selector reloading issues with applied filters

### LINKED ISSUE #1709: Same Cloud Project displays multiple times in project selector

- Cloud Tools for IntelliJ version: 17.GCS-RC2
- Google Cloud SDK version _(run `gcloud version` on the command line)_: 176.0.0
- OS: Linux

**What did you do?**
I wanted to select a different GCS bucket for my app. When I clicked the drop-down menu, I saw the same project listed 3 times.

I expected to see the project that I selected and then a list of the available GCP projects.
**What did you see instead?**

(screenshots are helpful)

(https://user-images.githubusercontent.com/11820506/31950228-26e829b8-b8a9-11e7-99d5-9ed18acd8686.png)

nkibler:
Fixes #1709.

The problem is that the filtering logic overrides the `getChildAt()` and `getChildCount()` methods in the `GoogleUserModelItem`. While it's important to do this for rendering purposes (so it only renders filtered children), when removing children, we never want to restrict the removal to just the filtered children. This PR adds overrides to the `remove()` and `removeAllChildren()` methods to not rely on the `getChildAt()` or `getChildCount()` methods and instead use the raw `children` vector directly.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/GoogleUserModelItem.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/GoogleUserModelItem.java
@@ -37,6 +37,7 @@ import java.util.stream.Stream;
 import javax.swing.SwingUtilities;
 import javax.swing.tree.DefaultMutableTreeNode;
 import javax.swing.tree.DefaultTreeModel;
+import javax.swing.tree.MutableTreeNode;
 import javax.swing.tree.TreeNode;
 import org.jetbrains.annotations.NotNull;
 
@@ -56,6 +57,7 @@ class GoogleUserModelItem extends DefaultMutableTreeNode {
   private volatile boolean isSynchronizing;
   private volatile boolean needsSynchronizing;
   private CloudResourceManager cloudResourceManagerClient;
+  private String filter;
 
   GoogleUserModelItem(@NotNull CredentialedUser user, @NotNull DefaultTreeModel treeModel) {
     this.user = user;
@@ -122,6 +124,7 @@ class GoogleUserModelItem extends DefaultMutableTreeNode {
 
   @SuppressWarnings("unchecked")
   void setFilter(String filter) {
+    this.filter = filter;
     children.forEach(
         child -> {
           if (child instanceof ResourceProjectModelItem) {
@@ -203,6 +206,8 @@ class GoogleUserModelItem extends DefaultMutableTreeNode {
               GoogleUserModelItem.this.add(item);
             }
 
+            // Re-applies the filter to the new child nodes.
+            setFilter(filter);
             treeModel.reload(GoogleUserModelItem.this);
           });
     } catch (InterruptedException ex) {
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/GoogleUserModelItem.java
```diff
@@ -225,6 +230,35 @@ public int getChildCount() {
...
+  @Override
+  public void removeAllChildren() {
+    if (children == null) {
+      return;
+    }
+
+    // This is different from the default implementation because it uses the raw children vector
+    // instead of relying on the getChildCount() method, which has been modified to only return
+    // the number of filtered children.
+    for (int i = children.size() - 1; i >= 0; i--) {
```
etanshaul: nit: `forEach` instead?

etanshaul:
lgtm. but could you point out the part relevant to the original bug (that the project was being displayed multiple times)?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/GoogleUserModelItem.java
nkibler: It would probably throw a `ConcurrentModificationException` since the loop modifies the underlying `children` vector. Also, I believe the point of reverse-iterating through the children is to prevent any shifting of positions that the `Vector` might do if I removed from beginning to end.

nkibler:
The relevant part is the override to `removeAllChildren()`. Here's the (unmodified) line that was causing issues: https://github.com/GoogleCloudPlatform/google-cloud-intellij/blob/d5b0ef5a0bf9264749249ffe63bf6b3e8c03be96/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/GoogleUserModelItem.java#L203

Before this change, the `removeAllChildren()` method would use a count and `getChildAt()` method that depended on the filtered children. After this change, this line to `removeAllChildren()` before we add the new ones on synchronization will actually remove all the children, regardless of the applied filter.

etanshaul:
makes sense. was there anything special you had to do to trigger the bug? Like have a filter then close and reopen the panel, or something like that? Just asking cause I wasn't able to trigger it.

nkibler:
Yeah, my repro steps were:

1. Open the project selector by any means.
2. Type text to apply a filter.
3. Press the _Synchronize_ button.
4. Modify the text to actually re-apply the filter after the synchronization and notice the duplicated nodes.

That last point was also fixed; the filter is now re-applied to the new nodes after synchronizing.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/GoogleUserModelItem.java
```diff
@@ -203,6 +206,8 @@ private void loadUserProjects() {
               GoogleUserModelItem.this.add(item);
             }
 
+            // Re-applies the filter to the new child nodes.
+            setFilter(filter);
```
etanshaul: should we check for null here? I know that ResourceProjectModelItem takes a nullable filter, but maybe this could same some cycles?

nkibler: Done.

PR #1713, opened 2017-10-25
Make text size consistent in the app engine application creation dialog

### LINKED ISSUE #1664: text size is not consistent in the app engine application creation dialog
screenshot from 2017-09-26 16 36 44

nbashirbello:
Fixes #1664


nbashirbello:
@etanshaul 
Are the text now the same size? I can't tell anymore lol

screen shot 2017-10-25 at 3 14 12 pm

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
```diff
@@ -74,6 +71,11 @@ public AppEngineApplicationCreateDialog(@NotNull Component parent, @NotNull Stri
     super(parent, false);
     this.gcpProjectId = gcpProjectId;
     this.userCredential = userCredential;
+
+    Font font = parent.getFont();
+    HTML_OPEN_TAG =
+        "<html><font face=\"" + font.getName() + "\" size=\\\"" + font.getSize() + "\\\">";
```
etanshaul: just curious, why the triple `\`? It looks like it would result in `size=\"12\"`. Also I don't think font size need double quotes around the number.

etanshaul:
I think so! definitely looks better than before

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/application/AppEngineApplicationCreateDialog.java
nbashirbello: The single quotes around the font size mean something a different: https://www.w3schools.com/tags/att_font_size.asp

etanshaul:
@nbashirbello could update this? Last I remember that the fix wasn't consistent across platforms right? If its not the correct fix then we can close the PR.

nbashirbello:
I think so, but I'll verify and update PR.

nbashirbello:
Fix on linux machine:
screenshot from 2017-11-21 14 21 56

Works well on linux without fix.

nbashirbello:
Text variation only occurs on mac and this fix isn't top-notch.

Issue #1714, opened 2017-10-26
The Select An App Engine Region dialog reopens once closed


nbashirbello:
**What did you do?**
In the App Engine Run/Debug configuration wizard
1. select a GCP project that does not have an App Engine application
2. The Region field would show an error message with a link to a dialog to create an App Engine application
3. Select that link and the Select An App Engine Region dialog would open up
4. In the  Select An App Engine Region dialog, click the Cancel button

**What did you expect to see?**
The Select An App Engine Region dialog should close and the  Run/Debug configuration wizard should be in focus

**What did you see instead?**
The Select An App Engine Region dialog closes and opens again. If you click the "Cancel" button again, it closes and does not reopen

<img width="794" alt="screen shot 2017-10-26 at 11 26 52 am" src="https://user-images.githubusercontent.com/6805942/32061789-9ae3fdd6-ba40-11e7-8c3e-1732174ead83.png">

nbashirbello:
This bugs seems to occur only the first time you open the Run/Debug configuration wizard in an IDE session

etanshaul:
that makes it even stranger

nkibler:
Fixed by #1718.

PR #1718, opened 2017-11-01
Fixes the ProjectSelector issues related to text changed listeners.

### LINKED ISSUE #1717: Project selector not responding properly to keyboard events
1) Tools -> Google Cloud Tools -> Setup on Cloud Source Repositories
2) Select a Cloud Project that has a cloud repository -> notice the cloud repository dropdown displays the repo
3) type something in the project selector to make the previous project invalid
4) delete what you did in step 3

=> The Cloud repository dropdown is not updated with the repo

Confirmed to be introduced in da734da35987b9c8496f67f38ad14f281b3966e4

Notes:

the listener is being attached to the project selector via:

```
projectSelector.getDocument().addDocumentListener
```

We need solidify (and make clear in the selector's API), the proper way to attach listeners that handles all events properly:
- text changes via manual typing
- changes by manually selecting from the tree
- etc.

Currently we are attaching them in multiple ways throughout the codebase.

nkibler:
Fixes #1717 and #1714.

There's several temporary, hacky solutions in this PR that solve a few issues:

1. **Issue:** In the GCR/GCS/deployment config editor windows, backspacing the `ProjectSelector` text so the text becomes a valid project does not properly update other dependent windows/dropdowns/etc. to read from this valid project.
  **Cause:** The listeners were attached directly to the text field document. This is also how the listener to apply the filter to the TreeModel is attached. We are not guaranteed an order here; if the filter listener does not run first, it will not add back the model item to the tree that matches the project so the other listeners will get `null` when trying to pull the selected project from the `ProjectSelector`.
  **Solution:** Add a `.addTextChangedListener()` method to the `ProjectSelector` and manually trigger these listeners, only after the filter listener runs first.
1. **Issue:** Even after the change above, the deployment config editor still does not update the region text if it is supposed to go from "Can't retrieve region for selected project" to the hyperlink text to create an App Engine app.
  **Cause:** Not 100% sure, but it seems like something internal to the `HyperlinkLabel` that we're using (part of the IntelliJ SDK). Somehow related to calling `.setText()`, then `.setHyperlinkText()`.
  **Solution:** Recreate the `HyperlinkLabel` any time we want to set the region text to the hyperlink text to create an App Engine app.
1. **Issue:** Closing the dialog to create an App Engine app causes it to re-open, but only on the first open since the deployment config editor was opened.
  **Cause:** Two listeners were being attached to the `HyperlinkLabel` by accident.
  **Solution:** Same as the solution about (recreating the `HyperlinkLabel` fixes this).

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineApplicationInfoPanel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineApplicationInfoPanel.java
@@ -114,31 +114,30 @@ public class AppEngineApplicationInfoPanel extends JPanel {
   }
 
   private void setMessage(Runnable messagePrinter, boolean isError) {
-    ApplicationManager.getApplication().invokeAndWait(() -> {
-      errorIcon.setVisible(isError);
-      messagePrinter.run();
-    }, ModalityState.stateForComponent(this));
-  }
+    ApplicationManager.getApplication()
+        .invokeAndWait(
+            () -> {
+              errorIcon.setVisible(isError);
 
-  /**
-   * Prints a message with a hyperlink.
-   */
-  private void setMessage(String beforeLinkText, String linkText, String afterLinkText) {
-    setMessage(() -> messageText.setHyperlinkText(beforeLinkText, linkText, afterLinkText), true);
+              // TODO(nkibler): Figure out what's causing the HyperlinkLabel to not refresh its view
+              // if we re-use the label here.
+              remove(messageText);
+              messageText = new HyperlinkLabel();
+              messageText.setOpaque(false);
+              add(messageText);
+
+              messagePrinter.run();
+            },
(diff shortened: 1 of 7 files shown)


etanshaul:
Thanks for doing this!

I tested out this branch and almost everything looks like its working again. One curious thing I was able to do is get the region indicator in this state:

image

I did this by manually typing in the project selector - adding characters then removing them (or vice versa)

etanshaul:
I think I was also able to break it this way:

1) select a project that has no region, the correct message is shown:
image


2) Select a project that does have a region (I clicked on one from the tree):
image

The region label is empty until click I on another widget on the form (after which, it gets populated).

I can't recall if this was a known limitation we discussed.

nkibler:
This should be fixed now, PTAL. Now, any time the text is set (regardless of whether it's the "Create App Engine app" message or not), we recreate the HyperlinkLabel. I couldn't repro those issues with this change, though I can't tell you exactly what was causing the blank message text.

Story of my life with this project selector. :)

PR #1728, opened 2017-11-03
Update CloudToolsBundle.properties

parkjam4:
The current error message is too long and get cut off.


etanshaul:
Thanks @parkjam4 lets merge this in when the checks are done so we can include it in the release

parkjam4:
Thanks!

PR #1731, opened 2017-11-03
Update project selector button area on login events

### LINKED ISSUE #1730: Sign in button still displays after I signed in.
The GCS project selector still shows a sign in button after I signed in.

gcs

etanshaul:
fixes #1730 

See the issue for details of the problem

### DIFF
--- a/common-lib/src/main/java/com/google/cloud/tools/intellij/util/ProjectUtil.java
+++ b/common-lib/src/main/java/com/google/cloud/tools/intellij/util/ProjectUtil.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.intellij.util;
+
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.project.ProjectManager;
+import com.intellij.util.messages.Topic;
+import java.util.stream.Stream;
+import org.jetbrains.annotations.NotNull;
+
+/** Project utility for various {@link Project} related utilities. */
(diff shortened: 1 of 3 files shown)


common-lib/src/main/java/com/google/cloud/tools/intellij/util/ProjectUtil.java
```diff
@@ -0,0 +1,42 @@
...
+  private ProjectUtil() {}
+
+  /** Return instance of this utility. */
+  public static ProjectUtil getInstance() {
+    return INSTANCE;
+  }
+
+  /** Subscribes all open projects to the given topic with the supplied handler. */
+  public <L> void subscribeAll(@NotNull Topic<L> topic, @NotNull L handler) {
+    Stream.of(ProjectManager.getInstance().getOpenProjects())
```
nkibler: nit: Consider using `Arrays.stream()` instead of `Stream.of()` here when working with streaming array sources. One main reason is:

```
List<String> stringList = new ArrayList<String>();

// This works:
Stream.of(stringList);

// This fails to compile:
Arrays.stream(stringList);
```

While this may seem insignificant, it prevents accidental mistakes where calling a method that you *think* returns an array that actually returns a list would silently succeed, returning you a `Stream<List<String>>` instead of a `Stream<String>`.

Another reason has to do with primitive types, where `Stream.of(new int[1])` returns you a `Stream<int[]>`, but `Arrays.stream(new int[1])` returns you an `IntStream`.

google-account-plugin/src/com/google/cloud/tools/intellij/login/ui/GoogleLoginEmptyPanel.java
```diff
@@ -53,14 +50,14 @@ public GoogleLoginEmptyPanel() {
...
-    addAccountButton.addActionListener(new ActionListener() {
-      @Override
-      public void actionPerformed(ActionEvent event) {
-        doLogin();
-      }
-    });
+    addAccountButton.addActionListener(event -> doLogin());
     addAccountButton.setHorizontalAlignment(SwingConstants.LEFT);
 
+    ProjectUtil.getInstance()
```
nkibler: Won't this leave a bunch of dangling subscribers for buttons that no longer exist?

etanshaul: yea. this needs some cleanup

etanshaul: a similar problem is that the subscribers keep getting registered multiple times (every time the project selector is initialized)

etanshaul:
This is the wrong fix. The problem has to do with the initialization sequence of the popup panel. Working on   a better approach.

PR #1732, opened 2017-11-05
Reinitialize project selector bottom panel

### LINKED ISSUE #1730: Sign in button still displays after I signed in.

etanshaul:
fixes #1730 

**What was causing the problem?**
After the changes to make the project selector searchable, the [logic was updated](https://github.com/GoogleCloudPlatform/google-cloud-intellij/blob/master/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/ProjectSelector.java#L417) to not recreate the project selector popup panel each time it was opened. Rather, its contents were just reinitialized.

The problem is that the bottom panel, which is created in `GoogleLoginEmptyPanel` was not also reinitialized. Because of this, the state of this bottom panel was not properly updated on login / logout events.

**Why does this PR fix the problem?**
Alongside reinitializing the popup panel, this PR adds a call to reinitialize the bottom panel as well. This will force the panel to be redrawn correctly according to the current login state when the project selector popup is opened.


etanshaul:
@nkibler ptal

Issue #1698, opened 2017-10-13
spurious app engine flexible detections


patflynn:
Any project with an app.yml or app.yaml in the source will cause the user to get an 'App Engine Flexible detected' notification.

Unfortunately this is not an unusual filename for users of Docker swarm and/or jhipster who aren't using Flex at all.  Ideally we'd inspect the file contents to validate that it's a app engine config file.

etanshaul:
just wondering if these the recommended names for these other frameworks? a quick google search for app.yaml / app.yml with docker swarm / jhipster doesn't result in anything relevant. Also searching for just app.yaml alone results in only GAE stuff.

patflynn:
I don't think it's a recommended name, but it's a likely name.  For
instance jhipster generates a compose file named app.yml when you generate
a microservice based app.

Anybody using swarm, kubernetes, compose, or some paas that's yaml based
would hit this if they called the yml app.(yaml|yml)

etanshaul:
We can do a simple check for the presence of `runtime: *` in the app.yaml since this line [is required](https://cloud.google.com/appengine/docs/flexible/nodejs/configuring-your-app-with-app-yaml#general).

PR #1737, opened 2017-11-07
Improve app engine flexible detection

ivanporty:
- Fixes issue #1698
- Replaces checking for simple presence of app engine flexible project file with actual checking for required YAML content ("runtime:")
- Adds unit test to test only valid app.yaml files are accepted when detecting app engine flexible

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
@@ -41,7 +41,8 @@ public class AppEngineFlexibleFrameworkDetector
         AppEngineFlexibleFacet, AppEngineFlexibleFacetConfiguration> {
 
   // app engine flex file names.
-  static final List<String> APP_ENGINE_FLEX_PROJECT_FILES = Collections.unmodifiableList(Arrays.asList("app.yaml",  "app.yml"));
+  static final List<String> APP_ENGINE_FLEX_PROJECT_FILES =
+          Collections.unmodifiableList(Arrays.asList("app.yaml",  "app.yml"));
   // required string in app engine flex YAML project file.
   static final String APP_ENGINE_REQUIRED_YAML = "runtime:";
 
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
@@ -78,7 +78,7 @@ public class AppEngineFlexibleFrameworkDetector
    */
   private static final class AppEngineFlexPattern
       extends ObjectPattern<FileContent, AppEngineFlexPattern> {
-    private static final Collection<String> APP_ENGINE_FLEX_CONFIG_FILES =
+    private static final ImmutableList<String> APP_ENGINE_FLEX_CONFIG_FILES =
         ImmutableList.of("app.yaml", "app.yml");
 
     private static final String APP_ENGINE_FLEX_REQUIRED_YAML_CONTENT = "runtime:";
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
@@ -86,7 +86,6 @@ public class AppEngineFlexibleFrameworkDetector
     private AppEngineFlexPattern withAppEngineFlexYamlContent() {
       return with(
           new PatternCondition<FileContent>("with-appengine-java-flexible") {
-            // app engine flex file names.
             private final Collection<String> APP_ENGINE_FLEX_CONFIG_FILES =
                 ImmutableList.of("app.yaml", "app.yml");
 
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
@@ -78,6 +78,10 @@ public class AppEngineFlexibleFrameworkDetector
    */
   private static final class AppEngineFlexPattern
       extends ObjectPattern<FileContent, AppEngineFlexPattern> {
+    private static final Collection<String> APP_ENGINE_FLEX_CONFIG_FILES =
+        ImmutableList.of("app.yaml", "app.yml");
+
+    private static final String APP_ENGINE_FLEX_REQUIRED_YAML_CONTENT = "runtime:";
 
     private AppEngineFlexPattern() {
       super(FileContent.class);
@@ -86,11 +90,6 @@ public class AppEngineFlexibleFrameworkDetector
     private AppEngineFlexPattern withAppEngineFlexYamlContent() {
       return with(
           new PatternCondition<FileContent>("with-appengine-java-flexible") {
-            private final Collection<String> APP_ENGINE_FLEX_CONFIG_FILES =
-                ImmutableList.of("app.yaml", "app.yml");
-
-            private static final String APP_ENGINE_FLEX_REQUIRED_YAML_CONTENT = "runtime:";
-
             @Override
             public boolean accepts(@NotNull FileContent fileContent, ProcessingContext context) {
               // checks for flex engine file names and then checks for required configuration line
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
@@ -42,7 +42,7 @@ public class AppEngineFlexibleFrameworkDetector
 
   // app engine flex file names.
   static final Collection<String> APP_ENGINE_FLEX_PROJECT_FILES =
-          Collections.unmodifiableList(Arrays.asList("app.yaml",  "app.yml"));
+      Collections.unmodifiableList(Arrays.asList("app.yaml", "app.yml"));
   // required string in app engine flex YAML project file.
   static final String APP_ENGINE_REQUIRED_YAML = "runtime:";
 
@@ -83,30 +83,30 @@ public class AppEngineFlexibleFrameworkDetector
    * IntelliJ API pattern class that checks for App Engine Flex project file presence and checks it
    * has required configuration lines to avoid spurious detection.
    */
-  private static class AppEngineFlexPattern extends ObjectPattern<FileContent, AppEngineFlexPattern> {
+  private static class AppEngineFlexPattern
+      extends ObjectPattern<FileContent, AppEngineFlexPattern> {
 
     private AppEngineFlexPattern() {
       super(FileContent.class);
     }
 
     private AppEngineFlexPattern withAppEngineFlexYAMLContent() {
-      return with(new PatternCondition<FileContent>("with-appengine-java-flexible") {
-        @Override
-        public boolean accepts(@NotNull FileContent fileContent, ProcessingContext context) {
-          // checks for flex engine file names and then checks for required configuration line inside.
-          boolean nameMatch = APP_ENGINE_FLEX_PROJECT_FILES.contains(fileContent.getFileName());
-          if (nameMatch) {
-            Scanner scanner = new Scanner(fileContent.getContentAsText().toString());
-            while (scanner.hasNextLine()) {
-              if (scanner.nextLine().startsWith(APP_ENGINE_REQUIRED_YAML))
-                return true;
+      return with(
+          new PatternCondition<FileContent>("with-appengine-java-flexible") {
+            @Override
+            public boolean accepts(@NotNull FileContent fileContent, ProcessingContext context) {
+              // checks for flex engine file names and then checks for required configuration line
+              // inside.
+              boolean nameMatch = APP_ENGINE_FLEX_PROJECT_FILES.contains(fileContent.getFileName());
+              if (nameMatch) {
+                Scanner scanner = new Scanner(fileContent.getContentAsText().toString());
+                while (scanner.hasNextLine()) {
+                  if (scanner.nextLine().startsWith(APP_ENGINE_REQUIRED_YAML)) return true;
+                }
+              }
+              return false;
             }
-          }
-          return false;
-        }
-      });
+          });
     }
-
   }
-
 }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
@@ -71,8 +71,8 @@ public class AppEngineFlexibleFrameworkDetector
   }
 
   /**
-   * IntelliJ API pattern class that checks for App Engine Flex project file presence and checks it has
-   * required configuration lines to avoid spurious detection.
+   * IntelliJ API pattern class that checks for App Engine Flex project file presence and checks it
+   * has required configuration lines to avoid spurious detection.
    */
   private static class AppEngineFlexFileCondition extends ObjectPattern<FileContent, FileContentPattern> {
 
@@ -85,7 +85,8 @@ public class AppEngineFlexibleFrameworkDetector
         @Override
         public boolean accepts(@NotNull FileContent fileContent, ProcessingContext context) {
           // checks for flex engine file names and then checks for required configuration line inside.
-          boolean nameMatch = fileContent.getFileName().equals("app.yaml") || fileContent.getFileName().equals("app.yml");
+          boolean nameMatch = fileContent.getFileName().equals("app.yaml") ||
+              fileContent.getFileName().equals("app.yml");
           if (nameMatch) {
             Scanner scanner = new Scanner(fileContent.getContentAsText().toString());
             while (scanner.hasNextLine()) {
@@ -93,7 +94,6 @@ public class AppEngineFlexibleFrameworkDetector
                 return true;
             }
           }
-
           return false;
         }
       });
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
@@ -18,10 +18,8 @@ package com.google.cloud.tools.intellij.appengine.facet.flexible;
 
 import com.google.cloud.tools.intellij.stats.UsageTrackerProvider;
 import com.google.cloud.tools.intellij.util.GctTracking;
-import com.google.common.collect.Lists;
 import com.intellij.facet.FacetType;
 import com.intellij.framework.detection.FacetBasedFrameworkDetector;
-import com.intellij.framework.detection.FileContentPattern;
 import com.intellij.openapi.fileTypes.FileType;
 import com.intellij.openapi.roots.ModifiableRootModel;
 import com.intellij.patterns.ElementPattern;
@@ -29,11 +27,13 @@ import com.intellij.patterns.ObjectPattern;
 import com.intellij.patterns.PatternCondition;
 import com.intellij.util.ProcessingContext;
 import com.intellij.util.indexing.FileContent;
-
-import java.util.*;
-
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.yaml.YAMLFileType;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Collection;
+import java.util.Scanner;
 
 /** Detects App Engine Flexible framework in a project. */
 public class AppEngineFlexibleFrameworkDetector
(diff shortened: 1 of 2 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
```diff
@@ -18,24 +18,34 @@
...
+import java.util.Collection;
+import java.util.Scanner;
+
 /** Detects App Engine Flexible framework in a project. */
 public class AppEngineFlexibleFrameworkDetector
     extends FacetBasedFrameworkDetector<
         AppEngineFlexibleFacet, AppEngineFlexibleFacetConfiguration> {
 
+  // app engine flex file names.
+  static final Collection<String> APP_ENGINE_FLEX_PROJECT_FILES =
```
etanshaul: APP_ENGINE_FLEX_PROJECT_FILES -> APP_ENGINE_FLEX_CONFIG_FILES?

Also is this non private just for test? Either add @VisibleForTesting or consider not using this field in your test.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
```diff
@@ -18,24 +18,34 @@
...
+      Collections.unmodifiableList(Arrays.asList("app.yaml", "app.yml"));
+  // required string in app engine flex YAML project file.
```
etanshaul: I think the comment is redundant probably

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
```diff
@@ -66,8 +76,37 @@ public FileType getFileType() {
...
+   * has required configuration lines to avoid spurious detection.
+   */
+  private static class AppEngineFlexPattern
+      extends ObjectPattern<FileContent, AppEngineFlexPattern> {
+
+    private AppEngineFlexPattern() {
+      super(FileContent.class);
+    }
+
+    private AppEngineFlexPattern withAppEngineFlexYAMLContent() {
```
etanshaul: nit: withAppEngineFlexYAMLContent -> withAppEngineFlexYamlContent

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
```diff
@@ -18,24 +18,34 @@
...
 /** Detects App Engine Flexible framework in a project. */
 public class AppEngineFlexibleFrameworkDetector
     extends FacetBasedFrameworkDetector<
         AppEngineFlexibleFacet, AppEngineFlexibleFacetConfiguration> {
 
+  // app engine flex file names.
+  static final Collection<String> APP_ENGINE_FLEX_PROJECT_FILES =
+      Collections.unmodifiableList(Arrays.asList("app.yaml", "app.yml"));
+  // required string in app engine flex YAML project file.
+  static final String APP_ENGINE_REQUIRED_YAML = "runtime:";
```
etanshaul: APP_ENGINE_FLEX_REQUIRED_YAML_CONTENT? Just to keep consistent with your pattern method naming

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
etanshaul: nitpicking, but ImmutableList.of might be more concise

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
```diff
@@ -66,8 +76,37 @@ public FileType getFileType() {
...
+          new PatternCondition<FileContent>("with-appengine-java-flexible") {
+            @Override
+            public boolean accepts(@NotNull FileContent fileContent, ProcessingContext context) {
+              // checks for flex engine file names and then checks for required configuration line
+              // inside.
+              boolean nameMatch = APP_ENGINE_FLEX_PROJECT_FILES.contains(fileContent.getFileName());
+              if (nameMatch) {
+                Scanner scanner = new Scanner(fileContent.getContentAsText().toString());
+                while (scanner.hasNextLine()) {
+                  if (scanner.nextLine().startsWith(APP_ENGINE_REQUIRED_YAML)) return true;
```
etanshaul: @nkibler I'm personally ok with it, but does google style guid allow {}'less statements like this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
nkibler: +1, use Guava's `ImmutableList` in most places. The syntax is shorter, it promotes immutability, and it guarantees order.

nkibler: +1, add `private` wherever you can to these static variables. If you really need to access them in your test (you should consider alternatives first), opening to package-private with `@VisibleForTesting` is ok.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
nkibler: It does not, sorry! http://go/java-style#s4.1.1-braces-always-used

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
nkibler: These static variables are only used by the inner `AppEngineFlexPattern` class; consider moving them inside the inner class to reduce the scope to where they are used.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
```diff
@@ -66,8 +76,37 @@ public FileType getFileType() {
...
-        FileContentPattern.fileContent().withName("app.yaml"),
-        FileContentPattern.fileContent().withName("app.yml"));
+    return new AppEngineFlexPattern().withAppEngineFlexYAMLContent();
+  }
+
+  /**
+   * IntelliJ API pattern class that checks for App Engine Flex project file presence and checks it
+   * has required configuration lines to avoid spurious detection.
+   */
+  private static class AppEngineFlexPattern
```
nkibler: nit: Use `final` by default to prevent unwanted inheritance, or document for inheritance if it's expected: http://go/java-practices/final

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetectorTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/** Tests {@link AppEngineFlexibleFrameworkDetector} for valid/invalid App Engine Flex detection. */
+public class AppEngineFlexibleFrameworkDetectorTest {
```
nkibler: nit: ditto about final

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetectorTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.jetbrains.yaml.YAMLFileType;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/** Tests {@link AppEngineFlexibleFrameworkDetector} for valid/invalid App Engine Flex detection. */
```
nkibler: nit: To prevent this comment from going stale if the tests change (or if the purpose of `AppEngineFlexibleFrameworkDetector` changes), consider just stating:

    /** Unit tests for {@link AppEngineFlexibleFrameworkDetector}. */

... which provides a rich link to the code under test.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetectorTest.java
```diff
@@ -0,0 +1,90 @@
...
+public class AppEngineFlexibleFrameworkDetectorTest {
+
+  private final String validAppEngineFlexYamlString =
```
nkibler: nit: This can be static.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetectorTest.java
```diff
@@ -0,0 +1,90 @@
...
+      AppEngineFlexibleFrameworkDetector.APP_ENGINE_REQUIRED_YAML + " java";
```
nkibler: Instead of using the constant in the file, use the actual value you would expect to pass. For example, change this to be "runtime: java".

This separation between test and code under test is important for several reasons. One is that if this constant were changed to something unexpected (i.e. a bug), the test would pass in its current form. However, if it's hardcoded to the expected value of "runtime: java" in the test, the developer would have to specifically change it in the test too, at which point they might realize the constant should not have changed.

It's a bit of a stretch in this case since how often do constants "accidentally" change? But consider the case where the field used contains logic (i.e. a `Function`, `Consumer`, etc.) or the shared piece of code is actually a method? It's easier to see how the test no longer validates the code under test does what it's supposed to do since it shares code with it.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetectorTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+  @Before
+  public void setUp() {
+    FileTypeRegistry mockFileTypeRegistry = mock(FileTypeRegistry.class);
```
nkibler: You should use the `@Mock` annotation and make this an instance field. Then use the `CloudToolsRule` so the mock is set-up automatically for you.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetectorTest.java
```diff
@@ -0,0 +1,90 @@
...
+    // mock file type check routine. getInstance() is static, has to replace it with IJ Getter mock.
+    FileTypeRegistry.ourInstanceGetter = () -> mockFileTypeRegistry;
+    when(mockFileTypeRegistry.getFileTypeByFile(any(VirtualFile.class)))
+        .thenReturn(YAMLFileType.YML);
+  }
+
+  @Test
+  public void testIncompleteAppYamlDetection() {
+    AppEngineFlexibleFrameworkDetector detector = new AppEngineFlexibleFrameworkDetector();
```
nkibler: This should be made an instance variable to remove bloat from each test method (since this is the class under test so you'll always need an instance of it).

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetectorTest.java
```diff
@@ -0,0 +1,90 @@
...
+  public void setUp() {
+    FileTypeRegistry mockFileTypeRegistry = mock(FileTypeRegistry.class);
+    // mock file type check routine. getInstance() is static, has to replace it with IJ Getter mock.
+    FileTypeRegistry.ourInstanceGetter = () -> mockFileTypeRegistry;
+    when(mockFileTypeRegistry.getFileTypeByFile(any(VirtualFile.class)))
+        .thenReturn(YAMLFileType.YML);
+  }
+
+  @Test
+  public void testIncompleteAppYamlDetection() {
```
nkibler: nit: Consider using the naming pattern of: _methodUnderTest_withParams_expectedResult_, where _withParams_ is optional. This helps understand what the test is doing without needing to read it. And if the test fails, it's even easier to know what's gone wrong.

For example, I might name this one: _createSuitableFilePattern_withInvalidFileName_doesNotMatch_

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetectorTest.java
```diff
@@ -0,0 +1,90 @@
...
+    // unaccepted file name pattern, although correct content.
+    MockVirtualFile invalidNameFile =
+        new MockVirtualFile("myApp.yaml", validAppEngineFlexYamlString);
+    FileContent wrongFile =
+        new FileContentImpl(
+            invalidNameFile, validAppEngineFlexYamlString, System.currentTimeMillis());
+    Assert.assertFalse(pattern.accepts(wrongFile));
+
+    // valid name, does not have required YAML contents.
+    String notAppEngineFlexYamlString =
```
nkibler: Break this out into its own test; you should limit your test methods to test a single piece of functionality and once you start asserting, you should not make any more calls to your code under test: http://go/unit-test-practices#dos-and-donts

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetectorTest.java
```diff
@@ -0,0 +1,90 @@
...
+        "spring:\n  application:\n  name: jhipsterSampleApplication";
+    for (String validFileName : AppEngineFlexibleFrameworkDetector.APP_ENGINE_FLEX_PROJECT_FILES) {
```
nkibler: Don't use the names from the valid names list in the code under test. Again, if there were a bug in this list, the test would not help and would blindly pass. Similarly, if this list were empty, the test wouldn't even validate anything.

Instead, use separate methods for trying names you would expect to pass (`app.yaml` and `app.yml`).

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetectorTest.java
```diff
@@ -0,0 +1,90 @@
...
+    AppEngineFlexibleFrameworkDetector detector = new AppEngineFlexibleFrameworkDetector();
+    ElementPattern<FileContent> pattern = detector.createSuitableFilePattern();
+
+    // unaccepted file name pattern, although correct content.
+    MockVirtualFile invalidNameFile =
+        new MockVirtualFile("myApp.yaml", validAppEngineFlexYamlString);
+    FileContent wrongFile =
+        new FileContentImpl(
+            invalidNameFile, validAppEngineFlexYamlString, System.currentTimeMillis());
+    Assert.assertFalse(pattern.accepts(wrongFile));
```
nkibler: It's up to you, but you can consider using the http://go/truth library to write your assertions. The statements almost always start with `assertThat()` and depending on the parameter passed in, the methods available take different forms.

For example, in this case, your assertion would be:

    assertThat(pattern.accepts(wrongFile)).isFalse();

If you choose not to do this, consider adding static imports for your `Assert` methods used here to reduce the code shown here.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetectorTest.java
```diff
@@ -0,0 +1,90 @@
...
+              invalidYamlFile, notAppEngineFlexYamlString, System.currentTimeMillis());
+      Assert.assertFalse(pattern.accepts(invalidYamlFileContent));
+    }
+  }
+
+  @Test
+  public void testValidAppYamlDetection() {
+    // test we require YAML files for detection.
+    AppEngineFlexibleFrameworkDetector detector = new AppEngineFlexibleFrameworkDetector();
+    Assert.assertEquals(YAMLFileType.YML, detector.getFileType());
```
nkibler: This should be its own test.

etanshaul:
@ivanporty - in case you haven't been pointed here yet:

It will help you set up the google style formatter which should take care of most of the annoying formatting things from now on.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
```diff
@@ -66,8 +69,45 @@ public FileType getFileType() {
...
+      extends ObjectPattern<FileContent, AppEngineFlexPattern> {
+
+    private AppEngineFlexPattern() {
+      super(FileContent.class);
+    }
+
+    private AppEngineFlexPattern withAppEngineFlexYamlContent() {
+      return with(
+          new PatternCondition<FileContent>("with-appengine-java-flexible") {
+            // app engine flex file names.
```
etanshaul: looks good.

nit: remove this comment? its pretty much repeating the variable name

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
```diff
@@ -66,8 +69,45 @@ public FileType getFileType() {
...
+
+            @Override
+            public boolean accepts(@NotNull FileContent fileContent, ProcessingContext context) {
+              // checks for flex engine file names and then checks for required configuration line
+              // inside.
+              boolean nameMatch = APP_ENGINE_FLEX_CONFIG_FILES.contains(fileContent.getFileName());
+              if (nameMatch) {
+                Scanner scanner = new Scanner(fileContent.getContentAsText().toString());
+                while (scanner.hasNextLine()) {
+                  if (scanner.nextLine().startsWith(APP_ENGINE_FLEX_REQUIRED_YAML_CONTENT)) {
```
etanshaul: should you trim the line before matching with the required content? I know yaml is indentation sensitive but I'm wondering if its still possible for it to be valid with leading spaces on the "runtime:" line.

ivanporty: Yes, I thought it's supposed to be on the first position only, but if it's not and it can be indented, I will trim the line.

etanshaul: I guess since its a top-level element it shouldn't be trimmed, on second thought

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
ivanporty: agree

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
etanshaul: it doesn't have to be in the first position, but it is a top-level element

ivanporty: Ok, then we leave it as it is, correct? Otherwise we'll be checking on nested element

etanshaul: yea i think its good as is

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/flexible/AppEngineFlexibleFrameworkDetector.java
```diff
@@ -76,32 +69,39 @@ public FileType getFileType() {
...
 
     private AppEngineFlexPattern() {
       super(FileContent.class);
     }
 
-    private AppEngineFlexPattern withAppEngineFlexYAMLContent() {
+    private AppEngineFlexPattern withAppEngineFlexYamlContent() {
       return with(
           new PatternCondition<FileContent>("with-appengine-java-flexible") {
+            private final Collection<String> APP_ENGINE_FLEX_CONFIG_FILES =
```
nkibler: Whoops, sorry for the confusion, I meant to say put these constants in the `AppEngineFlexPattern` class. I've never seen these declared in an anonymous class before and I'm not sure how they will be handled (i.e. are they re-initialized for every instance of this anonymous class?).

ivanporty: Agree, this is probably a bit too deep and plus it makes sense to make it static I think.

etanshaul:
@nkibler are you ok with review now so @ivanporty can merge it in?

nkibler:
Yeah whoops, I thought I approved it this morning but must have missed it. Sorry!

PR #1738, opened 2017-11-09
Adds the UI for the library selector

nkibler:
You can see the UI by un-commenting the added lines to `plugin.xml`, which will add the following menu action item:

screenshot from 2017-11-09 10 11 01

Clicking it brings up the following dialog in its default state:

screenshot from 2017-11-14 15 02 24

Selecting a library from the list on the left shows its details in the right panel:

screenshot from 2017-11-14 15 03 23

Notice how the selectable region is the entire row, which will trigger the row selection and the details pane to update. However, the checkbox is only selected if the user actually selects the box:

screenshot from 2017-11-14 15 04 41


patflynn:
Is there a way to select the library to view the details without checking
the box to add it?

nkibler:
Hmm I guess not, since selecting the checkbox text also selects the checkbox. Should we change this? We could potentially modify the checkboxes so they don't select/unselect when clicking the text.

patflynn:
I think it would be natural to want to click into details of an API without
necessarily selecting it.

Ideally the click target would be the box surrounding the label and
checkbox, and it would be visually differentiated as an indicator that it's
a click-target.

(hope that's doable. :))

nkibler:
PTAL, I've updated the checkboxes to only be selectable via the box itself. See the updated screenshots above.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/ui/GoogleCloudToolsIcons.java
```diff
@@ -48,13 +46,18 @@
   public static final Icon CLOUD_DEBUG_REACTIVATE_BREAKPOINT = load(
       "/icons/debugreactivatebreakpoint.png");
   public static final Icon CLOUD_DEBUG_DELETE_ALL_BREAKPOINTS = load("/icons/debugdeleteall.png");
+  public static final Icon LOADING = loadGif("/icons/loading.gif");
```
etanshaul: does the animation display properly?

etanshaul:
is it possible to add some tests for the UI panels?

etanshaul:
One thing I just noticed:

Go to settings -> plugins

Notice how this panel achieves the behavior we discussed (selecting the field without checking it, etc.). Its worth looking into if its doing it in a more standard way and if we can use the same concept.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/ui/GoogleCloudToolsIcons.java
nkibler: Yep, as long as we use the `ImageIcon` class directly in `loadGif()`. Using the `IconLoader.getIcon()` method in `.load()` doesn't work.

nkibler:
> is it possible to add some tests for the UI panels?

Yeah, I originally was hesitant because it has a dependency on the `CloudLibraries` utility, which can't be mocked out. But I'll pull that out of the panel class and ask the list of libraries to be provided on instantiation, which will easily allow for unit tests.

> Notice how this panel achieves the behavior we discussed (selecting the field without checking it, etc.). Its worth looking into if its doing it in a more standard way and if we can use the same concept.

Hmm I'll take a look and see if we can follow the same pattern.

etanshaul:
Just booted up the branch and the experience feels good.

Nit: the 'version' / 'satus' feels could use some margin / padding. They feel a bit crunched together.

etanshaul:
Also, currently it feels a bit strange to just have an 'ok' action after selecting the libraries. I'm wondering if we should

- Rename the 'ok' button to 'Add selected libraries' or something like that, or
- Have some sort of new dialog pop up after adding ok that contains more info (like showing the selecting libraries and saying that they will be added to the pom etc.

Along the same lines - I'm wondering if the 'ok' action (or whatever it ends up being called) should be disabled until the user checks one or more libs.

I'm fine with this being address in subsequent PRs; just wanted to bring it up.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiDetailsPanel.java
```diff
@@ -0,0 +1,259 @@
...
+
+    linksTextPane = new JTextPane();
+    linksTextPane.addHyperlinkListener(new BrowserOpeningHyperLinkListener());
+  }
+
+  /**
+   * Updates the UI elements in this panel to reflect the attributes in the {@link
+   * #currentCloudLibrary}.
+   */
+  private void updateUI() {
```
etanshaul: this only updates the "client" portion of the UI right? the name, icon etc. is updated by the caller of this. Maybe the name / description can be updated to more clearly indicate this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiDetailsPanel.java
```diff
@@ -0,0 +1,259 @@
...
+   * <p>If the given {@link CloudLibrary} is already displayed, nothing is done and no UI updates
+   * are made.
+   *
+   * @param library the {@link CloudLibrary} to display
+   */
+  void setCloudLibrary(CloudLibrary library) {
+    if (cloudLibrariesEqual(currentCloudLibrary, library)) {
+      return;
+    }
+    currentCloudLibrary = library;
```
etanshaul: library can't realistically be null, right?

nkibler: Right, if it's null, there's a bug somewhere so it should probably blow-up like this will.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiDetailsPanel.java
nkibler: Oh whoops, I forgot to remove the code above (in `.setCloudLibrary()`) when I added this method. The idea is that the setter should only set the `currentCloudLibrary` (if we need to) and call this method without doing any UI-related stuff, like it used to before this method.

PTAL, I've removed all the duplicate code in the calling method.

nkibler:
PTAL, I've added unit tests for the panel.

It required a lot of helper utilities that I've created, but they'll come in handy when we start working more with `CloudLibrary` objects.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -214,6 +214,15 @@ appengine.add.framework.support.choose.module.dialog.description={0} framework s
 appengine.add.framework.support.no.modules.message=All modules already have App Engine support, or no modules were found.
 appengine.add.framework.support.no.modules.title=Cannot add {0} framework support
 
+cloud.libraries.apireference.link=API Reference
+cloud.libraries.dialog.title=Add Cloud Libraries
+cloud.libraries.documentation.link=Documentation
+cloud.libraries.menu.action.text=Add Cloud libraries...
```
etanshaul: Is this all that changed in the UI since the last commit (updated button text)? Maybe update the screenshot?

LGTM otherwise. approving.

nkibler:
PTAL, I've updated to use the `JTable` approach and fixed a few UI-related bugs.

etanshaul:
As I think @paflynn previously suggested, I'm trying to figure out where the generic "no icon" icon comes from for services that don't have a dedicated one. On the web I see this:

image

It would be nice to add these in our UI too.

nkibler:
Yeah if you can get me the asset link, I can add it for those that are missing an icon. I'd rather add it here to this logic instead of the `libraries.json` file in the appengine-plugins-core.

etanshaul:
I found this: https://cloud.google.com/_static/images/cloud/products/logos/svg/launcher.svg

which is close

PR #1747, opened 2017-11-15
Integrate account plugin

### LINKED ISSUE #1735: Integrate Google Account Plugin features directly into Google Cloud Tools Plugin
Google Login was developed as a separate plugin back in the Android Studio days to support multiple Android Studio plugins that used this feature.  

The cost to maintain and the usability wart of having to install two separate plugins makes this separation undesirable for google-cloud-intellij. 

This should be fixed by incorporating the google login features directly into the cloud tools plugin and providing a simple migration path for existing users to upgrade and remove the deprecated plugin.

patflynn:
Fixes #1735

This is pretty much whole. It's real big although most of the changes are mechanical. 

I understand it well enough now to break it down into a feature branch as a set of PRs if that's easier.

Here's a breakdown of how it's done:

- I took the account plugin code and turned it into a regular java module under cloud tools.  Eventually we can pull cloud tools plugin up to the root of the project.
- I included the account plugin xml in cloud tools via optional depends on core idea.  This should make it available on all platforms. I hope.   The reason I did this instead of using <xi:include> is because <depends optional> will find the xml resources on the classpath while xi:include uses the file system. Eventually it would probably be better to package the module xmls into the plugin package so that we can easily xi:include them.
- I renamed all the classes that were exported from the account plugin as services or components so that they don't cause crashes on startup when the user has both the merged account code and the account lpugin installed at the same time.
- Things are still a little whack because there's no way to deregister the login widget from the toolbar (that I could find), so when the user has both plugins installed they will see two login widgets.  I've mitigated this by automatically uninstalling the account plugin at startup if it's present, and strongly prompting the user to restart.

screenshot from 2017-11-16 09 32 30


etanshaul:
:+1: Also some screenshots (I think only of that uninstall dialog) would be helpful too.

google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -295,17 +295,16 @@
...
   <depends>com.intellij.modules.lang</depends>
   <depends>com.intellij.modules.vcs</depends>
   <depends>com.intellij.modules.xml</depends>
   <depends>com.intellij.modules.xdebugger</depends>
   <depends>org.jetbrains.plugins.gradle</depends>
   <depends>org.jetbrains.plugins.yaml</depends>
-  <depends>com.google.gct.login</depends>
   <depends>Git4Idea</depends>
   <depends>org.jetbrains.idea.maven</depends>
+    <!--<xi:include href="/META-INF/google-account.xml" xpointer="xpointer(/idea-plugin/*)"/>-->
```
etanshaul: This was just experimental right? I guess it can be removed now?

patflynn:
PTAL at wording in message bundle. 

The wording in the screenshot felt too long so I changed to:
```
The Google Account plugin is no longer required and \
  has been uninstalled.  Please restart your IDE for it to take effect.  If you cancel you will \
  experience problems using the Google Cloud Tools plugin. \n\nOK to restart?
```

patflynn:
removed.

google-cloud-tools-plugin/google-account/src/com/google/cloud/tools/intellij/analytics/GoogleUsageTracker.java
```diff
@@ -111,7 +108,8 @@
   public GoogleUsageTracker() {
     analyticsId = UsageTrackerManager.getInstance().getAnalyticsProperty();
 
-    AccountPluginInfoService pluginInfo = ServiceManager.getService(AccountPluginInfoService.class);
+    AccountPluginInfoService pluginInfo = ServiceManager
```
nkibler: nit: This line break feels weird; did the auto-formatter do this? I would expect `ServiceManager` to be on the next line with the method call.

google-cloud-tools-plugin/google-account/src/com/google/cloud/tools/intellij/analytics/UsageTrackerConfigurableProvider.java
```diff
@@ -42,11 +41,7 @@ public boolean canCreateConfigurable() {
     // For now we can hide Google entirely if usage tracking isn't available as there are no
     // other Google related account settings in the IJ UI.
     // Create a sub-menu item for the cloud SDK and hide the usage tracker if not avaible
-    if (PlatformUtils.isIntelliJ() && UsageTrackerManager.getInstance()
-        .isUsageTrackingAvailable()) {
-      return true;
-    } else {
-      return false;
-    }
+    return PlatformUtils.isIntelliJ() && UsageTrackerManager.getInstance()
```
nkibler: nit: Ditto about formatting. Since we're touching all of these files it might be a good time to ensure the google-java-formatter is up-to-date and correctly running on these files.

google-cloud-tools-plugin/google-account/src/com/google/cloud/tools/intellij/login/IntegratedGoogleLoginService.java
```diff
@@ -18,17 +18,15 @@
...
 import java.util.Map;
 import java.util.Optional;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * Provides Google user authentication services.
  */
-public interface GoogleLoginService {
+public interface IntegratedGoogleLoginService {
```
nkibler: Will we have plans to name these back to remove the redundant `Integrated` prefix in the future?

google-cloud-tools-plugin/google-account/testSrc/com/google/cloud/tools/intellij/AccountPluginInitializationComponentTest.java
```diff
@@ -58,26 +58,10 @@
...
     registerService(AccountPluginInfoService.class, pluginInfoService);
     registerService(AccountPluginConfigurationService.class, pluginConfigurationService);
-    registerService(GoogleLoginService.class, googleLoginService);
+    registerService(IntegratedGoogleLoginService.class, googleLoginService);
     testComponent = new AccountPluginInitializationComponent();
   }
 
-  // TODO: factor out commonality in tests and application components for feedback initialization
-  @Test
-  public void testInitComponent_feedbackIsEnabled() {
```
nkibler: Any reason for removing these tests?

google-cloud-tools-plugin/google-account/testSrc/com/google/cloud/tools/intellij/login/integration/PluginXmlTest.java
```diff
@@ -32,7 +31,7 @@
     public void testWellFormed() throws SAXException, IOException {
         XMLReader reader = XMLReaderFactory.createXMLReader();
```
nkibler: nit: Since the line you touched was correctly updated to +2 spaces for indentation (instead of +4), can you fix the rest of the method here for consistency?

google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -14,7 +14,7 @@
   ~ limitations under the License.
   -->
 
-<idea-plugin version="2">
+<idea-plugin>
```
nkibler: Is the `version` attribute no longer needed? I remember seeing warnings during Gradle builds from the IDEA plugin saying that because no `version` attribute is specified, something wouldn't happen.

google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -449,7 +446,7 @@
     <GitHttpAuthDataProvider implementation="com.google.cloud.tools.intellij.vcs.GcpHttpAuthDataProvider" />
   </extensions>
 
-  <extensions defaultExtensionNs="com.google.gct.login">
+    <extensions defaultExtensionNs="com.google.gct.core">
```
nkibler: Revert this extra indentation?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/GoogleAccountPluginUninstaller.java
```diff
@@ -0,0 +1,80 @@
...
+import com.intellij.openapi.extensions.PluginId;
+import com.intellij.openapi.ui.MessageDialogBuilder;
+
+/**
+ * Google Login used to be implemented as its own plugin that was distributed separately, but
+ * depended on by the Cloud Tools plugin.  We have now incorporated login directly into the
+ * Cloud Tools plugin and this class checks whether the legacy Google Account plugin is still
+ * installed and removes it, prompting the user to restart.
+ *
+ * Keeping the old Google Account plugin installed causes problems such as duplicating the login
```
nkibler: nit: Use a `<p>` tag at the beginning of this statement so it's in a separate paragraph.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanelTest.java
```diff
@@ -67,7 +67,9 @@
 
   private GcsBucketContentEditorPanel editorPanel;
 
-  @TestService @Mock private GoogleLoginService loginService;
+  @TestService
+  @Mock
+  private IntegratedGoogleLoginService loginService;
```
nkibler: nit: Can you keep these all on one line? If not, you can move this into its own block below so the single-annotation fields are together for easier readability.

google-cloud-tools-plugin/google-account/src/com/google/cloud/tools/intellij/analytics/GoogleUsageTracker.java
patflynn: fixed.

google-cloud-tools-plugin/google-account/src/com/google/cloud/tools/intellij/analytics/UsageTrackerConfigurableProvider.java
patflynn: Should be.  I'm running the gradle rule.  Although it doesn't appear to be doing much so maybe it is misconfigured.

google-cloud-tools-plugin/google-account/src/com/google/cloud/tools/intellij/login/IntegratedGoogleLoginService.java
patflynn: no current plans but open to it once our users are sufficiently migrated over.

google-cloud-tools-plugin/google-account/testSrc/com/google/cloud/tools/intellij/login/integration/PluginXmlTest.java
patflynn: done

google-cloud-tools-plugin/resources/META-INF/plugin.xml
patflynn: hmm. I haven't seen that warning I'll take a look

google-cloud-tools-plugin/resources/META-INF/plugin.xml
patflynn: fixed

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanelTest.java
patflynn: fixed

google-cloud-tools-plugin/google-account/resources/META-INF/google-account.xml
```diff
@@ -0,0 +1,77 @@
...
+        com.google.cloud.tools.intellij.startup.AccountPluginInitializationComponent
+      </implementation-class>
+    </component>
+  </application-components>
+
+  <project-components>
+  </project-components>
+
+  <actions>
+    <action id="GoogleLogin.LoginService"
```
etanshaul: I would have thought using the same ID here as in the old account plugin would have causes exceptions on startup if the user chooses not to immediately restart (since the login widget isn't deregistered).

patflynn: yeah funnily enough it doesn't.  I didn't change it because for some reason the id had an effect on layout.  I may be crazy but JB may have whitelisted us for that right aligned spot on the toolbar.

google-cloud-tools-plugin/google-account/.gitignore
```diff
@@ -1,2 +1,3 @@
 /out
+/build
```
etanshaul: I dont' think we need this file now since the module now lives in the cloud tools plugin

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/GoogleAccountPluginUninstaller.java
nkibler: Oops sorry I didn't specify, but it should look like this:

    /**
     * <p>Keeping the old...
     * ...wrap like normal...
     */

google-cloud-tools-plugin/google-account/resources/META-INF/google-account.xml
etanshaul: i tested this out, and changing the ID had no effect. 

But if there are no exceptions then we can leave it. Presumably if the user manually set keyboard shortcuts on the old login action, they would lose the shortcut if we rename the ID (not that I think anyone actually did this).

etanshaul:
Just thinking out loud about our migration strategy given the current state of this:

Option A - Delete the account plugin from the repo alongside update to the cloud tools plugin
- Users will just be notified using the mechanism in place here - uninstalling the account plugin
- Question: does IntelliJ do any intelligent detection of plugins removed from their repo? e.g. warn the user that their installed plugin has been deleted with some action to remove it? I highly doubt this happens, but even if so it shouldn't be a problem because the Account Plugin uninstall check added here would run before.

Option B - Update the account plugin with an empty shell alongside update to cloud tools plugin
- Nice because after updating the plugins there is no longer any conflict
  - we could update the uninstall check here to just ignore the gutted account plugin - no need to warn the user if its not causing any conflicts
- New installers of the plugin are ok still since the old account plugin is no longer a dependency and would never be installed

I think B is the better approach since we can potentially have a seamless migration without ever even showing the new uninstall warning dialog for existing users.

patflynn:
We can definitely do option B but it doesn't preclude the need for the
check.  we have seen users fail to update both plugins at the same time
which is why we had the original compatibilty check in there.

etanshaul:
absolutely - we still need the check.

I'm just saying that we can update the check to only run if the account plugin has not been gutted.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/GoogleAccountPluginUninstaller.java
```diff
@@ -0,0 +1,81 @@
...
+          .yesNo(GctBundle.message("account.plugin.removal.requires.restart.title"),
+              GctBundle.message("account.plugin.removal.requires.restart.text"))
+          .yesText(GctBundle.message("OK"))
+          .noText(GctBundle.message("Cancel"))
+          .isYes()) {
+        UsageTrackerProvider.getInstance()
+            .trackEvent(GctTracking.ACCOUNT_PLUGIN_RESTART_DIALOG_YES_ACTION)
+            .ping();
+        ApplicationManagerEx.getApplicationEx().restart(true);
+        return; // presumably we never get here, but just to make sure we never ping no and yes.
```
etanshaul: nit, but cant you just do an `else {` here and then you don't really need the comment?

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/gcs/GcsBucketContentEditorPanelTest.java
etanshaul: is this fixed? still looks like the @Mock is on its own line

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/gcs/GcsBucketPanelTest.java
```diff
@@ -38,7 +38,9 @@
   @Rule public final CloudToolsRule cloudToolsRule = new CloudToolsRule(this);
   @TestFixture private IdeaProjectTestFixture testFixture;
 
-  @Mock @TestService private GoogleLoginService loginService;
+  @Mock
```
etanshaul: nit; same comment about annotations on the same line

patflynn:
argh. I fixed it and then intellij reformatted it again. I have the ij
google java formatter configured so not sure what's up. will look into it.

PR #1748, opened 2017-11-16
Converts the client library dialog into a wizard

etanshaul:
Some plumbing to turn the library flow into a wizard:

Step 1:
image

Step 2 (placeholder for upcoming API management panel):
image

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
@@ -21,24 +21,29 @@ import com.google.cloud.tools.intellij.util.GctBundle;
 import com.google.cloud.tools.libraries.CloudLibraries;
 import com.google.cloud.tools.libraries.json.CloudLibrary;
 import com.google.common.collect.ImmutableList;
+import com.intellij.ide.wizard.AbstractWizard;
+import com.intellij.ide.wizard.Step;
+import com.intellij.ide.wizard.StepAdapter;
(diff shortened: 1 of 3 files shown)


etanshaul:
looking into if/where unit tests make sense

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
```diff
@@ -48,23 +53,34 @@ public AddCloudLibrariesAction() {
...
 
     @Nullable
     @Override
-    protected JComponent createCenterPanel() {
+    protected String getHelpID() {
+      return null;
+    }
+  }
+
+  /** The wizard step encapsulating the client library selection panel */
```
nkibler: nit: period

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
```diff
@@ -58,6 +59,7 @@
 
   GoogleCloudApiSelectorPanel(List<CloudLibrary> libraries) {
     this.libraries = libraries;
+    panel.setPreferredSize(new Dimension(800, 600));
```
nkibler: I just realized: can this be part of the .form?

etanshaul: I wish. But for some unknown reason it renders differently when added to the form! (its wider, and the text doesn't wrap when resizing). No clue why yet.

PR #1750, opened 2017-11-17
Adds the module selector to the Cloud Library selector

nkibler:
This adds the module selector to the top of the library selector step:

module-selector-2

The dropdown defaults to the first, top-most module in the project, as defined by `Project.getSortedModules()`. If the user has no module in their project (which means they have no source and their project is likely misconfigured), they will see:

module-selector-1

... where the "Next" button is disabled and the module selector displays "[none]". If we really want, we can add more helpful text around this in the future, but it's a rare case that the user has absolutely nothing in their project but wants to use our dialog that I don't think it's worth investing more time into right now.

This PR also exposes some methods from the panels to get the selected module and the selected libraries; these will be used in a future PR to add the selected libraries to the selected module.


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
```diff
@@ -61,13 +64,31 @@ public void actionPerformed(AnActionEvent e) {
...
     }
 
+    /** Returns the set of selected {@link CloudLibrary CloudLibraries}. */
+    Set<CloudLibrary> getSelectedLibraries() {
+      return selectClientLibrariesStep.getSelectedLibraries();
+    }
+
+    @Override
+    protected boolean canGoNext() {
+      return selectClientLibrariesStep.getSelectedModule() != null;
```
etanshaul: what are you thoughts on adding warning messages in this PR? I'm ok with a separate PR if you think its best.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
```diff
@@ -61,13 +64,31 @@ public void actionPerformed(AnActionEvent e) {
...
+    private final SelectClientLibrariesStep selectClientLibrariesStep;
+    private final ManageCloudApisStep manageCloudApisStep;
+
     AddCloudLibrariesWizard(Project project) {
       super(GctBundle.message("cloud.libraries.dialog.title"), project);
-      addStep(new SelectClientLibrariesStep());
-      addStep(new ManageCloudApisStep());
+
+      selectClientLibrariesStep = new SelectClientLibrariesStep(project);
+      selectClientLibrariesStep.addModuleSelectionListener(event -> updateButtons());
```
etanshaul: why not add this listener directly in the `GoogleCloudApiSelectorPanel`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
```diff
@@ -89,7 +110,22 @@ protected String getHelpID() {
...
+      cloudApiSelectorPanel.addModuleSelectionListener(listener);
+    }
+
+    /** Returns the selected {@link Module}. */
+    Module getSelectedModule() {
+      return cloudApiSelectorPanel.getSelectedModule();
+    }
+
+    /** Returns the set of selected {@link CloudLibrary CloudLibraries}. */
+    Set<CloudLibrary> getSelectedLibraries() {
```
etanshaul: Is this idea that this wizard itself will be passed into subsequent steps?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
nkibler: What do you mean? The call to `updateButtons()` is specific to this `AbstractWizard<Step>`, which the `GoogleCloudApiSelectorPanel` has no knowledge of. IMO exposing a way to add a custom listener for when a module is selected (like this) prevents the panel from knowing too much about the context in which it's created.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
nkibler: @patflynn and I talked a bit about this this morning; we could either warn/error the user somehow to let them know they have to select a module, or we could default to the "top-most" module, which is rather easy. That way, a module would always be selected and we could remove this disabling logic.

WDYT?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
nkibler: My thoughts are that the surrounding wizard class (`AddCloudLibrariesWizard`) is responsible for the interplay between steps. If a step requires items from a previous step, for example, it's the wizard's job to pass those along because it understands the context in which these steps are created.

So in this case, if the `ManageCloudApisStep` requires the list of selected libraries from this step, the wizard will pull the selected libraries out of here and pass them to the next step.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
etanshaul: oops. I didn't realize that. sgtm

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
etanshaul: defaulting to the top module sgtm. this is really only something we could know for sure from a user study. but having a default should cover most cases imo

nkibler: Done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
```diff
@@ -124,6 +120,16 @@ private void createUIComponents() {
     modulesComboBox = new ModulesComboBox();
     modulesComboBox.fillModules(project);
 
+    ApplicationManager.getApplication()
+        .runReadAction(
+            () -> {
+              Module[] modules = ModuleManager.getInstance(project).getSortedModules();
+              if (modules.length > 0) {
```
etanshaul: what happens if there are no modules?

etanshaul: we should probably show some kind of message to the user in this case

nkibler: Can you have a project with no modules? I didn't think that was possible.

etanshaul: ah yea - you can manually delete all modules; but you don't have a meaningful project at that point (no source etc.). so i guess its fine to just show an empty dropdown

etanshaul: I'd make sure the "next" button is disabled though in this case; since I think you got rid of that override

nkibler: Done.

PR #1751, opened 2017-11-21
upgrade gradle version and bump error prone to keep it compatible.

patflynn:


loosebazooka:
don't want want to run `./gradlew tasks` or something to actually update the jar of the wrapper?

patflynn:
@loosebazooka I did not know about that!  Thanks PR pending

PR #1756, opened 2017-11-21
fix windows line endings

patflynn:
in gradle.bat


nkibler:
Can this be merged? You'll have to force it since the build is breaking because of the line-endings on this file.

patflynn:
I'm just going to close this until I figure what's going in with our Windows CI environment

PR #1759, opened 2017-11-27
Adds the Cloud library dependency writer

nkibler:
This PR includes support for adding dependencies of Cloud libraries to Maven modules via the `pom.xml`. After selecting some libraries and pressing the "Finish" button, the user is notified which libraries were successfully added and which ones were ignored:

screenshot from 2017-11-30 15-05-02

screenshot from 2017-11-30 15-05-31

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
@@ -34,7 +34,7 @@ import org.jetbrains.idea.maven.model.MavenId;
 import org.jetbrains.idea.maven.project.MavenProject;
 import org.jetbrains.idea.maven.project.MavenProjectsManager;
 
-/** A helper class that writes dependencies on Cloud libraries to a given module. */
+/** A helper class that adds dependencies of Cloud libraries to a given module. */
 final class CloudLibraryDependencyWriter {
 
   /** Prevents instantiation. */
@@ -50,7 +50,11 @@ final class CloudLibraryDependencyWriter {
    * @param libraries the set of {@link CloudLibrary CloudLibraries} to add
    * @param module the {@link Module} to add the libraries to
    */
-  static void addLibraries(Set<CloudLibrary> libraries, Module module) {
+  static void addLibraries(@NotNull Set<CloudLibrary> libraries, @NotNull Module module) {
+    if (libraries.isEmpty()) {
+      return;
+    }
+
     Project project = module.getProject();
     MavenProjectsManager projectsManager = MavenProjectsManager.getInstance(module.getProject());
     MavenProject mavenProject = projectsManager.findProject(module);
@@ -74,12 +78,13 @@ final class CloudLibraryDependencyWriter {
         // dependencies, maybe a warning?
         libraries
             .stream()
-            .map(CloudLibraryUtils::getJavaClientMavenCoordinates)
+            .map(CloudLibraryUtils::getFirstJavaClientMavenCoordinates)
             .filter(Optional::isPresent)
             .map(Optional::get)
             .map(CloudLibraryDependencyWriter::toMavenId)
             .filter(mavenId -> isMavenIdNotInDependencyList(mavenId, dependencies))
-            .forEach(mavenId -> MavenDomUtil.createDomDependency(model, null, mavenId));
+            .forEach(
+                mavenId -> MavenDomUtil.createDomDependency(model, /* editor= */ null, mavenId));
       }
     }.execute();
   }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
@@ -16,16 +16,25 @@
 
 package com.google.cloud.tools.intellij.apis;
 
+import com.google.cloud.tools.intellij.flags.PropertiesFileFlagReader;
+import com.google.cloud.tools.intellij.ui.GoogleCloudToolsIcons;
+import com.google.cloud.tools.intellij.util.GctBundle;
 import com.google.cloud.tools.libraries.json.CloudLibrary;
 import com.google.cloud.tools.libraries.json.CloudLibraryClientMavenCoordinates;
+import com.intellij.notification.Notification;
+import com.intellij.notification.NotificationDisplayType;
+import com.intellij.notification.NotificationGroup;
+import com.intellij.notification.NotificationType;
 import com.intellij.openapi.application.Result;
 import com.intellij.openapi.command.WriteCommandAction;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.project.Project;
 import com.intellij.util.xml.DomUtil;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.idea.maven.dom.MavenDomUtil;
 import org.jetbrains.idea.maven.dom.model.MavenDomDependency;
@@ -36,6 +45,14 @@ import org.jetbrains.idea.maven.project.MavenProjectsManager;
 
 /** A helper class that adds dependencies of Cloud libraries to a given module. */
 final class CloudLibraryDependencyWriter {
+
+  private static final NotificationGroup NOTIFICATION_GROUP =
+      new NotificationGroup(
+          new PropertiesFileFlagReader().getFlagString("notifications.plugin.groupdisplayid"),
+          NotificationDisplayType.BALLOON,
+          true,
+          null,
+          GoogleCloudToolsIcons.CLOUD);
 
   /** Prevents instantiation. */
   private CloudLibraryDependencyWriter() {}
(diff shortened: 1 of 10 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
```diff
@@ -0,0 +1,123 @@
...
+   * <p>For a module whose project manages its dependencies through Maven, the list of libraries
+   * will be added as dependencies to the {@code pom.xml}. For all other dependency management
+   * systems, the libraries will be downloaded and added directly to the module's classpath.
+   *
+   * @param libraries the set of {@link CloudLibrary CloudLibraries} to add
+   * @param module the {@link Module} to add the libraries to
+   */
+  static void addLibraries(Set<CloudLibrary> libraries, Module module) {
+    Project project = module.getProject();
+    MavenProjectsManager projectsManager = MavenProjectsManager.getInstance(module.getProject());
```
etanshaul: We have helpers for this in AppEngineProjectService#isMavenModule

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
```diff
@@ -0,0 +1,123 @@
...
+    MavenProject mavenProject = projectsManager.findProject(module);
+    if (mavenProject == null) {
+      // TODO(nkibler): Handle non-Maven projects.
+      return;
+    }
+
+    MavenDomProjectModel model =
+        MavenDomUtil.getMavenDomProjectModel(project, mavenProject.getFile());
+    if (model == null) {
+      // TODO(nkibler): Handle this error-state, maybe a warning?
```
etanshaul: I'm surprised this can be null. Basically, its determined to be a maven project but there is no pom.xml? I think a warning should be logged and some notice to the user (perhaps an error in the event log).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
etanshaul: Although this shouldnt be App Engine specific, so feel free to move / refactor this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
```diff
@@ -0,0 +1,123 @@
...
+   *
+   * <p>For a module whose project manages its dependencies through Maven, the list of libraries
+   * will be added as dependencies to the {@code pom.xml}. For all other dependency management
+   * systems, the libraries will be downloaded and added directly to the module's classpath.
+   *
+   * @param libraries the set of {@link CloudLibrary CloudLibraries} to add
+   * @param module the {@link Module} to add the libraries to
+   */
+  static void addLibraries(Set<CloudLibrary> libraries, Module module) {
+    Project project = module.getProject();
```
nbashirbello: Add check for an empty set?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
```diff
@@ -0,0 +1,123 @@
...
+   * Adds the given set of {@link CloudLibrary CloudLibraries} to the given {@link Module}.
+   *
+   * <p>For a module whose project manages its dependencies through Maven, the list of libraries
+   * will be added as dependencies to the {@code pom.xml}. For all other dependency management
+   * systems, the libraries will be downloaded and added directly to the module's classpath.
+   *
+   * @param libraries the set of {@link CloudLibrary CloudLibraries} to add
+   * @param module the {@link Module} to add the libraries to
+   */
+  static void addLibraries(Set<CloudLibrary> libraries, Module module) {
```
nbashirbello: @NotNull module?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
```diff
@@ -0,0 +1,123 @@
...
+        // TODO(nkibler): Inform the user when libraries were ignored because of existing
+        // dependencies, maybe a warning?
+        libraries
+            .stream()
+            .map(CloudLibraryUtils::getJavaClientMavenCoordinates)
+            .filter(Optional::isPresent)
+            .map(Optional::get)
+            .map(CloudLibraryDependencyWriter::toMavenId)
+            .filter(mavenId -> isMavenIdNotInDependencyList(mavenId, dependencies))
+            .forEach(mavenId -> MavenDomUtil.createDomDependency(model, null, mavenId));
```
etanshaul: nit: named (comment) param ( `/* editor */`)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
etanshaul: I think the user should get some sort of confirmation that the library was written; a simple write to the event log could work.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
nkibler: `MavenProjectsManager#isMavenizedModule()` exists too, but the existence of a `MavenProject` for the given `Module` means it's "Mavenized" and since we actually need to use the `MavenProject` later on, it's faster to null-check it here.

But I can change this to use one of the helpers if you feel it's more explanatory.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
nkibler: Done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
nkibler: Done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
nkibler: It might not be possible under normal circumstances. I can either not do this check and let it NPE or do what you said and error in the event log. WDYT?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
nkibler: Done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.Set;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.idea.maven.dom.MavenDomUtil;
+import org.jetbrains.idea.maven.dom.model.MavenDomDependency;
+import org.jetbrains.idea.maven.dom.model.MavenDomProjectModel;
+import org.jetbrains.idea.maven.model.MavenId;
+import org.jetbrains.idea.maven.project.MavenProject;
+import org.jetbrains.idea.maven.project.MavenProjectsManager;
+
+/** A helper class that writes dependencies on Cloud libraries to a given module. */
```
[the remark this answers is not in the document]
nkibler: Done, not sure what I was thinking with this one haha.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryUtils.java
```diff
@@ -0,0 +1,63 @@
...
+  /** Prevents instantiation. */
+  private CloudLibraryUtils() {}
+
+  /**
+   * Returns the first Java {@link CloudLibraryClient} in the given {@link CloudLibrary}, or {@link
+   * Optional#empty()} if none exists.
+   *
+   * @param library the {@link CloudLibrary} to return the first Java client for
+   */
+  static Optional<CloudLibraryClient> getJavaClient(CloudLibrary library) {
```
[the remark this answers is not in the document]
nkibler: Done.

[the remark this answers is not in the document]
nkibler: The JSON spec allows for a list of clients for each API. There's nothing preventing two clients from being in this list for the same language, even though it doesn't really make sense to have multiple. So this method is explicit about choosing the first Java client just in case.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
etanshaul: Lets let it explode so that we get notified since like you said this should not happen under normal circumstances.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
etanshaul: are you saying that `MavenProjectsManager#isMavenizedModule()` is not needed? I quickly looked through it and it looks like its doing some extra work (like checking the `ExternalSystemModulePropertyManager`).

etanshaul: on second look, it seems that `ExternalSystemModulePropertyManager` internally calling a now deprecated method. So I'm ok with what you had if its sufficient to determine that a module is "mavenized"

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
etanshaul: did you address this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryUtils.java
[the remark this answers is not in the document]
etanshaul: I'm find with choosing the first like you did here. Eventually, if it proves useful, we can list all of the clients and give the user the option to choose the desired one.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
```diff
@@ -0,0 +1,128 @@
...
+   *
+   * <p>Note that equality is tested via matching group IDs and artifact IDs. Equality of versions
+   * is not required. This prevents adding duplicate dependencies for the same artifact, but with a
+   * different version.
+   *
+   * @param mavenId the {@link MavenId} to check for existence in the given list of dependencies
+   * @param dependencies the list of {@link MavenDomDependency} objects that currently exist in the
+   *     DOM model
+   */
+  private static boolean isMavenIdNotInDependencyList(
```
etanshaul: ditto with informing the user if this is the case so the understand what the dependency wasn't added

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
nkibler: Done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
nkibler: PTAL, I've added a call to `MavenProjectsManager#isMavenizedModule()` first, mostly to generalize the `addLibraries()` method and delegate build-system dependent stuff from there.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
nkibler: Done.

nkibler:
PTAL, I've added notifications for added and ignored libraries. Take a look at the top post to see screenshots of what they look like.

PR #1763, opened 2017-11-30
Updates the annotation processor plugin to fix AutoValue resolutions

nkibler:
Without this, I've noticed that the `AutoValue` annotation isn't added to the classpath of the project properly, specifically within the `common-test-lib` module. This updates to the newest version of the annotation processor Gradle plugin and uses the new recommended plugin, which seems to resolve the issue.


etanshaul:
How was it working before? Did it break due to the account plugin refactoring?

nkibler:
I mean it's always "worked" in that the build works. But the class path
wasn't being configured correctly for IntelliJ to resolve the symbol. Since
we delegate builds in IntelliJ to Gradle, it doesn't break the build but
the symbol is red.

My second commit fixes the build that should have been broken from Ivan's
commit (since he's using AutoValue from the plugin module for the first
time) but I'm not sure why presubmit didn't catch it.

PR #1764, opened 2017-11-30
User project list loader service for project selector. 

ivanporty:
Based on GoogleUserModelItem, refactored, cleaned up, with increased unit test coverage.


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectLoader.java
```diff
@@ -0,0 +1,117 @@
...
+
+        response
+            .getProjects()
+            .stream()
+            // Filter out any projects that are scheduled for deletion.
+            .filter((project) -> !PROJECT_DELETE_REQUESTED.equals(project.getLifecycleState()))
+            // Add remaining projects to the set.
+            .forEach(allProjects::add);
+
+        while (!Strings.isNullOrEmpty(response.getNextPageToken())) {
```
nbashirbello: Can response.getNextPageToken() ever be empty?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectLoader.java
```diff
@@ -0,0 +1,117 @@
...
+          response =
+              cloudResourceManagerClient
+                  .projects()
+                  .list()
+                  .setPageToken(response.getNextPageToken())
+                  .setPageSize(PROJECTS_MAX_PAGE_SIZE)
+                  .execute();
+          allProjects.addAll(response.getProjects());
```
nbashirbello: Do we need to check for projects that are scheduled for deletion?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectLoader.java
ivanporty: Good question, there are no specific documentation on empty token in that API. It definitely returns null for no token, but what happens on empty string token is not clear. Should we assume empty value is a possible token as well? This is completely up to that API, does anyone knows?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectLoader.java
ivanporty: Good catch, we need on consequent page results as well. I refactored to filter deleted and no project ID projects in one pass.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectLoader.java
```diff
@@ -0,0 +1,113 @@
...
+
+/** Loads list of {@link Project} for a {@link CredentialedUser}. */
+class ProjectLoader {
+  private static final Logger LOG = Logger.getInstance(ProjectLoader.class);
+
+  private static final int PROJECTS_MAX_PAGE_SIZE = 300;
+  private static final String PROJECT_DELETE_REQUESTED = "DELETE_REQUESTED";
+
+  @SuppressWarnings("FutureReturnValueIgnored")
+  void loadUserProjectsInBackground(
```
etanshaul: Do you think the callback would be better represented using a future? In fact, ThreadUtil#executeInBackground does return a future.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/project/ProjectLoaderTest.java
```diff
@@ -0,0 +1,221 @@
...
+
+    projectLoader.loadUserProjectsInBackground(mockUser, mockResultCallback);
+
+    verify(mockResultCallback)
+        .onError(GctBundle.getString("project.selector.loader.couldnotconnect"));
+  }
+
+  @Test
+  public void runtimeException_resultsIn_exceptionMessage() throws IOException {
+    // remove IDEA logging which invokes AssertionError.
```
etanshaul: is this true for unit tests? I thought a different logger was implementation was loaded for test

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectLoader.java
ivanporty: I think standard Future is too simplistic for us - it does not allow notification when task is done, and we can't block UI. Do you mean Guava listenable future like https://github.com/google/guava/wiki/ListenableFutureExplained? From initial view it would work. Only question is will it be a bit harder to read code that deal with generic future rather than specific callback interface, but it's probably minor.
If you think it makes sense, I can convert it.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/project/ProjectLoaderTest.java
ivanporty: I checked again and it does stop the unit test when IntelliJ Logger.error() is called preventing the following logic to execute plus marking test as failed as a result of an error.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectLoader.java
nkibler: Technically you can add a listener to the returned `Future`, but it's not as great as `ListenableFuture`, hence why it was created in the first place.

I support changing the `ThreadUtil` class to use a listening decorator on its cached thread pool. The only reason I didn't do this is because up until now, we had no need for the returned future. Now that we do, it makes sense to provide a better interface to it via `ListenableFuture`.

I can send a PR if you want, or you can do it in this PR. It's up to you.

etanshaul: Just to point out. JDK 8 has the `CompletableFuture` which to my knowledge can accomplish the same thing. @nkibler any idea which is currently preferred internally?

nkibler: Seems like `ListenableFuture` is still the preferred type: http://go/java8walkthrough/redundancy#listenablefuture-vs-completablefuture

etanshaul: ah go/java8walkthrough/redundancy. pretty ambiguous, but seems to be a nod towards `ListenableFuture`

ivanporty: It should be a couple of line change in ThreadUtil, I think it makes sense to add it in this PR.

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/DelayedSubmitExecutorServiceProxy.java
```diff
@@ -68,8 +68,11 @@ public boolean awaitTermination(long timeout, @NotNull TimeUnit unit)
   }
 
   @Override
-  public void execute(@NotNull Runnable command) {
-    executor.execute(command);
+  public void execute(@NotNull Runnable task) {
+    if (this.task != null) {
```
etanshaul: can you explain the changes here? do the changes you made to use a ListeningExecutorService require this?

ivanporty: After the change GcsBucketContentEditorPanelTest started to fail - looks like MoreExecutors.listeningDecorator() calls execute() in that test so this essentially disabled effect of delayed executor. I made execute() to delay submission as well and it passes now.

PR #1766, opened 2017-12-01
Specifically set gradlew.bat to CRLF endings in gitattributes

nkibler:
This should fix it (I think). It undoes my last PR by using CRLF endings in `gradlew.bat` but forces it to stay that way in the `.gitattributes` file.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -23,4 +23,4 @@ CONTRIBUTORS text
 *.jpg binary
 *.jar binary
 
-gradlew.bat=crlf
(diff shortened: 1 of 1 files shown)


etanshaul:
looks like the windows build failed for some other transient reason.

nkibler:
Not sure how to fix this long-term, but I don't think this is the right way. At least master prevents the constant diffs, it's just that when we go to regenerate the wrapper, it might try to convert the line-endings back for now.

PR #1768, opened 2017-12-04
Update CHANGELOG.md

parkjam4:
I don't think you need 2 separate descriptions of the same change. Seems redundant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,15 +3,14 @@ This page documents production updates to Cloud Tools for IntelliJ. You can chec
 
 ## 17.12.1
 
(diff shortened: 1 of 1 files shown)


etanshaul:
fair enough - I did it twice just so it stands out more prominently. But its ok the way you have since you reordered the sections.

Don't forget to update plugin.xml as well to match this change please

etanshaul:
on second thought. Im leaning towards leaving it the way it was. I don't really mind the redundancy - I just want it to stand out. wdyt?

parkjam4:
Ok, I think this change is big enough to warrant extra attention. I'm ok to keep the text.

Issue #1770, opened 2017-12-04
Investigate test flakiness on Ubunu


etanshaul:
One instance (there have been others):
https://sponge.corp.google.com/target?id=c67fd7e4-44e0-4f20-8bf6-2dc518a42c28&target=google-cloud-intellij%2Fgcp_ubuntu%2Fpresubmit&searchFor=

com.google.cloud.tools.intellij.apis.GoogleCloudApiSelectorPanelTest > getSelectedLibraries_withOneSelected_returnsLibrary FAILED
    java.lang.IllegalArgumentException: Missing extension point: com.intellij.appStarter in area null

com.google.cloud.tools.intellij.apis.GoogleCloudApiSelectorPanelTest > getPanel_withOneLibrary_selectedLibrary_hasCheckboxAndPopulatedDetails FAILED
    java.lang.ExceptionInInitializerError

Caused by:
        java.lang.IllegalStateException: @NotNull method com/intellij/openapi/vfs/VirtualFileManager.getInstance must not return null

com.google.cloud.tools.intellij.apis.GoogleCloudApiSelectorPanelTest > getSelectedLibraries_withNoneSelected_returnsEmptySet FAILED
    java.lang.NoClassDefFoundError: Could not initialize class com.intellij.openapi.vfs.LocalFileSystem$LocalFileSystemHolder

com.google.cloud.tools.intellij.apis.GoogleCloudApiSelectorPanelTest > getPanel_withMultipleLibraries_noSelection_hasCheckboxesAndEmptyDetails FAILED
    java.lang.NoClassDefFoundError: Could not initialize class com.intellij.openapi.vfs.LocalFileSystem$LocalFileSystemHolder

com.google.cloud.tools.intellij.apis.GoogleCloudApiSelectorPanelTest > getSelectedModule_withNoneSelected_returnsDefaultModule FAILED
    java.lang.NoClassDefFoundError: Could not initialize class com.intellij.openapi.vfs.LocalFileSystem$LocalFileSystemHolder

com.google.cloud.tools.intellij.apis.GoogleCloudApiSelectorPanelTest > getPanel_withMultipleLibraries_multipleSelected_hasCheckboxesAndPopulatedDetails FAILED
    java.lang.NoClassDefFoundError: Could not initialize class com.intellij.openapi.vfs.LocalFileSystem$LocalFileSystemHolder

[...]

PR #1772, opened 2017-12-11
Fix 'File accessed outside allowed roots' CI error

etanshaul:
partial fix for #1770 

Fixes "java.lang.AssertionError: File accessed outside allowed roots:" errors.

See https://youtrack.jetbrains.com/issue/IDEA-129297

Update `DefaultAppEngineProjectServiceTest` to use the `CloudToolsRule`, and added the root fix to `AppEngineCodeInsightTestCase` (eventually this test could be updated to use the CloudToolsRule, but currently its deeply tied into the platforms test hierarchy so I left it as is).

### DIFF
--- a/google-cloud-tools-plugin/testSrc/com/intellij/appengine/AppEngineCodeInsightTestCase.java
+++ b/google-cloud-tools-plugin/testSrc/com/intellij/appengine/AppEngineCodeInsightTestCase.java
@@ -31,6 +31,7 @@ import com.intellij.openapi.vfs.LocalFileSystem;
 import com.intellij.openapi.vfs.VfsUtil;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.openapi.vfs.VirtualFileFilter;
+import com.intellij.openapi.vfs.newvfs.impl.VfsRootAccess;
 import com.intellij.testFramework.UsefulTestCase;
 import com.intellij.testFramework.builders.JavaModuleFixtureBuilder;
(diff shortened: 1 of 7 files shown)


etanshaul:
Also, somehow, this seems to consistently fix the CI overall.

google-cloud-tools-plugin/testSrc/com/intellij/appengine/AppEngineCodeInsightTestCase.java
```diff
@@ -61,6 +62,8 @@
   @Override
   protected void setUp() throws Exception {
     super.setUp();
+    // Fixes https://youtrack.jetbrains.com/issue/IDEA-129297. Only occurs in Jenkins.
+    VfsRootAccess.allowRootAccess(System.getProperty("user.dir"));
```
patflynn: are the invokeAndWait changes and this change related?

etanshaul: There are 2 things going on here:

1) `AppEngineInsightTestCase` was failing with that root access error - adding this `allowRootAccess` fixes the problem
2) I refactored `DefaultAppEngineProjectServiceTest` to use the new CloudToolsRule. This allowed me to 1) get rid of the `allowRootAccess` fix, and 2) forced me to wrap the read actions in an invokeAndWait so they are run on the EDT.

The reasons behind #2 are still unclear to me. I'm still investigating. I think I may roll back part 2 just to keep this PR concise.

PR #1777, opened 2017-12-11
Project selection dialog

ivanporty:
Final part of project selection change - dialog.
This looks like a large PR but it's actually not - at least 400 LOC are form XML/copyright headers/imports/visible for testing lines, and the rest cannot be split logically and should be straightforward to review.

You can review FilteredTextTableCellRenderer.java and ProjectListTableModelTest.java and their tests first - those are very small and independent, and continue with ProjectSelectionDialog.

Initial state - no logged in users:
sign_in_screen

Logged in users, no project selection:
dialog

Filtered projects - both in our field and IDEA speed search on table (they are synced):
dialog-filter

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectListTableModel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectListTableModel.java
@@ -46,21 +46,21 @@ class ProjectListTableModel extends AbstractTableModel {
         return projectList.get(row).getName();
       case PROJECT_ID_COLUMN:
         return projectList.get(row).getProjectId();
+      default:
+        return "";
     }
-
-    return "";
   }
 
   @Override
   public String getColumnName(int column) {
     switch (column) {
       case PROJECT_NAME_COLUMN:
-        return GctBundle.getString("project.selector.project.list.project.name.column");
+        return GctBundle.getString("cloud.project.selector.project.list.project.name.column");
       case PROJECT_ID_COLUMN:
-        return GctBundle.getString("project.selector.project.list.project.id.column");
+        return GctBundle.getString("cloud.project.selector.project.list.project.id.column");
+      default:
+        return "";
     }
-
-    return "";
   }
 
   String getProjectNameAtRow(int row) {
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectListTableModel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectListTableModel.java
@@ -67,6 +67,10 @@ class ProjectListTableModel extends AbstractTableModel {
     return getValueAt(row, ProjectListTableModel.PROJECT_NAME_COLUMN).toString();
   }
 
+  String getProjectIdAtRow(int row) {
+    return getValueAt(row, ProjectListTableModel.PROJECT_ID_COLUMN).toString();
+  }
+
   void setProjectList(List<Project> updatedList) {
     projectList.clear();
     projectList.addAll(updatedList);
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectListTableModel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectListTableModel.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.intellij.project;
+
+import com.google.api.services.cloudresourcemanager.model.Project;
+import com.google.cloud.tools.intellij.util.GctBundle;
+import java.util.ArrayList;
+import java.util.List;
+import javax.swing.table.AbstractTableModel;
+
+/** Model for project list table in {@link ProjectSelectionDialog}. */
+class ProjectListTableModel extends AbstractTableModel {
+  private static final int PROJECT_NAME_COLUMN = 0;
+  private static final int PROJECT_ID_COLUMN = 1;
+
+  private final List<Project> projectList = new ArrayList<>();
+
+  @Override
+  public int getRowCount() {
+    return projectList.size();
+  }
+
+  @Override
+  public int getColumnCount() {
+    return 2;
+  }
+
+  @Override
+  public Object getValueAt(int row, int column) {
+    switch (column) {
+      case PROJECT_NAME_COLUMN:
+        return projectList.get(row).getName();
+      case PROJECT_ID_COLUMN:
+        return projectList.get(row).getProjectId();
+    }
+
+    return "";
+  }
+
+  @Override
+  public String getColumnName(int column) {
+    switch (column) {
+      case PROJECT_NAME_COLUMN:
+        return GctBundle.getString("project.selector.project.list.project.name.column");
+      case PROJECT_ID_COLUMN:
+        return GctBundle.getString("project.selector.project.list.project.id.column");
+    }
+
+    return "";
+  }
+
+  String getProjectNameAtRow(int row) {
+    return getValueAt(row, ProjectListTableModel.PROJECT_NAME_COLUMN).toString();
+  }
+
+  void setProjectList(List<Project> updatedList) {
+    projectList.clear();
+    projectList.addAll(updatedList);
+    fireTableDataChanged();
+  }
+}
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
@@ -121,10 +121,10 @@ public class ProjectSelectionDialog {
                       .connect(dialogWrapper.getDisposable() /* disconnect once dialog is gone. */)
                       .subscribe(
                           GoogleLoginListener.GOOGLE_LOGIN_LISTENER_TOPIC,
-                          this::loadUsersAndProjects));
+                          this::loadAllProjects));
     }
 
-    loadUsersAndProjects();
+    loadAllProjects();
     setSelectedProject(cloudProject);
 
     DialogManager.show(dialogWrapper);
@@ -155,7 +155,7 @@ public class ProjectSelectionDialog {
   }
 
   @VisibleForTesting
-  void loadUsersAndProjects() {
+  void loadAllProjects() {
     Collection<CredentialedUser> credentialedUsers =
         Services.getLoginService().getAllUsers().values();
     if (credentialedUsers.isEmpty()) {
@@ -430,7 +430,7 @@ public class ProjectSelectionDialog {
 
     @Override
     public void actionPerformed(ActionEvent e) {
-      loadUsersAndProjects();
+      loadAllProjects();
     }
   }
 
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
@@ -120,8 +120,7 @@ public class ProjectSelectionDialog {
                       .getMessageBus()
                       .connect(dialogWrapper.getDisposable() /* disconnect once dialog is gone. */)
                       .subscribe(
-                          GoogleLoginListener.GOOGLE_LOGIN_LISTENER_TOPIC,
-                          this::loadAllProjects));
+                          GoogleLoginListener.GOOGLE_LOGIN_LISTENER_TOPIC, this::loadAllProjects));
     }
 
     loadAllProjects();
@@ -151,7 +150,7 @@ public class ProjectSelectionDialog {
   @VisibleForTesting
   CloudProject getSelectedProject() {
     CredentialedUser user = accountComboBox.getItemAt(accountComboBox.getSelectedIndex());
-    return CloudProject.create(getSelectedProjectName(), user.getEmail());
+    return CloudProject.create(getSelectedProjectName(), getSelectedProjectId(), user.getEmail());
   }
 
   @VisibleForTesting
@@ -333,6 +332,14 @@ public class ProjectSelectionDialog {
     int actualSelectedRow =
         projectListTable.getRowSorter().convertRowIndexToModel(projectListTable.getSelectedRow());
     return projectListTableModel.getProjectNameAtRow(actualSelectedRow);
+  }
+
+  @VisibleForTesting
+  String getSelectedProjectId() {
+    // row number change based on filtering state.
+    int actualSelectedRow =
+        projectListTable.getRowSorter().convertRowIndexToModel(projectListTable.getSelectedRow());
+    return projectListTableModel.getProjectIdAtRow(actualSelectedRow);
   }
 
   @VisibleForTesting
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
@@ -45,6 +45,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Stream;
 import javax.swing.AbstractAction;
@@ -144,6 +145,9 @@ public class ProjectSelectionDialog {
         // specified user is not in logged in user list, clear the account selection.
         accountComboBox.setSelectedItem(null);
       }
+    } else {
+      // no project set, show active user projects by default.
+      accountComboBox.setSelectedItem(Services.getLoginService().getActiveUser());
     }
   }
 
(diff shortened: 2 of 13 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectListTableModel.java
```diff
@@ -0,0 +1,75 @@
...
+  @Override
+  public Object getValueAt(int row, int column) {
+    switch (column) {
+      case PROJECT_NAME_COLUMN:
+        return projectList.get(row).getName();
+      case PROJECT_ID_COLUMN:
+        return projectList.get(row).getProjectId();
+    }
+
+    return "";
```
etanshaul: if you are using a switch, why not just use a `default` case?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectListTableModel.java
```diff
@@ -0,0 +1,75 @@
...
+  @Override
+  public String getColumnName(int column) {
+    switch (column) {
+      case PROJECT_NAME_COLUMN:
+        return GctBundle.getString("project.selector.project.list.project.name.column");
+      case PROJECT_ID_COLUMN:
+        return GctBundle.getString("project.selector.project.list.project.id.column");
+    }
+
+    return "";
```
etanshaul: same comment

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectorSignInPanel.java
```diff
@@ -0,0 +1,56 @@
...
+import java.awt.event.ActionEvent;
+import javax.swing.AbstractAction;
+import javax.swing.Action;
+import javax.swing.JButton;
+
+/** Sign in panel shown in {@link ProjectSelectionDialog} when no users are logged in. */
+class ProjectSelectorSignInPanel extends BaseGoogleLoginUi {
+
+  ProjectSelectorSignInPanel() {
+    super(GctBundle.message("select.project.signin"));
```
etanshaul: Didn't catch this last time time, but perhaps a better name here? Something along the lines of `cloud.project.selector.signin.message`

etanshaul:
Instead of "Filter Name/ID", how about "Search by name or ID"?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectorSignInPanel.java
ivanporty: good idea, changing here and adding cloud.project.selector to other strings instead of just project.selector

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectListTableModel.java
ivanporty: done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectListTableModel.java
ivanporty: done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -33,6 +95,334 @@
...
+
+  @VisibleForTesting
+  CloudProject getCloudProject() {
+    CredentialedUser user = accountComboBox.getItemAt(accountComboBox.getSelectedIndex());
+    return CloudProject.create(getSelectedProjectName(), user.getEmail());
+  }
+
+  @VisibleForTesting
+  void loadUsersAndProjects() {
+    Collection<CredentialedUser> credentialedUsers =
```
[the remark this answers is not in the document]
ivanporty: It does select active user by default, see line 152, but will keep the order of users in the dropdown as it is. I can change it to have the active user on top.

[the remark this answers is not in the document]
etanshaul: oops, didn't catch that. I think its fine as is.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -33,6 +95,334 @@
...
+
+    DialogManager.show(dialogWrapper);
+
+    int result = dialogWrapper.getExitCode();
+    return result == DialogWrapper.OK_EXIT_CODE ? getCloudProject() : null;
+  }
+
+  @VisibleForTesting
+  void setCloudProject(CloudProject cloudProject) {
+    this.cloudProject = cloudProject;
```
etanshaul: this also feels like an opportunity for refactoring. setting this as a member variable feels off to me.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -33,6 +95,334 @@
...
+    updateProjectAccountInformation();
```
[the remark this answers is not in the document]
ivanporty: In loadUsersAndProjects it loads all accounts and their projects and selects active (default) account - i.e. prepares default state of the dialog.
In updateProjectAccountInformation it changes the state of the dialog to match that of the passed CloudProject. How about setProjectSelectionTo(CloudProject)?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
ivanporty: We can remove this variable and just use it once to set the state of the account combobox and project list selection - I think we won't need to keep it after that. WDYT?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -33,6 +95,334 @@
    */
   @Nullable
   CloudProject showDialog(@Nullable CloudProject cloudProject) {
-    return cloudProject;
+    // create dialog wrapper for the form once, create services and subscribe to login changes once.
+    if (dialogWrapper == null) {
+      dialogWrapper = new ProjectSelectionDialogWrapper();
```
etanshaul: I'm trying to wrap my head around this. Why didn't you want this class to be a `DialogWrapper` itself?

etanshaul: in other words, why not just extend from DialogWrapper? As is it is, it is very tightly coupled to DialogWrapper anyway, so I'm not seeing the advantage. I could be wrong though.

ivanporty: I tried it first and inheriting from `DialogWrapper` brings a lot of internal UI implementation with it - I had to spy/mock/override a lot of internal methods, and it started 'leaking projects' because I used it - something was not disposing properly in mock test mode. It's much easier and cleaner to test and edit as a simple form, but we have to have this wrapper class to pass dialog wrapper buttons and additional actions. It is not very friendly to composition instead of inheritance.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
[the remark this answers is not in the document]
ivanporty: Removed updateProjectAccountInformation() altogether.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
ivanporty: done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -33,6 +103,346 @@
...
+
+  @VisibleForTesting
+  void setCloudProject(CloudProject cloudProject) {
+    if (cloudProject != null && !Strings.isNullOrEmpty(cloudProject.googleUsername())) {
+      Optional<CredentialedUser> loggedInUser =
+          Services.getLoginService().getLoggedInUser(cloudProject.googleUsername());
+      if (loggedInUser.isPresent()) {
+        selectedProjectsByAccount.put(loggedInUser.get(), cloudProject.projectName());
+        accountComboBox.setSelectedItem(loggedInUser.get());
+        updateProjectList(loggedInUser.get());
```
etanshaul: why is this needed here if the project list is already loaded (and cached) in #loadUsersAndProjects ?

etanshaul: oh..I think i see..thinking about this

etanshaul: actually, nm. If the cached list is already loaded, why do we need to load it again for this specific user?

ivanporty: It won't be loaded since it's cached, I sort of added it for increased readability so reader doesn't need to imply project will be selected because of the combox box event. We can remove it safely.

etanshaul: ah ok. lets remove it then. it had the opposite effect on me. made me confused. :)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -33,6 +103,346 @@
...
+              project ->
+                  project
+                      .getMessageBus()
+                      .connect(dialogWrapper.getDisposable() /* disconnect once dialog is gone. */)
+                      .subscribe(
+                          GoogleLoginListener.GOOGLE_LOGIN_LISTENER_TOPIC,
+                          this::loadUsersAndProjects));
+    }
+
+    loadUsersAndProjects();
```
etanshaul: I know we can go back and forth forever on naming, but just thinking after looking again..how about:

loadUsersAndProjects -> loadAllProjects
setCloudProject -> setActiveProject (I think the "cloud" part is implied enough)

etanshaul: actually not setActiveProject, but setSelectedProject

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
ivanporty: done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
ivanporty: sounds good, and I'll change the getter name to reflect that idea as well - i.e. `getSelectedProject`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -33,6 +103,348 @@
...
+        Services.getLoginService().getAllUsers().values();
+    if (credentialedUsers.isEmpty()) {
+      showSignInRequest();
+    } else {
+      hideSignInRequest();
+      cachedProjectList.clear();
+      accountComboBox.removeAllItems();
+      for (CredentialedUser user : credentialedUsers) {
+        accountComboBox.addItem(user);
+        updateProjectList(user);
```
etanshaul: This is what I was thinking, lmkwyt

- `loadAllProjects` is responsible for loading and caching the projects for all users. Thats is; it doesnt update the UI.
- `setSelectedProject` then updates the dropdown with the current CloudProject, and in turn, causes the table model to update (pulling the list of projects from the cache).
- the listener on the account combobox updates the table model like in the above bullet

Its similar, but I think this will separate concerns a bit more cleanly. Sorry if I am contradicting previous things I've said with these comments. The motivation is I think that updateProjectList is a bit too complicated and doing too much (loading projects, updating the UI..).

ivanporty: sounds reasonable, I'll do a simple refactor of this idea, trying to clean up `updateProjectList' a bit more

etanshaul:
Can be done in a separate PR later, but it might be nice to have some loading indicator when the projects are being refreshed (or when a new account is selected and the project list is loading).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -16,15 +16,85 @@
...
+  private RefreshAction refreshAction;
+
+  private ProjectListTableModel projectListTableModel;
+
+  private ProjectLoader projectLoader = new ProjectLoader();
+
+  private Map<CredentialedUser, ListenableFuture<List<Project>>> runningProjectLoaderJobs =
+      Maps.newHashMap();
+  // last project name selection made by user for an account.
+  private Map<CredentialedUser, String> selectedProjectsByAccount = Maps.newHashMap();
```
etanshaul: what do you think about combining selecteProjectsByAccount and cachedProjectList into a single mutlimap?

etanshaul: specifically, guava's multimap

ivanporty: you mean selected project is marked somehow in project list, i.e. we have Multimap<User,Project> and selected project is one of them?

etanshaul: Never mind, it may have been a bad idea.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -33,6 +103,348 @@
...
+    return result == DialogWrapper.OK_EXIT_CODE ? getSelectedProject() : null;
+  }
+
+  @VisibleForTesting
+  void setSelectedProject(CloudProject cloudProject) {
+    if (cloudProject != null && !Strings.isNullOrEmpty(cloudProject.googleUsername())) {
+      Optional<CredentialedUser> loggedInUser =
+          Services.getLoginService().getLoggedInUser(cloudProject.googleUsername());
+      if (loggedInUser.isPresent()) {
+        selectedProjectsByAccount.put(loggedInUser.get(), cloudProject.projectName());
```
etanshaul: Cloud project name is not unique; project ID is. I just want to make sure the logic here and elsewhere is aware of this. In other words, if this is actually using the cloud project name here, then I think this example usage will have problems (since it may not select the correct project).

ivanporty: so in `AppEngineDeploymentConfiguration`, the field is `cloudProjectName`, is it actually name or ID?

etanshaul: The main thing is that whenever we are keying on it, we must use the ID. But if we are displaying it (like in the new Cloud Project widget - [name] / [email]) we use the name.

ivanporty: sounds good, then let's add  `projectId` to keep the actual ID in `CloudProject` too

ivanporty:
Cleaned up `updateProjectList` a bit more, renamed to better understand what it does.
Also, added loading indicator - using small neat indicator in top right corner from IDEA table.
PTAL at changes.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -33,6 +104,363 @@
...
+      accountComboBox.removeAllItems();
+      for (CredentialedUser user : credentialedUsers) {
+        accountComboBox.addItem(user);
+        loadProjectList(user);
+      }
+    }
+  }
+
+  private void loadProjectList(CredentialedUser user) {
+    // check if cached, if not then check if already loading, and finally start loading if not.
```
etanshaul: could you promote top level comments like this that apply to the whole method to actually be javadoc method comments? I know some of these cases are for private methods, but I still think it makes sense.

etanshaul:
I dunno if its a bug or not, but the loader doesn't consistently show up. I just clicked the refresh button a bunch of times and sometimes it appeared and sometimes not.

Also, maybe move the loader to be beside the refresh button? It took me several seconds to find it even when it did appear.

etanshaul:
or even in the middle of table (if thats possible). wherever you think makes more sense.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -33,6 +104,363 @@
...
+  }
+
+  @VisibleForTesting
+  CloudProject getSelectedProject() {
+    CredentialedUser user = accountComboBox.getItemAt(accountComboBox.getSelectedIndex());
+    return CloudProject.create(getSelectedProjectName(), getSelectedProjectId(), user.getEmail());
+  }
+
+  @VisibleForTesting
+  void loadAllProjects() {
```
etanshaul: thanks for the refactoring. much easier to read now

etanshaul:
possible bug:

1) sign in to multiple accounts
2) select the second account in the list
3) click refresh

-> I'd expect the same account's projects to be refreshed. Instead it sent me back to the other account.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
ivanporty: sure, great idea.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
ivanporty: yeah it is much easier to read, good catch!

ivanporty:
Added clearly visible centered overlay progress bar, preserved account selection when refresh button is hit.

PR #1779, opened 2017-12-12
Upgrade gradle-intellij-plugin to latest version

etanshaul:
It just worked. They must have improved the multi-module support.

I had to also update the publish section since the old method is now deprecated.

TODOs

1) Test the plugin publish to make sure it still works
2) pluginId from the 'publish' section is now deprecated. Instead, the plugin id is pulled from the ID of plugin.xml. Does this mean we need to update the ID in the plugin.xml for the publish to work (and what are the effects of doing that)?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,6 @@ buildscript {
         maven { url "https://plugins.gradle.org/m2/" }
     }
     dependencies {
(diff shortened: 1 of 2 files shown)


etanshaul:
Tested out publish to alpha and it works properly.

etanshaul:
@patflynn I'll go ahead and merge since things look good to me but if you can recall the other reasons we delayed this then we can revisit those or roll this back if needed

PR #1782, opened 2017-12-13
App Engine local run deployment source cleanup and code consolidation

etanshaul:
First step at cleaning up AppEngineUtil and the deployment sources for App Engine local run.

- Consolidates the deployment source collection logic across local run and deploy
  - Previously, there was separate (yet ultimately identical) deployment source collection logic for local run vs deploy
  - Note: this does not bring in maven deployment sources to local run; only 'artifact' sources (native IntelliJ ones). This will be tackled separately
- Moves the local run UI logic out of the utility class

### DIFF
--- a/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineRunConfigurationEditor.java
+++ b/google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineRunConfigurationEditor.java
@@ -16,9 +16,10 @@
 
 package com.google.cloud.tools.intellij.appengine.server.instance;
 
+import com.google.cloud.tools.intellij.appengine.cloud.AppEngineArtifactDeploymentSource;
 import com.google.cloud.tools.intellij.appengine.util.AppEngineUtil;
 import com.google.cloud.tools.intellij.util.GctBundle;
-
+import com.google.common.collect.Lists;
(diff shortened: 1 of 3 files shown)


google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/server/instance/AppEngineRunConfigurationEditor.java
```diff
@@ -158,4 +160,29 @@ public void setAnchor(JComponent anchor) {
...
+              JList list, Artifact value, int index, boolean selected, boolean hasFocus) {
+            if (value != null) {
+              setIcon(value.getArtifactType().getIcon());
+              setText(value.getName());
+            }
+          }
+        });
+
+    List<AppEngineArtifactDeploymentSource> deploymentSources = Lists.newArrayList();
+    deploymentSources.addAll(AppEngineUtil.createArtifactDeploymentSources(myProject));
```
ivanporty: So `AppEngineUtil.createArtifactDeploymentSources(myProject)` essentially does the same as previous `collectAppEngineStandardArtifacts` implementation did, but using the standard way, correct?

etanshaul: exactly. my goal is to get test coverage here, but that will first require wider refactoring of AppEngineUtil (which I'll do separately)

Issue #1268, opened 2017-02-23
gradle google-java-format tasks don't appear to work


patflynn:
steps to test:

- make some random change to a java file that's clearly not google formatted (like make a line longer than 100 chars)
- then run `gradlew verifyGoogleJavaFormat`

observe that no errors are displayed

etanshaul:
obsolete. we are currently using spotless

PR #1783, opened 2017-12-13
Reformat from posix os take 3

patflynn:


patflynn:
testing whether running this from linux instead of windows will make any difference.

patflynn:
@ivanporty I'm thinking you should merge your project selector work before I push this through to minimize merge pain on your end.

patflynn:
Closes #1268

PR #1784, opened 2017-12-13
Add error prone and google java format plugins as project requirements

patflynn:
@etanshaul what do you think about this?

### DIFF
--- a/.idea/externalDependencies.xml
+++ b/.idea/externalDependencies.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ExternalDependencies">
+    <plugin id="Error-prone plugin" />
+    <plugin id="google-java-format" />
+  </component>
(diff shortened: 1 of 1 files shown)


etanshaul:
what actually happens? It forces you to install the plugins?

.idea/externalDependencies.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ExternalDependencies">
+    <plugin id="Error-prone plugin" />
+    <plugin id="google-java-format" />
+  </component>
+</project>
```
etanshaul: maybe newline here?

patflynn: IJ generated file when using the UI. gonna leave as is as I think it'll get reformatted any time it's changed from settings.

etanshaul: sounds good

PR #1786, opened 2017-12-14
Update GCS Bucket UI to use new project selector.

ivanporty:
Form and panel change, unit test update.
gcs-1

gcs-2

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
@@ -119,12 +119,11 @@ final class GcsBucketPanel {
       notificationLabel.setText(GctBundle.message("gcs.panel.bucket.listing.no.project.selected"));
     } else {
       CloudProject cloudProject = projectSelector.getSelectedProject();
-      String projectId = cloudProject.projectId();
       Optional<CredentialedUser> user =
           Services.getLoginService().getLoggedInUser(cloudProject.googleUsername());
 
       if (user.isPresent()) {
-        loadAndDisplayBuckets(projectId, user.get());
+        loadAndDisplayBuckets(cloudProject.projectId(), user.get());
       } else {
         notificationLabel.setText(
             GctBundle.message("gcs.panel.bucket.listing.error.loading.buckets"));
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
@@ -22,8 +22,9 @@ import com.google.cloud.storage.StorageException;
 import com.google.cloud.tools.intellij.analytics.UsageTrackerProvider;
 import com.google.cloud.tools.intellij.login.CredentialedUser;
 import com.google.cloud.tools.intellij.login.Services;
+import com.google.cloud.tools.intellij.project.CloudProject;
+import com.google.cloud.tools.intellij.project.ProjectSelector;
 import com.google.cloud.tools.intellij.resources.GoogleApiClientFactory;
-import com.google.cloud.tools.intellij.resources.ProjectSelector;
 import com.google.cloud.tools.intellij.ui.CopyToClipboardActionListener;
 import com.google.cloud.tools.intellij.util.GctBundle;
 import com.google.cloud.tools.intellij.util.GctTracking;
@@ -35,7 +36,6 @@ import com.intellij.openapi.fileEditor.FileEditorManager;
 import com.intellij.openapi.fileEditor.OpenFileDescriptor;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.vfs.VirtualFile;
-import com.intellij.ui.DocumentAdapter;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


etanshaul:
this point is not directly related to this PR, perhaps we can iterate now on the look and feel of this widget. something about it feels off to me:

- Even with the links its not entirely clear to me that it should be clicked
- It doesn't stand out enough from its surroundings
- The combination of the links and the `...` button is a bit confusing.

I'll also look myself to see if I can find an example of something similar to what I have in mind.

ivanporty:
We sort of started with a widget similar to what IDEA uses a lot - text field with a status and 'Browse' and then added hyperlinks - so that feels a bit off. I'm thinking on it too and try to come with a couple of screenshots.

etanshaul:
cool. i'll review this in the meantime since we can update the look of the widget separately

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
```diff
@@ -124,14 +115,16 @@ void refresh() {
...
     if (!Services.getLoginService().isLoggedIn()) {
       notificationLabel.setText(GctBundle.message("gcs.panel.bucket.listing.not.logged.in"));
-    } else if (StringUtils.isEmpty(projectSelector.getText())) {
+    } else if (projectSelector.getSelectedProject() == null) {
       notificationLabel.setText(GctBundle.message("gcs.panel.bucket.listing.no.project.selected"));
     } else {
-      String projectId = projectSelector.getText();
-      CredentialedUser user = projectSelector.getSelectedUser();
+      CloudProject cloudProject = projectSelector.getSelectedProject();
+      String projectId = cloudProject.projectId();
```
etanshaul: is this only used once? if so could you get rid of this variable extraction?

ivanporty: We also grab google user name from it

ivanporty: Or you mean remove projectId?

etanshaul: i meant getting rid of line 122 `String projectId = cloudProject.getProjectId()` since its only used once; you could just call `cloudProject.projectId` directly on line 127 no?

ivanporty: right, sounds good.

etanshaul: Also, I'm noticing the naming of the methods in CloudProject aren't standard pojo naming: `CloudProject#projectId` instead of `CloudProject#getProjectId`. Does this have something to do with `@AutoValue`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
```diff
@@ -79,14 +77,7 @@
...
 
-    projectSelector
-        .addTextChangedListener(
-            new DocumentAdapter() {
-              @Override
-              protected void textChanged(DocumentEvent event) {
-                refresh();
-              }
-            });
+    projectSelector.addProjectSelectionListener((selected) -> refresh());
```
etanshaul: this is so much nice now :)

ivanporty: it looks right :)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
ivanporty: Yes, it looks like that the default way AutoValue is used: https://g3doc.corp.google.com/third_party/java_src/auto/value/g3doc/index.md?cl=head

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
```diff
@@ -124,14 +115,15 @@ void refresh() {
...
-      String projectId = projectSelector.getText();
-      CredentialedUser user = projectSelector.getSelectedUser();
+      CloudProject cloudProject = projectSelector.getSelectedProject();
+      Optional<CredentialedUser> user =
+          Services.getLoginService().getLoggedInUser(cloudProject.googleUsername());
 
-      if (user != null) {
-        loadAndDisplayBuckets(projectId, user);
+      if (user.isPresent()) {
+        loadAndDisplayBuckets(cloudProject.projectId(), user.get());
```
etanshaul: I really don't like called `get` on an `Optional`. But I suppose its better than this:

```
userOption.map(user -> loadAndDisplayBuckets(...)).orElseGet(notificationLabel.setText(..));
```

wdyt?

I wish java had better ways to handle this. I guess we have to wait until pattern matching makes it into the jdk.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
etanshaul: sounds good

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
ivanporty: I would agree that multi-line `if` instead of one liner `ifPresent().orElse()` is easier to read.

etanshaul: well it wouldn't necessarily be a one liner. the formatter would probably put it on multiple lines.

Either way, using map purely for a side effect is ugly too. So I guess would leave it.

PR #1787, opened 2017-12-15
Convert setup on cloud repo dialog to use new project selector.

ivanporty:
setup-on-repo1

projectrepo2

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/SetupCloudRepositoryDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/SetupCloudRepositoryDialog.java
@@ -17,30 +17,28 @@
 package com.google.cloud.tools.intellij.vcs;
 
 import com.google.cloud.tools.intellij.login.CredentialedUser;
-import com.google.cloud.tools.intellij.resources.ProjectSelector;
+import com.google.cloud.tools.intellij.login.Services;
+import com.google.cloud.tools.intellij.project.CloudProject;
+import com.google.cloud.tools.intellij.project.ProjectSelector;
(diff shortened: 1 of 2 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/vcs/SetupCloudRepositoryDialog.java
```diff
@@ -99,44 +96,53 @@ protected String getDimensionServiceKey() {
 
   private void createUIComponents() {
     projectSelector = new ProjectSelector();
-    projectSelector.setMinimumSize(new Dimension(300, 0));
+    projectSelector.setMinimumSize(new Dimension(400, 0));
 
-    repositorySelector = new RepositorySelector(projectSelector.getText(),
-        projectSelector.getSelectedUser(), true /*canCreateRepository*/);
+    /* TODO(ivanporty) use 'default' project to init selector when available. */
+    repositorySelector = new RepositorySelector("", null, true /*canCreateRepository*/);
```
etanshaul: what effect does this have (initializing the RepositorySelector with empty string and null compared to before)? And should we maybe update the Repository Selector constructor if this is needed?

ivanporty: This is essentially what the previous code did - using empty, just created instance of ProjectSelector, so this has no effect on new functionality.
I plan to refactor RepositorySelector to use CloudProject in its constructor next PR, when updating 'clone from google cloud' form.

etanshaul: I see

PR #1790, opened 2017-12-18
 Get rid of redundant facet check on app engine local run; add artifact guidance wording

etanshaul:
This check is not needed because the facet is already confirmed prior to adding it to the target deployable dropdown (if it didn't have this facet, it would not have shown up in the dropdown to begin with):

image


ivanporty:
So in case we don't have this facet, we won't have this (implied) drop down, but just an error message - `Artifact isn't specified`, correct?

etanshaul:
correct

ivanporty:
So if I try to add an App Engine Standard target when project is App Engine Flex, it gives me empty drop-down for Artifact and `Artifact isn't specified` error. Is it clear enough to suggest what I did wrong here? Maybe expand that error a bit to make it easier to understand what is `Artifact` in this case and how to fix it?

etanshaul:
Yeah good idea. The usually is generally steered away from opening a local dev server run config for flex (they are blocked from the tools menu shortcut, and it shows up as hidden by default when manually adding the run config), but some detailed wording here is a good idea. I'll add it.

etanshaul:
@ivanporty @patflynn 

Added some guiding wording:

image

etanshaul:
On second look. Maybe I'll add something about requiring exploded-war artifacts too.

etanshaul:
Updated the wording and added some line breaks. let me know what you think.

image

etanshaul:
screenshot above is updated to reflect all the wording changes

ivanporty:
That's awesome, would be really helpful for a confused user!

PR #1793, opened 2017-12-19
Update deployment UI (flex and standard) to use new project selector

ivanporty:
deploy-flex

deploy-standard

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
@@ -26,6 +26,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.intellij.remoteServer.configuration.deployment.DeploymentSource;
 import com.intellij.ui.HyperlinkLabel;
+import com.intellij.ui.components.JBCheckBox;
 import com.intellij.ui.components.JBTextField;
 import java.util.Optional;
 import javax.swing.JCheckBox;
@@ -49,6 +50,7 @@ public final class AppEngineDeploymentConfigurationPanel {
   private JPanel appEngineCostWarningPanel;
   private HyperlinkLabel appEngineCostWarningLabel;
   private JLabel serviceLabel;
+  private JCheckBox hiddenValidationTrigger;
 
   private static final boolean PROMOTE_DEFAULT = false;
   private static final boolean STOP_PREVIOUS_VERSION_DEFAULT = false;
@@ -84,7 +86,13 @@ public final class AppEngineDeploymentConfigurationPanel {
     appEngineCostWarningLabel.addHyperlinkListener(new BrowserOpeningHyperLinkListener());
     appEngineCostWarningLabel.setHyperlinkTarget(CloudSdkAppEngineHelper.APP_ENGINE_BILLING_URL);
 
-    projectSelector.addProjectSelectionListener(this::refreshApplicationInfoPanel);
+    projectSelector.addProjectSelectionListener(
+        (selected) -> {
+          refreshApplicationInfoPanel(selected);
+          // manually trigger validate, hyperlinks events not caught by standard settings editor
+          // watcher.
+          triggerSettingsEditorValidation();
+        });
   }
 
   /**
@@ -152,6 +160,12 @@ public final class AppEngineDeploymentConfigurationPanel {
     }
   }
 
+  /** Triggers validation and marks settings as 'changed' using hidden component. */
+  // TODO(ivanporty) explore UserActivityProviderComponent usage.
+  public void triggerSettingsEditorValidation() {
+    hiddenValidationTrigger.doClick();
+  }
+
   /**
    * Updates the text of the panel as follows: if no project is selected, no message is displayed,
    * if the project represents a valid project, the project details are displayed, if the project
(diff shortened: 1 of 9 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -155,6 +154,17 @@ public void customize(
     editAppYamlButton.addActionListener(event -> openModuleSettings());
     dockerfileDirectoryPathLink.addHyperlinkListener(event -> openModuleSettings());
 
+    // we need to add a separate state change listener for project selector since hyperlinks are not
+    // caught by standard settings editor watcher.
```
etanshaul: what about the other potential solution for this that you mentioned? (the platform interface which will expose this to the editor watcher)

etanshaul: also, if we do go with this, can this be promoted to the common `AppEngineDeploymentConfigurationPanel` since this logic is duplicated for flex and standard

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
```diff
@@ -145,25 +141,32 @@ public void applyEditorTo(@NotNull AppEngineDeploymentConfiguration configuratio
...
-   * Updates the text of the panel as follows: if the project text box is empty, no message is
-   * displayed, if the project text represents a valid project, the project details are displayed,
-   * if the project text represents an invalid project, an error message is displayed.
+   * Updates the text of the panel as follows: if no project is selected, no message is displayed,
+   * if the project represents a valid project, the project details are displayed, if the project
+   * represents an invalid project, an error message is displayed.
    */
-  private void refreshApplicationInfoPanel() {
-    if (Strings.isNullOrEmpty(projectSelector.getText())) {
+  private void refreshApplicationInfoPanel(CloudProject selectedProject) {
```
etanshaul: why pass it in now instead of pulling it from the ProjectSelector as before?

ivanporty: With updated API project updated event is an updated `CloudProject` so it's easy to just pass it in as a parameter instead of calling it directly. It might be useful for more unit testing later as well. What do you think, does it feel it's less readable?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
```diff
@@ -110,15 +95,26 @@ protected void textChanged(DocumentEvent e) {
...
-    projectSelector.setText(configuration.getCloudProjectName());
     stopPreviousVersionCheckbox.setSelected(configuration.isStopPreviousVersion());
     deployAllConfigsCheckbox.setSelected(configuration.isDeployAllConfigs());
 
     if (configuration.getEnvironment() != null) {
       environmentLabel.setText(configuration.getEnvironment().localizedLabel());
     }
 
-    refreshApplicationInfoPanel();
+    // TODO(ivanporty) add project name to configuration and then use separate project ID field.
```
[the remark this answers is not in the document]
ivanporty: I mean here when we populate from `AppEngineDeploymentConfiguration`, we don't have project name there, just project ID. Unless we open dialog and load the list to figure out the project name, we will show project ID. So if we also keep project name and ID in `AppEngineDeploymentConfiguration`, we will show proper name in `ProjectSelector`.

I confirmed that `ProjectSelector#getText` was returning ID of the project, so deployment config using `getCloudProjectName()` is actually misleading, it's ID.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
ivanporty: It's called `UserActivityProviderComponent`, it must be supported, I tried it but it didn't work from the initial tries - they traverse container hierarchy and I think something is missing again. It will take more time than needed to debug this.

Good idea, I can put that checkbox into general panel.

etanshaul: sounds good. lets explore that separately then

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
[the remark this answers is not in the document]
ivanporty: Exactly, initially the project selector will show ID instead of a name. And then after dialog opens and load project list, it can show actual project name. That's what I thought too, it needs a separate PR since `AppEngineDeploymentConfiguration` is used in many places.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
```diff
@@ -85,20 +86,12 @@ public AppEngineDeploymentConfigurationPanel() {
...
-        });
-    projectSelector.addTextChangedListener(
-        new DocumentAdapter() {
-          @Override
-          protected void textChanged(DocumentEvent e) {
-            refreshApplicationInfoPanel();
-          }
+    projectSelector.addProjectSelectionListener(
+        (selected) -> {
+          refreshApplicationInfoPanel(selected);
```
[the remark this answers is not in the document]
ivanporty: `selectedCloudProject` sounds good, changed.

Issue #1792, opened 2017-12-18
Old loading icon in project selector needs to be removed


etanshaul:
I can't catch the screenshot fast enough (without modifying the code), but there are now 2 loader icons. The new big one in the center, and the old small one in the top right corner (I think you just forgot to remove it). Or was it intentional?

ivanporty:
It's kind of a 'free' perk since it's built-in into IDEA table, but it's really not necessary as we have large one showing. I'll take it out.

etanshaul:
ah. I didn't realize it was automatically added.

ivanporty:
Fixed.

Issue #1794, opened 2017-12-19
Add double-click support for project selector table.


ivanporty:
Double clicking on project in the list selects a project and closes the dialog.

ivanporty:
Fixed.

PR #1795, opened 2017-12-19
Few cosmetic enhancements and fixes in ProjectSelector.

### LINKED ISSUE #1791: Project selector should select account after sign in
1) start logged out of all google accounts
2) load up the new project selector and click to open a dialog
3) sign into your account

-> I would have expected the account I just signed into to be selected now. Instead, no account is selected  (it _is_ in the dropdown, I just need to manually choose it):

image

image

ivanporty:
Fixes #1791, #1792, #1794.
Also updates project list to render all matching cases bold regardless of case.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
@@ -269,13 +269,14 @@ public class ProjectSelectionDialog {
     FilteredTextTableCellRenderer filterRenderer = new FilteredTextTableCellRenderer();
     projectListTable.setDefaultRenderer(Object.class, filterRenderer);
 
-    DoubleClickListener tableDoubleClickListener = new DoubleClickListener() {
-      @Override
-      protected boolean onDoubleClick(MouseEvent event) {
-        dialogWrapper.clickDefaultButton();
-        return true;
-      }
-    };
+    DoubleClickListener tableDoubleClickListener =
+        new DoubleClickListener() {
+          @Override
+          protected boolean onDoubleClick(MouseEvent event) {
+            dialogWrapper.clickDefaultButton();
+            return true;
+          }
+        };
     tableDoubleClickListener.installOn(projectListTable);
 
     // filter rows based on text field content.
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
@@ -130,8 +130,7 @@ public class ProjectSelectionDialog {
                       .subscribe(
                           GoogleLoginListener.GOOGLE_LOGIN_LISTENER_TOPIC,
                           () -> {
-                            refreshDialog(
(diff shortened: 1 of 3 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -215,7 +217,12 @@ public void onFailure(Throwable throwable) {
   private void refreshDialog() {
     CredentialedUser currentUser = (CredentialedUser) accountComboBox.getSelectedItem();
     loadAllProjects();
-    accountComboBox.setSelectedItem(currentUser);
+    if (currentUser != null) {
+      accountComboBox.setSelectedItem(currentUser);
+    } else {
+      // no selection before, select active user by default.
```
etanshaul: Did you confirm that the most recently logged in user will be the active user?

The behavior I think we are after is that the _most recently logged in user_ is the one selected in the dropdown, regardless of if its the default user or not.

etanshaul:
I may be crazy, but where in this PR did you fix #1792 (removing the old loader)?

ivanporty:
Here, this line has been removed in the dialog code: ((JBTable) projectListTable).setPaintBusy(loading);
this was the call to set table as 'loading'

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
ivanporty: So, when we reload, we take the currently selected user, and after reload, set it back as selected so the state is preserved. If nothing was selected (no signed in users first), we check for null and set user to active - which I think is the most recently logged in user.

etanshaul: Yeah thats what I wanted to be sure of - that the most recently logged in user will always be selected (even if I am already logged in to one or more accounts). I think, that be adding a new account I'd want that new account selected on successful login (even if there are other accounts already logged in).

ivanporty: Got it, updated - so manual refresh now keeps the selection, but if we change accounts, i.e. add new active one, dialog will switch to new active one and load its project list.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -211,11 +217,21 @@ public void onFailure(Throwable throwable) {
...
 
-  /** Preserves currently selected account and refreshes accounts and project lists. */
-  private void refreshDialog() {
+  /**
+   * Preserves currently selected account if requested and refreshes accounts and project lists.
+   *
+   * @param keepPreviousSelection true to keep current user selected after refresh. false to select
+   *     active user instead.
+   */
+  private void refreshDialog(boolean keepPreviousSelection) {
```
etanshaul: would it be possible to directly pass in the user to select rather than having a flag like this?

ivanporty: good idea, let me rework it that way,

PR #1796, opened 2017-12-20
API management features UI setup

etanshaul:
I'm splitting up the API management features into smaller PRs.

This part sets up the UI in the second step of the wizard per the design ([go/api-management-intellij](http://go/api-management-intellij)) for the API management panel. Note that it does not query the service management API to get the actual enabled / disabled status of the API; this will come next.

Step 1 (from earlier PR): Select the libraries you wish to add:
image

Step 2: management panel state with no Cloud Project yet selected:
image

Step3: management panel after selecting a Cloud Project:
image


etanshaul:
@ivanporty @patflynn ptal. I know this could use some look and feel updates (I'm thinking about how exactly to improve it).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelector.java
```diff
@@ -155,7 +155,8 @@ private void updateCloudProjectSelection(CloudProject selection) {
         GoogleLoginIcons.getScaledUserIcon(ACCOUNT_ICON_SIZE, loggedInUser.orElse(null)));
   }
 
-  private void notifyProjectSelectionListeners() {
+  @VisibleForTesting
+  public void notifyProjectSelectionListeners() {
```
etanshaul: @ivanporty I really don't want to do this, but it was a quick and dirty way to make things testable. Thinking of a better way.

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/apis/TestCloudLibrary.java
```diff
@@ -100,6 +105,8 @@ public static TestCloudLibraryClient createEmpty() {
       return create("", "", "", "", "", "", TestCloudLibraryClientMavenCoordinates.createEmpty());
```
ivanporty: It looks like think this is where spotlessJava check fails - we probably need to reformat spaces in this file.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryRenderer.java
```diff
@@ -0,0 +1,52 @@
...
+import javax.swing.border.EmptyBorder;
+import javax.swing.table.DefaultTableCellRenderer;
+
+/**
+ * Custom {@link javax.swing.table.TableCellRenderer TableCellRenderer} for {@link CloudLibrary}
+ * objects.
+ */
+final class CloudLibraryRenderer extends DefaultTableCellRenderer {
+
+  private static final Border NO_FOCUS_BORDER = new EmptyBorder(5, 5, 5, 5);
```
ivanporty: IDEA suggests to replace default empty border with the following call:
`JBUI.Borders..empty(5)`.
I think this is related with how they treat border in component in case look and feel changes (i.e. Idea to Darcula etc.)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryRenderer.java
```diff
@@ -0,0 +1,52 @@
...
+
+  @Override
+  public Component getTableCellRendererComponent(
+      JTable table, Object value, boolean isSelected, boolean hasFocus, int row, int column) {
+    Component component =
+        super.getTableCellRendererComponent(table, value, isSelected, hasFocus, row, column);
+    setBorder(NO_FOCUS_BORDER);
```
ivanporty: Please correct me if I'm wrong, do we set the `NO_FOCUS_BORDER` only to remove "selected" state of the cell in the table? What do you think if we just pass `isSelected` and `hasFocus` as always `false` instead into `super.getTableCellRendererComponent`? It might be a bit easier to understand and we can directly return the result of `super` call.

ivanporty:
As a side note, it probably takes some time to understand a difference between adding a library into a project (assuming it is added as a plain old JAR library and its dependencies) and then enabling the library (this time we mean we enable that on cloud side). Can we also ask our UX person about this? I.e. compare the two wizard pages with one page and one table with just one more column added - i.e.

| Library | API Added for Module | API Enabled on Server |
|-----------|-------------------------------|--------------------------------|
| Name   |  Checkbox                   | Checkbox                     |

Second column could be enabled only when project is selected and otherwise rendered disabled.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelector.java
etanshaul: @ivanporty to make this better I'm thinking of the following. let me know your thoughts:

- changing `ProjectSelector#handleOpenProjectSelectionDialog` -> `ProjecSelector#showDialog()` (just to make things clearer as a public method)
- making the above part of the public api of the selector
- reverting the change here (line 159) and using the showDialog method for unit testing

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiManagementPanel.java
```diff
@@ -0,0 +1,278 @@
...
+    apisScrollPane.setVisible(true);
+    selectProjectLabel.setVisible(false);
+  }
+
+  /** Initializes the table models of the API tables based on the user's library selection. */
+  // TODO(eshaul) its very important that we add an equals method to CloudLibrary so that the set
+  // operation is reliable
+  void init(List<CloudLibrary> allLibraries, Set<CloudLibrary> selectedLibraries) {
+    selectedApisTable.setModel(new ApisTableModel(selectedLibraries));
+    selectedApisTable
```
ivanporty: It feels a bit off that we mix the code for updating the model of the tables with data and then resize the column.
How about we initialize the table with empty model in `createUIComponents` and apply column sizing/formatting in there as well, and use `init` for data operations only?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryRenderer.java
etanshaul: this was code brought over from a previous PR (I just gave it its own class since its now reused).

But that being said, I'll try your suggestion.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiManagementPanel.java
etanshaul: sgtm

etanshaul:
And to your higher level point:

I completely agree that there is some cognitive overhead involved in understanding the difference between the first step (add Client _Libraries_) and the second step (managing the Cloud _APIs_). 

Your suggestion of collapsing them into a single table is interesting. We could probably fix the current multi-step solution with better wording too. I'll think about this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiManagementPanel.java
```diff
@@ -0,0 +1,278 @@
...
+        .getColumn(CLOUD_API_NAME_COL)
+        .setMinWidth(MIN_CLOUD_API_COL_WIDTH);
+    allApisTable
+        .getColumnModel()
+        .getColumn(CLOUD_API_ENABLEMENT_COL)
+        .setMinWidth(MIN_CLOUD_API_ENABLEMENT_COL_WIDTH);
+  }
+
+  JPanel getPanel() {
+    return panel;
```
ivanporty: `panel` as a variable and getter sound a bit too generic? Maybe name it after the class itself meaning it contains everything in it - i.e. `googleCloudApiManagementPanel` or `contentPanel`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiManagementPanel.java
```diff
@@ -0,0 +1,278 @@
...
+    selectedApisTable = new ApisTable();
+    selectedApisTable.setTableHeader(null);
+
+    allApisTable = new ApisTable();
+    allApisTable.setTableHeader(null);
+  }
+
+  private static final class ApisTable extends JBTable {
+    ApisTable() {
+      setDefaultRenderer(CloudLibrary.class, new CloudLibraryRenderer());
```
ivanporty: Do we only create separate `ApisTable` class to `setDefaultRenderer`? Maybe leave it as general JBTable and just call the methods directly in  `createUIComponents`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiManagementPanel.java
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    @Override
+    public boolean isCellEditable(int rowIndex, int columnIndex) {
+      return false;
+    }
+
+    @Override
+    public Object getValueAt(int rowIndex, int columnIndex) {
+      return columnIndex == 0
```
ivanporty: How about using `switch` here? It's pretty clear from `switch` how many columns we have and how we assign them values whereas here it looks like we only provide separate value for column 0 here. And also we might use constants instead of number (i.e. `CLOUD_API_NAME_COL`) maybe? Same idea about `getColumnClass`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiManagementPanel.java
etanshaul: sure. I'll change it to contentPanel

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiManagementPanel.java
```diff
@@ -0,0 +1,278 @@
...
+          ? librariesToEnabledStatus.keySet().toArray()[rowIndex]
+          : librariesToEnabledStatus.values().toArray()[rowIndex];
+    }
+
+    @Override
+    public void setValueAt(Object aValue, int rowIndex, int columnIndex) {}
```
ivanporty: Can you elaborate a bit more please why we need these 3 `TableModel` methods overriden with empty implementations? It might be easier to inherit from `AbstractTableModel` - then we won't have to do it and also won't have to override `isCellEditable()`.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiManagementPanel.java
```diff
@@ -0,0 +1,278 @@
...
+   */
+  private static final class CloudApiEnablementRenderer extends DefaultTableCellRenderer {
+
+    private final JButton enableButton =
+        new JButton(GctBundle.message("cloud.apis.management.enable.button"));
+
+    CloudApiEnablementRenderer() {
+      super();
+
+      setOpaque(false);
```
ivanporty: We probably don't need the first `setOpaque` since the default table renderer is not opaque.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiManagementPanel.java
```diff
@@ -0,0 +1,278 @@
...
+      enableButton.setOpaque(false);
+    }
+
+    @Override
+    public Component getTableCellRendererComponent(
+        JTable table, Object value, boolean isSelected, boolean hasFocus, int row, int column) {
+      Couple<Color> colors = UIUtil.getCellColors(table, isSelected, row, column);
```
ivanporty: Since we use `DefaultTableCellRenderer` we actually might not need these colors set explicitly, this should be done for us by implementation.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiManagementPanel.java
etanshaul: yea good point. that class is a lightweigt

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiManagementPanel.java
etanshaul: thanks - i forgot to get rid of that after refactoring

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiManagementPanel.java
```diff
@@ -0,0 +1,286 @@
...
+      boolean isEnabled;
+      if (value instanceof String) {
+        isEnabled = Boolean.parseBoolean((String) value);
+      } else {
+        isEnabled = (Boolean) value;
+      }
+
+      if (isEnabled) {
+
+        component.setBorder(IdeBorderFactory.createEmptyBorder());
```
ivanporty: Probably we can try the same idea as for another renderer - pass `isSelected` and `hasFocus` as always false and remove the explicit border call.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
```diff
@@ -158,6 +158,7 @@ private void createUIComponents() {
     CloudLibraryTable(List<CloudLibrary> libraries) {
       super(new CloudLibraryTableModel(libraries));
 
+      setSelectionMode(ListSelectionModel.SINGLE_SELECTION);
```
ivanporty: Double line it seems?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiManagementPanel.java
etanshaul: actually the main reason is that I am creating two of these api tables which means that with an `ApisTable` implementation I can then set the renderer only once (encapsulated in the table). wdyt?

ivanporty: Makes sense then, thanks

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
etanshaul: oops. fixed

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelector.java
ivanporty: It still feels a bit too open - project selector calling into dialog is internal implementation. The only other idea I might have is keep the listener on project selector as a field and provide `@VisibleForTesting` access to it.
or, even simpler, why don't we just directly call `showApiEnablementUi()` instead - that's the method that gets called on project selection change, correct?

etanshaul: calling showApiEnablementUi directly wouldn't be much of a test since the test in question is just asserting that the fields set in showApiEnablementUi are actually set. The idea is to test that when a cloud project is correctly set that the panel UI is correctly initialized.

The way I justified exposing showDialog in my head was that I could sort of imagine a UX where we'd want to automatically open up the dialog for the user. But I don't really see that happening.

I could see exposing the listeners instead.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryRenderer.java
etanshaul: Actually CloudLibraryRenderer is used in two contexts, one of which requires selection (the first step of the wizard). So I don't think your strategy will work.

etanshaul:
@ivanporty your point about the confusion between client libraries and cloud api management really got me thinking.

I think we can radically simplify things and also do more automatically for the user. What do you think of this:

screen shot 2017-12-21 at 12 57 52 pm

Notice the new mocked up checkbox in the description field. The idea here is:
- This will be checked by default for APIs corresponding to the libraries the user chooses to add to the project. Otherwise it will either be disabled or hidden.
- The question mark will show some more info explaining why GCP requires APIs to be enabled possibly with a link to the docs
- If checked, it will automatically enable the Cloud API (if its not already enabled). 
- If we do something like this I need to think of how to integrate the service account flow here

etanshaul:
@ivanporty @patflynn sorry for the change of direction here..but let me know what you think of the above

ivanporty:
Yes, from my point of view (i.e. as a relatively unexperienced user of Google Cloud APIs) this looks much easier. The defaults are aligned - we enable API in both places we select. If this has billing consequences we can confirm that in final confirmation message, correct?

It essentially the same as just adding another checkbox into the table but then we will need to show the header and appropriate column name and tooltip - then we can only show 'Enable API on Cloud' checkbox when Enable for module is on and otherwise column is empty. I think both ways are better than original.

etanshaul:
- Yes, we can have billing messages as appropriate
- I prefer nesting the checkbox into the details pane as in my mock as it simplifies the table (plus the interaction of disabling the API enable checkbox would be weird if it was in a table column). Also this will naturally lend itself to adding service account controls here too.
- Yes, it is totally possible to have the API enabled on GCP while the user hasn't yet added the client library (in fact this will always happen for certain APIs that are enabled by default in certain envs). This is not a problem in my mind and we don't really need to expose the nuances to the user - checking the box will enable it (if not yet enabled). I can improve the wording to make it clear.

etanshaul:
closing in favor of a different approach (as I outlined in the last couple of comments). New PR coming shortly

PR #1800, opened 2017-12-22
Update cloud debug project selector, added project number.

ivanporty:
debug-no-project

debug-project-attach

debug-config-editor

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugRunConfiguration.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugRunConfiguration.java
@@ -19,7 +19,6 @@ package com.google.cloud.tools.intellij.debugger;
 import com.google.api.client.repackaged.com.google.common.base.Strings;
 import com.google.cloud.tools.intellij.analytics.UsageTrackerProvider;
 import com.google.cloud.tools.intellij.util.GctTracking;
-import com.intellij.execution.ExecutionException;
 import com.intellij.execution.Executor;
 import com.intellij.execution.RunManager;
 import com.intellij.execution.configurations.ConfigurationFactory;
@@ -138,7 +137,7 @@ public class CloudDebugRunConfiguration extends LocatableConfigurationBase
   /**
    * Returns either cached state (if we were previously watching this state in this or the last IDE
    * session) Or it returns a partially valid state, which will later be filled in by the {@link
-   * com.google.gct.idea.debugger.CloudDebuggerRunner}
+   * CloudDebuggerRunner}
    *
    * @param executor the execution mode selected by the user (run, debug, profile etc.)
    * @param environment the environment object containing additional settings for executing the
@@ -147,8 +146,7 @@ public class CloudDebugRunConfiguration extends LocatableConfigurationBase
   @Nullable
   @Override
   public RunProfileState getState(
-      @NotNull Executor executor, @NotNull ExecutionEnvironment environment)
-      throws ExecutionException {
+      @NotNull Executor executor, @NotNull ExecutionEnvironment environment) {
     if (processState == null) {
       return new CloudDebugProcessState(null, null, cloudProjectName, null, getProject());
     }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugSettingsEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugSettingsEditor.java
@@ -20,6 +20,7 @@ import com.google.cloud.tools.intellij.debugger.ui.CloudDebugRunConfigurationPan
 import com.google.cloud.tools.intellij.login.CredentialedUser;
 import com.google.cloud.tools.intellij.login.Services;
 import com.google.cloud.tools.intellij.project.CloudProject;
+import com.google.common.base.Strings;
 import com.intellij.openapi.options.SettingsEditor;
 import java.util.Optional;
 import javax.swing.JComponent;
@@ -35,11 +36,21 @@ public class CloudDebugSettingsEditor extends SettingsEditor<CloudDebugRunConfig
 
   CloudDebugSettingsEditor() {
     settingsPanel = new CloudDebugRunConfigurationPanel();
+    settingsPanel
+        .getProjectSelector()
+        .addProjectSelectionListener(
+            (selectedProject) -> {
+              // settings editor does not see all the changes by default, use explicit notification.
+              settingsPanel.triggerValidation();
+            });
   }
 
   @Override
   protected void applyEditorTo(@NotNull CloudDebugRunConfiguration runConfiguration) {
-    runConfiguration.setCloudProjectName(settingsPanel.getCloudProject().projectId());
+    runConfiguration.setCloudProjectName(
+        Optional.ofNullable(settingsPanel.getProjectSelector().getSelectedProject())
+            .map(CloudProject::projectId)
+            .orElse(""));
   }
 
   @NotNull
@@ -50,13 +61,20 @@ public class CloudDebugSettingsEditor extends SettingsEditor<CloudDebugRunConfig
 
   @Override
   protected void resetEditorFrom(@NotNull CloudDebugRunConfiguration runConfiguration) {
-    settingsPanel.setCloudProject(
-        CloudProject.create(
-            runConfiguration.getCloudProjectName(),
-            runConfiguration.getCloudProjectName(),
-            null,
-            Optional.ofNullable(Services.getLoginService().getActiveUser())
-                .map(CredentialedUser::getEmail)
-                .orElse("")));
+    String projectId = runConfiguration.getCloudProjectName();
+    if (Strings.isNullOrEmpty(projectId)) {
+      settingsPanel.getProjectSelector().setSelectedProject(null);
+    } else {
+      settingsPanel
+          .getProjectSelector()
+          .setSelectedProject(
+              CloudProject.create(
+                  projectId,
+                  projectId,
+                  null,
+                  Optional.ofNullable(Services.getLoginService().getActiveUser())
+                      .map(CredentialedUser::getEmail)
+                      .orElse("")));
+    }
   }
 }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugSettingsEditor.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugSettingsEditor.java
@@ -17,8 +17,11 @@
 package com.google.cloud.tools.intellij.debugger;
 
 import com.google.cloud.tools.intellij.debugger.ui.CloudDebugRunConfigurationPanel;
-import com.intellij.openapi.options.ConfigurationException;
+import com.google.cloud.tools.intellij.login.CredentialedUser;
+import com.google.cloud.tools.intellij.login.Services;
+import com.google.cloud.tools.intellij.project.CloudProject;
 import com.intellij.openapi.options.SettingsEditor;
+import java.util.Optional;
 import javax.swing.JComponent;
 import org.jetbrains.annotations.NotNull;
 
@@ -30,14 +33,13 @@ public class CloudDebugSettingsEditor extends SettingsEditor<CloudDebugRunConfig
 
   private final CloudDebugRunConfigurationPanel settingsPanel;
 
-  public CloudDebugSettingsEditor() {
+  CloudDebugSettingsEditor() {
     settingsPanel = new CloudDebugRunConfigurationPanel();
   }
 
   @Override
-  protected void applyEditorTo(CloudDebugRunConfiguration runConfiguration)
-      throws ConfigurationException {
-    runConfiguration.setCloudProjectName(settingsPanel.getProjectName());
+  protected void applyEditorTo(@NotNull CloudDebugRunConfiguration runConfiguration) {
+    runConfiguration.setCloudProjectName(settingsPanel.getCloudProject().projectId());
   }
 
   @NotNull
@@ -47,7 +49,14 @@ public class CloudDebugSettingsEditor extends SettingsEditor<CloudDebugRunConfig
   }
 
   @Override
-  protected void resetEditorFrom(CloudDebugRunConfiguration runConfiguration) {
-    settingsPanel.setProjectName(runConfiguration.getCloudProjectName());
+  protected void resetEditorFrom(@NotNull CloudDebugRunConfiguration runConfiguration) {
+    settingsPanel.setCloudProject(
+        CloudProject.create(
+            runConfiguration.getCloudProjectName(),
+            runConfiguration.getCloudProjectName(),
+            null,
+            Optional.ofNullable(Services.getLoginService().getActiveUser())
+                .map(CredentialedUser::getEmail)
+                .orElse("")));
   }
 }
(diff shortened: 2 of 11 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugSettingsEditor.java
```diff
@@ -47,7 +60,21 @@ protected JComponent createEditor() {
   }
 
   @Override
-  protected void resetEditorFrom(CloudDebugRunConfiguration runConfiguration) {
-    settingsPanel.setProjectName(runConfiguration.getCloudProjectName());
+  protected void resetEditorFrom(@NotNull CloudDebugRunConfiguration runConfiguration) {
+    String projectId = runConfiguration.getCloudProjectName();
```
ivanporty: Previous code was setting *just project ID* (disguised as project name again in code) and never set user name (email) in `CloudDebugRunConfiguration`. As far as I understand, previous ProjectSelector would load all projects anyway and would search for any project of any user with matching ID? We don't do it anymore and load on demand only, plus match projects strictly per user account, should we add user name into `CloudDebugRunConfiguration` as well?

etanshaul: are you basically asking if IDs are unique across users? if so, the answer is yes. project IDs are unique across all of GCP as far as I understand it

ivanporty: Yes, there are a few places where both ID and user name is stored, and here only project ID is stored. It looks like the code assumed it is enough to find project match later, but this needs loading of all projects for all available accounts. Do you think it makes sense to add project name and user name in cloud debug configuration as well to avoid re-loading all projects?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugRunConfiguration.java
```diff
@@ -147,8 +146,7 @@ public void setProcessState(@Nullable CloudDebugProcessState processState) {
   @Nullable
   @Override
   public RunProfileState getState(
-      @NotNull Executor executor, @NotNull ExecutionEnvironment environment)
-      throws ExecutionException {
+      @NotNull Executor executor, @NotNull ExecutionEnvironment environment) {
```
etanshaul: just curious - why did you get rid of the `throws`?

etanshaul:
what do you mean by "adding project number" in this PR? Are screenshots that help illustrate this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugRunConfiguration.java
ivanporty: This method does not throw any checked exceptions (probably it used to?).

ivanporty:
Project number added into `CloudProject` and `ProjectSelector` now provides it when selection is made - it's internal, nothing in UI is affected.

etanshaul:
Maybe I'm being dumb, but what is project number and what is it used for internally? I understand that we have a project ID and a project name, but number is eluding me.

ivanporty:
Yeah, that surprised me as well - apparently this is a unique long number for each project. Cloud debugger uses it to attach I believe, not project ID/user name combination.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/CloudProject.java
```diff
@@ -16,18 +16,29 @@
...
+  public static CloudProject create(
+      String projectName, String projectId, Long projectNumber, String googleUsername) {
+    return new AutoValue_CloudProject(projectName, projectId, projectNumber, googleUsername);
   }
 
   public abstract String projectName();
 
   public abstract String projectId();
 
+  /** See {@link Project#getProjectNumber()} */
```
etanshaul: nit: period at the end; and maybe add similar javadoc to the other methods in `CloudProject` for consistency

etanshaul:
ah. I see. I had completely forgotten about that

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugSettingsEditor.java
```diff
@@ -30,14 +34,23 @@
...
+            });
   }
 
   @Override
-  protected void applyEditorTo(CloudDebugRunConfiguration runConfiguration)
-      throws ConfigurationException {
-    runConfiguration.setCloudProjectName(settingsPanel.getProjectName());
+  protected void applyEditorTo(@NotNull CloudDebugRunConfiguration runConfiguration) {
+    runConfiguration.setCloudProjectName(
+        Optional.ofNullable(settingsPanel.getProjectSelector().getSelectedProject())
```
etanshaul: do you think it would be better separation to have the settings panel return the CloudProject instead of the selector widget itself?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/CloudProject.java
ivanporty: sounds good.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugSettingsEditor.java
ivanporty: yes, probably cleaner, I'll change that.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugSettingsEditor.java
etanshaul: Previously, it would have to reload all the projects too right? If so I would avoid any optimization here to keep the PR concise (we could do it separately if we wanted).

etanshaul: Also any updates to the persisted configuration requires consideration of backwards compatibility with already stored configs. I definitely wouldn't want to go down that path in this PR since its meant to just be updating the project selector to the new widgets.

etanshaul: I'd add a comment / todo on this line just to make it clear why you are setting projectId to projectName

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugSettingsEditor.java
```diff
@@ -47,7 +60,21 @@ protected JComponent createEditor() {
...
+    String projectId = runConfiguration.getCloudProjectName();
+    if (Strings.isNullOrEmpty(projectId)) {
+      settingsPanel.getProjectSelector().setSelectedProject(null);
+    } else {
+      settingsPanel
+          .getProjectSelector()
+          .setSelectedProject(
+              CloudProject.create(
+                  projectId,
+                  projectId,
```
etanshaul: also a comment / todo here on why you are setting the projectId twice

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugSettingsEditor.java
```diff
@@ -47,7 +60,21 @@ protected JComponent createEditor() {
...
+                  null,
+                  Optional.ofNullable(Services.getLoginService().getActiveUser())
+                      .map(CredentialedUser::getEmail)
+                      .orElse("")));
```
etanshaul: is it valid to call `create` with an empty string for the user name?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/ui/CloudAttachDialog.java
```diff
@@ -397,7 +389,7 @@ private boolean stash() {
       GitHandlerUtil.doSynchronously(
           handler, GitBundle.getString("stashing.title"), handler.printableCommandLine());
     } finally {
-      DvcsUtil.workingTreeChangeFinished(project, token);
+      token.finish();
```
etanshaul: mind explaining this change just so I understand the reasons / differences?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/ui/CloudDebugRunConfigurationPanel.java
```diff
@@ -43,11 +46,16 @@ public JPanel getMainPanel() {
...
 
-  public void setProjectName(String name) {
-    elysiumProjectId.setText(name);
+  public void triggerValidation() {
+    hiddenValidationTrigger.doClick();
+  }
+
+  private void createUIComponents() {
+    hiddenValidationTrigger = new JBCheckBox();
+    hiddenValidationTrigger.setVisible(false);
```
etanshaul: Not for this PR by any means, but I wonder if instead of duplicating this "hack" everywhere, we can create a base panel class that encapsulates this..like a `HiddenValidationTriggeringPanel` or something..

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugSettingsEditor.java
ivanporty: Sure, there are no changes or optimizations in this PR, we can still use the same project ID it already holds, it just affects that project selector won't show the project name.

ivanporty: agree.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/ui/CloudAttachDialog.java
ivanporty: Sure, it's an easy one - IDEA looks through the docs and sources and suggests removing deprecated code - this method has been deprecated and recommends calling token.finish() instead in Javadoc and sources. It actually calls `token.finish()` in deprecated `workingTreeChangeFinished`. I know it's not a change aligned with this PR, just minor clean up in the same code, what do you think?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/ui/CloudDebugRunConfigurationPanel.java
ivanporty: Probably or even better make that interface that `SettingsEditor` is supposed to recognize work and implement it in `ProjectSelector`. I'll spend some time debugging it and see if it can actually work.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugSettingsEditor.java
ivanporty: It is a valid case for `ProjectSelector` - when user name is empty, it will show `Select Google Cloud project` prompt assuming project couldn't be found or bound to logged in user.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/ui/CloudAttachDialog.java
etanshaul: I'm fine with it. Probably worth testing out just to be certain they didn't break anything.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/ui/CloudDebugRunConfigurationPanel.java
etanshaul: Yea ideally that would be the solution.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/ui/ProjectDebuggeeBinding.java
```diff
@@ -128,26 +126,44 @@ public CloudDebugProcessState getInputState() {
...
+      }
+      projectSelector.setSelectedProject(
+          CloudProject.create(
+              this.inputState.getProjectName(), // TODO(ivanporty) add separate project name/ID
+              this.inputState.getProjectName(),
+              projectNumber,
+              this.inputState.getUserEmail()));
+
+      // update the state here as well
+      refreshDebugTargetList(projectSelector.getSelectedProject());
```
etanshaul: just wondering why this needs to happen now, but it wasn't needed previously?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/ui/CloudAttachDialog.java
etanshaul: ah nm. looks like the deprecated method was just doing this anyway

etanshaul: which you did already say in your original post if I had read it more carefully :)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/ui/ProjectDebuggeeBinding.java
ivanporty: That's again that tricky `ProjectSelector` thing - old `ProjectSelector` had a text field which fired text changed events all the time - so when this code called `setText()`, `ProjectSelector` would immediately fire text changed event, and this class would receive event and update its state as if new project has been set. We don't generate events when `ProjectSelector` is updated with project explicitly so we update the panel explicitly which makes it easier to read.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/ui/CloudAttachDialog.java
ivanporty: My fault, TLDR applies! :)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/ui/ProjectDebuggeeBinding.java
etanshaul: I see. Do you think it would make sense to update the project selector so that it fired an event when the project selector is updated programmatically via setSelectedProject?

etanshaul: If you do think it makes sense, I think it would also make certain unit tests easier (I recall running into this previously).

ivanporty: Currently the use case is (1) we set the project so project selector updates UI, and we already know what this project is. (2) we listen to project selection change made by user. 
If we fire event when we call `setSelectedProject` it means we would have to disconnect listeners in some cases - otherwise it could be a loop of 'project changed' -> 'update' -> 'change project' -> listener again etc.

etanshaul: ok sounds good. nevermind then.

PR #1802, opened 2017-12-27
Adds API enablement in the IDE to the cloud library management features

### LINKED ISSUE #1807: Enable GCP APIs from the IDE

etanshaul:
fixes #1807

@ivanporty @patflynn 

This greatly simplifies the approach of integrating API enablement in the cloud library management workflows. Instead of a new panel entirely in a wizard flow, API enablement is integrated into the cloud library screen:

Select cloud libraries, and configure the API management section. (Service account related things will appear here too in the next PR):
image

Tooltip info on API enablement:
image

Select a cloud project on which to enable this APIs:
image
image

APIs are being enabled on GCP:
image

Notification to the user:
image


ivanporty:
To understand a flow a bit better, when we hit 'OK' in 'Add Cloud Libraries' dialog, a user is asked for a project selection in a new dialog, correct? Is this only when new APIs need to be enabled on GCP, and if not, project selection is not required? What happens if someone decides he doesn't want to select a project now and hits 'Cancel' - does it go back to 'Add Cloud Libraries' again?

etanshaul:
> To understand a flow a bit better, when we hit 'OK' in 'Add Cloud Libraries' dialog, a user is asked for a project selection in a new dialog, correct? Is this only when new APIs need to be enabled on GCP, and if not, project selection is not required?

Correct - if there are no APIs to be enabled (and in the future - no service account modifications to be made) then the user will not be shown this dialog

> What happens if someone decides he doesn't want to select a project now and hits 'Cancel' - does it go back to 'Add Cloud Libraries' again?

If the user hits cancel. Then currently it will close everything. I was thinking about this too. Do you think it would be better to send them back to the 'Add cloud libs' dialog?

ivanporty:
It feels it would be a bit more natural to get to the previous step instead of closing everything - maybe a usr just realized API enabled per project and needs to double check what he selected - if we keep the original 'Add Cloud Libraries' with the current state, it might be a bit easier.

etanshaul:
agree. I'll update it to that behavior.

ivanporty:
Oh, by the way, won't we also want to show confirmation we added libraries as dependencies after dialog is done? I.e. another notification, same style as API on GCP, with module name and library names.

etanshaul:
you mean after the libs are added to the POM? that is also already shown as a bubble (and in the event log). I jsut didn't show the screenshot here.

ivanporty:
That's great, thanks!

ivanporty:
Another thing - library dependencies are added to the module and stays with it, but we may choose an arbitrary project when we run/deploy the actual service/app. We may add a warning or validation check for deployment later, something like 'Module XYZ uses Cloud ABC API, but selected project FOO does not have this API enabled on Google Cloud.' And then maybe suggest immediately open add cloud libraries dialog to enable missing APIs.

etanshaul:
Yes. I agree. I think @patflynn brought up this point too earlier. In the deployment config it would be a good enhancement to add this type of notification / integration into the API management.

etanshaul:
I added some tooltip help text to explain what this enablement thing is all about. See the updated screenshots

ivanporty:
Awesome, crystal clear!

patflynn:
I like it.  Could we also inline the API enablement project selection into
the the same panel?

etanshaul:
@patflynn I thought about that initially. The trouble there is that we don't want a project selector to show for each library. Even if the selection is remembered it feels like we are giving the option of selecting a different project per library, which we are not.

etanshaul:
I guess we could show the project selector at the top somewhere (outside of the details panel). I'd just need to make it clear that a project selection is not required unless the user is enabling an API or updating service accounts.

etanshaul:
I think I prefer the dialog so I'll stay with that for now. But we can give it a closer look when I demo this.

etanshaul:
@ivanporty I updated the flow to keep the library dialog open while the user is selecting a cloud project for API management. If the user then cancels, he/she will be returned to the library dialog. Selecting a cloud project and clicking 'ok' will close all dialogs and perform all actions.

ivanporty:
Nice, it feels transparent and easy to see what's happening to the project and on the cloud/server side!

For future maybe, I think in cases like this, where we call project selection explicitly and we actually need to make a decision right away that depends on project selected/not selected, we might skip the ProjectSelector and call project selection dialog directly - so instead of additional dialog we can pop up actual account/project list, it could be even more straightforward.

etanshaul:
@ivanporty I agree and I started down the path of refactoring the selector to enable this. But then I realized that in the dialog that pops up I will probably have some additional information including service account info etc.

ivanporty:
Cool, we may extend that dialog a bit and provide a few hooks later - like change title, provide additional labels or maybe change button names, so it can be tailored for a specific need.

patflynn:
+1 fine with dialog.

etanshaul:
@ivanporty, @patflynn, this is ready for code review now

google-cloud-tools-plugin/build.gradle
```diff
@@ -61,6 +61,7 @@ dependencies {
         exclude group: 'com.google.guava', module: 'guava'
         exclude group: 'com.google.oauth-client', module: 'google-oauth-client'
     }
+    compile 'com.google.apis:google-api-services-servicemanagement:v1-rev374-1.22.0'
```
ivanporty: Just wondering, is it google-api-services-* practice to specify very exact revisions and versions in dependencies?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
```diff
@@ -67,12 +71,25 @@
...
     this.libraries = libraries;
     this.project = project;
+
+    apiManagementMap =
+        libraries
+            .stream()
+            .collect(
+                Collectors.toMap(
+                    Function.identity(),
+                    lib -> new CloudApiManagementSpec(SHOULD_ENABLE_API_DEFAULT)));
```
ivanporty: Would you mind elaborating a bit on when we might need not to enable an API by default? As I understand, this is for cloud side enable?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
```diff
@@ -147,7 +181,7 @@ private void createUIComponents() {
                 int selectedIndex = model.getMinSelectionIndex();
                 CloudLibrary library =
                     (CloudLibrary) cloudLibrariesTable.getModel().getValueAt(selectedIndex, 0);
-                detailsPanel.setCloudLibrary(library);
+                detailsPanel.setCloudLibrary(library, apiManagementMap.get(library));
```
ivanporty: So this essentially sets a library and its "enable on cloud" state to details panel, and `apiManagementMap` is always pre-populated with the set of libraries and they match, so no chance to get null here, correct? This maybe a candidate for immutable map?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
```diff
@@ -67,12 +71,25 @@
   private JLabel modulesLabel;
   private ModulesComboBox modulesComboBox;
 
+  private final Map<CloudLibrary, CloudApiManagementSpec> apiManagementMap;
```
ivanporty: Feels like immutable map after filling with initial state?

google-cloud-tools-plugin/build.gradle
etanshaul: I guess so. I don't really have more info on that other than this is the way they specify the versions of their binaries.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
etanshaul: I think that 'enabled' is a sensible default. But I guess its conceivable that we could determine (after seeing user patterns) that it makes more sense to have the user opt-into having the API enabled (e.g. due to billing ramifications etc.).

Was your point just about exposing that boolean as a member variable?

ivanporty: Ah I see. I was just wondering when we actually would want to make the change in that constant, it makes sense.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiDetailsPanel.java
```diff
@@ -187,6 +211,9 @@ private void updateUI() {
                 linksTextPane.setText(joinLinks(links));
               });
     }
+
+    apiManagementPanel.setVisible(true);
```
ivanporty: Is it hidden by default?

etanshaul: it is. that is so that it is not visible when no library is selected in the table. I think a better way would be to make the entire details panel itself hidden until a library is selected. i'll update to that. wdyt?

ivanporty: Agree, it will be more consistent.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
etanshaul: correct. I made it immutable. thanks

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -0,0 +1,131 @@
...
+   *
+   * @param libraries the set of {@link CloudLibrary CloudLibraries} to enable on GCP
+   * @param cloudProject the {@link CloudProject CloudProject} on which to enable the APIs
+   */
+  static void enableApis(Set<CloudLibrary> libraries, CloudProject cloudProject) {
+    Optional<CredentialedUser> user =
+        Services.getLoginService().getLoggedInUser(cloudProject.googleUsername());
+
+    if (!user.isPresent()) {
+      LOG.warn("Cannot enable APIs: logged in user not found.");
```
ivanporty: This error is unlikely to happen, but maybe show this is in error balloon as well for consistency with other errors here?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
etanshaul: done (the instance created in the constructor is now immutable)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
etanshaul: hmm. I actually don't think this error should ever happen unless something really goes wrong. I think changing it to log.error instead so that the user sees an ide plugin error and it gets reported to us makes sense. wdyt?

ivanporty: sounds good, it does look like a hypothetical error only.

etanshaul: changed

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
ivanporty: Maybe change interface to explicit `ImmutableMap` so it's easy to see/catch attempts to modify?

etanshaul: unless you feel strongly about it, I think its ok: if I were to change it I'd probably have to update all the getters too, and also update the `libraries` list too for consistency. I'm not sure if we need to expose the implementation as part of the api.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanelTest.java
```diff
@@ -322,6 +325,90 @@ public void getSelectedLibraries_withSomeSelected_returnsLibraries() {
...
+  public void getApisToEnable_shouldEnableSelectedLibraryByDefault() {
+    CloudLibrary library = LIBRARY_1.toCloudLibrary();
+
+    GoogleCloudApiSelectorPanel panel =
+        new GoogleCloudApiSelectorPanel(ImmutableList.of(library), testFixture.getProject());
+    JTable table = panel.getCloudLibrariesTable();
+    checkCheckbox(table, 0);
+
+    Set<CloudLibrary> apisToEnable = panel.getApisToEnable();
+    assertThat(apisToEnable).containsExactly(library);
```
ivanporty: This test depends on value of that `SHOULD_ENABLE_API_DEFAULT` constant, correct?

etanshaul: yes

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanelTest.java
```diff
@@ -322,6 +325,90 @@ public void getSelectedLibraries_withSomeSelected_returnsLibraries() {
...
+    CloudLibrary library2 = LIBRARY_2.toCloudLibrary();
+
+    GoogleCloudApiSelectorPanel panel =
+        new GoogleCloudApiSelectorPanel(
+            ImmutableList.of(library1, library2), testFixture.getProject());
+
+    Map<CloudLibrary, CloudApiManagementSpec> apiManagementMap = panel.getApiManagementMap();
+
+    panel.getDetailsPanel().setCloudLibrary(library1, apiManagementMap.get(library1));
+    panel.getDetailsPanel().getEnableApiCheckbox().setSelected(true);
```
ivanporty: BTW, regarding this test case, how about disabling that checkbox at all if library is not enabled? Otherwise a user might get a feeling he is able to enable Cloud API on GCP side without adding libraries to the module and project - the checkbox is still available.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiDetailsPanel.java
etanshaul: updated

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
ivanporty: That's fine with me as well.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanelTest.java
etanshaul: good point. thank you for reminding me of this.

etanshaul: updated for this behavior: https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1802/commits/42898ba73c9eabfab7c92ead3734e8dcc3bf04fe

etanshaul:
@ivanporty I'm working on fixing the test leakages now. I pushed a couple changes from the review since you last approved if you wanted to look again.

ivanporty:
@etanshaul Sure, I think that was the reason I had to craft a dialog as a wrapper in `ProjectSelectionDialog` - `DialogWrapper` was leaking in each test used and I was not yet familiar with IDEA test setup and project tracking enough for a fix it. I'll take a look at the new changes

etanshaul:
ah interesting. i'll see if I can figure out how to get it working as a DIalogWrapper

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
```diff
@@ -184,17 +184,29 @@ private void createUIComponents() {
...
               }
             });
+    cloudLibrariesTable
+        .getModel()
+        .addTableModelListener(
+            e -> {
+              int row = e.getFirstRow();
+              int column = e.getColumn();
+              TableModel model = (TableModel) e.getSource();
+              boolean isSelected = (boolean) model.getValueAt(row, column);
```
ivanporty: It feels a bit over complicated here. We have a specific column that we need to check for a boolean value, and we can check for it on currently selected row. Getting first changed row and changed column is very generic and it feels the code might check for many more combinations, plus it explicitly casts any value from table to boolean. How about just getValueAt(getSelectedRow(), [apiSelectedColumn])?

Also, maybe we can use less generic name for `isSelected`? It feels very UI specific.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanelTest.java
ivanporty: If we updated checkboxes to be disabled, do we still need to set them selected here? Could we also test they are disabled in this case?

etanshaul: ah thanks. I forgot to update these tests after adding the disabled feature

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
etanshaul: updated

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanelTest.java
etanshaul: is it truly redundant to set the checkbox to selected even if its disabled? This test is still testing that panel returns no APIs to enable when the checkboxes are disabled+checked.

etanshaul: (I agree that I still need to add assertions to confirm the enabled/disabled states)

ivanporty: Agree, selection makes tests more thorough

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
```diff
@@ -183,21 +185,26 @@ private void createUIComponents() {
...
-                    (CloudLibrary) cloudLibrariesTable.getModel().getValueAt(selectedIndex, 0);
-                boolean isSelected =
-                    (boolean) cloudLibrariesTable.getModel().getValueAt(selectedIndex, 1);
+                    (CloudLibrary)
+                        cloudLibrariesTable.getModel().getValueAt(selectedIndex, CLOUD_LIBRARY_COL);
+                boolean addLibrary =
+                    (boolean)
+                        cloudLibrariesTable
+                            .getModel()
+                            .getValueAt(selectedIndex, CLOUD_LIBRARY_SELECT_COL);
```
ivanporty: Looks like an identical code here (190:194) and in lines 204:207, could we use a common method instead, something like `updateManagementUI()`?

etanshaul:
@ivanporty addressed the last few comments I think

ivanporty:
Awesome, looks great!

etanshaul:
@ivanporty to unblock progress on API work I removed the test of the DialogWrapper instance that was leading to leaked projects. The test itself was very light anyway, but once I figure out the underlying cause and a solution we can reintroduce this.

PR #1809, opened 2018-01-04
remove old account plugin modules compiler config

patflynn:
not sure why only my os x workspace does this update, but it appears to be valid since those modules no longer exist.


etanshaul:
Strange, my ide never complained about it even on OSX. But it’s valid like you said.

PR #1810, opened 2018-01-04
Remove IJ binary gcs cache - test

### LINKED ISSUE #1798: Gradle sync re-downloads idea distribution every time

etanshaul:
Seems to improve build speed. Testing out on Kokoro.

Also fixes #1798

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,7 +61,6 @@ subprojects {
         version = ideaVersion
         updateSinceUntilBuild = false
         downloadSources = true
-        intellijRepo = intellijRepoUrl
     }
 
     configurations.runtime {
--- a/gradle.properties
+++ b/gradle.properties
@@ -16,7 +16,6 @@
 
 ideaEdition = IU
 ideaVersion = 2017.2
-intellijRepoUrl = https://storage.googleapis.com/cloud-tools-for-java-team-kokoro-build-cache/idea-distributions
 javaVersion = 1.8
 ijPluginRepoChannel = alpha
 version = 17.12.3-SNAPSHOT


patflynn:
it's odd that this would cause the re-downloads all of a sudden.   The problem with not using our GCS cache is less about performance and more about build flakiness.  Using the default JB binary repo causes relatively frequent CI failures due to failed downloads.

etanshaul:
ah ok. I'll see if I can pinpoint exactly the cause of the re-download

patflynn:
another option is to manually copy in our build cache using gsutil at build start. Should speed things up significantly but may require more maintenance

etanshaul:
yeah that would speed things up. But i'd still like to figure why swapping out the intellijRepoUrl from:

`https://www.jetbrains.com/intellij-repository` (the default value)
to
`https://storage.googleapis.com/cloud-tools-for-java-team-kokoro-build-cache/`

leads to the binary not being cached

etanshaul:
Possible reason: I'm seeing several 403s in the logs downloading the artifact metadata (i.e. `ideaIU-2017.2.pom`)

etanshaul:
I quickly revisited this issue this morning and noticed that the slowdown (redownload of the artifact metadata each time) happened ever since we introduced the mirror repo.

@patflynn I know its not a full solution, but to speed up local development, wdyt about defaulting the IntelliJ repo to the default JetBrains one, and then only in the relevant Kokoro environments setting it to our GCS mirror?

patflynn:
works for me.

etanshaul:
closing. will use the other approach mentioned above

Issue #1808, opened 2018-01-03
Manage service accounts from the IDE in the client library / GCP API experience


etanshaul:

etanshaul:
complete

PR #1813, opened 2018-01-05
Moves project selection to the parent library selector panel

etanshaul:
In support of service account work I'm doing (#1808)

**What does this PR do?**
- Moves the project selector out of the popup dialog as implemented in https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1802 and into the parent client library panel
- Repurposes the popup dialog into a confirmation dialog - confirming the GCP API changes about to be made on the user's project

**Why?**

I realized to support service accounts as I intended, we need to know the target cloud project earlier. I intend to add service account features into the right details panel (see the 2nd screenshot below) and these require knowledge of the cloud project.

**Screenshots**

library not selected and cloud project not chosen:
image

selections made:
image

After clicking 'OK' on the library selection panel, with APIs to enable on GCP:
image

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.tools.intellij.apis;
 
+import com.google.cloud.tools.intellij.project.CloudProject;
 import com.google.cloud.tools.intellij.ui.GoogleCloudToolsIcons;
 import com.google.cloud.tools.intellij.util.GctBundle;
 import com.google.cloud.tools.libraries.CloudLibraries;
@@ -96,6 +97,10 @@ public final class AddCloudLibrariesWizardAction extends DumbAwareAction {
       return selectClientLibrariesStep.getSelectedLibraries();
     }
 
+    CloudProject getCloudProject() {
+      return selectClientLibrariesStep.getCloudProject();
+    }
+
     Set<CloudLibrary> getApisToEnable() {
       return selectClientLibrariesStep.getApisToEnable();
     }
@@ -115,24 +120,24 @@ public final class AddCloudLibrariesWizardAction extends DumbAwareAction {
      * Overrides {@link DialogWrapper#doOKAction()} to first check if there are any APIs to enable
      * on GCP.
      *
-     * <p>If so, the {@link CloudApiManagementDialog} is opened allowing the user to select a cloud
-     * project and confirm. If the user cancels, the user is returned to this parent dialog.
+     * <p>If so, the {@link CloudApiManagementConfirmationDialog} is opened confirming the API
+     * changes to be made. If the user cancels, the user is returned to this parent dialog.
      * Otherwise, it is closed and the default {@link DialogWrapper#doOKAction()} is invoked.
      */
     @Override
     protected void doOKAction() {
+      CloudProject cloudProject = getCloudProject();
       Set<CloudLibrary> apisToEnable = getApisToEnable();
 
-      if (!apisToEnable.isEmpty()) {
-        CloudApiManagementDialog managementDialog = new CloudApiManagementDialog(project);
+      if (cloudProject != null && !apisToEnable.isEmpty()) {
+        CloudApiManagementConfirmationDialog managementDialog =
+            new CloudApiManagementConfirmationDialog(project, apisToEnable);
         DialogManager.show(managementDialog);
 
         if (managementDialog.isOK()) {
           ProgressManager.getInstance()
               .runProcessWithProgressSynchronously(
-                  () ->
-                      CloudApiManager.enableApis(
-                          apisToEnable, managementDialog.getCloudProject(), project),
+                  () -> CloudApiManager.enableApis(apisToEnable, cloudProject, project),
                   GctBundle.message("cloud.apis.enable.progress.title"),
                   true /*canBeCanceled*/,
                   project);
@@ -175,6 +180,10 @@ public final class AddCloudLibrariesWizardAction extends DumbAwareAction {
     /** Returns the set of selected {@link CloudLibrary CloudLibraries}. */
     Set<CloudLibrary> getSelectedLibraries() {
       return cloudApiSelectorPanel.getSelectedLibraries();
+    }
+
+    CloudProject getCloudProject() {
+      return cloudApiSelectorPanel.getCloudProject();
     }
 
     Set<CloudLibrary> getApisToEnable() {
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
@@ -26,14 +26,17 @@ import com.google.cloud.tools.intellij.resources.GoogleApiClientFactory;
 import com.google.cloud.tools.intellij.ui.GoogleCloudToolsIcons;
 import com.google.cloud.tools.intellij.util.GctBundle;
 import com.google.cloud.tools.libraries.json.CloudLibrary;
-import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
 import com.intellij.notification.Notification;
 import com.intellij.notification.NotificationDisplayType;
 import com.intellij.notification.NotificationGroup;
 import com.intellij.notification.NotificationType;
 import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.progress.ProgressIndicator;
+import com.intellij.openapi.progress.ProgressManager;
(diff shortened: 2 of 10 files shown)


ivanporty:
What do you think about making project selector a bit longer? It might need more space if project id/user email and name are large.

etanshaul:
Yea agreed. I'll update

etanshaul:
wdyt of this? the left/right alignment might be a little weird, but it seems to make things stand out more

image
image

etanshaul:
oh nm - its still not fixed size. let me fix this. sorry

etanshaul:
as discussed offline. we'll go with the above right-aligned look

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
```diff
@@ -115,24 +120,24 @@ protected String getHelpID() {
...
+            new CloudApiManagementConfirmationDialog(project, apisToEnable);
         DialogManager.show(managementDialog);
 
         if (managementDialog.isOK()) {
           ProgressManager.getInstance()
               .runProcessWithProgressSynchronously(
-                  () ->
-                      CloudApiManager.enableApis(
-                          apisToEnable, managementDialog.getCloudProject(), project),
+                  () -> CloudApiManager.enableApis(apisToEnable, cloudProject, project),
```
ivanporty: We call `runProcessWithProgressSynchronously` with `true /*canBeCanceled*/`, but I cannot see anything that is done on cancel, can it truly cancelled as of now?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
```diff
@@ -16,6 +16,7 @@
```
ivanporty: General comment about this class - it starts feeling a bit confusing for me - it's an action, inside of the action is a Wizard, as a nested class, then another nested class is a wizard step, and this steps returns actual UI, which is a separate class.  Could we refactor a bit later - for example, separate action class, remove wizard step and combine wizard and its only step as one simple dialog?

etanshaul: yes. that is a todo I've had. I'll remove the wizard in a separate PR

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesWizardAction.java
etanshaul: let me check how this works. I thought that if canceled, than the process is terminated. But since its not set up as a transaction, this really wouldn't make sense

etanshaul: yea - good catch - canceling doesn't do anything now. I'll look into supporting it.

ivanporty:
I'm fine with proposed UI changes and code change (addressing refactoring in following PR). 
@patflynn, what do you think about UI updates?

etanshaul:
just fyi - I'm still adding in proper progress indication / cancel functionality into the api enable dialog

ivanporty:
Oh cool, I thought it's in the next PR, I'll review those as well.

patflynn:
This is good, and yet something seems a little off to me.  I wonder if
users are really going to understand what that gcp project selector
represents.

etanshaul:
@ivanporty @patflynn sorry I forgot to ping about this. This should be ready for review. I updated the progress indicator to show fraction complete. Also the progress is truly cancelable now: after the user cancels, the APIs that were "skipped" are reported back in the event log.

I'd like to make the other changes from the demo comments in separate PRs to keep things manageable. Let me know if you disagree.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -57,6 +60,9 @@ private CloudApiManager() {}
   /**
    * Enables the supplied set of {@link CloudLibrary CloudLibraries} on GCP.
    *
+   * <p>Configures the {@link ProgressIndicator} to display the progress of the tasks. Also notifies
+   * the user of the success / failure of API enablement via messages on the event log.
+   *
    * @param libraries the set of {@link CloudLibrary CloudLibraries} to enable on GCP
    * @param cloudProject the {@link CloudProject CloudProject} on which to enable the APIs
```
ivanporty: nit: since we provide parameters doc, shall we add `project` parameter short description as well?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -69,67 +75,99 @@ static void enableApis(Set<CloudLibrary> libraries, CloudProject cloudProject, P
...
+
+        setApiEnableProgress(
+            progress,
+            library.getName(),
+            cloudProject.projectName(),
+            (double) i / libraryList.size());
+        enableApi(library, cloudProject, user.get());
+
+        enabledApis.add(library);
+      } catch (IOException e) {
```
ivanporty: Does `enableApi` call use same `IOException` to identify I/O and all other (i.e. auth or billing) errors? Or it's expected this call always succeeds if I/O doesn't fail?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
etanshaul: oops. fixed.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -69,67 +75,99 @@ static void enableApis(Set<CloudLibrary> libraries, CloudProject cloudProject, P
...
+          notifyApiEnableSkipped(Sets.difference(libraries, enabledApis), project);
+          return;
+        }
+
+        setApiEnableProgress(
+            progress,
+            library.getName(),
+            cloudProject.projectName(),
+            (double) i / libraryList.size());
+        enableApi(library, cloudProject, user.get());
```
ivanporty: What could be an average expected time to enable APIs on GCP? It looks like we do this synchronously here, it's probably fine if this is quick, but may be annoying if it takes longer than 10-20 seconds for example. Could we also look at `runProcessWithProgressAsynchronously` in `ProgressManager`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
etanshaul: These use the generic (autogenerated?) apis to manage things on gcp. So yes, I believe that it rolls all exceptions into an IOException.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
etanshaul: from my tests. each api takes about a second or less to enable

I could explore doing this asynchronously in a separate PR though since that may be a better model for this regardless.

ivanporty: sounds good, just to be sure if some user runs into issue and we freeze his IDE for a few minutes

etanshaul: +1. There is always the cancel button escape hatch though, just to point that out. I still agree though.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
ivanporty: So if this happens, all problematic APIs will be shown in error notification and error stacktraces printed in log, correct? And they can either retry from IDE or go to GCP console to fix.

etanshaul: Correct. 

They can retry from the IDE, but there is not yet a good story for this (since currently the enablement is tightly coupled to client library addition). So it won't exactly be very clear how to retry. This goes along with what @elharo brought up - we do plan on having a standalone mechanism for api enablement.

ivanporty:
LGTM, and we could probably add a few enhancements for items we plan for next PRs - i.e. refactor wizard, async progress so it's easier to see. 
Thanks, I think this dialog is pretty awesome!

etanshaul:
thanks for the review. looks like the the windows CI is acting up. will merge once fixed.

PR #1820, opened 2018-01-09
Client library no longer a wizard

### LINKED ISSUE #1817: Client library refactor to not be a wizard
It not longer needs to be a wizard as its only a single step dialog

etanshaul:
fixes #1817 

also went ahead and enabled the action in plugin.xml

### DIFF
--- a/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
+++ b/google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
@@ -218,6 +218,7 @@ cloud.libraries.documentation.link=Documentation
 cloud.libraries.menu.action.text=Add Cloud libraries...
 cloud.libraries.menu.action.description=Select the Cloud libraries to download and include in your application
 cloud.libraries.module.selector.label=Module:
+cloud.libraries.ok.button.text=Add Client Library
 cloud.libraries.source.link=Source
 cloud.libraries.status.label=Status: {0}
 cloud.libraries.version.label=Version: {0}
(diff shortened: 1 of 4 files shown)


google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -218,6 +218,7 @@ cloud.libraries.documentation.link=Documentation
 cloud.libraries.menu.action.text=Add Cloud libraries...
 cloud.libraries.menu.action.description=Select the Cloud libraries to download and include in your application
 cloud.libraries.module.selector.label=Module:
+cloud.libraries.ok.button.text=Add Client Library
```
etanshaul: oh, and I snuck in this button text change too

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesAction.java
```diff
@@ -68,52 +63,53 @@ public void actionPerformed(AnActionEvent e) {
...
 
     private final Project project;
-    private final SelectClientLibrariesStep selectClientLibrariesStep;
-
-    AddCloudLibrariesWizard(Project project) {
-      super(GctBundle.message("cloud.libraries.dialog.title"), project);
+    private final GoogleCloudApiSelectorPanel cloudApiSelectorPanel;
 
-      this.project = project;
+    AddCloudLibrariesDialog(Project project) {
```
ivanporty: This is great, so much easier to read! Could we also consider making `AddCloudLibrariesDialog` its own top-level class, since it does not have any real dependencies on `AddCloudLibrariesAction` and it's pretty large?

etanshaul: yes. good idea.

etanshaul:
@ivanporty done - extract AddCloudLibraryDialog to top level class

ivanporty:
`AddCloudLibrariesDialog` is a neat little class, great, thanks!

PR #1821, opened 2018-01-10
Make ProjectSelector use last selected project by default.

### LINKED ISSUE #1812: Keep the last selected project for project selector
Current project selector is stateless and does not keep the last selected project.
Most of the activities in the GCP plugin require project and might use the same active project user selected last.
Some of them keep the state (deployment, run configuration), however many do not (GCS browser, source control support, API selection etc.)

Let's persist the last selected project in project selector and pre-populate it whenever it's used in UI.

ivanporty:
Saves project selector last selected cloud project, per IDE project.
Fixes #1812

Default behaivoir of the project selector doesn't change.

New methods to support last selection:
`setIdeProject` - to set IDE project which project selector works with.
This has to be done explicitly - I've spent considerable amount of time trying to figure out is it possible to find the project of the "current" window, but there are a lot of caveats in all of them. Similar conclusions from Jetbrains forum.

`loadActiveCloudProject` -  Loads active `CloudProject`, if IDE project has been specified.
This must be called when UI is ready and all listeners are attached, so we cannot figure out this moment automatically without introducing implied behavior and complexity.

Saving is done automatically when selection is made (if IDE project is set).

Setup on Cloud Repo, Clone from Cloud Repo, Browse GCS, Attach to Debugger all use last selected project.
Run/Deployment targets (Std, Flex, Cloud Debug) show last selected if no selection is made before, or otherwise use their own saved project.

Edge case: project is saved, but user is not logged in.
Works: project and account are shown, but UI asks to login to perform actions (i.e. browse buckets or setup repo).
Once logged in, all functionality just works - I think this is pretty good.

All screenshots stay the same, so this PR can be best tried/tested directly in UI, with multiple projects opened in IDEA.

Minor fix: made (...) button focusable and accessible from keyboard for project selector.


etanshaul:
so if the developer forgets to call `setIdeProject` then this functionality will just silently not work?

etanshaul:
also, the setIdeProject is simply so that we can persist these values at the project level correct? If we stored them at the IDE level instead, then it wouldn't be necessary, right?

ivanporty:
Yes, correct. Unless we require `Project` to be always set in constructor - I thought about this but this means all project selectors in IDEA forms have to be custom constructed, quite annoying.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/flexible/AppEngineFlexibleDeploymentEditor.java
```diff
@@ -236,6 +238,11 @@ private void reloadAppYamls(Project project) {
   protected void resetEditorFrom(@NotNull AppEngineDeploymentConfiguration configuration) {
     commonConfig.resetEditorFrom(configuration);
 
+    // if cloud project was not set at all, use active cloud project.
+    if (commonConfig.getProjectSelector().getSelectedProject() == null) {
```
etanshaul: I was kinda thinking that instead of having each consumer check for thie new 'default' project this, that this could be handled internally by the project selector.

i.e. if the project persisted in the specific config by the user is null (or not set), then the project selector itself will be responsible for showing its default project (if one exists).

do you think this is doable / makes sense?

patflynn: ditto.

ivanporty: I think I can just refactor this code to be called less explicitly, i.e. when settings editor is ready, always load active project first, not in `resetEditor` checking for null.

If we load active project automatically in setSelectedProject(null), we must be sure everything is ready - UI, state, listeners so active project will be properly loaded and all listeners triggered. This is a very strong side effect vs. just doing it explicitly when one as owner of UI knows it is ready. WDYT?

ivanporty:
@etanshaul @patflynn Should be ready for review. I refactored ProjectSelector to enforce passing IDE Project in constructor and custom created it for all forms. Custom checking for empty projects are gone and `setIdeProject` is removed for public access. 

So now ProjectSelector needs to be explicitly created with a IDE project (or null) and once all UI has been set up, call `loadActiveCloudProject` to initialize.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelector.java
```diff
@@ -73,6 +87,22 @@ public void setSelectedProject(CloudProject cloudProject) {
     }
   }
 
+  /**
+   * Loads active {@link CloudProject}, if IDE {@link Project} has been specified. This should be
+   * called when project selector and the related UI are completely initialized. Calls listeners to
+   * notify on new project selection.
+   */
+  public void loadActiveCloudProject() {
```
etanshaul: so the calls to `loadActiveCloudProject` calls get invoked separately from the calls to `setSelectedProject` which typically get called from platform managed methods like `resetEditorFrom`. Is there a possibility for UI flickering here (first showing the active project, then updating it to the panel specific persisted project)?

etanshaul: also, is there any case where we wouldn't want to load the active/default project?

I'm wondering if we can make the call to `loadActiveCloudProject` implicit in the ProjectSelector itself rather than having the clients call it. In other words, have the ProjectSelector do this itself as part of initialization.

ivanporty: Depends on particular UI design and how long it takes to update it. We can delay `loadActiveCloudProject` and do the explicit check in `resetEditorFrom` for such forms (run/deployment and cloud debugger) to make sure we don't do double work.

etanshaul: my main interest with this is really to see if we can have the ProjectSelector handle all default/active project logic itself rather than having the clients have to explicitly call it (whether it be on init or in `resetEditorFrom`). 

for instances of the selector where we wouldn't want to show an active project (if such a case exists), I guess we could flag this on initialization of the project selector.

ivanporty: As discussed offline, it could be complicated to properly define a moment when default project needs to be loaded without knowing details of surrounding UI initialization. I'll change forms with their own selected project not to load default/active project without checking on existing project first.

ivanporty:
@etanshaul I changed forms with their own selected project not to load default/active project without checking on existing project first. Should be good to review again.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugSettingsEditor.java
```diff
@@ -70,6 +69,8 @@ protected void resetEditorFrom(@NotNull CloudDebugRunConfiguration runConfigurat
               Optional.ofNullable(Services.getLoginService().getActiveUser())
                   .map(CredentialedUser::getEmail)
                   .orElse("")));
+    } else {
+      settingsPanel.setSelectedCloudProject(null);
```
etanshaul: why setting to null and not the active project?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
```diff
@@ -107,7 +107,7 @@ public void mousePressed(MouseEvent event) {
           }
         });
 
-    refresh();
```
etanshaul: what if there is no active project? will refresh still get called by the project selector listener?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ActiveCloudProjectHolder.java
```diff
@@ -0,0 +1,82 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Holds active cloud project, one per IDE project. Active cloud project is the last cloud project
+ * selected by a user in most of the GCP related UI reported by {@link ProjectSelector}.
+ *
+ * <p>Must only be used on Swing/IDE EDT.
+ */
+public class ActiveCloudProjectHolder {
```
etanshaul: how about ActiveCloudProjectManager instead? Sounds a bit more "normal" to me than "Holder".

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ActiveCloudProjectHolder.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  void setActiveCloudProject(
+      @NotNull CloudProject activeCloudProject, @NotNull Project ideProject) {
+    PropertiesComponent propertiesComponent = PropertiesComponent.getInstance(ideProject);
+
+    propertiesComponent.setValue(PROJECT_ACCOUNT_KEY, activeCloudProject.googleUsername());
+    propertiesComponent.setValue(PROJECT_NAME_KEY, activeCloudProject.projectName());
+    propertiesComponent.setValue(PROJECT_ID_KEY, activeCloudProject.projectId());
+    if (activeCloudProject.projectNumber() != null) {
```
etanshaul: sorry if I've asked this before. but just so I understand: project number is nullable but the other values are not?

etanshaul:
looks good! just a few questions and comments

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugSettingsEditor.java
ivanporty: This editor does not have access to project selector itself, it can just pass the cloud project. And active cloud project is internal to `project` and only used internally by `ProjectSelector`. If you think we can make it cleaner, please let me know. Idea was to hide as many details as possible in `loadActiveProject()`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ActiveCloudProjectHolder.java
ivanporty: Yeah, `Manager` is kind of very frequent name. It just felt this class does not really manage anything, just holds the active project per IDE project. I think `Manager` is fine as well, let me know if `Holder` seems off or not clear.

etanshaul: Holder just feels strange to me. It is sort of a manager of the persistence. I'll leave it up to you.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ActiveCloudProjectHolder.java
ivanporty: Mostly for existing run/deploy/debug configurations compatibility. Only one that sets project number is debug configuration. Existing run/deploy configurations will have that as null until we expand them to contain complete project information (name, number, id, account).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/ui/CloudDebugRunConfigurationPanel.java
```diff
@@ -52,7 +57,12 @@ public CloudProject getSelectedCloudProject() {
   }
 
   public void setSelectedCloudProject(CloudProject cloudProject) {
-    projectSelector.setSelectedProject(cloudProject);
+    if (cloudProject != null) {
+      projectSelector.setSelectedProject(cloudProject);
+    } else {
+      // project unset or not configured yet, use active one.
```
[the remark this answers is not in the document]
ivanporty: I agree, pretty obvious code.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/gcs/GcsBucketPanel.java
ivanporty: Good catch, it will not set 'Unselected project' message in this case. Fixing.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/CloudDebugSettingsEditor.java
ivanporty: I like it, this will read better! Thanks.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/debugger/ui/CloudDebugRunConfigurationPanel.java
```diff
@@ -52,7 +57,12 @@ public CloudProject getSelectedCloudProject() {
...
+      projectSelector.loadActiveCloudProject();
```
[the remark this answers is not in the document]
ivanporty: Mmm, can you please expand more on that? We do call `loadActiveCloudProjec` explicitly here if we don't have existing project, is something else missing?

PR #1826, opened 2018-01-10
Set GCS mirror for idea binaries for CI only

### LINKED ISSUE #1798: Gradle sync re-downloads idea distribution every time

etanshaul:
fixes #1798 

See https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1810#issuecomment-355587005.

Configures the build to use the GCS idea mirror only on Kokoro. Locally, it will use the default jetbrains repo which will prevent the maven metadata from being re-downloaded each time when doing a gradle sync from the IDE.


etanshaul:
@patflynn @ivanporty lmyt about this. with this change, gradle syncs run on the order of seconds instead of minutes

Issue #1816, opened 2018-01-09
UI improvements to API enablement dialog


etanshaul:
From demo feedback:

- uncheck add to GCP when disabled
- change OK action to more specific wording ("add lib to module")
- in confirmation dialog confirm module and GCP project to be added to
- some notification when users choose not to enable an API on GCP?

etanshaul:
closed via #1819, #1828 ..

PR #1828, opened 2018-01-10
Client library selector UI improvements

etanshaul:
More updates for #1816 

changes:

Wording on module selector to clarify what its for:
image

Ok action wording (was done in a previously merged PR):
image

Using a stronger 'warning' icon to draw attention to this field (also updated the text a bit):
image

Showing the cloud project on which the APIs will be enabled in the confirmation dialog:
image

The "ok" action will be disabled until there is at least one client library selected and a module is chosen (wasn't yet implemented in the above screenshots):
image


etanshaul:
I feel like I'm may be missing one or two things that were brought up. @patflynn @ivanporty ptal at these changes and also let me know if you can remember other changes that we discussed.

ivanporty:
I think we also discussed showing which APIs are added to module but are not selected for enabling on GCP side, in confirmation dialog.

ivanporty:
I feel that "stronger" red warning is a bit too much maybe when I just opened a dialog - i.e. I want to add a library, open the dialog and immediately see something that looks like I have a serious issue. WDYT?

etanshaul:
> I feel that "stronger" red warning is a bit too much maybe when I just opened a dialog - i.e. I want to add a library, open the dialog and immediately see something that looks like I have a serious issue. WDYT?

Fair point. Especially if I implement your other point (showing apis that are added to the module but not enabled) then the stronger attention isn't needed as much.

etanshaul:
@ivanporty I added confirmation into the existing dialog when the user chooses to add libraries to the module, but does not enable them on GCP. Here are the scenarios:

1) All libraries will have the APIs enabled on GCP
image

2) Some will be enabled, some not
image

3) None will be enabled
image

ivanporty:
Looks very clear, thanks! give me a minute to review the code, I think this is great!

Issue #1829, opened 2018-01-11
Disable flaky IDEA 'leaked unit test projects' check until we figure out its usefulness


ivanporty:
Time to time unit tests fail with TooManyProjectLeakedException from internal IDEA project test system.
The behaivour is non-deterministic: 
* rarely happen on local Ubuntu/Mac builds
* never happens on Windows
* more frequently happens with Ubuntu, especially when IDE Project is used in new tests.

I spent a couple of hours checking for solution and it doesn't seem obvious.
Here is the code IDEA uses:
```java
  private static final boolean LOG_PROJECT_LEAKAGE_IN_TESTS = Boolean.parseBoolean(System.getProperty("idea.log.leaked.projects.in.tests", "true"));
  private static final int MAX_LEAKY_PROJECTS = 5;


    if (ApplicationManager.getApplication().isUnitTestMode()) {
      //noinspection AssignmentToStaticFieldFromInstanceMethod
      TEST_PROJECTS_CREATED++;
      //noinspection TestOnlyProblems
      checkProjectLeaksInTests();
    }

  @TestOnly
  private void checkProjectLeaksInTests() {
    if (!LOG_PROJECT_LEAKAGE_IN_TESTS || getLeakedProjects().count() < MAX_LEAKY_PROJECTS) {
      return;
    }

    if (Math.random() >= 0.05) {
      return; // Check for leaked projects ~ every 20 tests
    }

    for (int i = 0; i < 3 && getLeakedProjects().count() >= MAX_LEAKY_PROJECTS; i++) {
      GCUtil.tryGcSoftlyReachableObjects();
    }

    System.gc();

    if (getLeakedProjects().count() >= MAX_LEAKY_PROJECTS) {
      List<Project> copy = getLeakedProjects().collect(Collectors.toCollection(UnsafeWeakList::new));
      myProjects.clear();
      if (ContainerUtil.collect(copy.iterator()).size() >= MAX_LEAKY_PROJECTS) {
        throw new TooManyProjectLeakedException(copy);
      }
    }
  }
```
From the code:
* It's only used for unit-tests test project tracking. No relation to production issues
* It's based on Math.random (heavily depends on JVM/OS)
* It's based on System.gc() (again, total dependency on JVM implementation)

I don't see any usefulness in tracking why test project reference was not GCed after each (x*random) unit test.

[...]

PR #1830, opened 2018-01-11
Disables flaky leaked project IDEA check.

ivanporty:
See details in #1829.
Internal system property is set to disable the checks since it may take considerable amount of time to investigate existing unit tests approach for creating test projects.

### DIFF
--- a/google-cloud-tools-plugin/build.gradle
+++ b/google-cloud-tools-plugin/build.gradle
@@ -108,7 +108,7 @@ test {
         showCauses true
     }
     doLast{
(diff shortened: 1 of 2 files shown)


google-cloud-tools-plugin/build.gradle
```diff
@@ -108,6 +108,7 @@ test {
         showCauses true
     }
     doLast{
+       println "IDEA leaked check: " + System.getProperty('idea.log.leaked.projects.in.tests')
```
etanshaul: nit: "..leaks project check"

ivanporty: done, thanks

PR #1833, opened 2018-01-16
Disallow hiding account hyperlink for ProjectSelector.

### LINKED ISSUE #1831: The GCP Project field in the Add Cloud Libraries dialog doesn't properly display long project and user names
**What did you do?**
- Go to Tools -> Google Cloud Tools -> Add Cloud libraries...
- When the Add Cloud Libraries dialog opens, select a GCP project that has a long name

The full project name/username in the GCP Project field

screen shot 2018-01-16 at 1 46 47 pm

**What did you see instead?**
In the GCP Project field, only the project name is visible, it seems like the username is on a new line and you can see the top part of the username

screen shot 2018-01-16 at 1 46 38 pm

ivanporty:
Fixes #1831.

Seems like this issue only reproduces on Mac.
Replaced FlowLayout with horizontal BoxLayout to avoid wrapping components to the next row if there is no enough space.
This layout is not supported by GUI designer so the change is in `createUIComponents()`.

@nbashirbello for some reason I am not able to add you to reviewers, can you please add yourself or maybe @etanshaul can take a look why the reviewers list is so short?

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelector.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelector.java
@@ -30,6 +30,7 @@ import java.awt.Component;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import javax.swing.BoxLayout;
 import javax.swing.Icon;
 import javax.swing.JPanel;
 import javax.swing.UIManager;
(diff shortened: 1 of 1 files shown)


etanshaul:
I thought we removed the test project leak check?

ivanporty:
No, I believe no changes needed in form file and it uses form specified layout by default.

Yes, we did added this property to remove test project leak. It does work when the property is set to VM explicitly - I tested it on/off in IDEA test runner and it's easy to see the effect.
I think this has something to do with Gradle way of running unit tests. Looks like it does tests differently and not all tests run with this property. Hard to say why, need to investigate.

PR #1834, opened 2018-01-16
Disable leaked test projects check Gradle way.

ivanporty:
Seems for some reason Gradle test runner loads classes before the system property is passed to VM. 

Looks like we need to explicitly specify the property before the test target runs.

Testing: do not review yet.

### DIFF
--- a/google-cloud-tools-plugin/build.gradle
+++ b/google-cloud-tools-plugin/build.gradle
@@ -102,7 +102,7 @@ artifacts {
 }
 
 test {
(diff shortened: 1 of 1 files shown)


ivanporty:
Please take a look, I found this as a solution to pass a system variable into test task before VM is forked for test, and it seems to work reliably and the variable is set before classes are loaded. I can see random 'leaked project errors' on my local Ubuntu and Kokoro Ubuntu if this is removed. If this line is in, tests seem to pass quickly and reliably.

PR #1835, opened 2018-01-17
Service account mgmt

etanshaul:
fixes part of #1808 

Please only comment on the UI for now (screenshots below). NOT ready for code review.

This PR will adds service account role display to the client library dialog. The next PR will actually add the roles to an existing service account / create a new service account; these controls will appear in the confirmation dialog that pops up.

New service account checkbox (library not selected / cloud project not chosen):
image

Unchecked by default (doesn't have to be though) (library and cloud project selected):
image

Service account roles for the selected API are shown:
image

Each API shows its own set of roles (or none if there aren't any):
image

Clicking the minus button removes the role from the list and the removal is remembered (plus is not yet implemented):
image


ivanporty:
I'm not particularly familiar with service account roles, so all information is more than welcome.
When we enable API, 
> service account checkbox unchecked by default
do we assume this is not required for adding libraries; and defaults are reasonably well pre-set so this checkbox is not enabled by default?

ivanporty:
> Service account roles for the selected API are shown:

Are all roles assigned to service account by default, i.e. nothing to add initially, only to remove?

nbashirbello:
Could the list of roles be selected using checkboxes instead so the users would always know all the roles available?

ivanporty:
I agree with @nbashirbello, we probably can integrate add/remove functionality in the table itself. It should be even easier to implement rather than present choices when a user clicks `+` button.

etanshaul:
Weather or not the updating the service account is required depends on many factors (is the user running locally? is this an App Engine environment? does the API itself require any roles to be granted? etc.). But long story short, a service account key _is_ required for local development to hit certain GCP APIs. We definitely could have the checkbox on by default.

etanshaul:
@nbashirbello @ivanporty regarding using checkboxes to select APIs:

What you are saying definitely makes sense. My original intent was to uncluttered the UI as much as possible - i.e. don't show the roles that the user is not interested in. Adding checkboxes also slightly complicates the UI with extra controls. The idea behind the '+' button was to pop open a simple list panel with the set of roles available not already in the main list. 

Also to point out - the initial default state is to show all the roles available for the API, so the user will see this initially.

lmkwyt

ivanporty:
I don't feel very strong going specifically with checkbox list right now - we can definitely start as is, it is working and obvious how to use it. We can then assess UX flow a bit better once we start using the actual feature.

patflynn:
What does the help text (presumably from clicking the question mark) say?

I'm a little concerned that this UI begs a lot of questions that aren't
obviously answered. As an end-user I'd be confused about what service
account it's talking about and what the context is (how it's hooked up at
runtime).

nbashirbello:
Is the preferred case that the user have all the roles enabled for an API or to have 1 specific role, e.g Admin enabled for the API? If the preferred case is to have 1 role enabled, that means the user has to delete n-1 roles to have the preferred scenario. If roles where selected/unselected by checkboxes, we could initially only select the preferred role(s) so that the user would be able to set up the API in the preferred way with fewer clicks.

ivanporty:
Good point from @patflynn about what exactly service account this panel is talking about. As a user I would probably like to see what's the relation between `API + GCP Project + user project module ~ service account`. 

I think it's a very clear dialog when it comes to adding a library dependency and automatically enabling it on GCP side. It would be good if it stays that straightforward with service account as well.

etanshaul:
The text I chose for the service account UI isn’t the greatest. I’ll tweak things to make it clearer and update the screenshots. I think with better text it will be straightforward. If, however, you have suggestions for bigger UI changes here (more than just wording and hints) let me know.

@nbashirbello there is no concept of a “preferred” role. It’s up to the user’s specific use-case. All we have is a mapping from API to set of associated roles. My thinking that by putting them all in front of the user like this it will at least give them a clear opportunity to remove the ones they don’t require.

patflynn:
How about a little UI brainstorm Monday?

etanshaul:
@ivanporty @nbashirbello @patflynn 

Still not ready for code review (code needs cleanup and refactoring) but the functionality and the UI is complete. Confirmation of service account key download (and actual download) work will be submitted as a separate PR.

Please take a look at the screenshots below. We can iterate on the wording / documentation if you think something is unclear. But if you have a more general concern let me know:

**Library is selected - service account checkbox on by default**
image
Note that the role selections will be remembered as the user switches between different libraries

**API with no associate roles - role table is hidden**
image

patflynn:
Oh I see the issue with the 'download a new service account key' checkbox
now... Basically it doesn't make sense for it to be scoped at the library
level since you're creating a single service account key with all the
permissions across the APIs. Is that right?

etanshaul:
Exactly, the download will occur in the confirmation dialog which is scoped to all selected APIs (next PR).

etanshaul:
Or, are you suggesting to move all the service account UI to the confirmation dialog and not even show it at the individual library level? If so, I think either approach could work.

patflynn:
@etanshaul not really suggesting anything at this stage, but I do think it'll be puzzling to be able to check and uncheck the box for different libraries.

etanshaul:
I see you’re point. Let’s discuss again today briefly in the meeting today. I’ll explain what the other approach looks like.

etanshaul:
Based on the discussion today we are modying the UI. Since this thread is getting long and the changes have deviated, I'm closing and will open a clean PR.

PR #1847, opened 2018-01-25
Exclude version of protobuf incompatible with error feedback code 

etanshaul:

PR #1849, opened 2018-01-25
Exclude version of protobuf incompatible with error feedback code

etanshaul:
see #1847


etanshaul:
please approve again. 

this has just been released as a patch on the 17.2.2 release: https://plugins.jetbrains.com/plugin/8079-google-cloud-tools

so now I am merging it into master

PR #1851, opened 2018-01-26
Adding new 2017.3 idea codestyle metadata to vcs

etanshaul:
see https://confluence.jetbrains.com/display/IDEADEV/New+project+code+style+settings+format+in+2017.3


patflynn:
are these styles google style compatible?

etanshaul:
These files are a port from the old codeStyleSettings.xml that lived in the root of the .idea folder (2017.3 refactored how the styles are stored). I think that the google-java-format plugin just overrides these.

I confirmed that when the google-java-format plugin is enabled that the files are styled correctly.

patflynn:
lgtm

PR #1852, opened 2018-01-26
Hides add libraries action for non-Maven projects (no Maven modules) & Java 7 AE standard projects.

### LINKED ISSUE #1843: Disable / hide library UI for unsupported build systems & Java 7 App Engine projects
Until support is added for native / gradle, this UI should be somehow hidden/disabled for users.

ivanporty:
Fixes #1843.

As a side note, this is extremely useful functionality (even just enabling APIs on GCP side without adding dependencies into local project), should we consider just showing an additional message that APIs have been enabled, show 
`dependependencies {
    compile(...) ` 

block for Gradle project to copy/paste, suggest downloading JARs for native, and advise to use Maven/gradle? (i.e. instead of hiding the action completely)


etanshaul:
absolutely that is the plan (I should create an issue for that). but this issue was meant as an interim solution until that is implemented

etanshaul:
https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1853

If we can get that one done before the release, then this PR will not be necessary. So we can just hold off on merging this one for now and see.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesAction.java
```diff
@@ -35,6 +39,23 @@ public AddCloudLibrariesAction() {
         GoogleCloudToolsIcons.CLOUD);
   }
 
+  @Override
+  public void update(AnActionEvent e) {
+    boolean enableAddLibraries = false;
+    Project project = e.getProject();
+    if (project != null) {
+      // check if this project has any 'maven-style' modules.
+      for (Module module: ModuleManager.getInstance(project).getModules()) {
```
etanshaul: `DefaultAppEngineProjectService` has an isMavenModule method. My point isn't that we should use that instead, but I see that it also does a check for `MavenProjectsManager#findProject` not being null. I'm wondering if that is redundant.

ivanporty: Hard to say from the source code, but let's do that in one place, always a good thing.

etanshaul: sure. I don't like that that is an "app engine" utility. But we can make note to refactor that.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesAction.java
```diff
@@ -35,6 +39,23 @@ public AddCloudLibrariesAction() {
         GoogleCloudToolsIcons.CLOUD);
   }
 
+  @Override
+  public void update(AnActionEvent e) {
+    boolean enableAddLibraries = false;
```
etanshaul: might be cleaner over all if you made it more functional:

```
...findAny().map( //set visible true).orElseGet( //set visible to false);
```

Up to you if you think its cleaner

etanshaul: or better yet: 

```
setVisible(..anyMatch(module -> isMavenModudel(module)));
```

ivanporty: Interesting exercise, not sure if it's more readable, but looks neat :)

etanshaul: might be even more readable if you take out the project optional stuff from that and just have 

```
if (e.getProject() == null) {
  setVisible(false);
  return;
}

// functional craziness
```

ivanporty: That's what I wanted initially and then I thought "sooner or later someone will definitely want *everything* wrapped in one functional style run if it's already there".

etanshaul: lol. i was really fine with what you had originally. I just like to make things functional for fun. Go with whichever you prefer.

ivanporty: I think let's leave it functional here - I strongly suspect we'll get rid of this in a few days or a week anyway once we have Gradle/native message/wording added into libraries UI. WDYT?

etanshaul:
as discussed offline, can you look into adding the check for App Engine + java 7 projects into this (or in a separate PR if you think it makes more sense)? We should have all the utilities in place already to get this information about the project.

ivanporty:
Please review the changes, added support for GAE Standard + Java 7 and disabled action instead of hiding, using helpful wording for enabling action with Maven/Java8.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -217,6 +217,7 @@ cloud.libraries.dialog.title=Add Cloud Libraries
 cloud.libraries.documentation.link=Documentation
 cloud.libraries.menu.action.text=Add Cloud libraries...
 cloud.libraries.menu.action.description=Select the Cloud libraries to download and include in your application
+cloud.libraries.menu.action.disabled.description=Maven project and Java 8 required to enable selection of the cloud libraries
```
etanshaul: The Java 8 part of this message only applies if its App Engine standard. So perhaps split this message up into several cases or just make the wording more general?

@patflynn wdyt?

etanshaul: see the "supported platforms" section of the [client library github page](https://github.com/GoogleCloudPlatform/google-cloud-java). I think it supports my comment. So I would go ahead and make the change.

ivanporty: sounds good, I added separate messages.

patflynn:
I would be specific about the concrete reason it's disabled.  i.e. for App
Engine Standard Java 7 'These libraries are not compatible with App Engine
Standard Java 7, use the App Engine SDK instead', and for non-Maven 'Only
Maven based projects are supported at this time."

ivanporty:
Updated wording, please check.

PR #1861, opened 2018-01-29
Bump idea dev version to 2017.3

etanshaul:


etanshaul:
2017.3 is the latest GA version

PR #1863, opened 2018-01-30
Update release notes

etanshaul:


etanshaul:
if I missed any features/fixes you can think of, please let me know

PR #1865, opened 2018-01-30
Add Deploy config advanced flags ui

### LINKED ISSUE #1743: Extend App Engine deployment configuration
Certain flags, such as `enable_jar_classes` that are currently not exposed in the UI are potentially useful to users. See https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1658

Need to investigate which flags to expose (do we expose them all? custom fields vs generic flag field?).

nbashirbello:
Fixes #1743


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/standard/AppEngineStandardStage.java
```diff
@@ -68,11 +72,22 @@ public void onOutputLine(String line) {
 
     // TODO determine the default set of flags we want to set for AE standard staging
     DefaultStageStandardConfiguration stageConfig = new DefaultStageStandardConfiguration();
-    stageConfig.setEnableJarSplitting(true);
```
nbashirbello: @etanshaul Do we want to set the default value of enable_jar_splitting to true, since at some point we started setting that value to true internally?

etanshaul: I think it was pointed out (by @chanseokoh?) that changing this default value could cause some changes on how the project runs.

@patflynn do you know if we have any guidance on what defaults we should use, or is it up to us to come up with sensible defaults?

nbashirbello:
screen shot 2018-02-01 at 10 59 46 am

screen shot 2018-02-01 at 11 02 05 am

etanshaul:
that looks cool!

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
```diff
@@ -52,6 +54,18 @@
   private HyperlinkLabel appEngineCostWarningLabel;
   private JLabel serviceLabel;
   private JCheckBox hiddenValidationTrigger;
+  private JTextField compileEncodingTextField;
```
etanshaul: is it not possible to refactor this so that this UI fields live in `AppEngineStandardDeploymentEditor` instead of in this common panel?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/standard/AppEngineStandardDeploymentEditor.java
```diff
@@ -65,13 +65,33 @@ public AppEngineStandardDeploymentEditor(Project project, AppEngineDeployable de
   @Override
   protected void resetEditorFrom(@NotNull AppEngineDeploymentConfiguration configuration) {
     commonConfig.resetEditorFrom(configuration);
+    commonConfig.getCompileEncodingTextField().setText(configuration.getCompileEncoding());
```
etanshaul: if the fields live in the standard editor class like I mentioned in the other comment, then you wouldn't have to expose the UI fields with getters like this, they would just be part of this class

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfigurationPanel.java
etanshaul: you could have the tabbed panel field live in the common panel, but then all of the inner components of the "advanced" tab live in `AppEngineStandardDeploymentEditor`

patflynn:
What are valid values for the compile encoding?   Can we make it a
combobox?  What about the jar_splitting_excludes what is the value in there
supposed to look like?

nbashirbello:
@patflynn I don't know all the valid values for compile_encoding and jar_splitting_excludes. The only documentation I found on them was this https://github.com/GoogleCloudPlatform/gcloud-maven-plugin#application-deployment-goal, and it doesn't mention the valid values.

ivanporty:
Nice design! Are we sure we won't have more boolean properties in the future? They are sitting nicely in a grid now but if we keep adding them then it might grow beyond being beautiful. 

Also, it feels good they use the same property name one'd use with SDK itself, but slightly strange for UI - maybe we should add small question mark which brings us to the description page? I don't know about the amount of work, but adding a tooltip for each property with short description of the property would be even more cool.

patflynn:
I wouldn't worry. I don't think we'll be adding any more staging params.

patflynn:
@nbashirbello 
it sounds like the excludes are only applicable when enable jar splitting is enabled.

Perhaps we should only show it when enable jar splitting is enabled, and have some documentation to indicate that it should be a comma separated lists of file suffixes to exclude with an example?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/standard/AppEngineStandardStagingPropertiesPanel.java
```diff
@@ -0,0 +1,106 @@
...
+  private JCheckBox deleteJspsCheckBox;
+  private JCheckBox enableJarClassesCheckBox;
+  private JCheckBox disableJarJspsCheckBox;
+  private JCheckBox enableJarSplittingCheckBox;
+  private JCheckBox enableQuickstartCheckBox;
+  private JCheckBox disableUpdateCheckCheckBox;
+  private JTextField jarSplittingExcludesTextField;
+  private JPanel mainPanel;
+
+  public JTextField getCompileEncodingTextField() {
```
etanshaul: are these exposed on for testing? if so `@VisibleForTesting`

etanshaul: and probably  move them down below the methods that have the core functionality, for readability

etanshaul:
@nbashirbello thanks for the refactoring. Just some minor comments from me.

to follow up on @patflynn's comment: if the `excludes` field is only available if "jar splitting" is checked, then perhaps we can make this even clearer in the UI by having it be a child component (i.e. visually nested underneath it)?

etanshaul:
on second though, if we have some helper wording around this, then there is no need really to visually nest it like I suggested above

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
```diff
@@ -130,6 +139,78 @@ public boolean isStagedArtifactNameLegacy() {
     return stagedArtifactNameLegacy;
   }
 
+  @Attribute("compileEncoding")
+  public String getCompileEncoding() {
```
ivanporty: There are many staging properties, do we know if it's possible to introduce a helper bean, like `StagingProperties`, and annotate it with `@Attribute`? It could possibly make it a bit simpler to understand major properties in the deployment configuration then.

etanshaul:
@nbashirbello we discussed in the meeting today updating the labels of the advanced staging properties to something more human readable. E.g.:

`enable_jar_classes` -> `Enable JAR Classes`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/standard/AppEngineStandardStage.java
etanshaul: Lets cross reference with the doc and gcloud default behavior to determine what defaults to set for these.

nbashirbello:
Added a help icon in the staging tab that currently links to  https://github.com/GoogleCloudPlatform/gcloud-maven-plugin#application-deployment-goal. Would update later to show a file with the tablized version of the flags with their description

screen shot 2018-02-06 at 2 55 47 pm

ivanporty:
Looks nice, maybe we can think of placing the help icon to the right bottom corner? Now it kind of creates a large margin between the tab and the entire flags UI.

nbashirbello:
screen shot 2018-02-06 at 4 27 53 pm

nbashirbello:
When "Enable JAR Splitting" is disabled and there are no existing values for "JAR Splitting excludes"

screen shot 2018-02-11 at 4 27 00 pm

When "Enable JAR Splitting" is disabled and there is a value for "JAR Splitting excludes"

screen shot 2018-02-11 at 4 29 10 pm

briandealwis:
My 2¢: it seems odd to have a text field that's controlled by what appears to be a sub-button.

etanshaul:
@briandealwis that can probably be solved by making the text field a child (visually indented) of the checkbox. wdyt? I'm open to other suggestions.

briandealwis:
Makes sense to me!

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
etanshaul: this sounds like a good suggestion. maybe add a todo to revisit this in a future PR?

briandealwis:
Does it make sense for the user to enable the up-to-date check?

nbashirbello:
The staging flags will soon be configurable from the appengine-web.xml so UI support is no longer need.

PR #1870, opened 2018-01-30
Fix IDEA style table search indexes validation.

### LINKED ISSUE #1869: Exception may occasionally occur if using IDEA style table search in Project Selector
IDEA style built-in table search (just start typing in the table) supports case of matching a record even if typed filter text only matches partially and standard table filter does not match anything. This can generate model size and selection index not expected by project selection dialog and generate `ArrayIndexOutOfBoundsException` in the logs.

ivanporty:
Fixes #1869

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
@@ -331,8 +331,8 @@ public class ProjectSelectionDialog {
       // reflect this text in the filter text field.
       @Override
       protected void selectElement(Object element, String selectedText) {
-        filterTextField.setText(selectedText);
         super.selectElement(element, selectedText);
+        filterTextField.setText(selectedText);
       }
     };
   }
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -359,11 +359,14 @@ private void hideSignInRequest() {
...
-    if (projectListTable.getSelectedRow() >= 0) {
+    int selectedRow = projectListTable.getSelectedRow();
+    // make explicit visible row check due to non-standard behaivor of IDEA's TableSpeedSearch
+    int visibleRowCount = projectListTable.getRowSorter().getViewRowCount();
+    if (selectedRow >= 0 && selectedRow < visibleRowCount) {
       dialogWrapper.setOKActionEnabled(true);
       selectedProjectsByAccount.put(
           (CredentialedUser) accountComboBox.getSelectedItem(), getSelectedProjectName());
-    } else {
+    } else if (selectedRow < 0 /* nothing selected. */) {
```
etanshaul: nit: move comment out of if clause (just put it next to the `{`)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -359,11 +359,14 @@ private void hideSignInRequest() {
...
   private void validateProjectSelection() {
-    if (projectListTable.getSelectedRow() >= 0) {
+    int selectedRow = projectListTable.getSelectedRow();
+    // make explicit visible row check due to non-standard behaivor of IDEA's TableSpeedSearch
+    int visibleRowCount = projectListTable.getRowSorter().getViewRowCount();
+    if (selectedRow >= 0 && selectedRow < visibleRowCount) {
       dialogWrapper.setOKActionEnabled(true);
       selectedProjectsByAccount.put(
           (CredentialedUser) accountComboBox.getSelectedItem(), getSelectedProjectName());
-    } else {
```
etanshaul: just so I understand; if none of these conditions are met, this just represents a non-existent index and nothing should be done? We don't want to disable the ok action in this case?

ivanporty: It could be an existing index but generated by IDEA table speed search and out of bounds for normal table filtering. So we check explicitly for no selection (<0) to avoid disabling ok action when some selection is present.

PR #1871, opened 2018-01-30
Announce new feature at top of changelog

etanshaul:
wdyt?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ This page documents production updates to Cloud Tools for IntelliJ. You can chec
 
 ## 18.1.2
 
-You can now add client libraries for accessing Google Cloud APIs to your project. Just select 'Tools > Google Cloud Tools > Add Cloud Libraries ...'.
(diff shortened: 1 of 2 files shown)


ivanporty:
Sounds great, also, could we also match our change log to our releases page? I.e. it seems a bit empty comparing to some projects I used before: compare with https://github.com/facebook/rocksdb/releases for example. If every release could have a list of changes along with assets, that'd be awesome.

etanshaul:
agree. that would be awesome. I don't know how to tie into that. But we should look into it.

etanshaul:
I guess we could do it manually until we figure out how to automate it. Starting with this release would be nice.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ This page documents production updates to Cloud Tools for IntelliJ. You can chec
 
 ## 18.1.2
 
+You can now add client libraries to your project for accessing Google Cloud APIs. Just select 'Tools > Google Cloud Tools > Add Cloud Libraries ...'.
```
patflynn: My only concern is for App Engine users.   Java 7 won't work, and Java 8 will be a bad experience.

etanshaul: Fair point. Do you suggest not including this or rewording it to make it clear?

PR #1872, opened 2018-01-31
When creating a file from template, do not overwrite file if it already exists

### LINKED ISSUE #1788: adding App Engine support wipes out existing appengine-web.xml

- Cloud Tools for IntelliJ version: 17.12.2 
- Google Cloud SDK version _(run `gcloud version` on the command line)_:
Google Cloud SDK 182.0.0
app-engine-java 1.9.59
app-engine-python 1.9.63
bq 2.0.27
core 2017.12.01
gcloud 
gsutil 4.28
- OS: OSX 10.13.1

**What did you do?**

1. **Tools** --> **Google Cloud Tools** --> **Add App Engine Support** --> **Google App Engine Standard**
2. select Maven module I wish to add GAE support to
3. Maven module already had src/main/webapp/appengine-web.xml

Some sort of warning at least of any destructive changes it would make. 

**What did you see instead?**
Maven module already had `src/main/webapp/WEB-INF/appengine-web.xml`, local db, etc. Without any feedback, it removed / overwrote `appengine-web.xml` and the local db (`appengine-generated/` files I think). It did not overwrite other GAE config files: `cron.xml`, `datastore-indexes.xml`, `dispatch.xml`, `queue.xml`, etc.

`appengine-web.xml` was replaced with: 
```xml
<?xml version="1.0" encoding="utf-8"?>
<appengine-web-app xmlns="http://appengine.google.com/ns/1.0">
  <threadsafe>true</threadsafe>
  <runtime>java8</runtime>
</appengine-web-app>
```

I understand logic in doing that if no `appengine-web.xml` exists, but not overwriting an existing one; it should either warn or leave it. In my case, my `appengine-web.xml` already had both those params, so it was effectively just removing the rest of the configuration.

This is most problematic due to #1625 and the repeated need to add/remove GAE support, and re-import the project, to try to keep the GAE Run Configuration working. It adds recovering the old `appengine-web.xml` as an extra step to do whenever you re-import.

nbashirbello:
Fixes #1788

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/AppEngineStandardSupportProvider.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/AppEngineStandardSupportProvider.java
@@ -108,15 +108,17 @@ public class AppEngineStandardSupportProvider extends FrameworkSupportInModulePr
   public static VirtualFile createFileFromTemplate(
       final String templateName, final VirtualFile parent, final String fileName) {
     parent.refresh(false, false);
+    VirtualFile file = parent.findChild(fileName);
+    if (file != null) {
+      return file;
+    }
+
     final FileTemplate template =
         FileTemplateManager.getDefaultInstance().getJ2eeTemplate(templateName);
     try {
       final String text =
           template.getText(FileTemplateManager.getDefaultInstance().getDefaultProperties());
-      VirtualFile file = parent.findChild(fileName);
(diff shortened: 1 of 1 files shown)


etanshaul:
likely unrelated to this change, but I'm noticing that the appengine-web.xml is not being created using the new project wizard. looking into it to see whats happening

etanshaul:
it was unrelated. it was due to changes in the web project wizard in 2017.3

etanshaul:
Let’s still hold off on merging until after the release please

PR #1878, opened 2018-01-31
Add credentialed user in the CSR link

### LINKED ISSUE #1877: Sending the wrong credentials to the Console
NOTE: This is an extreme edge case, so it may not be worth pursuing, but I thought I should file the issue anyway.
I wanted to import from source control.  In the Set up on Google Cloud Source Repositories dialog, I changed accounts to one that did not have a Cloud repo. When I click Create a new cloud repo button, I get to the Console and see an error: "You do not have sufficient permissions to view this page." In the top right corner, I see the incorrect account displayed. The first account I signed into IntelliJ. See the attached screenshot.

edge2

Steps to reproduce:
1. Sign into IntelliJ
2. Select VCS > Import into Version Control > Set up on Cloud Source Repositories
3. In the Set up on Google Cloud Source Repositories dialog, click the ellipse and click add account
4. Select a new account.
5. Return to Set up on Google Cloud Source Repositories dialog
6. Click the Create a new Cloud Repository button
7. See error in Console.

Looks like IntelliJ is sending the first account and not the one I selected because it does not happen if I run through these steps again but in reverse order.

etanshaul:
fixes #1877

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
@@ -66,8 +66,8 @@ public class RepositorySelector extends CustomizableComboBox implements Customiz
   private static final Logger logger = Logger.getInstance(RepositorySelector.class);
   private static final int SELECTOR_HEIGHT = 140;
   private static final int SELECTOR_WIDTH = 400;
-  private static final String PANETHEON_CREATE_REPO_URL_PREFIX =
-      "https://console.cloud.google.com/code/develop/repo?project=";
+  private static final String PANETHEON_CREATE_REPO_URL_PATTERN =
+      "https://console.cloud.google.com/code/develop/repo?project=%s&authuser=%s";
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/resources/RepositorySelector.java
```diff
@@ -66,8 +66,8 @@
   private static final Logger logger = Logger.getInstance(RepositorySelector.class);
   private static final int SELECTOR_HEIGHT = 140;
   private static final int SELECTOR_WIDTH = 400;
-  private static final String PANETHEON_CREATE_REPO_URL_PREFIX =
-      "https://console.cloud.google.com/code/develop/repo?project=";
+  private static final String PANETHEON_CREATE_REPO_URL_PATTERN =
```
ivanporty: Typo in `pantheon`?

etanshaul: thanks! fixed

PR #1879, opened 2018-01-31
ProjectSelector prompts to login if no users are logged in.

### LINKED ISSUE #1875: Project selector padding
image

looks like between the "/" and the avatar image could use a bit of padding

### LINKED ISSUE #1876: Project selector logged out view still shows cloud project and user name
1) Sign in to your account
2) verify that the selector shows the project, avatar, and email
3) sign out
4) look at the project selector widget:

image

It should show the signed out message. Instead it still shows the last signed in email and project name (and the avatar is now the generic one).

ivanporty:
Fixes #1876.
Fixes #1875.

When no users are signed in, project selector will show prompt to sign in instead of account name and generic icon. This prompt leads directly to account sign in page. Once signed in, project selector behavior changes to default.

projectselector-sign-in

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelector.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelector.java
@@ -17,12 +17,16 @@
 package com.google.cloud.tools.intellij.project;
 
 import com.google.cloud.tools.intellij.login.CredentialedUser;
+import com.google.cloud.tools.intellij.login.GoogleLoginListener;
+import com.google.cloud.tools.intellij.login.IntegratedGoogleLoginService;
 import com.google.cloud.tools.intellij.login.Services;
 import com.google.cloud.tools.intellij.login.ui.GoogleLoginIcons;
 import com.google.cloud.tools.intellij.util.GctBundle;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.project.ProjectManager;
 import com.intellij.openapi.ui.FixedSizeButton;
 import com.intellij.ui.HyperlinkLabel;
 import com.intellij.ui.components.JBLabel;
@@ -30,6 +34,7 @@ import java.awt.Component;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Stream;
 import javax.swing.Icon;
 import javax.swing.JPanel;
 import javax.swing.UIManager;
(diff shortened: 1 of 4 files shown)


etanshaul:
Haven't reviewed yet, but does this also fix [my last comment on the issue](https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1876#issuecomment-362065876)?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelector.java
```diff
@@ -64,9 +69,32 @@
...
+                      notifyProjectSelectionListeners();
+                    });
+          }
+        };
+    Stream.of(ProjectManager.getInstance().getOpenProjects())
+        .forEach(
+            project -> {
+              project
+                  .getMessageBus()
+                  .connect()
```
etanshaul: should we pass in a disposable so that the subscriber get detached when closed?

ivanporty: I tried to think what would be that disposable in case of universal component and couldn't find a solution that wouldn't fail - i.e. we cannot count on being inside of the dialog or tie to the IDE project - those could be flaky. Open to suggestions or hints.

etanshaul: hmm I see. for the deployment window as an example, each time its opened a new project selector instance is created. So if we don't pass in a disposable, then the subscribers will never get killed and will be stuck orphaned subscribers - does that sound correct?

ivanporty: Sure, we need to get rid of that listener. It seems like a new case, I don't see it in our code yet - do you remember if it's possible to unregister a message bus listener?

etanshaul: `MessageBusConnection` has a `disconnect` method. That can be called. Its likely that I have also made this error previously in our code (not passing in a disposable or explicitly disconnecting). We can (and should) fix this in a separate pr. WDYT?

ivanporty: Sounds good, let's do the disconnect - I think we can count on an UI component being removed from the container event to do that cleanly. I'll file a separate PR for that.

etanshaul:
is this ready for review after all the discussion? if not, just ping me and ill get on it.

ivanporty:
Yep, it's ready to go - I just added change for active project (not to set it when no users are logged in) as we discussed in the issue.

etanshaul:
works great! thanks for getting to it so fast.

One minor thing:
1) start signed in
2) open the gcs browser
3) sign out
4) notice the project selector gets updated properly
5) sign in
6) The project selector gets updated again and everything works, except one minor thing:

screen shot 2018-01-31 at 4 48 49 pm

The avatar is now missing

ivanporty:
Yep, `CredentialedUser` does not return an icon and full user name right after log in. There's some delay associated with it, I've seen the same with the login in the project selection dialog. I guess we need to check a bit deeper what happens behind the scenes there (in separate PR?).

etanshaul:
yep. separate PR sounds good. This is a minor inconvenience.

PR #1886, opened 2018-02-02
Disconnect message bus from project selector on component removal

### LINKED ISSUE #1881: Clean up message bus listener for GOOGLE_LOGIN_LISTENER_TOPIC in ProjectSelector
ProjectSelector now uses message bus and GOOGLE_LOGIN_LISTENER_TOPIC event to listen for login changes, but does not detach itself or specify a disposable to automatically detach.

ivanporty:
Fixes #1881.


etanshaul:
so having this in addition to a disposable (for the cases where we use disposables) won't matter (no interference or anything)?

ivanporty:
I think message bus connection is local to the code that calls `connect()` so it sort of "owns" its lifetime. I looks to be working fine so far.

PR #1887, opened 2018-02-05
Add service account management in the client library UI

etanshaul:
Clean PR based on discussion and work on #1835 

Adds Service Account management to the GCP API management confirmation dialog.


etanshaul:
**Library screen is unchanged**
image

**Confirmation dialog with an API that has no roles**
image

**Confirmation dialog with APIs that have roles**
image

**User chooses not to create a service account / download key**
image

etanshaul:
@patflynn @ivanporty please look at the updated UX above.

This is based on the mockups I demod today. Yes I know it feels a little busy and I'm thinking of how to improve that.

Almost ready for code review, will update here.

ivanporty:
That's great, much easier to feel the connection between APIs and service accounts.
Do we enable 'Service Account' checkbox by default?

Nit: could we align text fields for "Service Account Name" and "Download Path" too?

etanshaul:
@ivanporty I was actually working on the alignment like you mentioned. Screenshots updated.

Yes, service account checkbox is selected by default.

Also, ready for code review

ivanporty:
For service account name, do we think it's viable to show what name is generated if nothing is specified (if it's possible to know at all)? Or maybe we're going to show the resulting service account name in final notification bubble?

etanshaul:
@ivanporty yes - sorry forgot to mention - I'd like to confine this PR mostly to agreeing on the UI. The next PR will have the service account generation (including setting a default name, key downloading, etc.)

etanshaul:
(screenshot updated again with slight change - added some margin between the download path field and the role table)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -137,6 +140,28 @@ private static void enableApi(
...
+
+    if (!user.isPresent()) {
+      LOG.error("Cannot fetch service account roles: logged in user not found.");
+      return ImmutableList.of();
+    }
+
+    Iam iam = GoogleApiClientFactory.getInstance().getIamClient(user.get().getCredential());
+
+    try {
+      return iam.roles().list().execute().getRoles();
```
ivanporty: Could we consider doing this asynchronously as well, or this call usually takes short time and role list is mostly not large?

etanshaul: Seems to be very quick. But yes I agree that all of these calls should ultimately be asynchronous. I’ll open an issue to track it separately.

etanshaul:
@patflynn - would be great if you could take a quick look at the UX too (screenshots above).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
```diff
@@ -126,6 +133,29 @@ protected void doOKAction() {
...
 
+  /**
+   * Fetches all the available roles for the {@link CloudProject} and returns the set of roles
+   * corresponding to the selected client libraries.
+   *
+   * @param apis the set of {@link CloudLibrary apis} selected
+   * @return the set of {@link Role roles} corresponding to the selected apis
+   */
+  private Set<Role> getServiceAccountRolesRoles(Set<CloudLibrary> apis) {
+    List<Role> serviceAccountRoles = CloudApiManager.getServiceAccountRoles(getCloudProject());
```
ivanporty: Looks like we have to get a complete set of roles for the user and we only need a smaller subset of those to  get list of roles for selected APIs? I briefly looked myself too and couldn't find anything more specific, so it seems it's currently not possible to obtain list of roles per API/resource?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
```diff
@@ -126,6 +133,29 @@ protected void doOKAction() {
...
+    Set<String> roleIds =
+        apis.stream()
+            .map(CloudLibrary::getServiceRoles)
+            .filter(Objects::nonNull)
+            .flatMap(Collection::stream)
+            .collect(Collectors.toSet());
+
+    return serviceAccountRoles
+        .stream()
+        .filter(role -> roleIds.contains(role.getName()))
```
ivanporty: Do we need `contains()` or `equals()` to match role names in this filter, can they be slightly different in names so `contains()` is required?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -137,6 +140,28 @@ private static void enableApi(
         .execute();
   }
 
+  /**
+   * Fetches the list of {@link Role} for the supplied {@link CloudProject} by querying the Iam API.
+   */
+  static List<Role> getServiceAccountRoles(CloudProject cloudProject) {
```
ivanporty: What do you think about using valid `CredentialedUser` as a parameter for this method? It seems we are getting list of all roles for the full account rather than for a specific project so it may make this method a bit cleaner.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -137,6 +140,28 @@ private static void enableApi(
...
+      LOG.error("Cannot fetch service account roles: logged in user not found.");
+      return ImmutableList.of();
+    }
+
+    Iam iam = GoogleApiClientFactory.getInstance().getIamClient(user.get().getCredential());
+
+    try {
+      return iam.roles().list().execute().getRoles();
+    } catch (IOException e) {
+      e.printStackTrace();
```
ivanporty: Could we log this exception instead of printing its stacktrace?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
```diff
@@ -126,6 +133,29 @@ protected void doOKAction() {
...
   }
 
+  /**
+   * Fetches all the available roles for the {@link CloudProject} and returns the set of roles
+   * corresponding to the selected client libraries.
+   *
+   * @param apis the set of {@link CloudLibrary apis} selected
+   * @return the set of {@link Role roles} corresponding to the selected apis
+   */
+  private Set<Role> getServiceAccountRolesRoles(Set<CloudLibrary> apis) {
```
ivanporty: Nit: in general it feels this method is pure API management logic and might be a better fit for `CloudApiManager` class?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
```diff
@@ -42,6 +67,17 @@
...
+  private JScrollPane serviceAccountDetailsPane;
+  private JTable roleTable;
+  private JScrollPane rolePane;
+  private JPanel rolePanel;
+
+  private final Set<Role> roles;
+  private static final boolean UPDATE_SERVICE_ACCOUNT_DEFAULT = true;
 
   /**
    * Initializes the Cloud API management confirmation dialog.
```
ivanporty: Could we also have `@param` javadoc descriptions for all parameters since we already have them for first two?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
```diff
@@ -81,4 +138,110 @@ private void populateLibraryList(JList<String> list, Set<CloudLibrary> libraries
...
+    public Object getValueAt(int rowIndex, int columnIndex) {
+      if (columnIndex == 0) {
+        return roleMap.keySet().toArray()[rowIndex];
+      }
+      return roleMap.values().toArray()[rowIndex];
+    }
+
+    @Override
+    public void setValueAt(Object value, int rowIndex, int columnIndex) {
+      if (columnIndex == 0) {
```
ivanporty: It might be a bit excessive protection here - table model should keep its contract not to call this for columns not set in `isCellEditable()`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
```diff
@@ -81,4 +138,110 @@ private void populateLibraryList(JList<String> list, Set<CloudLibrary> libraries
...
+        throw new RuntimeException();
+      }
+
+      Role role = (Role) roleMap.keySet().toArray()[rowIndex];
+      roleMap.put(role, (boolean) value);
+
+      TableModelEvent event = new TableModelEvent(this, rowIndex);
+      Stream.of(listenerList.getListeners(TableModelListener.class))
+          .forEach(listener -> listener.tableChanged(event));
```
ivanporty: It might be more complicated than in needs to - it's possible table will update this cell itself after editing and event is not needed at all? Or if it is needed, there are a set of useful methods in `AbstractTableModel` -
 we can try `fireTableCellUpdated(rowIndex, columnIndex);`, it could be easier to read.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
```diff
@@ -81,4 +138,110 @@ private void populateLibraryList(JList<String> list, Set<CloudLibrary> libraries
...
+    }
+
+    @Override
+    public int getColumnCount() {
+      return 2;
+    }
+
+    @Override
+    public Class<?> getColumnClass(int columnIndex) {
+      return columnIndex == 0 ? Role.class : Boolean.class;
```
ivanporty: This is super simple table with 2 columns but to make it even simpler and easier to change later should we put column indexes in constants maybe and use `switch`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
```diff
@@ -81,4 +138,110 @@ private void populateLibraryList(JList<String> list, Set<CloudLibrary> libraries
...
+    }
+
+    @Override
+    public boolean isCellEditable(int rowIndex, int columnIndex) {
+      return columnIndex == 1;
+    }
+
+    @Override
+    public Object getValueAt(int rowIndex, int columnIndex) {
+      if (columnIndex == 0) {
```
ivanporty: ditto: switch and constants

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
etanshaul: Oops that was for debugging. Thanks

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
etanshaul: This is actually doing post filtering based on the user’s selected libraries. I.e given the set of libraries the user selected, filter out the corresponding Role objects from the list returned by the API. Does your point still hold?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
etanshaul: I couldn’t find anything either.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
etanshaul: I think it needs to be contains because roleIds is a list of roles. Did I misunderstand your point?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
```diff
@@ -68,6 +117,14 @@
 
     populateLibraryList(apisToEnableList, apisToEnable);
     populateLibraryList(apisNotSelectedToEnableList, apisNotToEnable);
+
+    newServiceAccountCheckbox.addActionListener(
+        e -> serviceAccountDetailsPane.setVisible(((JCheckBox) e.getSource()).isSelected()));
```
ivanporty: Nit: this event handler feels a bit hidden in the setup code, maybe we could call a method, something like `handleNewServiceAccountClick()`? Might make it a bit easier to find what happens.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
etanshaul: I don’t think it is needed. I’ll check. Thanks

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialogTest.java
```diff
@@ -0,0 +1,138 @@
...
+                      librariesNotToEnable,
+                      ImmutableSet.of());
+
+              assertThat(dialog.getApisToEnablePanel().isVisible()).isFalse();
+              assertThat(dialog.getApisNotSelectedToEnablePanel().isVisible()).isTrue();
+            });
+  }
+
+  @Test
+  public void serviceAccountUi_whenSomeRolesExist_isVisible() {
```
ivanporty: We could also add a test later which checks that role table properly populates and sets default state for roles, WDYT?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
ivanporty: Oops sorry, I forgot it's a list, not a string, sorry.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
etanshaul: Looks like I need to fix the name of that method too :)

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialogTest.java
etanshaul: Good idea

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
ivanporty: I mean it could also be used in getting set of roles for a set of APIs not selected by user but for some other future cases, but it's very subtle reasoning. It's completely up to you, it's good as it is.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
etanshaul: I'd like to isolate the Credential management to the API class itself. That being said, some refactoring I'd like to do in a different PR is to make this class non-static and do the credential management only once.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
etanshaul: i think its ok without a switch; but im updating to use constants for the indexes

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
etanshaul: done

etanshaul:
@ivanporty I think I've either fixed or commented on all your points. PTAL.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialogTest.java
etanshaul: done

ivanporty:
Excellent, thanks, looks great!

patflynn:
:+1:  I like it.  For service account name what do you guys think of defaulting to the module name?  In GCP they don't have a context that they can use to guess an appropriate name, but perhaps we do.

etanshaul:
Yes, I like using the module name. How about module name plus a random string. e.g. `my-module-[some random string]` to avoid collisions (if they do it multiple times on their same module)

Henhouse creates something like this:
`starting-account-j0v866a5l5q5`

etanshaul:
(or timestamp instead of a random string)

etanshaul:
merged - I'll add the default service account name logic (plus other validation) in the next PR

etanshaul:
Actually, service account names don't have to be unique. So the module name would work just fine. The key itself though can be the module name plus a timestamp.

PR #1890, opened 2018-02-06
Managed SDK service implementation

ivanporty:
Implements:
* SDK service interface 
* background installation 
* low level logging
* cancellation and other housekeeping.

Not here, for the following PRs:
* UI and integration
* SDK auto-updates

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
@@ -36,7 +36,12 @@ public interface CloudSdkService {
 
   SdkStatus getStatus();
 
-  boolean installAutomatically();
+  /**
+   * Asks SDK service to attempt install and prepare Cloud SDK if it's not ready for use yet.
+   *
+   * @return true if install is supported and started, false if install is not supported.
+   */
+  boolean install();
 
   void addStatusUpdateListener(SdkStatusUpdateListener listener);
 
@@ -47,6 +52,7 @@ public interface CloudSdkService {
     NOT_AVAILABLE
   }
 
+  /** Interface to receive SDK service updates like changes from installing to available. */
   interface SdkStatusUpdateListener {
     void onSdkStatusChange(CloudSdkService sdkService, SdkStatus status);
   }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
@@ -45,6 +45,8 @@ public interface CloudSdkService {
 
   void addStatusUpdateListener(SdkStatusUpdateListener listener);
 
+  void removeStatusUpdateListener(SdkStatusUpdateListener listener);
+
   enum SdkStatus {
     READY,
     INSTALLING,
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
@@ -27,6 +27,9 @@ public interface CloudSdkService {
     return ServiceManager.getService(CloudSdkService.class);
   }
 
+  /** Called when this service becomes primary choice for serving Cloud SDK. */
+  void activate();
+
   @Nullable
   Path getSdkHomePath();
 
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
@@ -35,6 +35,11 @@ public class DefaultCloudSdkService implements CloudSdkService {
   private PropertiesComponent propertiesComponent = PropertiesComponent.getInstance();
   private static final String CLOUD_SDK_PROPERTY_KEY = "GCT_CLOUD_SDK_HOME_PATH";
 
+  @Override
+  public void activate() {
+    // TODO track event that custom SDK is activated and used.
+  }
+
   @Nullable
   @Override
   public Path getSdkHomePath() {
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
@@ -74,7 +74,7 @@ public class DefaultCloudSdkService implements CloudSdkService {
   }
 
   @Override
-  public boolean installAutomatically() {
+  public boolean install() {
     return false; /* not supported. */
   }
 
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
@@ -82,4 +82,9 @@ public class DefaultCloudSdkService implements CloudSdkService {
   public void addStatusUpdateListener(SdkStatusUpdateListener listener) {
     /* not supported. */
   }
+
+  @Override
+  public void removeStatusUpdateListener(SdkStatusUpdateListener listener) {
+    /* not supported. */
+  }
 }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
@@ -84,7 +84,7 @@ public class ManagedCloudSdkService implements CloudSdkService {
   @Override
   public boolean install() {
     // check if installation is already running first, to prevent multiple jobs running.
-    if (runningInstallationJob == null) {
+    if (runningInstallationJob == null || runningInstallationJob.isDone()) {
       runningInstallationJob =
           ThreadUtil.getInstance().executeInBackground(this::installSynchronously);
       Futures.addCallback(
@@ -92,9 +92,9 @@ public class ManagedCloudSdkService implements CloudSdkService {
           new FutureCallback<Path>() {
             @Override
             public void onSuccess(Path path) {
-              logger.info("Managed Google Cloud SDK successfully installed at: " + path);
-
-              runningInstallationJob = null;
+              if (path != null) {
+                logger.info("Managed Google Cloud SDK successfully installed at: " + path);
+              }
             }
 
             @Override
@@ -102,8 +102,6 @@ public class ManagedCloudSdkService implements CloudSdkService {
               if (t instanceof InterruptedException) {
                 logger.info("Managed Google Cloud SDK installation cancelled.");
               }
-
-              runningInstallationJob = null;
             }
           });
     }
@@ -140,6 +138,7 @@ public class ManagedCloudSdkService implements CloudSdkService {
   /**
    * Checks for Managed Cloud SDK status and creates/installs it if necessary.
    *
+   * @return Path of the installed managed SDK, null if error occurred while installing.
    * @throws InterruptedException if Managed Cloud SDK has been interrupted by {@link
    *     #cancelInstall()}
    */
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
@@ -140,6 +140,7 @@ public class ManagedCloudSdkService implements CloudSdkService {
   }
 
   @VisibleForTesting
+  // TODO(ivanporty) move into ThreadUtil common test code.
   void invokeOnApplicationUIThread(Runnable runnable) {
     ApplicationManager.getApplication().invokeLater(runnable);
   }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
@@ -16,7 +16,20 @@
 
 package com.google.cloud.tools.intellij.appengine.sdk;
 
+import com.google.cloud.tools.intellij.util.ThreadUtil;
+import com.google.cloud.tools.managedcloudsdk.ManagedCloudSdk;
+import com.google.cloud.tools.managedcloudsdk.ManagedSdkVerificationException;
+import com.google.cloud.tools.managedcloudsdk.ManagedSdkVersionMismatchException;
+import com.google.cloud.tools.managedcloudsdk.MessageListener;
+import com.google.cloud.tools.managedcloudsdk.UnsupportedOsException;
+import com.google.cloud.tools.managedcloudsdk.install.SdkInstaller;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Lists;
+import com.intellij.openapi.diagnostic.Logger;
+import java.lang.ref.WeakReference;
 import java.nio.file.Path;
+import java.util.Collection;
+import java.util.Iterator;
 import org.jetbrains.annotations.Nullable;
 
 /**
@@ -25,6 +38,19 @@ import org.jetbrains.annotations.Nullable;
  */
 // TODO(ivanporty) implementation coming in the next PR
 public class ManagedCloudSdkService implements CloudSdkService {
+  private static final Logger logger = Logger.getInstance(ManagedCloudSdkService.class);
+
+  private ManagedCloudSdk managedCloudSdk;
+
+  private SdkStatus sdkStatus;
+
+  private final Collection<WeakReference<SdkStatusUpdateListener>> weakRefListeners =
+      Lists.newArrayList();
+
+  /** Called when this service becomes primary choice for serving Cloud SDK. */
+  public void activate() {
+    install();
+  }
 
   @Nullable
   @Override
@@ -41,10 +67,74 @@ public class ManagedCloudSdkService implements CloudSdkService {
   }
 
   @Override
-  public boolean installAutomatically() {
+  public boolean install() {
+    ThreadUtil.getInstance().executeInBackground(this::installSynchronously);
     return true;
   }
 
   @Override
-  public void addStatusUpdateListener(SdkStatusUpdateListener listener) {}
+  public void addStatusUpdateListener(SdkStatusUpdateListener listener) {
+    weakRefListeners.add(new WeakReference<>(listener));
+  }
+
+  @VisibleForTesting
+  ManagedCloudSdk createManagedSdk() throws UnsupportedOsException {
+    return ManagedCloudSdk.newManagedSdk();
+  }
+
+  private void updateStatus(SdkStatus sdkStatus) {
+    this.sdkStatus = sdkStatus;
+    notifyListeners(this, sdkStatus);
+  }
+
+  private void notifyListeners(CloudSdkService sdkService, SdkStatus status) {
+    Iterator<WeakReference<SdkStatusUpdateListener>> refIterator = weakRefListeners.iterator();
+    while (refIterator.hasNext()) {
+      SdkStatusUpdateListener listener = refIterator.next().get();
+      if (listener == null /* GC-ed */) {
+        refIterator.remove();
+      } else {
+        listener.onSdkStatusChange(sdkService, status);
+      }
+    }
+  }
+
+  /** Checks for Managed Cloud SDK status and creates/installs it if necessary. */
+  private void installSynchronously() {
+    if (managedCloudSdk == null) {
+      try {
+        managedCloudSdk = createManagedSdk();
+      } catch (UnsupportedOsException osex) {
+        managedCloudSdk = null;
+        updateStatus(SdkStatus.NOT_AVAILABLE);
+        return;
+      }
+    }
+
+    boolean managedSdkInstalled = false;
+    try {
+      managedSdkInstalled = managedCloudSdk.isInstalled();
+    } catch (ManagedSdkVerificationException | ManagedSdkVersionMismatchException e) {
+      // TODO ??
+      logger.error("Error while checking Cloud SDK status", e);
+    }
+
+    if (!managedSdkInstalled) {
+      SdkInstaller sdkInstaller;
+      try {
+        sdkInstaller = managedCloudSdk.newInstaller();
+      } catch (UnsupportedOsException e) {
+        updateStatus(SdkStatus.NOT_AVAILABLE);
+        return;
+      }
+
+      MessageListener messageListener = rawString -> {};
+
+      try {
+        sdkInstaller.install(messageListener);
+      } catch (Exception ex) {
+        ex.printStackTrace();
+      }
+    }
+  }
 }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
@@ -29,6 +29,7 @@ import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.diagnostic.Logger;
 import java.nio.file.Path;
 import java.util.Collection;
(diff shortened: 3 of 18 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -16,35 +16,218 @@
...
+    }
+  }
+
+  @VisibleForTesting
+  ManagedCloudSdk createManagedSdk() throws UnsupportedOsException {
+    return ManagedCloudSdk.newManagedSdk();
+  }
+
+  @VisibleForTesting
+  void setLogger(Logger logger) {
```
etanshaul: Take a look at `UsefulTestCase`. Could you, instead of exposing a setter like this, do `Logger.setFactory` and pass in your own implementation if you need to?

etanshaul: Also I'm wondering about the general approach of asserting log statements here. For the "operation successful" case for example, wouldn't it be better to assert that success occurred instead (the correct callback was called or whatever)?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
```diff
@@ -35,6 +35,11 @@
   private PropertiesComponent propertiesComponent = PropertiesComponent.getInstance();
   private static final String CLOUD_SDK_PROPERTY_KEY = "GCT_CLOUD_SDK_HOME_PATH";
 
+  @Override
+  public void activate() {
```
etanshaul: is this just to be used for metrics?

ivanporty: For `ManagedSDKService`, it's a notification to check if installation is needed (i.e. first time install, broken existing managed SDK etc.). For default service, nothing except metrics so far.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: You might be right that it's too close testing even including logs. Success and failure of installation are tested via public APIs a lot in other unit tests (i.e. `activate_service_sdkInstalled_status_ready()` etc.). We can remove logging checks completely and check for messages when UI arrives.

ivanporty: As for `Logger.setFactory`, should we consider moving this as a default to `CloudToolsRule`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
etanshaul: do you mind elaborating a bit? how is this different than `install`; what is the workflow that calls `active` / `install`; etc. `install` is intuitive to me, but `activate` isn't in this context

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: do you mean so that it sets the `TestLoggerFactory`? I think that would make sense

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java
ivanporty: We are going to have two implementations - custom and managed SDK services. When a user goes to settings and changes implementation, `activate()` is called for newly selected and applied service. `install()` is called by it only by `ManagedSDKService`. `install()` will also be called by deployment / run configuration explicitly if status is not `READY` in attempt to prepare SDK immediately without going to settings.

etanshaul: oh i see! i forgot that the user can switch between managed and custom

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -16,35 +16,218 @@
...
   public SdkStatus getStatus() {
-    return SdkStatus.NOT_AVAILABLE;
+    return sdkStatus;
   }
 
   @Override
-  public boolean installAutomatically() {
+  public boolean install() {
+    // check if installation is already running first, to prevent multiple jobs running.
+    if (runningInstallationJob == null) {
```
etanshaul: ListenableFuture has `isDone` and `isCancelled`. Can these be used instead of checking for null and nulling it out?

ivanporty: Sounds good, this will make it cleaner, thanks.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -16,35 +16,217 @@
...
+    return sdkStatus;
   }
 
   @Override
-  public boolean installAutomatically() {
+  public boolean install() {
+    // check if installation is already running first, to prevent multiple jobs running.
+    if (runningInstallationJob == null || runningInstallationJob.isDone()) {
+      runningInstallationJob =
+          ThreadUtil.getInstance().executeInBackground(this::installSynchronously);
```
etanshaul: this is a little weird that you are calling `executeInBackground` and passing a runnable with a method called `installSynchronously`. Don't these two terms conflict?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -16,35 +16,217 @@
...
+
+  /**
+   * Checks for Managed Cloud SDK status and creates/installs it if necessary.
+   *
+   * @return Path of the installed managed SDK, null if error occurred while installing.
+   * @throws InterruptedException if Managed Cloud SDK has been interrupted by {@link
+   *     #cancelInstall()}
+   */
+  @VisibleForTesting
+  Path installSynchronously() throws InterruptedException {
```
etanshaul: this is a pretty long method. is it possible to refactor for clarity?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: My idea was `install()` starts background job, so `installSynchronously()` runs as is, but you're right, it sounds like a loaded name. With your next comment in mind, how about `installManagedSdk()`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: Good idea. How about: `installManagedSdk()` -> calling two smaller methods `installSdk()` and `installAppEngineJava()`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: yeah that sounds good. also can `installManagedSdk` be private? since you are using ThreadUtils to kick off the job, you can always swap out a synchronous executor for unit tests so you can test the visible `install` method instead

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: Filed in a separate issue/PR.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: It was private until I included test for interrupted exception which needs to be handled properly and means explicit process cancel. So I need access to it to mock and make it throw that exception (see `interruptedInstall_status_notAvailable` unit test). I'm open to suggestions if we can make it via public API.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: yea. that would already be a nice improvement

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: ManagedCloudSdk is part of the core library right? Could you, instead, find a way to pass in a mock of this and then ultimately mock it up so that sdkInstall#install throws the InterruptedException?

ivanporty: Cool, this should work, thanks!

ivanporty:
@etanshaul I think your review suggestions are addressed, please let me know of you have anything else. Thanks!

etanshaul:
I'm on it

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkServiceTest.java
```diff
@@ -0,0 +1,223 @@
...
+  @Spy private ManagedCloudSdkService sdkService;
+
+  @Mock private ManagedCloudSdk mockManagedCloudSdk;
+
+  @Mock private CloudSdkService.SdkStatusUpdateListener mockStatusUpdateListener;
+
+  @Before
+  public void setUp() throws UnsupportedOsException {
+    doReturn(mockManagedCloudSdk).when(sdkService).createManagedSdk();
+    sdkService.setLogger(new TestInMemoryLogger());
```
etanshaul: Is this custom logger still needed?

ivanporty: Looks like we need this to avoid using standard IJ logging system - when it logs exceptions it prevents some tests to finish successfully. I plan to remove that when I fix and update the PR to use test logging factory for all `CloudToolsRule` tests.

etanshaul: thats what I suspected. Its those `log#error` calls which throw exceptions on the non-test logger. Sounds good. Maybe a `todo` here to remove this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
```diff
@@ -27,6 +27,9 @@ static CloudSdkService getInstance() {
     return ServiceManager.getService(CloudSdkService.class);
   }
 
+  /** Called when this service becomes primary choice for serving Cloud SDK. */
+  void activate();
```
etanshaul: nit: how about setActive instead of activate?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -16,35 +16,234 @@
...
+ * <p>Some methods ({@link #install()} do their work on background thread but listeners methods are
+ * always called on EDT thread. This class is not thread-safe, must be used on UI thread.
  */
-// TODO(ivanporty) implementation coming in the next PR
 public class ManagedCloudSdkService implements CloudSdkService {
+  private Logger logger = Logger.getInstance(ManagedCloudSdkService.class);
+
+  private ManagedCloudSdk managedCloudSdk;
+
+  private SdkStatus sdkStatus = SdkStatus.NOT_AVAILABLE;
```
etanshaul: its a little odd to see the status being set like this as a member variable. What happens if the SDK was already installed; what updates the status to READY? I wonder if this can be refactored a bit for clarity.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -16,35 +16,234 @@
...
+          new FutureCallback<Path>() {
+            @Override
+            public void onSuccess(Path path) {
+              if (path != null) {
+                logger.info("Managed Google Cloud SDK successfully installed at: " + path);
+              }
+            }
+
+            @Override
+            public void onFailure(Throwable t) {
```
etanshaul: why not let the exceptions trickle up to this callback instead of catching them all in the install runnable?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -16,35 +16,234 @@
...
+    }
+
+    updateStatus(SdkStatus.INSTALLING);
+
+    Path installedManagedSdkPath;
+    try {
+      installedManagedSdkPath = installSdk();
+    } catch (InterruptedException iex) {
+      updateStatus(SdkStatus.NOT_AVAILABLE);
+      throw iex;
```
etanshaul: follow up to my other comment: it seems a little bit inconsistent that you are throwing this one, but swallowing the other ones and returning null instead.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkServiceTest.java
ivanporty: good idea, I'll add `todo`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -16,35 +16,234 @@
...
+          CommandExitException, SdkInstallerException, UnknownArchiveTypeException {
+    boolean managedSdkInstalled = false;
+    try {
+      managedSdkInstalled = managedCloudSdk.isInstalled();
+    } catch (ManagedSdkVerificationException | ManagedSdkVersionMismatchException ex) {
+      logger.error("Error while checking Cloud SDK status, will attempt to re-install", ex);
+    }
+
+    if (!managedSdkInstalled) {
+      MessageListener sdkInstallListener = logger::trace;
```
etanshaul: trace for debugging?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: It does feel filled with various exceptions, I discussed with Appu which ones means what and it's not pretty.
There are a few things here why this happens:
* InterruptedException is not an error but sign that process was gracefully cancelled so that's the reason it goes out to the top and will be logged in UI
* ManagedSdkVerificationException | ManagedSdkVersionMismatchException are not fatal and means SDK was there but it failed to execute status check on them - it just means we go ahead and re-install it. Those happen for both base SDK and App Engine Java and we need to log different messages about them.
* Other exceptions are fatal.

So it's possible `onFailure()` can become really ugly method, especially when we need to distinguish between SDK / App Engine Java component errors. WDYT?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: SDK installation logs a lot, including all of the CLI style progress (i.e. adding `=` into new line etc.). But still it's probably important info, I agree we can make it `debug`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: See my next comment please.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
ivanporty: I think this is good, except maybe seeing `setActive` could imply we want to pass a property there, like `setActive(true|false)`, could be a bit misleading?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: So currently the progress of the installation is simply done in the background and logged to the logger (which the user will probably never look at) correct? What do you think about enhancing this (later) to pipe the progress somewhere visible in the IDE. Something along the lines of what it does when syncing gradle:

screen shot 2018-02-08 at 11 29 36 am

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
etanshaul: yeah thats true. I'm ok leaving it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: Yes, that'd be great, I believe Appu just finished new progress bar with meaningful user-visible steps  instead of dumping all of CLI messages, we'll use it to provide step-by-step progress information when we provide UI.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: Status is updated to `READY` when `activate()` is called, so before that we're not checking SDK and assume we're not available to avoid extra work and possible errors. Maybe move this assignment to constructor?

etanshaul: so just at a higher level:

when is activate called: Lets say the user is using a managed SDK. Its installed, and activated, and therefore the status is updated to READY. What happens the next time the user boots up the IDE from scratch?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: hmm thinking about this

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: So when IDE is booted, plugin init code is called, and it checks which SDK service is active, then creates it and calls `activate()`. SDK service checks its status.
Initially I was thinking about doing this in a constructor but it's a lot of work which is not what constructors are usually for.

etanshaul: can you point me to that initialization code?

etanshaul: or is that coming next?

ivanporty: Correct, in next PRs, something like:
```java
CloudToolsPluginInitializationComponent {
...
SdkService sdkService = createSdkServiceFromSettings(...);
sdkService.activate();
CloudSdkService.setInstance(sdkService);
```

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -16,35 +16,234 @@
...
+  ManagedCloudSdk createManagedSdk() throws UnsupportedOsException {
+    return ManagedCloudSdk.newManagedSdk();
+  }
+
+  @VisibleForTesting
+  void setLogger(Logger logger) {
+    this.logger = logger;
+  }
+
+  @VisibleForTesting
```
etanshaul: Could this (having to expose this for testing) be solved by expanding what we have in ThreadUtils? Specifically having a swappable executor that runs on the EDT?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: got it. so just to confirm: active / install will have to be expanded in the next PRs to initialize the status (and skip install if already READY)?

ivanporty: correct, init code will just do `activate()`, and Managed SDK service will skip install if current SDK appears to be valid (i.e. isInstalled() == true)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: Yes I thought that too, it's pretty useful considering our way of testing UI related things. I searched for existing code in IJ testing framework that might do this but didn't find it. We can add something similar to ThreadUtil. How about a separate PR for that?

etanshaul: yeah separate pr is fine. maybe add a todo on it then

ivanporty: done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: > InterruptedException is not an error but sign that process was gracefully cancelled so that's the reason it goes out to the top and will be logged in UI

why not just do the log.info in the same place rather then letting the exception propagate out then? I just think there is a slight readability / consistency issue here

etanshaul: ah I think I see part of your dilemma. you need to handle equivalent exceptions differently if its installing the sdk vs installing the java component

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: sounds good. im ok with this how it is for now then. Will take another look in the next PRs when its expanded.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: Yes, that was the main reason - same exceptions for two different processes (we can combine them but then can loose exact understanding of what happened.)
As for handling `InterruptedException` it's handled on top level since I expect to call UI (notification bubble) for it, so I didn't want to call UI notification code from `install()` method family. I guess that would be cleaner with UI updates in place but that could have made PR too big.

ivanporty:
thanks, I agree, we might come up with a finalized solution once we have update and UI code in place.

PR #1891, opened 2018-02-06
Add validation and default values to service account management form

etanshaul:
- Add validation on the service account name and download path fields. Uses the built in dialog wrapper validation mechanism. When first opening the dialog, if in an error state, then the message is not shown until clicking "ok'
image
image

- defaults the service account name to the name of the module:
image

- attempts to default the download path to the user's home directory; if this fails, it will leave it empty (and the above validation will warn the user)

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
@@ -160,10 +160,13 @@ public class CloudApiManagementConfirmationDialog extends DialogWrapper {
         return new ValidationInfo(
             GctBundle.message("cloud.apis.management.dialog.serviceaccount.name.error"),
             serviceAccountNameTextField);
-      } else if (StringUtils.isEmpty(serviceKeyPathSelector.getText())
-          || !isValidDirectory(serviceKeyPathSelector.getText())) {
+      } else if (StringUtils.isEmpty(serviceKeyPathSelector.getText())) {
         return new ValidationInfo(
-            GctBundle.message("cloud.apis.management.dialog.serviceaccount.key.path.error"),
+            GctBundle.message("cloud.apis.management.dialog.serviceaccount.key.path.empty.error"),
+            serviceKeyPathSelector);
+      } else if (!isValidDirectory(serviceKeyPathSelector.getText())) {
+        return new ValidationInfo(
+            GctBundle.message("cloud.apis.management.dialog.serviceaccount.key.path.invalid.error"),
             serviceKeyPathSelector);
       }
     }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
@@ -39,6 +39,7 @@ import com.intellij.util.ui.UIUtil;
 import java.awt.Component;
 import java.awt.Graphics;
 import java.awt.event.ActionListener;
+import java.io.File;
 import java.util.Comparator;
 import java.util.Map.Entry;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


etanshaul:
And to make it explicit:

Most of the time, the dialog will not open up in an error state and user can just click "ok" immediately. The only time that may require user action is if the platform dependent home path cannot be resolved.

ivanporty:
I like the UI, the flow looks logical, great! Code review starts soon as well.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
```diff
@@ -127,12 +136,38 @@
...
 
+  @Nullable
+  @Override
+  protected ValidationInfo doValidate() {
+    if (newServiceAccountCheckbox.isSelected()) {
+      if (StringUtils.isEmpty(serviceAccountNameTextField.getText())) {
+        return new ValidationInfo(
+            GctBundle.message("cloud.apis.management.dialog.serviceaccount.name.error"),
+            serviceAccountNameTextField);
+      } else if (StringUtils.isEmpty(serviceKeyPathSelector.getText())) {
```
nbashirbello: What happens if the service key path is not valid? Like when a user types in an invalid path

etanshaul: yeah good point I should probably check to make sure its a valid directory too

etanshaul: done: e1f86da

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
```diff
@@ -143,6 +180,11 @@ private ActionListener newServiceAccountClickHandler() {
     return e -> serviceAccountDetailsPane.setVisible(((JCheckBox) e.getSource()).isSelected());
   }
 
+  private String getServiceKeyPathSelectorDefaultText() {
```
ivanporty: Nit: it feels from this method name like we are getting just a text , although it's a default path, maybe something like `getDefaultServiceKeyPath()`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
```diff
@@ -153,6 +195,11 @@ private void createUIComponents() {
     roleTable = new ServiceAccountRolesTable(roles);
   }
 
+  private static boolean isValidDirectory(String path) {
+    File file = new File(path);
+    return file.exists() && file.isDirectory() && file.canWrite();
```
ivanporty: We could consider slightly shorter option with `Files.isWritable(path)` that combines all calls together.

etanshaul: I thought of that, but that takes in a `Path` which would we need to create via `Paths#get` which then leads to catching exceptions. Do you still think its better? or is there a shorter way you were thinking?

ivanporty: Oh you mean a user could enter a total nonsense and `Paths.get` would throw an exception? Yeah, then probably it's easier to read without exceptions.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialogTest.java
```diff
@@ -166,4 +172,111 @@ public void roleTable_whenRolesExist_isPopulated_andAllSelectedByDefault() {
               assertThat(dialog.getSelectedRoles()).containsExactlyElementsIn(roles);
             });
   }
+
+  @Test
+  public void serviceAccount_whenFieldsCorrectlyPopulated_showsNoValidationError() {
+    ApplicationManager.getApplication()
+        .invokeAndWait(
```
ivanporty: What usually happens when a test running inside of `invokeAndWait` thread fails? Will it properly propagate failed assertions to main test thread?

etanshaul: yes it will

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
etanshaul: yea exactly. ok I'll leave it then

etanshaul:
@nbashirbello any other comments? I made the update to validate the path like you suggested

PR #1893, opened 2018-02-07
Create service account and download key from the IDE

etanshaul:
NOT yet ready for code review. You can review the screenshots for now.

- Creates a new service account based on the name specified by the user
- Adds the selected roles to this service account
- Creates a json key and saves it to the user's filesystem at the chosen path

**User configures the API as before, choosing the service account details and roles:**
screen shot 2018-02-07 at 5 43 17 pm

**After enabling the APIs, the service account is created with the following progress shown:**
screen shot 2018-02-07 at 5 44 01 pm

**After successful completion, the user is shown an info dialog:**
screen shot 2018-02-07 at 5 44 17 pm

**A notification bubble / event log message is shown as well:**
screen shot 2018-02-07 at 5 45 41 pm

**The key is now on the user's system at the chosen location:**
screen shot 2018-02-07 at 5 47 07 pm

**The service account is created on GCP**
screen shot 2018-02-07 at 5 49 10 pm


ivanporty:
Looks great! As a minor thoughts, I can think of a few things:
* I guess we can change the default icon to GCP one in the confirmation dialog for service account key file?
* It clearly says a user needs to add an environment variable into his local dev server  - is this the only or most used case? I mean one can deduce that this variable only applies to this case.
* Could we provide small copy icon next to variable / key file name to quickly copy variable and its value name? I see this frequently in GCP API pages, also in Git clone, and it helps a lot.
* Probably this is a larger effort, maybe for next PRs or separate features, could we add that variable into existing local dev server targets that exist in the current IDE project?

etanshaul:
replies inline:

- It clearly says a user needs to add an environment variable into his local dev server - is this the only or most used case? I mean one can deduce that this variable only applies to this case.
I need to check this again. It also ties into the local dev experience that @nbashirbello is working on.

- Could we provide small copy icon next to variable / key file name to quickly copy variable and its value name? I see this frequently in GCP API pages, also in Git clone, and it helps a lot.
Good Idea

- I guess we can change the default icon to GCP one in the service account confirmation dialog?
I'm not sure about that. But I'll check

- Probably this is a larger effort, maybe for next PRs or separate features, could we add that variable into existing local dev server targets that exist in the current IDE project?
Agreed. This is again part of what @nbashirbello is looking into

etanshaul:
@ivanporty @nbashirbello ready for code review please

@ivanporty while they are all still valid, I haven't yet addressed you're initial UX review points. I think they all can be added in separate PRs as they are all enhancements to the current functionality.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -119,6 +146,174 @@ static void enableApis(Set<CloudLibrary> libraries, CloudProject cloudProject, P
...
+
+      setProgress(
+          progress,
+          GctBundle.message("cloud.apis.service.account.download.key.progress.message"),
+          step / numSteps);
+      Path keyPath = writeServiceAccountKey(serviceAccountKey, downloadDir, cloudProject);
+
+      notifyServiceAccountCreated(project, name, keyPath);
+    } catch (IOException e) {
+      e.printStackTrace();
```
etanshaul: working updating exception handling

etanshaul:
As discussed today, I'm going to update the confirmation dialog that pops up after downloading the service account key. It won't have the default canned dialog behavior (with the Java icon) and it will allow the user to easily copy the env vars to the clipboard

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -119,6 +146,175 @@ static void enableApis(Set<CloudLibrary> libraries, CloudProject cloudProject, P
...
+
+      showProgress(
+          progress,
+          GctBundle.message("cloud.apis.service.account.download.key.progress.message"),
+          step / numSteps);
+      Path keyPath = writeServiceAccountKey(serviceAccountKey, downloadDir, cloudProject);
+
+      notifyServiceAccountCreated(project, name, keyPath);
+    } catch (IOException e) {
+      LOG.warn(
```
ivanporty: Could we also show this error to the user directly, using the same notification/bubble approach?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -194,14 +411,31 @@ private static void notifyApiEnableSkipped(Set<CloudLibrary> apis, Project proje
...
+
   private static String joinApiNames(Set<CloudLibrary> apis) {
     return apis.stream().map(CloudLibrary::getName).collect(Collectors.joining("<br>"));
   }
 
-  private static void setApiEnableProgress(
-      ProgressIndicator indicator, String apiName, String cloudProjectName, double fraction) {
-    indicator.setText(
-        GctBundle.message("cloud.apis.enable.progress.message", apiName, cloudProjectName));
+  private static void showProgress(ProgressIndicator indicator, String message, double fraction) {
```
ivanporty: Nit: it feels like this method shows the progress (like it wasn't showing before), maybe `updateProgress()`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -119,6 +146,175 @@ static void enableApis(Set<CloudLibrary> libraries, CloudProject cloudProject, P
...
+          progress,
+          GctBundle.message("cloud.apis.service.account.create.key.progress.message"),
+          step / numSteps);
+      step++;
+      ServiceAccountKey serviceAccountKey = createServiceAccountKey(user.get(), serviceAccount);
+
+      showProgress(
+          progress,
+          GctBundle.message("cloud.apis.service.account.download.key.progress.message"),
+          step / numSteps);
```
ivanporty: Nit: If I get it right, here `step == numSteps` when last action is run, so it shows download message when progress bar is already full? Maybe add another step and fill progress bar after download is complete?

etanshaul:
I updated the service account confirmation dialog:

image

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -119,6 +146,175 @@ static void enableApis(Set<CloudLibrary> libraries, CloudProject cloudProject, P
...
+      return;
+    }
+
+    ProgressIndicator progress = ProgressManager.getInstance().getProgressIndicator();
+
+    try {
+      int numSteps = roles.isEmpty() ? 3 : 4;
+      double step = 1;
+
+      showProgress(
```
ivanporty: It looks like this method is running on thread provided by `ProgressManager.runProcessWithProgressSynchronously()`, could you recall if progress bar allow calling its method from non-UI threads? I.e. `showProgress()`calls `setText()` etc. from here.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
etanshaul: I'm now actually debating changing this to a log.error so that an IDE level error surfaces and we get notified. This would be in the case of an API failure so it seems like a good case for us to be alerted. wdyt?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -119,6 +146,175 @@ static void enableApis(Set<CloudLibrary> libraries, CloudProject cloudProject, P
...
+      Set<Role> roles,
+      CloudProject cloudProject)
+      throws IOException {
+    CloudResourceManager resourceManager =
+        GoogleApiClientFactory.getInstance().getCloudResourceManagerClient(user.getCredential());
+
+    Policy existingPolicy =
+        resourceManager
+            .projects()
+            .getIamPolicy(cloudProject.projectId(), new GetIamPolicyRequest())
```
ivanporty: Looks like this is a policy for a whole cloud project, I guess there is no way to get a policy and bindings only for a specific service account, correct?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
etanshaul: what about changing the starting index to 0 instead (like it was before for enablement). that way the 1st step will show the progress as completely unstarted, and the last step will be 1 less than full. wdyt? 

For some reason, I initially wanted to change it so that the first step didn't show as unstarted but I guess it can be looked at both ways.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -119,6 +146,175 @@ static void enableApis(Set<CloudLibrary> libraries, CloudProject cloudProject, P
...
+
+    List<Binding> additionalBindings =
+        roles
+            .stream()
+            .map(
+                role -> {
+                  Binding binding = new Binding();
+                  binding.setRole(role.getName());
+                  binding.setMembers(
+                      ImmutableList.of(
```
ivanporty: Nit: feels slightly complex, maybe a simple helper method which describes what that list is for, something like `createPolicyBindingsForServiceAccount()`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
etanshaul: thats a good question. The platform isn't raising any EDT related exceptions like it typically does so I am assuming that ProgressManager internally handles doing things on the correct thread.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -119,6 +146,175 @@ static void enableApis(Set<CloudLibrary> libraries, CloudProject cloudProject, P
...
+   * @param downloadDir the {@link Path} on the file system to download the ky
+   * @param cloudProject the {@link CloudProject} associated with this service account
+   * @return the {@link Path} to the service account key that was written
+   * @throws IOException if the an IO error occurs when writing the file
+   */
+  private static Path writeServiceAccountKey(
+      ServiceAccountKey key, Path downloadDir, CloudProject cloudProject) throws IOException {
+    Path keyPath =
+        Paths.get(
+            downloadDir.toString(), cloudProject.projectName() + "-" + getTimestamp() + ".json");
```
ivanporty: What if we consider using actual service account name for a file name (i.e. name we generated from `createServiceAccountId()`?

ivanporty: For `cloudProject.projectName()`, will it always be compatible with file name, i.e. we won't suddenly generate `Paths.get()` exception?

etanshaul: yeah that sounds reasonable. I went with the cloud project name since that is probably the most immediately recognizable name to the user. But it creates ambiguity when there are multiple service accounts. 

What about a combination of the cloud project name and the service account name?

etanshaul: good question. ill check what restriction there is on projectname

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
ivanporty: Oh you mean this would look like plugin level error when it shows report to google? Looks OK, except won' they be offered an option to disable plugin in this dialog, could be a bit harsh?

etanshaul: thats true. and we probably don't want to treat things that may fail as a result of transient network issues as an IDE level error. I'll look into surfacing it to the user as a warning instead.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
ivanporty: I think it's good either way - adding 1 more extra step or showing first step as 0 complete, so I leave it up to you. Probably initial value of 0 is a bit easier to read.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
ivanporty: Combination sounds good to me, could be a bit long, but least amount of ambiguity for sure.

ivanporty:
Nice dialog! The icon sort of looks very familiar as copy action, do we need the long description at all? Maybe just leave an icon and provide a tooltip with text?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
ivanporty: Awesone, this makes it much easier to use!

etanshaul:
which long description do you mean? the "copy to clipboard" wording?

ivanporty:
Yep,  "copy to clipboard", not really long, but still adds to that line length visibly.

etanshaul:
how about just "copy". I without the wording it might look a little weird (unless you think users are used to it?)

etanshaul:
I think it looks find without the wording:

image

Lets go with that.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
etanshaul: my understanding is that the Policy is at the cloud project level. The bindings (where the roles are defined), however, can be scoped to a service account. This is why in the next few lines of code, the new bindings with the new roles are scoped to the service account (meaning it is the service account that will be requesting access to the resources).

I will double check this though.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
etanshaul: updated

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyDisplayDialog.java
```diff
@@ -0,0 +1,59 @@
...
+  private static final String CREDENTIAL_ENV_VAR_KEY = "GOOGLE_APPLICATION_CREDENTIALS";
+
+  ServiceAccountKeyDisplayDialog(@Nullable Project project, String downloadPath) {
+    super(project);
+    init();
+
+    setTitle(GctBundle.message("cloud.apis.service.account.key.downloaded.title"));
+    downloadPathLabel.setText(downloadPath);
+
+    String credentialEnvVar = CREDENTIAL_ENV_VAR_KEY + "=" + downloadPath;
```
ivanporty: Nit: maybe `String.format()` and bake the whole assignment string into a constant?

etanshaul: sounds good.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -162,6 +363,27 @@ private static void enableApi(
...
+   *
+   * <p>Must be less than 30 characters.
+   *
+   * @param prefix the name chosen by the user for the service account
+   * @return an ID that is a combination of the prefix and a timestamp
+   */
+  private static String createServiceAccountId(String prefix) {
+    int maxLen = 30;
+    String timestamp = getTimestamp();
+    int maxPrefixLen = maxLen - timestamp.length() - 1;
```
ivanporty: UX Nit: we could possibly define max prefix length as a constant based on timestamp length and limit text field a user types service account name into to that length so they won't be surprised by trimming.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
etanshaul: I decided to keep cloud project as the name of the key for the following reasons:
- its too long with cloud project and service account, like you said
- the web console also generates keys with the same strategy (cloud project name + some unique id).

sound ok to you?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
etanshaul: and to add to this, it appears as though there is a restriction to the pattern that the service account name can take.

I think I'll create a new issue to update this validation (including a regex on the service account field) to fix in a separate PR.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
ivanporty: Sounds good.

etanshaul:
to organize my thoughts on what is remaining from the review comments:

- surface api failures as messages to the user instead of just logs
- defer service account name field regex validation to separate PR - file issue
- Check is the CloudProject name could potentially break service account key file writing (if the name could contain characters that would be invalid)
- for adding roles to service account - double check if its possible to fetch an existing policy at the service account level instead of the cloud project level

hopefully that covers everything

ivanporty:
I think you covered everything in your list, thanks!

P.S. How did you make checkbox list, this is great!?

etanshaul:
the checkboxes are a new feature release sometime last year I think: https://github.com/blog/1375-task-lists-in-gfm-issues-pulls-comments

etanshaul:
@ivanporty ptal again.

I think I addressed all the checkboxes above. 

For #3 the console prevents creating projects with slashes; so I think that covers your point (it also limits the characters to numbers, letters, spaces, quotes, and `!`.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudApiManagerTest.java
```diff
@@ -0,0 +1,192 @@
...
+  @Mock private ServiceAccounts.Create serviceAccountCreate;
+  @Mock private Keys.Create keysCreate;
+
+  @Mock Notifications notifications;
+
+  private ServiceAccount serviceAccount;
+  private ServiceAccountKey serviceAccountKey;
+  private Policy policy;
+
+  @TestDirectory(name = "downlaodDir")
```
ivanporty: Nit: download misspelled?

ivanporty:
For #4 I guess there's no way to do it, right? Otherwise, looks great!

etanshaul:
yeah for #4 its simply the way IAM works; project and service account are different concepts. `serviceAccount` is like a user: we fetch the policy for our cloud project, and then grant this "user" access to the set of roles we want and apply that to our resource (the cloud project in this case).

PR #1894, opened 2018-02-12
Managed SDK update and notifications

ivanporty:
PR contains:
* clean up of exception handling - we discussed with @etanshaul it doesn't feel right. I refactored the code to avoid fine-grained exception handling in favor of unified approach. Logs should give enough additional details.
* added SDK update support
* added initial IDE notifications/balloons support

Not in this PR:
* progress bars (next PR, pending better support in `appengine-plugins-core`)
* automatic update scheduling
* UI for settings

How to see in action:
You can add
`new ManagedCloudSdkService().activate();` line into `CloudToolsPluginInitializationComponent.initComponent()` method to install managed SDK and see notifications.
This will just install it, it will not be used for deployments yet.


etanshaul:
quickly skimmed this and looks good. I'll do a full review today.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -123,9 +102,9 @@ public void removeStatusUpdateListener(SdkStatusUpdateListener listener) {
     statusUpdateListeners.remove(listener);
   }
 
-  public void cancelInstall() {
-    if (runningInstallationJob != null) {
-      runningInstallationJob.cancel(true);
+  public void cancelInstallOrUpdate() {
```
etanshaul: this will get called by code in your next PR when you implement progress bars, correct?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -145,71 +124,57 @@ void invokeOnApplicationUIThread(Runnable runnable) {
...
-      installAppEngineJavaComponent();
-    } catch (InterruptedException iex) {
-      updateStatus(SdkStatus.NOT_AVAILABLE);
-      throw iex;
-    } catch (Exception ex) {
-      logger.error("Error while installing managed Cloud SDK App Engine Java component", ex);
-      updateStatus(SdkStatus.NOT_AVAILABLE);
-      return null;
+    if (managedSdkBackgroundJob == null || managedSdkBackgroundJob.isDone()) {
+      updateStatus(SdkStatus.INSTALLING);
```
etanshaul: there doesn't need to be an SdkStatus#UPDATING?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -145,71 +124,60 @@ void invokeOnApplicationUIThread(Runnable runnable) {
...
-    } catch (Exception ex) {
-      logger.error("Error while installing managed Cloud SDK", ex);
-      updateStatus(SdkStatus.NOT_AVAILABLE);
-      return null;
+  /**
+   * Runs managed SDK install/update code on background job and handles success/errors.
+   *
+   * @param jobType Install/update
+   * @param managedSdkTask Task to execute.
+   * @return True of task started, false if Managed SDK is not supported at all.
```
etanshaul: nit:

"True [if] task started.."

also

true/false -> {@code true/false}

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -236,6 +196,33 @@ private void installAppEngineJavaComponent()
...
+      managedCloudSdk.newUpdater().update(sdkUpdateListener);
+    }
+
+    return managedCloudSdk.getSdkHome();
+  }
+
+  /**
+   * Checks SDK status (installed, up-to-date, has component), catching non-fatal exceptions which
+   * does not mean SDK cannot be updated or re-installed. Other exceptions are propagated to the
+   * caller.
```
etanshaul: Is there example here of these exceptions being propagated to and caught by the caller? or is that in the next PRs?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -248,4 +235,58 @@ private void updateStatus(SdkStatus sdkStatus) {
...
+
+    private void handleJobCancellation() {
+      logger.info("Managed Google Cloud SDK install/update cancelled.");
+      // interrupted install means not available, but interrupted update means SDK still available.
+      switch (jobType) {
+        case INSTALL:
+          updateStatus(SdkStatus.NOT_AVAILABLE);
+          break;
+        case UPDATE:
+          updateStatus(SdkStatus.READY);
```
etanshaul: can we relatively safely assume that when an update is cancelled, the SDK is still in a "Ready" state (and not corrupted or something)?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -55,12 +52,13 @@
...
 
-  private volatile ListenableFuture<Path> runningInstallationJob;
+  private volatile ListenableFuture<Path> managedSdkBackgroundJob;
 
   @Override
   public void activate() {
     // TODO track event that custom SDK is activated and used.
 
+    initManagedSdk();
     install();
```
etanshaul: I may have asked the same thing in the last PR (sorry); but why does activate call `install`? What if its already installed?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: yes, exactly

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: So far I haven't thought of a specific use case for install vs. update for SDK service external users. `INSTALLING` says it's busy until `READY`. The only difference might be for messages we're showing to the user - how about we leave it like this for now, and when we implement UI checking for status, decide if we need additional status?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: Good question, let's ask Appu about that.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: `install()` checks first if it's already installed and then checks if App Engine Component is already installed. It may sound a bit ambigious, maybe introduce `checkInstallStatus()` method or something  like that, or bake that into `initManagedSdk()`?

etanshaul: is the idea that it will also update it if its out of date (and auto-update is enabled)?

I think extracting it out into the init method makes sense. It will make it more explicit and easier to trace whats going on.

ivanporty: Yes, I think we planned to do update as well when plugin starts or managed SDK is activated first time. I think it's a good idea to add a new method.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: Appu and Mark said most of the time this should true, but on Windows it might break badly in some case. I'm going to add an additional check here for `SDK.isInstalled()` to ensure SDK still in good shape, wdyt?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: There's a few examples in the unit tests (IOException etc.). Do you mean we might add them into javadoc?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: agree.

the next question I guess is: if it breaks badly in windows like you said, are we even in a recoverable state? In other words, can we just call `update` or `install` again later and hope it works?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: I didn't mean add them to the javadoc. I just wanted get an idea of how we going to handle these exceptions that are not caught (in non-test code).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: I think it should work - if it doesn't return true in `isInstalled()`, we set status to `NOT_AVAILABLE` and call `install()` next time we need managed SDK - this should erase broken SDK and re-install it fresh.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: As of now all error handling is in `ManagedSdkJobListener.onFailure()`

etanshaul: oh I missed that. looks good

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: sounds good

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -248,4 +234,66 @@ private void updateStatus(SdkStatus sdkStatus) {
...
+    public void onFailure(Throwable t) {
+      if (t instanceof InterruptedException) {
+        logger.info("Managed Google Cloud SDK install/update cancelled.");
+
+        ManagedCloudSdkServiceUiPresenter.getInstance().notifyManagedSdkJobCancellation(jobType);
+      } else {
+        logger.warn("Error while installing/updating managed Cloud SDK", t);
+
+        ManagedCloudSdkServiceUiPresenter.getInstance()
+            .notifyManagedSdkJobFailure(jobType, t.toString());
```
etanshaul: I think this makes sense to visibly show the user the throwable message here right? Otherwise it would be hard to diagnose without opening the logs. I'm going back and forth on this since I have a similar dilemma in my other PR.

ivanporty: Exactly, plus hopefully we will not hit these errors frequently on major platforms, and exception messages are pretty clear.

Issue #1896, opened 2018-02-13
Support a broader set of Jetbrains IDEs


patflynn:
related to #1472 and #1552 

Our plugin should be structured such that we can target as many Jetbrains SKUs as possible.

PR #1900, opened 2018-02-13
Merge core with account module

patflynn:
More incremental work towards #1896.

Merging these two modules turns out to be reasonable since we don't plan on supporting the account IDE integration independently of cloud and this simplifies our module structure greatly as they currently reach into each other a fair bit due to analytics and plugin info services.


ivanporty:
Looks great, I guess project selector and coming managed SDK can easily be moved into core plugin as well in next steps.

patflynn:
yep!

PR #1910, opened 2018-02-14
Improve service account name/id validation

### LINKED ISSUE #1895: Service account name needs more validation
There are character and length restrictions on this field that we can validate in the UI.

Currently, the name field is trimmed if it exceeds the length. And we do not validate acceptable characters.

etanshaul:
fixes #1895

- refactors the service account ID generator method
- improves the validation of the service account name field:
  - adds a max allowable characters check
  - validates against a regex (found in the output of the API call)


etanshaul:
working on some test coverage for this

etanshaul:
ready for review

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
```diff
@@ -174,9 +174,23 @@ Path getServiceAccountKeyDownloadPath() {
...
+            GctBundle.message(
+                "cloud.apis.management.dialog.serviceaccount.name.len.error",
+                CloudApiManager.SERVICE_ACCOUNT_NAME_MAX_LEN),
+            serviceAccountNameTextField);
+      } else if (!CloudApiManager.SERVICE_ACCOUNT_ID_PATTERN.matcher(name).matches()) {
         return new ValidationInfo(
-            GctBundle.message("cloud.apis.management.dialog.serviceaccount.name.error"),
+            GctBundle.message(
+                "cloud.apis.management.dialog.serviceaccount.name.regex.error",
+                CloudApiManager.SERVICE_ACCOUNT_ID_PATTERN.toString()),
```
ivanporty: It could be a bit unfriendly to show a regular expression when error is saying service account name is not valid, could we show something similar in wording, i.e. 'contains only characters, numbers etc.'?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -87,6 +88,10 @@
   private static final String SERVICE_ACCOUNT_ROLE_REQUEST_PREFIX = "serviceAccount:";
   private static final String SERVICE_ACCOUNT_KEY_FILE_NAME_FORMAT = "%s-%s.json";
 
+  private static final int SERVICE_ACCOUNT_ID_MAX_LEN = 30;
+  static final int SERVICE_ACCOUNT_NAME_MAX_LEN = 100;
```
ivanporty: Could you please expand a bit on the difference between ID and NAME length here and their use in validation?

etanshaul: Sure: these are 2 separate components of a Service Account. It has a name (which doesn't need to be unique, can be up to 100 characters, and can have any characters). The id is restricted to 30 chars, must be unique (hence the timestamp), and is subject to the regex in this PR. Does that answer the question?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
etanshaul: I thought about this, but there are other rules, like it cannot start with a number etc. This could get unwieldy to put in words so I just used the regex (since these are developers anyway). wdyt?

ivanporty: Might be a good question for UX person... I can image a scenario when something is wrong with the name, someone is shown a regex which he doesn't understand and it's kind of guessing game.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
ivanporty: Yes, thanks.
So if I understand correctly: we allow name to be at max 100 chars, then check for valid characters, then format it as `name-timestamp`, and then finally trim that to 30 chars?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
etanshaul: hmm. yeah I definitely see your concern. I'm tempted to leave it. @patflynn any opinion on this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
etanshaul: oh. i forgot to mention: the name is used as the seed for the id. So the name is maxed at 100. But then the ID is created based off of the name (capped at 30 characters including a timestamp that is added).

so yes, you are correct.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
patflynn: I kind of like a regular expressions expressivity, but I agree with @ivanporty that we can't expect our users to understand a regular expression.  I'd be OK with a verbal description and/or link to docs on what is valid.

etanshaul: ok. I'll use wording instead. I don't think this exists in the docs (i'll look). I discovered this pattern by executing (failed) API calls and seeing the message in the API response which printed out this regex.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
ivanporty: So it looks like minus the timestamp, we only include ~20 chars from the original name into ID, correct? But we still use the name for the service account for presentation purposes, i.e. its visible in GCP Console?

etanshaul: yes. I'm also following the strategy used by the web console (of including the filtered name as part of the id). We could potentially not include the name and bypass all of this regex validation mess.

ivanporty: Ok, got it, thanks, I think it's still good to keep at least part of name in the ID - probably much easier to identify later.

etanshaul: Yeah. I'll update the user-visible message to not use the regex and we'll leave in this naming strategy for now.

etanshaul:
@ivanporty @patflynn updated the user visible message:
image

etanshaul:
worth mentioning: I didn't not cover ALL the things that are allowed (since the wording would get too complicated), but I think that's ok.

patflynn:
:+1:

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManagementConfirmationDialog.java
```diff
@@ -174,9 +174,22 @@ Path getServiceAccountKeyDownloadPath() {
   @Override
   protected ValidationInfo doValidate() {
     if (createNewServiceAccountCheckbox.isSelected()) {
-      if (StringUtils.isEmpty(serviceAccountNameTextField.getText())) {
+      String name = serviceAccountNameTextField.getText().trim().toLowerCase();
```
ivanporty: Nit: just call `getServiceAccountName()`?

etanshaul: ah yeah. thats much better and safer

etanshaul: done - also did the same for service account path

PR #1913, opened 2018-02-15
We only need one initialization component for now.

patflynn:
The reason it lives outside of core is the end state can either be:
1- uber plugin module that depends on core+everything else with one initialization component handling everything. (+ this is nice to see what's going on)
or
2- initialization component in each vertical module.

We definitely don't want it in core as core should not depend on anything upstream. (verticals or uber plugin)


patflynn:
I'm actually second-guessing myself here. I was really thinking that all the initialization would happen in the uber-plugin, but maybe it'll be more manageable for the application components of each module to handle their own business (including login and analytics setup in the core)... 

You guys have thoughts on this?

etanshaul:
just so I understand - would what you are second-guessing affect this PR? In other words, are you now thinking that maybe the account plugin should have its own initialization?

patflynn:
em. kinda yeah.  Let's say I'm on the fence. I guess I just thought I'd waste everybody's time.

ivanporty:
> maybe the account plugin should have its own initialization?
As far as I understand, we don't have account plugin, it's part of `core`? Do we talk here about account login component of `core` then and the separate initialization component for it?

patflynn:
yeah sorry each module will often have it's own plugin.xml so I have the
tendency to think of them as plugins.

PR #1916, opened 2018-02-15
move cloud source repos classes out to its own package (and rename it from vcs)

patflynn:
- feel free to disagree on the final source/repos package name.
- vcs IMO is just too generic.


etanshaul:
Is the intent to have anything live in the `source` package, or everything under `source.repo`? If its the later, what about just combining to `sourcerepo`?

patflynn:
Yeah I considered that but I really hate java package names with words conjoined like that with no clear separator. :)

I'm half tempted to duplicate 'cloud' under intellij for GCP related resources.
e.g.
c.g.c.t.intellij.cloud.repos
c.g.c.t.intellij.cloud.storage
c.g.c.t.intellij.cloud.appengine
c.g.c.t.intellij.cloud.debugger
c.g.c.t.intellij.cloud.model|resources

etanshaul:
i like that actually. makes more sense to me to have packages that are there for name only higher up in the tree.

etanshaul:
plus the duplicated cloud sort of makes sense still: the first `cloud` is just the name of the tool

patflynn:
we could also use 'gcp'. let's chat offline.

ivanporty:
second `cloud` seems to be a bit vague, reads like this is some IntelliJ Cloud related? I like `gcp`

patflynn:
one gross result:
package com.google.**cloud**.tools.intellij.**cloud**.appengine.**cloud**;

patflynn:
while the idea is good I think coupling it to this work is going to be bad news as it's a huge amount of file movement.  How about I just rename source.repos to 'csr' and be consistent with the 'gcs' package for now?

ivanporty:
`csr` sounds good to me

etanshaul:
sgtm

patflynn:
updated.

patflynn:
@etanshaul just waiting on an approval for this one.

patflynn:
next PR should pull out enough to core that I can extract a feature module.

Issue #1921, opened 2018-02-22
Add BOM support to client library integration


etanshaul:
Overall issue to track work around assistance with the diamond dependency problem via BOM support.

PR #1924, opened 2018-02-26
Managed library dependency synchronization

etanshaul:
Part of the BOM integration effort (#1921)

Add synchronization for Maven projects allowing us to keep track of the set of managed dependencies added to the project (either via our UI, or externally).

Detailed design with bigger picture coming.

@ivanporty @patflynn @nbashirbello ptal


etanshaul:
ok tests are clean now

ptal @ivanporty @nbashirbello

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
```diff
@@ -0,0 +1,164 @@
...
+import org.jetbrains.idea.maven.project.MavenProjectsManager;
+import org.jetbrains.idea.maven.project.MavenProjectsManager.Listener;
+
+/**
+ * A {@link ProjectComponent} that maintains, in memory, the current set of {@link CloudLibrary}
+ * that are configured on the user's project.
+ */
+public class CloudLibraryProjectState implements ProjectComponent {
+
+  private static final Logger logger = Logger.getInstance(AddCloudLibrariesAction.class);
```
ivanporty: Looks like this logger might not be used?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
```diff
@@ -0,0 +1,164 @@
...
+                syncManagedProjectLibraries();
+              }
+            });
+  }
+
+  /**
+   * Returns an optional set of {@link CloudLibrary} currently configured on the given {@link
+   * Module}.
+   */
+  Set<CloudLibrary> getManagedLibraries(Module module) {
```
ivanporty: Looks like for now this method is only used in the tests, what is its further target, do we plan to maintain state of cloud libraries for a project depending on their version compatibility? I.e. for now it seems we derive this mapping from POM, the next step would be to enforce specific versions?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
```diff
@@ -0,0 +1,164 @@
...
+ */
+public class CloudLibraryProjectState implements ProjectComponent {
+
+  private static final Logger logger = Logger.getInstance(AddCloudLibrariesAction.class);
+
+  private final Project project;
+  private Map<Module, Set<CloudLibrary>> moduleLibraryMap = Maps.newHashMap();
+  private List<CloudLibrary> allLibraries;
+
+  private CloudLibraryProjectState(Project project) {
```
ivanporty: Interesting, so the IDE is able to instantiate this class from private constructor?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
etanshaul: correct. this is the main entry point into the state of the managed libraries for a given module. It will be used to control versions, like you said. Also for BOM stuff in the UI (where we are scoped to a specific module).

etanshaul: and yeah its only in tests now, but will be used in all those areas in subsequent PRs

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
etanshaul: yeah, I thought it was odd too (I initially had it public). But whatever they are doing allows for it to be private.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
```diff
@@ -0,0 +1,161 @@
...
+   *
+   * <p>If the project is not a maven project, then the listener simply isn't fired. If the project
+   * is not a maven project, but later becomes one, then the listener will be fired as expected and
+   * the managed libraries will be synchronized.
+   */
+  @Override
+  public void projectOpened() {
+    // todo (eshaul) handle Gradle & native projects and ensure that double syncing won't occur
+    MavenProjectsManager.getInstance(project)
+        .addManagerListener(
```
ivanporty: What would happen to the listener when project is closed and then re-opened, or re-loaded - does it have a well defined behavior? Does it need to be explicitly removed?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibrariesService.java
```diff
@@ -0,0 +1,44 @@
...
+  static CloudLibrariesService getInstance() {
+    return ServiceManager.getService(CloudLibrariesService.class);
+  }
+
+  /** Returns an immutable list of {@link CloudLibrary libraries}. */
+  List<CloudLibrary> getCloudLibraries() {
+    try {
+      return ImmutableList.copyOf(CloudLibraries.getCloudLibraries());
+    } catch (IOException e) {
+      logger.error(e);
```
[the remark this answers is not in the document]
etanshaul: this is a pretty catastrophic situation since we bundle the json ourselves. my thinking was that this is an appropriate situation for the plugin level error (and the feedback notification to us). wdyt?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
etanshaul: thanks, removed.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
ivanporty: So when we add library version control, we won't be able to cleanly derive this mapping from POM itself, and that's the reason we will keep this as a separate mapping?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
etanshaul: good question. My assumption was that the platform would be smart enough to close these "maven project" listeners when the project is closed. But i guess its dangerous to assume anything.

Ill test it out

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibrariesService.java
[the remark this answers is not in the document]
ivanporty: Agree, looks like broken build.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
etanshaul: sorry not sure I follow.

when a library is added (either via our UI and indirectly to the pom) or directly to the pom, then the maven project listener will be fired and force a resynchronization of this mapping. This sync will be called at the same time the user does a manual maven sync (or automatic if they have that setting enabled). This way, our knowledge of the state of the managed libraries is also in sync with the maven integration itself.

I'm not sure if I addressed your question.

ivanporty: Oh got it, we just update it on every listener update, I confused that with some cached state different from pom.xml file state. It's clear, thanks!

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryProjectStateTest.java
```diff
@@ -0,0 +1,242 @@
...
+            });
+  }
+
+  @Test
+  public void managedLibraries_withOnlyUnmanagedDependencies_isEmptyOptional() {
+    when(librariesService.getCloudLibraries())
+        .thenReturn(ImmutableList.of(LIBRARY.toCloudLibrary()));
+
+    ApplicationManager.getApplication()
+        .invokeAndWait(
```
ivanporty: Interesting, what is the reason for having `invokeAndWait` here, mostly because of `writeDependenciesToPom` or sync with `runReadAction` calls in `CloudLibraryProjectState`?

etanshaul: I believe its because the write action needs to be executed from the EDT

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryProjectStateTest.java
```diff
@@ -0,0 +1,242 @@
...
+            () -> {
+              Module module = createNewModule(id);
+              MavenId unmanagedDependency = new MavenId("my-group", "my-artifact", "1.0");
```
ivanporty: Unmanaged dependency is one without mapping to a cloud library?

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryProjectStateTest.java
```diff
@@ -0,0 +1,242 @@
...
+                  ModifiableModuleModel model =
+                      ModuleManager.getInstance(testFixture.getProject()).getModifiableModel();
+                  Module module = null;
+                  try {
+                    module = moduleBuilder.createModule(model);
+                  } catch (IOException
+                      | ModuleWithNameAlreadyExists
+                      | JDOMException
+                      | ConfigurationException e) {
+                    e.printStackTrace();
```
ivanporty: Could we rather throw `AssertionError` here since this is completely unexpected in a unit test?

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryProjectStateTest.java
etanshaul: yes. perhaps I could have used better naming. open to any suggestions to make it clearer.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryProjectStateTest.java
etanshaul: nice catch; fixed

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryProjectStateTest.java
ivanporty: Maybe `nonCloudLibraryDependency`?

etanshaul: sgtm. I guess I should also change `managedDependency` to `cloudLibraryDependency` to keep things consistent

etanshaul: and the names of the test methods themselves too

ivanporty: I agree, that would sound extremely clear to me!

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryProjectStateTest.java
```diff
@@ -0,0 +1,242 @@
...
+    setModuleNameAndRoot(testFixture.getProject().getBasePath());
+  }
+
+  @After
+  public void tearDown() {
+    MavenServerManager.getInstance().shutdown(true);
+  }
+
+  @Test
+  public void managedLibraries_withNoDependenies_isEmptyOptional() {
```
ivanporty: I think method name mentions `Optional` but actually tests for just collection size, probably can remove "Optional"?

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryProjectStateTest.java
```diff
@@ -0,0 +1,242 @@
...
+              Module module = createNewModule(id);
+              MavenId unmanagedDependency = new MavenId("my-group", "my-artifact", "1.0");
+              writeDependenciesToPom(module, ImmutableList.of(unmanagedDependency));
+
+              assertThat(state.getManagedLibraries(module)).isEmpty();
+            });
+  }
+
+  @Test
+  public void managedLibraries_withUnmanagedAndManagedDependencies_isPresent() {
```
ivanporty: Maybe instead of `_isPresent` -> `_isValidSet` or `_returnValidSet`? Same for the next method.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryProjectStateTest.java
etanshaul: Oops. It’s used to return an optional before refactoring :). I’ll fix these

etanshaul: fixed

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryProjectStateTest.java
etanshaul: updated

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryProjectStateTest.java
etanshaul: updated (to _returnsValidCloudLibrary)

etanshaul:
yeah. was just checking that now before merging. thanks for the thorough review.

etanshaul:
confirmed that there are no listener leaks

PR #1926, opened 2018-02-27
Use application wide message bus for project selector.

### LINKED ISSUE #1904: Sign in from CSR checkout doesn't complete
- Cloud Tools for IntelliJ version: master
- OS: linux

**What did you do?**
1. run gradlew runIde
1. make sure all projects are closed
1. make sure all your are fully google logged out from the IDE.
1. from welcome dialog select 'checkout from version control'
1. pick google cloud 
1. try to login from the  login widget
1. go through the browser song and dance until you see the successful login landing page
1. observe:
screenshot from 2018-02-14 10-10-46
 is still there waiting for login to complete.
1. I had to restart the process to find that the user was logged in. 
1. adding a new second logged completes except the logged in user does not appear in the project selector until you restart the process as well.

ivanporty:
Fixes #1904.

Project selection components now listen to application wide login events instead of set of open projects (which caused event not to be delivered in case of no open projects).

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
@@ -29,7 +29,7 @@ import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
-import com.intellij.openapi.project.ProjectManager;
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.ui.ComboBox;
 import com.intellij.openapi.ui.DialogWrapper;
 import com.intellij.openapi.ui.ValidationInfo;
@@ -50,7 +50,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.stream.Stream;
 import javax.swing.AbstractAction;
 import javax.swing.Action;
 import javax.swing.JButton;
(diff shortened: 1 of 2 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -122,17 +121,12 @@ CloudProject showDialog(@Nullable CloudProject cloudProject) {
...
-              project ->
-                  project
-                      .getMessageBus()
-                      .connect(dialogWrapper.getDisposable() /* disconnect once dialog is gone. */)
-                      .subscribe(
-                          GoogleLoginListener.GOOGLE_LOGIN_LISTENER_TOPIC,
-                          () -> {
-                            refreshDialog(Services.getLoginService().getActiveUser());
-                          }));
+      ApplicationManager.getApplication()
```
etanshaul: I guess its ok to use the Application wide bus because you are still passing in the disposable? I.e. it will always get cleaned when the dialog is closed.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelector.java
```diff
@@ -91,14 +87,9 @@ public ProjectSelector(@Nullable Project ideProject) {
...
         };
-    messageBusConnections =
-        Stream.of(ProjectManager.getInstance().getOpenProjects())
-            .map(p -> p.getMessageBus().connect())
-            .collect(Collectors.toSet());
-    messageBusConnections.forEach(
-        connection ->
-            connection.subscribe(
-                GoogleLoginListener.GOOGLE_LOGIN_LISTENER_TOPIC, googleLoginListener));
+    messageBusConnection = ApplicationManager.getApplication().getMessageBus().connect();
```
etanshaul: how will this behave if you have multiple projects opened?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
ivanporty: Yep, I think it's the same mechanics.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelector.java
ivanporty: It gets called for all listeners regardless of how many projects are opened, or no projects are opened at all, so should be good.

etanshaul: just so I understand: since we currently can't be signed into different accounts in different projects (right?) then this is ok. since a logout in one project should be a logout in all projects

etanshaul: oh. nm. I see now

ivanporty: Right, it's completely generic event, i.e. no correlation between IDE projects and login state.

Issue #1903, opened 2018-02-14
Creating a new AE standard Java app using the new module wizard (non EE) creates an app with no appengine-web.xml


patflynn:
this is in master.

steps to reproduce:
1. start IDE
1. new project
1. Java -> web application -> Google App Engien Standard (w/ a java 8 JDK)
1. set cloud sdk to something valid
1. next
1. observe in the src tree that there is no appengine-web.xml
1. observe you get a notification that your app engine module doesn't support java 8 and should be java 7 (odd right?)
1. observe that the run configuration seem to be OK about everything 
1. try to run locally and observe:
`Error running 'Google App Engine Standard Local Server': Error running Google App Engine Dev Server. Exception occurred : com.google.cloud.tools.appengine.api.AppEngineException Exception message : java.nio.file.NoSuchFileException: /usr/local/google/home/paflynn/IdeaProjects/delete2/out/artifacts/delete2_war_exploded/WEB-INF/appengine-web.xm`

patflynn:
this seems to be happening for Java EE module wizard too.

etanshaul:
Just a guess: but is it also not creating a web.xml (since it’s now using
servlet 3+)? This might be causing the logic that generates the app
engine-web.xml to fail since it can’t find the web.xml.

etanshaul:
@nbashirbello mind looking into this since its related to resent work you were doing?

nbashirbello:
@etanshaul is right.

web.xml is not getting generated so IntelliJ's WebFacetImpl#getWebXmlDescriptor returns null when we call it to determine the parent for the appengine-web.xml.

etanshaul:
closed via 1927

PR #1927, opened 2018-02-27
Add default location for appengine-web.xml when creating new GAE projects

nbashirbello:
Part of the fix for #1903


nbashirbello:
Not ready for review

patflynn:
I wonder to what extent 'add framework support' is affected by this bug.
Also what the right location to place an appengine-web.xml file is in a
springboot project with no web.xml.

etanshaul:
fwiw. a springboot maven project's typical location for the appengine-web.xml is `src/main/webapp/WEB-INF`

nbashirbello:
The 'add framework support' does not create the appengine-web.xml too but this fix solves that.

nbashirbello:
@etanshaul 
So should the default path be src/main/webapp/WEB-INF for modules with the Spring facet for UI or CI?

etanshaul:
I don't think we should have spring specific logic.

Perhaps just searching for the WEB-INF directory of a given module? @patflynn wdyt?

patflynn:
could we use the web facet configuration?

etanshaul:
Looks like we should at least be able to get the web resource directory:

screen shot 2018-02-28 at 1 57 42 pm

nbashirbello:
Just trying to think this through. So to get the directory for the appengine-web.xml;
1) use the web.xml directory if its exists
2) use the nth directory from the list of web resource directories (if count is more than 1, what happens?)
3) use a default path

etanshaul:
1) agree
2) there should still only be 1 WEB-INF folder per module. Its the WEB-INF you are ultimately looking for not the web resource directory root
3) just thinking out loud here, but its ultimately the App Engine staging that needs the ability to locate the appengine-web.xml. So I guess it would be an ok heuristic to just try and put it in a well known location if 1 & 2 fail.

nbashirbello:
After off-line conversation with @etanshaul, considering this strategy

1. use web.xml directory if it exists
2. search for WEB-INF folder and use that if it exists
3. use this well know location as the default: src/main/webapp/WEB-INF

@patflynn Thoughts?

patflynn:
+1. that makes sense.  For community is there a web facet?

etanshaul:
hmm good question. I was about to say this doesn't apply to CE because you can't deploy or create new AE standard projects. But we do provide workarounds for local-run.

We need to check if we support AE standard framework support in CE / AE standard facet (which I don't think we do).

But to answer the question - no, there is no web facet in CE.

patflynn:
It may make sense to create look for WEB-INF in source roots for CE as a
place to generate the file.

nbashirbello:
In IC, the GAE Standard framework is not available either via the New Project wizard or via the Add Framework dialog. 

(Screenshots are from IC)

screen shot 2018-03-05 at 2 04 44 pm

screen shot 2018-03-05 at 2 07 39 pm


The GAE Standard facet is available in the Module Setting dialog but adding the GAE Standard facet that does not include adding the appengine-web.xml.

screen shot 2018-03-05 at 2 05 31 pm

screen shot 2018-03-05 at 2 05 56 pm

etanshaul:
@nbashirbello should we still hold off on review?

nbashirbello:
Oh please review.

nbashirbello:
@etanshaul What should we do about the Add Google App Engine Standard menu item in IC? Should we hide it?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/impl/AppEngineStandardCommunityWebIntegration.java
```diff
@@ -114,6 +91,12 @@ public void addDescriptor(
             PackagingElementFactory.getInstance().createFileCopy(descriptor.getPath(), null));
   }
 
+  @NotNull
+  @Override
+  public String getDefaultAppEngineWebXmlPath() {
```
etanshaul: this isn't really a path? more like a directory name?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineStandardUltimateWebIntegration.java
```diff
@@ -74,10 +80,48 @@ public ArtifactType getAppEngineWebArtifactType() {
...
+      VirtualFile webInfPath = getWebInfPath(webFacet);
+      if (webInfPath != null) {
+        return webInfPath;
+      }
+    }
+
+    return super.suggestParentDirectoryForAppEngineWebXml(module, rootModel);
+  }
+
+  private VirtualFile getWebInfPath(@NotNull WebFacet webFacet) {
```
etanshaul: mind adding a javadoc that describes the strategy you are using?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineStandardUltimateWebIntegration.java
```diff
@@ -74,10 +80,48 @@ public ArtifactType getAppEngineWebArtifactType() {
...
+      return null;
+    }
+
+    for (WebRoot webRoot : webRoots) {
+      VirtualFile parent = webRoot.getFile();
+      if (parent.getPath().endsWith("/WEB-INF")) {
+        return parent;
+      }
+
+      VirtualFile child = parent.findChild("WEB-INF");
```
etanshaul: some places "WEB-INF" is hardcoded, others you have a helper method. maybe just move it to a private static final variable somwhere?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineStandardUltimateWebIntegration.java
```diff
@@ -74,10 +80,48 @@ public ArtifactType getAppEngineWebArtifactType() {
...
+
+  private VirtualFile getWebInfPath(@NotNull WebFacet webFacet) {
+    List<WebRoot> webRoots = webFacet.getWebRoots();
+    if (webRoots.isEmpty()) {
+      return null;
+    }
+
+    for (WebRoot webRoot : webRoots) {
+      VirtualFile parent = webRoot.getFile();
+      if (parent.getPath().endsWith("/WEB-INF")) {
```
etanshaul: is this Windows safe?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineStandardUltimateWebIntegration.java
```diff
@@ -196,6 +240,12 @@ public void registerFrameworkInModel(
         .setFacet(AppEngineStandardFacetType.ID, appEngineStandardFacet);
   }
 
+  @NotNull
+  @Override
+  public String getDefaultAppEngineWebXmlPath() {
+    return "src/main/webapp/WEB-INF";
```
etanshaul: didn't we say we were not going to use this strategy of assume this path? This is really only valid in special cases (maven directory structure).

google-cloud-tools-plugin/ultimate/testSrc/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineStandardUltimateWebIntegrationTest.java
```diff
@@ -0,0 +1,139 @@
...
+import com.intellij.openapi.vfs.VirtualFile;
+import com.intellij.testFramework.PlatformTestCase;
+import com.intellij.util.descriptors.ConfigFile;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import org.jetbrains.annotations.NotNull;
+
+/** Tests for {@link AppEngineStandardUltimateWebIntegration}. */
+public class AppEngineStandardUltimateWebIntegrationTest extends PlatformTestCase {
```
ivanporty: Could we probably think about using `CloudToolsRule` and `@Mock/@TestService` here (it could be a bit cleaner to read and in line with the latest tests), or this test requires using `PlatformTestCase`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/AppEngineStandardWebIntegration.java
```diff
@@ -42,21 +44,35 @@
...
+  private static final Logger LOG = Logger.getInstance(AppEngineStandardWebIntegration.class);
 
   public static AppEngineStandardWebIntegration getInstance() {
     return ServiceManager.getService(AppEngineStandardWebIntegration.class);
   }
 
   @Nullable
-  public abstract VirtualFile suggestParentDirectoryForAppEngineWebXml(
-      @NotNull Module module, @NotNull ModifiableRootModel rootModel);
+  public VirtualFile suggestParentDirectoryForAppEngineWebXml(
```
etanshaul: can you explain a bit (just so I understand) the idea behind moving the implementation to the base integration class? Before, the Ultimate implementation didn't fall back to this strategy (by calling super), but in the PR it is. Just trying to wrap my head around it.

etanshaul:
@nbashirbello sorry missed your question about hiding the menu item in IC:

yes, it should be hidden for IC. but I think in a separate PR to keep things clean.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/AppEngineStandardWebIntegration.java
```diff
@@ -42,21 +44,35 @@
...
+      @NotNull Module module, @NotNull ModifiableRootModel rootModel) {
+    final VirtualFile root = ArrayUtil.getFirstElement(rootModel.getContentRoots());
```
ivanporty: Looks like `getContentRoots()` returns a collection, would it be enough in all cases to just work with the first content root?

ivanporty: Also here, just for clarification on IDE model here, I think there could be source and resources content roots, are they all returned here or it's only source ones?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineStandardUltimateWebIntegration.java
```diff
@@ -74,10 +80,48 @@ public ArtifactType getAppEngineWebArtifactType() {
...
+      }
+
+      VirtualFile child = parent.findChild("WEB-INF");
+      if (child != null) {
+        return child;
+      }
+    }
+
+    try {
+      return VfsUtil.createDirectoryIfMissing(webRoots.get(0).getFile(), "WEB-INF");
```
ivanporty: It looks similar to the code used in standard integration, what would be the reason for not using `getDefaultAppEngineWebXmlPath` here but rather use hard-coded name instead?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineStandardUltimateWebIntegration.java
```diff
@@ -74,10 +80,48 @@ public ArtifactType getAppEngineWebArtifactType() {
...
+    } catch (IOException ioe) {
+      LOG.info(ioe);
```
ivanporty: Could we raise log level to `WARN` here?

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/facet/standard/impl/AppEngineStandardCommunityWebIntegrationTest.java
```diff
@@ -0,0 +1,70 @@
...
+import com.intellij.openapi.application.WriteAction;
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.roots.ModifiableRootModel;
+import com.intellij.openapi.vfs.VirtualFile;
+import com.intellij.testFramework.PlatformTestCase;
+import org.jetbrains.annotations.NotNull;
+
+/** Tests for {@link AppEngineStandardCommunityWebIntegration}. */
+public class AppEngineStandardCommunityWebIntegrationTest extends PlatformTestCase {
```
ivanporty: Possibly same idea as for `AppEngineStandardUltimateWebIntegrationTest`, consider using `CloudTestRule` and `@Mock` annotations?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineStandardUltimateWebIntegration.java
ivanporty: Could we use `parent.getName()` instead of string comparison for the whole path maybe? It usually supposed to return the name of the actual last file in the path.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineStandardUltimateWebIntegration.java
nbashirbello: This is the path that would be used when then are no web resource directories.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/facet/standard/AppEngineStandardWebIntegration.java
nbashirbello: This method creates the appengine-web.xml directory from a static path from the root of the project. 

In IC - this is only what it did using the WEB-INF path 
In IU - if there was no web.xml, this method returned null (hence why the appengine-web.xml was not getting generated in #1903). With the PR, if web.xml doesn't exist and there are no web resource directories, then it creates the parent of the appengine-web.xml in a from a static path too.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineStandardUltimateWebIntegration.java
etanshaul: If we have no indication that this is the right path (i.e. no web roots, not web-inf) I'm not sure that its valid to assume that this is a good path (especially since it may not be a Maven project). @patflynn wdyt?

patflynn:
yes I agree. For cases where there's no web facet there's probably no good
answer, although if you can detect maven vs gradle you could follow
convention there.  We may not want to get that complicated in this PR
though.  Perhaps just placing it under the first source root in WEBINF/?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/appengine/facet/impl/AppEngineStandardUltimateWebIntegration.java
nbashirbello: I'm confused. My initial default path was "web/WEB-INF". I thought you wanted it changed to "src/main/webapp/WEB-INF".

etanshaul: sorry - its possible Im contradicting myself. Here is the strategy as I see it (lmk if it makes sense to you):

1) if the web.xml exists use this directory
2) find the web roots - get the WEB-INF from it, and place it here
3) if there is no WEB-INF in the web root, create it and use this
4) we really don't have a good guess at this point so just fail and log (later in a different PR we can maybe do what @patflynn suggested about inspecting if its a Maven or Gradle project and putting it in that well known location)

etanshaul: and just to point out - its arguable that step 1 should actually come after step 2 & 3. I'm open to either approach.

nbashirbello:
PR is ready for review. Thanks

etanshaul:
Thanks. I just tested it with a few scenarios and it works as expected:

- New AE standard project wizard (creates the appengine-web.xml even without a web.xml)
- Various Spring boot projects

@ivanporty any comments on this?

ivanporty:
reviewing now

ivanporty:
Looks good to me as well, scenarios work fine

PR #1929, opened 2018-02-28
Managed SDK UI (behind feature flag)

ivanporty:
PR contains purely UI for choosing between Managed/Custom SDK. Feature flag is used to show managed SDK UI (set `feature.managed.sdk = true` in `config.properties` file to enable).

### Managed SDK UI:
managedsdk-1

managedsdk-custom

### Default UI (managed SDK feature disabled).

Settings:

old-sdk-settings

Local Dev Server settings (re-used panel):

old-sdk-applocal


etanshaul:
nit: instead of just "managed" what about something more descriptive (to indicate that its automatically managed (installed) by the plugin (like you have for custom)? 

Something like "Managed (Installation managed by the plugin)" or something along those lines?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -45,6 +54,11 @@
   private JTextPane warningMessage;
   private JPanel cloudSdkPanel;
   private JLabel warningIcon;
+  private JRadioButton managedRadioButton;
+  private JRadioButton customRadioButton;
+  private JCheckBox enableAutomaticUpdatesCheckbox;
```
etanshaul: what about explicitly setting its default check/uncheck value and having a member boolean here containing that default? that way its easy to see what the default is and to change it later if we wanted.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -45,6 +54,11 @@
   private JTextPane warningMessage;
   private JPanel cloudSdkPanel;
   private JLabel warningIcon;
+  private JRadioButton managedRadioButton;
```
etanshaul: what determines if "managed" or "custom" will be selected by default? is this for a different PR?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
ivanporty: Sure, good idea.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
ivanporty: Yes, for the next PR - I'll add settings that will check and select proper value.

ivanporty:
I agree, we can have better name... `Managed (installed by plugin)`? I'm open for suggestions here, can't think of really clean label so far.

etanshaul:
yeah I can't really think of anything either. I think that will be ok for now (`Managed (installed by plugin)`).

etanshaul:
I just thought to glance at the JB UI guide that @patflynn linked to. Specifically the rules around text and capitalization:

https://jetbrains.github.io/ui/text/capitalization/

They suggest to use sentences for radio buttons / checkboxes and have first-word-capitalized-only capitalization for these.

Not saying necessarily we should do this. But perhaps it would solve the other problem (of unclear wording).

ivanporty:
Probably right. Let me change that too.

ivanporty:
sdk-ui-2-managed

sdk-ui-custom-2

etanshaul:
I think the guide was suggesting not just about the capitalization, but also making them sentences.

etanshaul:
(only if you think it would help though)

ivanporty:
`Managed and installed by plugin` and
`Custom local installation`?

etanshaul:
hmm. if we go that way we might as well go all the way and make it a proper sentence. i.e.:

`Allow the plugin to manage and install the Cloud SDK`
`Use a custom local installation`

ivanporty:
Sounds good to me.

ivanporty:
sdk-ui-sentence

etanshaul:
nit: "Automatic updates" as a sentence for consistency?

etanshaul:
for example "Automatically update the SDK"

ivanporty:
sdk-ui-final

PR #1931, opened 2018-03-01
Managed SDK UI settings

ivanporty:
Persists settings for Managed / Custom SDK using `PropertiesComponent`, adding unit tests to ensure proper state between UI and settings.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
@@ -50,7 +50,6 @@ import org.jetbrains.annotations.NotNull;
 /** Reusable panel for configuring the path to the Cloud SDK from various contexts. */
 @SuppressWarnings("FutureReturnValueIgnored")
 public class CloudSdkPanel {
-  public static final boolean MANAGED_SDK_AUTOMATIC_UPDATES_DEFAULT = true;
 
   private TextFieldWithBrowseButton cloudSdkDirectoryField;
   private JTextPane warningMessage;
@@ -76,8 +75,6 @@ public class CloudSdkPanel {
     warningMessage.addHyperlinkListener(new BrowserOpeningHyperLinkListener());
     warningIcon.setVisible(false);
     warningIcon.setIcon(RunConfigurations.ConfigurationWarning);
-
-    enableAutomaticUpdatesCheckbox.setSelected(MANAGED_SDK_AUTOMATIC_UPDATES_DEFAULT);
 
     checkManagedSdkFeatureStatus();
 
@@ -202,8 +199,10 @@ public class CloudSdkPanel {
       CloudSdkService.getInstance().setSdkHomePath(getCloudSdkDirectoryText());
     }
 
-    CloudSdkServiceUserSettings.getInstance()
-        .setUserSelectedSdkServiceType(selectedCloudSdkServiceType);
+    CloudSdkServiceUserSettings sdkServiceUserSettings = CloudSdkServiceUserSettings.getInstance();
+    sdkServiceUserSettings.setUserSelectedSdkServiceType(selectedCloudSdkServiceType);
+
+    sdkServiceUserSettings.setEnableAutomaticUpdates(enableAutomaticUpdatesCheckbox.isSelected());
 
     // settings are applied and saved, clear modification status
     settingsModified = false;
@@ -226,6 +225,8 @@ public class CloudSdkPanel {
     CloudSdkService sdkService = CloudSdkService.getInstance();
     setCloudSdkDirectoryText(
         sdkService.getSdkHomePath() != null ? sdkService.getSdkHomePath().toString() : "");
+
+    enableAutomaticUpdatesCheckbox.setSelected(sdkServiceUserSettings.getEnableAutomaticUpdates());
 
     // reset modified flag too so user won't see this as changed state.
     settingsModified = false;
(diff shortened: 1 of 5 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -65,15 +66,17 @@
       "https://cloud.google.com/sdk/docs/"
           + "#install_the_latest_cloud_tools_version_cloudsdk_current_version";
 
+  private boolean settingsModified;
```
etanshaul: I wish there was a better way to do this. I remember we discussed something previously where there was a way to make dialogs aware of changes to certain swing components that it wasn't registering. But I do see that its complicated by the fact that this is just a panel that is included in various contexts. Do you know if its at all possible to update this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -172,43 +172,73 @@ private void invokePanelValidationUpdate(Runnable runnable) {
...
 
   public void reset() {
-    CloudSdkService sdkService = CloudSdkService.getInstance();
+    CloudSdkServiceUserSettings sdkServiceUserSettings = CloudSdkServiceUserSettings.getInstance();
+
+    CloudSdkServiceUserSettings.CloudSdkServiceType selectedSdkServiceType =
+        sdkServiceUserSettings.getUserSelectedSdkServiceType();
+    switch (selectedSdkServiceType) {
+      case MANAGED_SDK:
+        managedRadioButton.doClick();
```
etanshaul: why doClick instead of setSelected?

ivanporty: I use `ActionListener` to listen for their actions, but `ActionListener`s are not called after `setSelected` - I think `ItemListener` can be used instead but they are a bit more verbose to use. Does it feel slightly confusing? We can use selection listeners and methods instead.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
ivanporty: You're right, I think IDE uses something for `SettingsEditor` to track standard components state, but this is abstract `Configurable` with `createComponent()` and I think it's not tracked by default. Looks pretty ugly.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
etanshaul: just curios why you need the listeners to be called from the reset method? I see that later you are defaulting the other properties (like auto update).

ivanporty: That's sort of unified way to set up UI state properly - `reset()` is called when panel is opened, and UI state is reset based on settings - selecting radio button brings the state of components into desired state (i.e. enables/disables based on chosen SDK).

etanshaul: Ah yeah. There are other things like enable / disable controls which happen via the listener. Makes sense. 

I’ll leave the doClick up to you since I don’t feel too strongly either way. Personally I tend to just use the other kind of listener that gets fired on setSelected.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
ivanporty: It's called `UserActivityWatcher`, it looks like it's working pretty good with standard components we have here in panel, and simplified code a bit. We still use the flag to mark the state changes between `apply()`/`reset()` but it's easier to read now.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
ivanporty: I tried `setSelected` but forgot it only fires events when state is actually changed - i.e. false -> true or vice versa. In this case we want the action to be called for sure, so `doClick()` works better.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -283,6 +284,11 @@ private void createUIComponents() {
   }
 
   private void initEvents() {
+    // track all changes in UI to report settings changes.
+    UserActivityWatcher activityWatcher = new UserActivityWatcher();
```
etanshaul: this is cool!

ivanporty: Yeah we can use it directly when we need something like that, pretty simple.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanelTest.java
```diff
@@ -114,17 +134,165 @@ public void testCheckSdk_validSdk() {
...
+  @Test
+  public void customSdk_choice_apply_validSettings() {
+    ApplicationManager.getApplication()
+        .invokeAndWait(
+            () -> {
+              // use non-spy panel as spy messes up with UI event thread field updates.
+              CloudSdkPanel sdkPanel = new CloudSdkPanel();
+              sdkPanel.reset();
+              sdkPanel.getCustomRadioButton().doClick();
+              String customSdkPath = "/home/gcloud";
```
etanshaul: is there any validation at the moment that the custom path is actually a valid gcloud directory?

ivanporty: Yes, there is, it's still there from original implementation. You can take a look at first test methods of `CloudSdkPanelTest` for examples I think.

etanshaul: aha. ok cool

PR #1937, opened 2018-03-06
Cloud SDK service management

ivanporty:
Adds `CloudSdkServiceManager` class as a IDE service which will manage instances of Cloud SDK service (between custom and managed), depending on a user selection.

`CloudSdkService.getInstance()` will now use this class to return current choice for SDK service.

Removed `CloudSdkService.setSdkHomePath()`, this is set in settings for custom SDK.

### Testing Managed SDK
Plugin will now use Managed SDK if selected in settings. You can test it after enabling feature flag in `config.properties`:
> feature.managed.sdk=false -> true

then go to Settings -> Google to select Managed SDK.
Once applied, managed SDK will install (you should see the progress in background). Once installed (success popup), it will be used everywhere.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
@@ -30,9 +30,8 @@ public class CloudSdkServiceManager {
   }
 
   public CloudSdkService getCloudSdkService() {
-    CloudSdkServiceType currentSdkServiceType =
-        CloudSdkServiceUserSettings.getInstance().getUserSelectedSdkServiceType();
-    return supportedCloudSdkServices.get(currentSdkServiceType);
+    return supportedCloudSdkServices.get(
+        CloudSdkServiceUserSettings.getInstance().getUserSelectedSdkServiceType());
   }
 
   /** Callback when a user selected and applied a new cloud sdk type. */
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
@@ -36,6 +36,16 @@ public class CloudSdkServiceManager {
 
   /** Callback when a user selected and applied a new cloud sdk type. */
   public void onNewCloudSdkServiceTypeSelected(CloudSdkServiceType newServiceType) {
-    supportedCloudSdkServices.get(newServiceType).activate();
+    if (supportedCloudSdkServices.containsKey(newServiceType)) {
+      supportedCloudSdkServices.get(newServiceType).activate();
+    } else {
+      throw new UnsupportedCloudSdkTypeException(newServiceType.name());
+    }
+  }
+
+  private static class UnsupportedCloudSdkTypeException extends RuntimeException {
+    private UnsupportedCloudSdkTypeException(String message) {
+      super(message);
+    }
   }
 }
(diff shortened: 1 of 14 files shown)


etanshaul:
may be premature, but was testing this out and noticed the following:

1) enabled the flag
2) opened up the deploy dialog, configured the server/sdk and left it on managed
3) closed the sdk dialog

the deploy dialog is in an error state:

image

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceType.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.intellij.appengine.sdk;
+
+/** Supported {@link CloudSdkService} types. */
+public enum CloudSdkServiceType {
```
etanshaul: something intuitively feels like this whole strategy could be more more OO. But I guess since we are persisting strings and we then need some way to return the appropriate Service from the string there isn't really a better approach.

ivanporty:
So for the error I think managed SDK didn't get installed yet by the time - I suggest for now, until all checks are in place, use Settings -> Google and then wait for successful Managed SDK installation confirmation, then we can test existing scenarios. Blocking deployments is one if the next PRs.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceType.java
ivanporty: Agree, it feels more of the settings things. We can probably use class name / generic settings object per SDK interface, but then again, have to specify them directly from Cloud SDK settings UI. I feel it would be easy to generalize that if we need to (i.e. more implementations of SDK service), but for now it doesn't feel we'll have any more anytime soon.

etanshaul:
Oh ok. I didn't notice that the SDK was being installed in the background (this is what should have been happening in that scenario?). I'll check again.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  public CloudSdkService getCloudSdkService() {
+    return supportedCloudSdkServices.get(
+        CloudSdkServiceUserSettings.getInstance().getUserSelectedSdkServiceType());
+  }
+
+  /** Callback when a user selected and applied a new cloud sdk type. */
+  public void onNewCloudSdkServiceTypeSelected(CloudSdkServiceType newServiceType) {
+    supportedCloudSdkServices.get(newServiceType).activate();
```
etanshaul: should you check to make sure that get actually returns a service? (for example, if we add a new type but forget to put it in the map)

etanshaul:
regarding you last comment about the error:

I checked again and saw no indication that the SDK is being installed in the background. Also you said:

>  I suggest for now, until all checks are in place, use Settings -> Google and then wait for successful Managed SDK installation confirmation, then we can test existing scenarios

How to I check for this? I don't see any progress from that panel.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
ivanporty: Sure, good idea.

ivanporty:
Interesting, it might be already installed - do you see successful message in event log or popup? We can take a look offline when you have a minute.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
ivanporty: I'll also throw an exception when this happens since this would be a serious misconfiguration.

ivanporty:
Thanks! @nbashirbello if you have anything, please let me know, I'll continue with refactoring existing UI panels to make sure Cloud SDK configuration and validation is done in one point (global settings).

PR #1938, opened 2018-03-08
Fetch available BOM versions and show them in UI

etanshaul:
progress towards #1921 

- Adds service to fetch BOM versions from Maven Central (using Aether abstractions)
- Populates the BOM versions (the latest n of them) in the UI
- Hides this with a feature flag (`feature.bom`)
- If no versions can be fetched from Maven Central, then the BOM UX is hidden (with the idea that instead we will fall back to the versions from the static json metadata with no BOM support).

Does **not** have any of the following (will come in subsequent PRs):
- Adding the BOM to the pom.xml
- syncing with existing BOM
- updating the versions in the UX for each library to match what is in the selected BOM

image


etanshaul:
@ivanporty @nbashirbello @patflynn ptal

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
```diff
@@ -99,6 +108,12 @@
     panel.setPreferredSize(new Dimension(800, 600));
 
     projectSelector.loadActiveCloudProject();
+
+    if (ServiceManager.getService(PluginInfoService.class).shouldEnable(GctFeature.BOM)) {
+      populateBomVersions();
+    } else {
+      hideBomUI();
```
ivanporty: When the feature is off, and components are hidden, does the panel adjust itself properly, i.e., there is no large empty spaces?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
```diff
@@ -224,6 +249,27 @@ private void updateManagementUI() {
...
+   * Populates the BOM {@link JComboBox} with the fetched versions. If there are no versions
+   * returned, then the BOM UX is hidden.
+   */
+  // TODO (eshaul): make async with loader icons
+  private void populateBomVersions() {
+    List<Version> bomVersions =
+        CloudApiMavenService.getInstance().getBomVersions(NUM_BOM_VERSIONS_TO_SHOW);
+
+    if (bomVersions.isEmpty()) {
+      hideBomUI();
```
ivanporty: So per your description, this will hide BOM versions and UI controls, and instead show individual versions and show them in library details? In this case, do we also plan to add individual versions to POM file until BOM is available, correct?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
```diff
@@ -224,6 +249,27 @@ private void updateManagementUI() {
...
+    } else {
+      bomVersions.forEach(bomComboBox::addItem);
```
ivanporty: Will this always show proper version display name, i.e. `toString()` method is stable for `Version` and no separate renderer is needed?

patflynn:
what do you guys think about dropping the 'client' in the client libraries
label?

etanshaul:
@patflynn just to confirm. do you mean both labels? or just "Google Cloud Client LIbraries Version". I could see dropping it from both.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiMavenService.java
```diff
@@ -0,0 +1,74 @@
...
+import org.jetbrains.idea.maven.aether.ArtifactRepositoryManager;
+import org.jetbrains.idea.maven.aether.ProgressConsumer;
+
+/** An application service providing Cloud API Maven functionality. */
+public class CloudApiMavenService {
+  private static final Logger logger = Logger.getInstance(CloudApiManager.class);
+
+  private static final String GOOGLE_CLOUD_JAVA_BOM_GROUP_NAME = "com.google.cloud";
+  private static final String GOOGLE_CLOUD_JAVA_BOM_ARTIFACT_NAME = "google-cloud-bom";
+  private static final String GOOGLE_CLOUD_JAVA_BOM_VERSION_CONSTRAINT = "[0,)";
```
ivanporty: Interesting math notation, is this what they used in Aether? Is this required or we can fetch all versions regardless of constraint?

etanshaul:
oops test is broken cause I forgot to enable the feature flag for testing. fix coming

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiMavenService.java
```diff
@@ -0,0 +1,74 @@
...
+
+    try {
+      List<Version> availableVersions =
+          repositoryManager.getAvailableVersions(
+              GOOGLE_CLOUD_JAVA_BOM_GROUP_NAME,
+              GOOGLE_CLOUD_JAVA_BOM_ARTIFACT_NAME,
+              GOOGLE_CLOUD_JAVA_BOM_VERSION_CONSTRAINT,
+              ArtifactKind.ARTIFACT);
+
+      availableVersions.sort(Comparator.reverseOrder());
```
ivanporty: Will this always work, i.e. `Version` implements `Comparable`?

ivanporty:
Agree, library sounds clear enough without "client"

patflynn:
Good point.  Right now it's a bit weird that it's inconsistent between the
two.
maybe something like:
window title: "Add Google Cloud Libraries"
1st label: 'Add library to module'
2nd label: 'library version'

I think I'd also be OK with prefacing everything with 'Google Cloud'.

etanshaul:
@patflynn 

I like the suggestions. Regarding prefacing everything with Google Cloud - the 2nd label imo must have that preface otherwise its a bit weird. But I'm also OK prefacing them all with it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
etanshaul: yes. Ive confirmed it

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiMavenService.java
```diff
@@ -0,0 +1,74 @@
...
+
+  static CloudApiMavenService getInstance() {
+    return ServiceManager.getService(CloudApiMavenService.class);
+  }
+
+  /**
+   * Returns the available Google Cloud Java client library BOM versions from Maven Central.
+   *
+   * @param limit the number of results to the n latest BOM versions
+   * @return returns the {@link Version versions} of the BOMs
```
ivanporty: Probably worth mentioning this is reversely sorted list.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
etanshaul: yes. an empty return from the BOM service call will do the following:

1) hide the BOM label and dropdown (done in this PR)
2) in the client lib selector UI, show the versions from the metadata json instead of from the BOM (effectively already in place because fetching lib versions based on the BOM is not yet implemented)
3) add the dependencies to the pom.xml as if there is no BOM support at all (like we already do)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
etanshaul: yes, looks like the toString method itself is smart enough to render the version properly

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanelTest.java
```diff
@@ -235,6 +244,37 @@ public void getPanel_withMultipleLibraries_multipleSelected_hasCheckboxesAndPopu
...
+    GoogleCloudApiSelectorPanel panel =
+        new GoogleCloudApiSelectorPanel(ImmutableList.of(), testFixture.getProject());
+
+    JComboBox<Version> bomComboBox = panel.getBomComboBox();
+
+    assertThat(panel.getBomSelectorLabel().isVisible()).isTrue();
+    assertThat(panel.getBomComboBox().isVisible()).isTrue();
+    assertThat(bomComboBox.getItemCount()).isEqualTo(3);
+
+    for (int i = 0; i < bomComboBox.getItemCount(); i++) {
```
ivanporty: Nit: remembering Java readability from Nate :) it could be cleaner to move this to a separate test case and test against explicit list of values rather than a loop.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiMavenService.java
etanshaul: yes, version implements comparable

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiMavenService.java
etanshaul: yes this is Aether notation.

The service call requires a version to be passed in. So we must pass in some pattern to indicate that we want them all. I toyed with the idea of using this notation itself to limit the results instead of limiting them later client-side, but we would have to pre-suppose a version syntax (which is a bad idea I think).

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanelTest.java
```diff
@@ -574,4 +622,26 @@ private static void assertDetailsShownForLibrary(
...
+  private static class TestVersion implements Version {
+    private final String name;
+
+    public TestVersion(String name) {
+      this.name = name;
+    }
+
+    @Override
+    public int compareTo(@NotNull Version o) {
+      return 0;
```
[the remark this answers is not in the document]
etanshaul: the sorting is done inside of the BOM service method. in these tests we are mocking out that entire method (so no sorting is done).

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanelTest.java
etanshaul: sounds good

etanshaul: the ghost of Nate haunts us :)

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanelTest.java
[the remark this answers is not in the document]
ivanporty: cool, got it

etanshaul:
wording updates:
image

ivanporty:
We have two places with plural "Libraries" and one with "Add .. Library to Module", is this intentional?

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanelTest.java
etanshaul: I removed the loop. but didn't see much value in moving to a separate test case..lmk if you disagree

etanshaul:
Yes that is intentional:

The plural case is because that BOM version applies to ALL the client libraries. Its an overarching version that can be thought of loosely as the version of the client libraries.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanelTest.java
ivanporty: LGTM

ivanporty:
Sounds good. I just thought a user might be able to add multiple versions to module, and title says "Libraries", but it's pretty clear anyway.

etanshaul:
side note: I put a `todo` in the code to fetch the bom version asynchronously.

in my testing, however, its instantaneous enough to not be noticeable in the UI.

etanshaul:
yes - that is a bit confusing. its the difference between the BOM version and the individual library version.

If you have any ideas to make it even clearer, feel free

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiMavenService.java
etanshaul: what do you think about moving the sorting outside of this method. seems like sorting is the responsibility of the client that is displaying them, not the service itself.

then we can even unit test this here in this pr

ivanporty: Sounds like a great idea, it does make two things here and sorting is needed for UI only.

etanshaul:
@ivanporty ptal at 98035d2:

- Moved sorting out to the caller of the service
- Moved limited out to the caller (using the same reasoning as for the first point)
- Added test coverage for BOM version limiting

etanshaul:
updated to test reverse order sorting too

etanshaul:
I think I addressed all comments so far. lmk if I missed anything

Issue #673, opened 2016-05-05
Auto download cloud sdk if not installed


etanshaul:
If the user doesn't have the cloud sdk pre-installed it would reduce a lot of friction if we had an option (perhaps from within the cloud configuration) to download and install it for them.

patflynn:
I would add that potentially you could completely hide the Cloud SDK dependency if you can resolve it entirely under the covers in the plugin.

etanshaul:
@ivanporty using this as the overall issue to track this effort, unless there is a dup?

ivanporty:
sounds good

PR #1940, opened 2018-03-09
Consolidate Cloud SDK UI in general settings panel

ivanporty:
This PR removes nested Cloud SDK configuration panels from App Engine Server configuration, Local Dev  Server configuration, App Engine Standard/Flexible Support panels.

All of those re-used same `CloudSdkPanel` and changed the same settings.
Going forward with Managed SDK these settings would mostly be unneeded.
The only place to configure Cloud SDK is Settings -> Google -> Cloud SDK.

#### Local Dev Server Settings
local-dev-settings

#### Adding App Engine Standard/Flexible Support
app-engine-add-std-support

add-gae-flex-support

#### App Engine Server
Note: if we have Managed SDK feature off, this panel will contain a link to Cloud SDK settings to quickly set SDK path for users who used to find it here. Once Managed SDK is released and feature is on, this link will be removed to avoid unnecessary pointing to usage of Google Cloud SDK .

gae-server-config

Part of #673.


etanshaul:
just a general question:

Lets say the cloud sdk is invalid (for example, if they use a custom installation which is not valid). What will the user see from these screens? Presumably some kind of messaging or links to help remedy the situation?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
```diff
@@ -35,17 +39,40 @@
...
   private String displayName = GctBundle.message("appengine.name");
   private JPanel mainPanel;
   private JTextPane appEngineMoreInfoLabel;
-  private CloudSdkPanel cloudSdkPanel;
 
   /** Initialize the UI. */
-  public AppEngineCloudConfigurable() {
-    appEngineMoreInfoLabel.setText(
+  AppEngineCloudConfigurable() {
+    // in case managed SDK is not available yet, provide direct link to SDK settings to allow
```
etanshaul: Ah. I guess this almost answers my other question: can you maybe show screenshots of examples of when this link is shown (and any warning messages in case the custom sdk is misconfigured)?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
```diff
@@ -35,17 +39,40 @@
...
+          GctBundle.message("appengine.cloud.sdk.settings", PSEUDO_GOOGLE_SDK_LINK));
+      System.out.println(messageBuilder.toString());
+      messageBuilder.append("<p/>");
+    }
+    messageBuilder.append(
         GctBundle.message("appengine.more.info", MORE_INFO_URI_OPEN_TAG, MORE_INFO_URI_CLOSE_TAG));
-    appEngineMoreInfoLabel.addHyperlinkListener(new BrowserOpeningHyperLinkListener());
+
+    appEngineMoreInfoLabel.setText(messageBuilder.toString());
+    appEngineMoreInfoLabel.addHyperlinkListener(
```
etanshaul: Wondering if it is cleaner to overload the `appEngineMoreInfoLabel` for 2 use-cases like this instead of creating a new label used just for the cloud sdk link? I'm open to it, but just want to put that option out there.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
```diff
@@ -69,18 +96,14 @@ public JComponent createComponent() {
 
   @Override
   public boolean isModified() {
-    return cloudSdkPanel.isModified();
+    return false;
```
etanshaul: won't this always return false anyway if theres nothing to configure from this screen now? Even if so, you may be right that its better to return an explicit false.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
ivanporty: Sure, attaching screenshots.

ivanporty:
Regarding invalid SDK path, the deployment target view will still have the same error validation message we used to have before:

sdk-server-misconfig

Implying the server is misconfigrued and we can fix this looking at the server settings.
When updated server settings panel is opened, it contains a link to settings, and clicking it will immediately bring settings panel where fixing SDK path is possible:

sdk-settings-custom-bad-path

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
ivanporty: Oh you mean have the separate text with placeholders for feature based scenario vs. standard scenario? It looks like we use the same label here, can you please clarify if I missed your point a bit.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
etanshaul: i don't think there is an adapter as far as I know. I think returning an explicit false like you have is fine since it makes the contract clear.

etanshaul:
For the message:

> Google Cloud SDK settings can be found ...

What do you think of making it more of an action. Something like:

> Go to x > y to configure the Google Cloud SDK

Also, what do you think about have the error/warning message about a misconfigured cloud sdk here (in the server panel) too? The reason I ask is because its slightly odd to show the warning on the deploy panel, then have the user click on "server" and not see the error here (it almost seems like it disappeared).

ivanporty:
Excellent idea, we can quickly grab an SDK error message in the panel as well and show it - this would definitely connect the dots better until we have managed SDK flow complete. Let me do that.

ivanporty:
As we discussed, SDK validation status is shown on app engine server panel as well so user can see the logical flow between deployment error and app engine server panel suggestion to use the settings to fix an error.
@etanshaul PTAL.

sdk-custom-error-settings

etanshaul:
nice! it looks like that font issue is fixed here too (unless that was a linux screenshot)

ivanporty:
Yep, that fix was merged, so the fonts should be good!

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineCloudConfigurable.java
```diff
@@ -92,12 +122,23 @@ public boolean canCheckConnection() {
...
-    return cloudSdkPanel;
-  }
-
-  private void createUIComponents() {
-    cloudSdkPanel = new CloudSdkPanel();
+  /**
+   * Checks for custom SDK validation errors and shows error label if they exist. To be removed once
+   * managed SDk feature is rolled out.
+   */
+  private void updateSdkValidationLabel() {
```
etanshaul: is it possible to cover this in tests? like check for the visibility of the label depending on the validation status of the SDK.

ivanporty: Sure, good idea!

PR #1942, opened 2018-03-12
Remove reference to fields becoming private in EAP.

### LINKED ISSUE #1941: EAP build failing

ivanporty:
Fixes #1941.

Repository fields are private in upcoming releases, so we can use shorter version of constructor that takes two public repositories by default. I quickly tested and it seems to work fine.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiMavenService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiMavenService.java
@@ -21,7 +21,6 @@ import com.intellij.jarRepository.JarRepositoryManager;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.diagnostic.Logger;
 import java.util.List;
-import org.eclipse.aether.repository.RemoteRepository;
 import org.eclipse.aether.version.Version;
 import org.jetbrains.idea.maven.aether.ArtifactKind;
 import org.jetbrains.idea.maven.aether.ArtifactRepositoryManager;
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiMavenService.java
```diff
@@ -45,13 +44,9 @@ static CloudApiMavenService getInstance() {
    * @return returns the {@link Version versions} of the BOMs
    */
   List<Version> getBomVersions() {
-    List<RemoteRepository> remoteRepositories =
-        ImmutableList.of(ArtifactRepositoryManager.MAVEN_CENTRAL_REPOSITORY);
```
etanshaul: is MAVEN_CENTRAL_REPOSITORY still used after this? wondering how it knows which remote repo to access with this update

ivanporty: They use two central repos - MAVEN_CENTRAL_REPOSITORY and JBOSS_COMMUNITY_REPOSITORY in this constructor. As an alternative, we may create Maven central repo directly from URL ourselves and use it instead so we can have more control, wdyt?

etanshaul: > we may create Maven central repo directly from URL ourselves and use it instead so we can have more control, wdyt?

Could you elaborate on that?

ivanporty: Sure, they create the repository like that in their code:
```java
private static final RemoteRepository MAVEN_CENTRAL_REPOSITORY = createRemoteRepository(
    "central", "http://repo1.maven.org/maven2/"
  );
```

we can take this approach and manually create the same repository in our code without relying on their selection of public repositories for `ArtifactRepositoryManager`

etanshaul: ah. I just looked at the code - so by default they use maven central and jboss..I think we want to exclude jboss.

so yeah - what you suggested sounds good to me

ivanporty:
Their process in not so good it seems. Someone just changed that as part of using proxies - non-relevant change at all

PR #1943, opened 2018-03-12
Convert App Engine Region styled text to system font.

### LINKED ISSUE #1664: text size is not consistent in the app engine application creation dialog

ivanporty:
Fixes #1664 

Introduces utility to convert styled document to default system font keeping other attributes (i.e. hyperlinks).

gae-region-font

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/ui/FontUtils.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/ui/FontUtils.java
@@ -34,14 +34,15 @@ public class FontUtils {
    * @param styledDocument {@link StyledDocument} to update, see {@link javax.swing.JTextPane}
    */
   public static void convertStyledDocumentFontToDefault(StyledDocument styledDocument) {
-    for (int ch = 0; ch < styledDocument.getLength(); ch++) {
-      AttributeSet defaultStyle = styledDocument.getCharacterElement(ch).getAttributes();
+    for (int i = 0; i < styledDocument.getLength(); i++) {
+      AttributeSet defaultStyle = styledDocument.getCharacterElement(i).getAttributes();
       MutableAttributeSet updatedFontAttrbutes = new SimpleAttributeSet(defaultStyle);
       Font defaultFont = UIManager.getFont("Label.font");
       StyleConstants.setFontFamily(updatedFontAttrbutes, defaultFont.getFamily());
       StyleConstants.setFontSize(updatedFontAttrbutes, defaultFont.getSize());
 
-      styledDocument.setCharacterAttributes(ch, 1, updatedFontAttrbutes, true);
+      styledDocument.setCharacterAttributes(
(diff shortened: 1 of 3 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/ui/FontUtils.java
```diff
@@ -0,0 +1,47 @@
...
+public class FontUtils {
+
+  /**
+   * Converts font for all characters in {@link StyledDocument} to system default font. Does not
+   * change or affect newly added characters.
+   *
+   * @param styledDocument {@link StyledDocument} to update, see {@link javax.swing.JTextPane}
+   */
+  public static void convertStyledDocumentFontToDefault(StyledDocument styledDocument) {
+    for (int ch = 0; ch < styledDocument.getLength(); ch++) {
```
etanshaul: super nit: maybe use the standard `i` here instead for readability?

etanshaul:
also, should you apply it here to the other case (the SDK panel)? Or were you planning on doing that separately?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/ui/FontUtils.java
etanshaul: seems crazy we have to do this character by character. but I guess we can thank Swing for that

ivanporty: Sounds good, I thought `ch` for "character" but it might not be obvious since it's index, not a char.

ivanporty:
Yep, I'll apply that to SDK panel as well, once we merge that, I can merge that branch and use the same utility from there

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/ui/FontUtils.java
ivanporty: lol, you can do the whole text at once but then we'll loose hyperlink formatting - and looking for hyperlink and default styles and altering them might work but I think they are not mutable as is, so this is the best idea just to change fonts for all styles in the document :)

etanshaul:
oh I forgot its not merged. No problem them. You can just put in a todo on the code in the Sdk panel to apply it later if you prefer.

ivanporty:
You're probably right, I can change the style for SDK panel here - it will merge nicely anyway, it doesn't relate to UI changes in that PR.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/ui/FontUtils.java
etanshaul: yeah - imo if you wanted to use `ch` or something contextual it should be `charIndex` or something like that (to indicate its an index). but I think `i` is fine too so whatever you prefer.

PR #1946, opened 2018-03-14
Ask for a specific account sign-in even if another account is already signed in.

### LINKED ISSUE #1883: Switching accounts doesn't update project selector (in GCS view)
1) Sign in to 2 different accounts
2) open the GCS browser
3) swap between the 2 accounts (in the top right widget)

notice that the project selector stays on the original accont

ivanporty:
Fixes #1883.

In this screenshot, another account is already signed in, but it's not the one assigned to project selector, so it asks to sign in with that specific account to work with this specific project (or select a different one with another account).

gcs-different-account

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelector.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelector.java
@@ -234,9 +234,9 @@ public class ProjectSelector extends JPanel {
     Optional<CredentialedUser> loggedInUser =
         loginService.getLoggedInUser(selection.googleUsername());
 
-    if (!loginService.isLoggedIn()) {
+    if (!loggedInUser.isPresent()) {
       accountInfoLabel.setHyperlinkText(
-          GoogleCloudCoreMessageBundle.message("cloud.project.selector.not.signed.in"));
+          GoogleCloudCoreMessageBundle.message("cloud.project.selector.not.signed.in", selection.googleUsername()));
     } else if (loggedInUser.isPresent()) {
       accountInfoLabel.setHyperlinkText(
           String.format(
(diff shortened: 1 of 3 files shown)


etanshaul:
nit (and I could be wrong), but I think it should be "Sign in **to** your blah@blah.com Google account"

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelector.java
```diff
@@ -234,9 +234,10 @@ private void updateCloudProjectSelection(CloudProject selection) {
     Optional<CredentialedUser> loggedInUser =
         loginService.getLoggedInUser(selection.googleUsername());
 
-    if (!loginService.isLoggedIn()) {
+    if (!loggedInUser.isPresent()) {
```
etanshaul: does this still cover the case where the user is not logged in (not just logged in to a different account)?

ivanporty: Yes, this would be the same logic - logged in user will not be present.

etanshaul: right, but in that case, there won't be an email to show in the message right?

ivanporty:
About "Sign in to/with" I followed with what we used before - we say "Sign in with you Google account" when no users at all are signed in, do you think it makes sense to change both of them?

etanshaul:
I think I prefer "to" over "with". But doesn't really matter to me..@parkjam4 any opinion on this?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/project/ProjectSelector.java
ivanporty: In case we have a cloud project already saved (i.e. deployment) it does not matter if not users are logged in, we would know which account we need to sign in to get access to that specific project.

etanshaul: aha. so these branch of logic is exclusively for the case where there is a stored cloud project (with a username stored too). sounds good.

parkjam4:
In this case, I think it makes more sense to say sign in with. You are signing in using your Google account, and not signing in to your Google account directly.

etanshaul:
good point! ok disregard my dumb suggestion then :)

PR #1951, opened 2018-03-16
Adding a plugin.xml for core unit tests 

patflynn:
since the plugin fragments from our modules don't get used to setup fixture environments by default


etanshaul:
I'm not sure I understand why this is required; mind explaining a bit more? Also was there a test that was previously broken due to this change being missing?

patflynn:
The intellij base fixture classes configure the fixtures from the plugin.xml, but these common modules don't actually have a plugin.xml, just a fragment, so we need to create one in the test resources and import the modules fragment. 

No tests were hurt in the meantime, but future test moves will require this.

etanshaul:
got it thanks.

PR #1953, opened 2018-03-16
move project selector package into the core module

patflynn:
Key aspect of this move is that we start including .form files in the build output for cross module inclusion of nested-forms.


google-cloud-tools-plugin/build.gradle
```diff
@@ -25,7 +25,7 @@ apply plugin: 'org.jetbrains.intellij'
 intellij {
 
     pluginName = 'google-cloud-tools'
-    plugins 'Groovy', 'gradle', 'git4idea', 'properties', 'junit', 'maven', 'yaml'
+    plugins 'Groovy', 'git4idea', 'gradle', 'properties', 'junit', 'maven', 'yaml'
```
ivanporty: I wonder is it just ordering or changing their order does change something in the build process?

ivanporty: I wonder is it just ordering or changing their order does change something in the build process?

PR #1954, opened 2018-03-16
Fetch managed dependency versions from selected BOM and display them

etanshaul:
progress towards #1921 

- Refactors the maven aether service to use raw aether calls instead of the intellij abstractions. This is because the abstractions don't expose all the necessary services (including the ability to fetch the managed library versions from the BOM)
- Fetches managed library versions from the BOM, asynchronously, and displays them in the UI

The version shown for the client library is dynamically fetched based on the selected BOM version:
screen shot 2018-03-16 at 3 59 59 pm

If there is no BOM version (e.g. the initial call fails), the version will be the one from the static json metadata (as before).

Note: I'm leaving any service call optimizations for a later PR. This includes caching the bomVersion -> managed lib versions on disk (since we don't need to fetch this data every time).


etanshaul:
closing for now as there is a bit more refactoring to do. will reopen when ready

etanshaul:
note: I need to update the feature flag control so that this stuff is hidden when appropriate

etanshaul:
disregard my last comment - it correctly works as before when the BOM feature is disabled by flag (this because the UI falls back to the old scheme (pulling the version from the static json) when there is no BOM).

etanshaul:
@nbashirbello @patflynn @ivanporty this is ready for review

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanelTest.java
```diff
@@ -303,6 +307,48 @@ public void getPanel_withNoAvailableBomVersions_hidesBomUi() {
...
+    assertThat(panel.getDetailsPanel().getVersionLabel().getText())
+        .isEqualTo("Version: " + libVersion);
+  }
+
+  @Test
+  public void getPanel_withNoVersionReturnedFromBomQuery_fallsBackToAndDisplaysStaticVersion() {
+    // TODO (eshaul): remove once feature is released
+    when(pluginInfoService.shouldEnable(GctFeature.BOM)).thenReturn(true);
+
+    when(mavenService.getBomVersions()).thenReturn(ImmutableList.of(newTestVersion(BOM_VERSION)));
```
ivanporty: What would happen if the BOM version list is returned empty instead - would it have the same behavior?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiMavenService.java
```diff
@@ -16,24 +16,48 @@
...
 import org.jetbrains.idea.maven.aether.ArtifactRepositoryManager;
-import org.jetbrains.idea.maven.aether.ProgressConsumer;
 
-/** An application service providing Cloud API Maven functionality. */
+/** Aether-based application service providing Cloud API Maven functionality. */
 public class CloudApiMavenService {
   private static final Logger logger = Logger.getInstance(CloudApiManager.class);
 
+  private static RepositorySystem SYSTEM;
+  private static RepositorySystemSession SESSION;
```
ivanporty: These two fields feel like they may be final as well since they are initialized in constructor and effectively constants?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiDetailsPanel.java
```diff
@@ -229,6 +238,52 @@ private void updateUI() {
...
+  void updateManagedLibraryVersionFromBom(String bomVersion) {
+    if (currentCloudLibrary.getClients() != null) {
+      CloudLibraryUtils.getFirstJavaClient(currentCloudLibrary)
+          .ifPresent(
+              client -> {
+                CloudLibraryClientMavenCoordinates coordinates = client.getMavenCoordinates();
+
+                if (coordinates != null) {
+                  versionLabel.setIcon(GoogleCloudCoreIcons.LOADING);
+                  versionLabel.setText("");
```
ivanporty: What would happen if the following code returns empty optional without the version? Will the version label remain empty and showing `loading`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
```diff
@@ -231,7 +238,12 @@ private void createUIComponents() {
                 CloudLibrary library =
                     (CloudLibrary)
                         cloudLibrariesTable.getModel().getValueAt(selectedIndex, CLOUD_LIBRARY_COL);
-                detailsPanel.setCloudLibrary(library, apiManagementMap.get(library));
+                detailsPanel.setCloudLibrary(
```
ivanporty: Nit: pretty long event handling call here as part of component initialization, maybe it could read better as a separate event handler method?

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanelTest.java
etanshaul: good eye. this mock is a red herring: its irrelevent to this test. The part that's important is the call to `setCloudLibrary` with the passed in BOM version (in this case null).

I'll remove this mock from both of the new test cases

etanshaul: done

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiMavenService.java
etanshaul: done (had to make then non-static as well)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
etanshaul: done

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiDetailsPanel.java
etanshaul: yeah good catch. if it fails to fetch the version I think we should probably just hide the loader and show a blank version. The reason I think this is better than falling back to the version from the json file is because, in this situation, there _is_ a BOM. So its not 100% essential for us to know the version (its really just for display for the user) since the pom dependency itself won't have a version.

I can update this and add a test to cover this situation. wdyt?

etanshaul: updated

ivanporty: I agree, good reasoning.

etanshaul:
thanks. holding off on merging until after the release

PR #1956, opened 2018-03-19
Fix sonar issues.

Vaisman:
Try changed on try-with-resources according sonar static analyzer.

Tested:
* run gradlew test on Windows
* install Cloud Tools on IntelliJ


etanshaul:
@Vaisman will go ahead and merge this in. thanks again.

PR #1963, opened 2018-03-22
Managed SDK integration and activation

ivanporty:
Part of #673.

This PR enables managed SDK to be used as a default choice for deployments.
Deployment workflow is refactored to include ability of Managed SDK to complete install or re-install a broken SDK before a deployment. If Managed SDK is used, a deployment is blocked until its installation is complete or fails, or a user cancels the wait. All SDK preconditions and waiting is done in new `CloudSdkPreconditionsSupport` class.

Multiple repeated Cloud SDK validation checks are consolidated in `CloudSdkPreconditionsSupport` as well instead of being scattered around in multiple helpers, runners and deployments.

Managed SDK version is updated to the latest (0.5.1) to include all fixes and progress listener updates.

#### Testing
* Make sure Managed SDK is selected in settings.
* While it's installing initially, attempt one or few simultanous deployments to ensure blocking and unblocking.
* Once installed and ready, it won't be blocking, so to test you'd need to delete Managed SDK to start over.

#### Screenshots: Deployment running, SDK still installing -

screenshot from 2018-03-22 11-23-15

#### Outcome 1: Installation complete, deployment continues.
screenshot from 2018-03-22 11-23-32

#### Outcome 2: User decides to cancel and postpone deployment for a while, installation still continues, deployment cancels with a message in the log:

screenshot from 2018-03-23 09-06-43

#### Outcome 3: Installation fails, deployment fails, warning about invalid SDK and a link to settings is shown:

screenshot from 2018-03-22 11-32-08

#### Local Dev Server: waiting, SDK still installing
screenshot from 2018-03-23 10-24-16

#### Outcome 1: install complete, local run starts
screenshot from 2018-03-23 10-24-39

#### Outcome 2: install failed with an error, local run fails

screenshot from 2018-03-23 10-28-09

#### Outcome 3: waiting cancel by user, local run cannot be continued

screenshot from 2018-03-23 10-35-42


etanshaul:
this is great. will start review.

for the third screenshot, could you elaborate on that flow? I.e. when does that notification popup show up, after the user cancels deployment? Why wouldn't that notification show up for the first case (where its just waiting for it to complete) - I'm not saying it should, but just confused why its in case 3, but not 1?

ivanporty:
Sure. Case 3 is where attempt to install or re-install an SDK resulted in an error - in that case error notification with option to open the settings is shown. When user cancels deployment, only warning that SDK should be installed completely is shown since the installation still runs normally.

etanshaul:
Yeah I meant the third screenshot. 

> When user cancels deployment, only warning that SDK should be installed completely is shown since the installation still runs normally.

This is what I'm confused about. Why do you warn in this case? This warning is not shown when deployment is not cancelled and is just blocked and waiting, right?

ivanporty:
Yes, this is sort of a normal case, I agree. We might just print an info string ("user cancelled deployment") into deployment log, WDYT? On the other hand, deployment panel clearly shows deployment as failed but I guess this is expected if one cancels its own deployment process.

etanshaul:
Yeah showing a "cloud sdk installation needs to complete" notification when the user cancels deployment doesn't really make sense to me. 

I'm totally cool with enhancing the deployment output to indicate that the user cancelled the deployment (seems like an enhancement for a different PR since its not directly related?).

ivanporty:
Sure, let's remove that notification. Printing a string into deployment log instead should be 2-3 lines long, maybe too little for a separate PR, if you think it's ok, I can just replace that warning with this cancel message and we'll be done.

etanshaul:
For the third case (where installation failed), wdyt about having a more specific message (saying that installation failed, not just "invalid"). Or is this generic on purpose (used by other things etc.)?

And even to take this further - if deployment fails due to failed Cloud SDK installation, we could even have wording in the deployment console indicating the reason deployment failed.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPreconditionsSupport.java
```diff
@@ -0,0 +1,199 @@
...
+   * @param project Project to which runnable belongs.
+   * @param runnable Runnable to run after Cloud SDK is ready.
+   * @param deploymentLog Log to print SDK statuses.
+   * @param callback Callback to report errors.
+   */
+  public void runAfterCloudSdkPreconditionsMet(
+      @Nullable Project project,
+      Runnable runnable,
+      LoggingHandler deploymentLog,
+      DeploymentOperationCallback callback) {
```
etanshaul: seems strange that this is tied to a `DeploymentOperationCallback` when its meant to be a generic installation manager. what about cases like blocking for local run?

etanshaul: also, and I may be off, but could the logic in this class (for blocking operations until install complete) be folded into the `CloudSdkServiceManager`? Seems like this would be a natural place for it.

google-cloud-tools-plugin/resources/config.properties
```diff
@@ -16,7 +16,7 @@
 
 # Place cloud tools plugin configuration here.
 feature.appengine.flex=true
-feature.managed.sdk=false
+feature.managed.sdk=true
```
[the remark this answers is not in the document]
ivanporty: I thought that might be easier for us to catch possible errors or issues if we enable earlier before the release and start working with it for a couple of weeks. We can always disable it for a release if needed.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPreconditionsSupport.java
ivanporty: I agree, I thought about putting most of the supportive functionality into a new structure of interface, something like `CloudSdkPreconditionCallback` since we don't know anything about deployment, runnable or logger, but decided it'd be easier to see what exactly goes into that method so the changes are easier to digest. If you think that makes sense, I'll refactor that. Open to suggestions on callback name too.

ivanporty: Good idea about `CloudSdkServiceManager`, I'll do that.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPreconditionsSupport.java
```diff
@@ -0,0 +1,199 @@
...
+
+    SdkStatus sdkStatus = cloudSdkService.getStatus();
+    boolean installInProgress = sdkStatus == SdkStatus.INSTALLING;
+    // if not already installing and still not ready, attempt to fix and install now.
+    if (!installInProgress && sdkStatus != SdkStatus.READY && cloudSdkService.supportsInstall()) {
+      cloudSdkService.install();
+      installInProgress = true;
+    }
+
+    CountDownLatch installationCompletionLatch = new CountDownLatch(1);
```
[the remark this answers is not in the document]
ivanporty: It definitely could be done this way, but since we need to poll `ProgressManager` for cancellation status, we'll end up spinning a loop checking for this boolean value and then sleeping for a while, and `await` method is a bit cleaner to understand maybe. It almost feel the same way, but maybe `CountDownLatch` a bit better implies what exactly the deployment is waiting for by looking at the listener code.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
```diff
@@ -36,12 +36,11 @@ static CloudSdkService getInstance() {
 
   SdkStatus getStatus();
 
-  /**
-   * Asks SDK service to attempt install and prepare Cloud SDK if it's not ready for use yet.
-   *
-   * @return true if install is supported and started, false if install is not supported.
-   */
-  boolean install();
+  /** Returns true if install is supported and started, false if install is not supported. */
```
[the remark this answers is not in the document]
ivanporty: I think the idea is if install is *generally* supported, and it returns `false` if install is not possible at all (i.e. Managed SDK failed to initialize at all). If it's generally supported, we may try to fix invalid state by re-installing it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPreconditionsSupport.java
```diff
@@ -0,0 +1,199 @@
...
+    // check the status of SDK after install.
+    SdkStatus postInstallSdkStatus = cloudSdkService.getStatus();
+    switch (postInstallSdkStatus) {
+      case INSTALLING:
+        String message = GctBundle.message("appengine.deployment.error.sdk.still.installing");
+        callback.errorOccurred(message);
+        showCloudSdkNotification(
+            message, NotificationType.WARNING, false /* no settings needed for this case. */);
+        return;
+      case NOT_AVAILABLE:
```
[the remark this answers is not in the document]
ivanporty: For now it means it was never installed or failed to install / was cancelled. For custom path SDK it means the path is empty. Managed SDK does not have `INVALID` status per se, since we always attempt to re-install if it's not `READY`.

google-cloud-tools-plugin/resources/config.properties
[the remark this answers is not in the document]
etanshaul: yeah. I'm all for enabling it now. Was just making sure there weren't any other PRs that this depends on.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkService.java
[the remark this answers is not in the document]
etanshaul: hmm so the `supportsInstall` means that its generally supported _and_ the Managed SDK is initialized, right? If so, do you think this justifies a different name?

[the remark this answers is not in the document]
ivanporty: Correct, I think `isInstallReady` sounds good.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPreconditionsSupport.java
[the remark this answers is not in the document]
etanshaul: I see. sgtm.

so now you are blocking for 1 sec, then polling to see if the sdk installation is cancelled. For the user, this means that they (just to make sure I understand):

1) click cancel to stop the sdk installation
2) a max of 1 second later the user will be notified that the deployment failed

Is that correct?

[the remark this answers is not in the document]
ivanporty: Yep, correct. Better way would be to trigger the latch on cancel itself, but I didn't initially find the way to properly do this, `onCancel()` method in `Task` does not get called.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPreconditionsSupport.java
etanshaul: regarding the callback - doesn't it have to be something generic? What about blocking for local-run - don't we need that implemented before we mark this complete (and permanently enable the feature flag)?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPreconditionsSupport.java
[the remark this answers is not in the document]
etanshaul: I may just be misunderstanding something with the last comment but, the `checkIfCanceled` in your latch count loop is checking if the cloud SDK installation is canceled, right? Not if deployment is canceled.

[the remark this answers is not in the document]
ivanporty: It does check if deployment is canceled - it checks the progress instance assigned to this `Task` thread, where we are waiting for Cloud SDK to become ready.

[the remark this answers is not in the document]
etanshaul: Oh. my bad, I think I had this confused. wdyt about a quick review meeting after you do any pending refactorings to discuss in more detail?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPreconditionsSupport.java
ivanporty: I think local run goes some other path, let me check it.

ivanporty: Ah yes, I missed `CloudSdkStartupPolicy` where the checks are done too before running on local app dev server. Let me include that into the new flow as well.

ivanporty:
Checking out task list. Review list:
- [x] do not show notification on user cancel
- [x] rename supportsInstall()
- [ ] move CloudSdkPreconditionsSupport to CloudSdkServiceManager to reduce class complexity.
- [ ] more error details for notification if SDK not available/invalid.
- [ ] use generic way to communicate with SDK Preconditions checker instead of deployment interfaces
- [ ] support local run blocking

ivanporty:
@etanshaul I think all of the comments are addressed, including blocking of the local run, except for the extended error messages for notifications - I think this needs a concept of an extended error code from Managed SDK installation as well, different from standard error messages we used for validating Cloud SDK path, so it's better to do in a separate PR (I filed issue for that).

etanshaul:
thanks. I'll start review today

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -162,6 +162,12 @@ appengine.deployment.deployallconfigs.checkbox.label=Update dos, dispatch, cron,
 appengine.deployment.flexible.nogaesupport=No modules with Google App Engine flexible framework support found.
 appengine.deployment.error.cancelled=Deployment process was cancelled.
 appengine.deployment.staging.artifact.as=Staging artifact as: {0}
+appengine.deployment.status.preparing.sdk=Waiting for Google Cloud SDK installation to complete...
+appengine.deployment.error.sdk.still.installing=Google Cloud SDK with App Engine Java needs to be completely installed to perform this action. Please wait for installation to complete.
```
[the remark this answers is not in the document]
etanshaul: forgot to update this one?

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -162,13 +162,20 @@ appengine.deployment.deployallconfigs.checkbox.label=Update dos, dispatch, cron,
...
 appengine.promoteinfo.url=https://console.cloud.google.com/appengine/versions
 appengine.run.server.name=Google App Engine Standard Local Server
 appengine.run.server.artifact.missing=Artifact to deploy isn't specified. If no artifacts are available, ensure the following: <br /> 1) An 'exploded-war' artifact is configured for your desired deployment. <br />2) The App Engine standard facet is applied to the corresponding module(s) you wish to run. You can use the menu action under 'Tools -> Google Cloud Tools -> Add App Engine support -> Google App Engine Standard'.
-appengine.run.server.sdk.misconfigured.panel.message=The Cloud SDK is misconfigured. To fix, click the 'Configure' button next to the application server.
-appengine.run.server.sdk.misconfigured.message=The Cloud SDK is misconfigured. To fix, reconfigure the application server.
+appengine.run.server.sdk.misconfigured.panel.message=The Cloud SDK is misconfigured. To fix, use Google -> Cloud SDK settings.
+appengine.run.server.sdk.misconfigured.message=The Cloud SDK is misconfigured. To fix, use Google -> Cloud SDK settings.
```
etanshaul: this and the one above look to be the same message?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
```diff
@@ -43,6 +74,176 @@ public void onNewCloudSdkServiceTypeSelected(CloudSdkServiceType newServiceType)
...
+   * @param project Project to which runnable belongs.
+   * @param runnable Optional runnable to run after Cloud SDK is ready. This runnable will be run on
+   *     UI thread.
+   * @param progressMessage Message to show in progress dialog to identify which process is
+   *     starting, i.e. deployment or local server.
+   * @param callback Callback to report errors and log progress.
+   * @return blocking latch on which another thread may wait for Cloud SDK precondition checks to
+   *     finish.
+   */
+  public CountDownLatch runAfterCloudSdkPreconditionsMet(
```
etanshaul: wdyt about getting rid of all of the "precondition" wording? Its really just doing a blocking install and then executing a callback after. How about:

- method: just `install` (I'm also second guessing this being here in the Manager instead of the service since the service has an `install` too. It may be better to just have this as an overloaded install method in the service itself. wdyt?)
- callback name: `SdkInstallCallback`. Callback variable name: `afterSdkInstall`.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
```diff
@@ -43,6 +74,176 @@ public void onNewCloudSdkServiceTypeSelected(CloudSdkServiceType newServiceType)
...
+      CloudSdkPreconditionCheckCallback callback,
+      boolean userCancelled) {
+    // check the status of SDK after install.
+    SdkStatus postInstallSdkStatus = cloudSdkService.getStatus();
+    switch (postInstallSdkStatus) {
+      case INSTALLING:
+        String message =
+            GctBundle.message(
+                userCancelled
+                    ? "appengine.deployment.error.cancelled"
```
etanshaul: why is there still deployment specific stuff in here if its now meant to be generic (i.e. also handling local run)?

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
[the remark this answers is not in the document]
ivanporty: This actually is a different one - I didn't found a way to cleanly cancel local server run except for throwing execution exception, that's the message for that exception. We can take a look offline on Monday, I'm open to some cues how to get access to their server console without actually running it.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
ivanporty: Yep. First one is shown for local server config panel, second one for error when server run failed. Considering that server will have additional error message with link to settings from common precondition check, I think we can remove the second part.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
ivanporty: Good point, we need another message as a parameter for user cancel message then. I'll update.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
ivanporty: Well, `install` in service actually does installation of SDK if possible. This method could call `install` but only when a few factors are met, then it blocks waiting and executes code after that, again optionally. What it does all the time is runs task with progress, checks SDK status, shows errors and runs code if SDK is ready. maybe `runWhenSdkReady`?

Callback name is nice, thanks.

Also, I feel a bit bad ignoring return value in one place and using in another, what if CountDownLatch to call on completion would be passed in as an optional parameter instead?

etanshaul: For some reason in my head I pictured you passing a runnable callback to the SdkService(/Manager) that would handle all the SDK business then execute the callback upon completion.

I see now that you are instead returning a latch that is used _by the caller_ to block until ready. (so my original comment here may not be applicable).

Lets take a closer look at this during the in person meeting

ivanporty:
To do list after the team review meeting:
- [ ] refactor blocking method into two for different blocking strategies
- [ ] make error messages less strict if Managed SDK can still be re-installed, suggest to just re-run
- [ ] investigate IDE level dependency on running task to make deployment wait for install task instead of manual sync

etanshaul:
I'd add "renaming callbacks" to that list :)

ivanporty:
@etanshaul @patflynn I cleaned up the code to make it a bit more generic and easier to digest and I think addressed review notes and suggestions from our meeting. Couple of them may deviate from this PR so I filed separate issues for tracking them. PTAL, should be ready for review.

etanshaul:
thanks. will start on this soon.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: was this comment addressed? If I'm not mistaken it looks like there are still duplicated messages.

etanshaul: that is, 
`appengine.run.server.sdk.misconfigured.panel.message` and
`appengine.run.server.sdk.misconfigured.message `

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
```diff
@@ -235,12 +237,18 @@ public int hashCode() {
...
-        CloudSdkValidator.getInstance().validateCloudSdk();
-    if (!sdkValidationResult.isEmpty()) {
-      CloudSdkValidationResult result = Iterables.getFirst(sdkValidationResult, null);
-      throw new RuntimeConfigurationError(
-          GctBundle.message("appengine.flex.config.server.error", result.getMessage()));
+    // do not check SDK if it supports dynamic install - the deployment runner will block itself
+    // until installation is done.
+    CloudSdkService cloudSdkService = CloudSdkService.getInstance();
+    SdkStatus sdkStatus = cloudSdkService.getStatus();
+    if (sdkStatus != SdkStatus.READY && !cloudSdkService.isInstallReady()) {
```
etanshaul: so you only check for an invalid SDK if the sdk is not ready and the service is no ready to install? could you explain this? Im having a hard time understanding.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineRuntimeInstance.java
```diff
@@ -84,18 +83,42 @@ public void deploy(
...
-                          task.getProject(),
-                          GctBundle.message("appengine.deployment.status.deploying"),
-                          true /* canBeCancelled */,
-                          null /* backgroundOption */) {
-                        @Override
-                        public void run(@NotNull ProgressIndicator indicator) {
-                          ApplicationManager.getApplication().invokeLater(deployRunner);
-                        }
-                      });
+              CloudSdkLogger cloudSdkLogger =
```
etanshaul: wdyt about un-inlining this to make this method cleaner?

etanshaul: just to confirm too; CloudSdkLogger is used to add _Cloud SDK specific_ log messaging to whatever task is being blocked by its installation (e.g. deploy or local-run). This is why you need to have the caller specify this callback (since its specific messaging related to the task, and a specific way to indicate errors. Does that sound correct?

I feel like maybe CloudSdkLogger still isn't the perfect name since it also handles failing the task (e.g. `callback.errorOccured`.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
```diff
@@ -43,6 +77,196 @@ public void onNewCloudSdkServiceTypeSelected(CloudSdkServiceType newServiceType)
...
+   * @param project Project to which runnable belongs.
+   * @param runAfterSdkReady Runnable to execute after Cloud SDK is ready. This runnable will be
+   *     executed on the application UI thread.
+   * @param progressMessage Message to show in progress dialog to identify which process is
+   *     starting, i.e. deployment or local server.
+   * @param sdkLogger Logger for errors etc.
+   */
+  public void runWhenSdkReady(
+      @Nullable Project project,
+      @NotNull Runnable runAfterSdkReady,
```
etanshaul: nit: usually start the callback name with something like `after`, `on`, `before` etc. So in this case, something like `afterSdkReady`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
```diff
@@ -43,6 +77,196 @@ public void onNewCloudSdkServiceTypeSelected(CloudSdkServiceType newServiceType)
...
+   * Blocks current thread until Cloud SDK is ready for all operations. If process results in error
+   * or user cancel, calls {@link CloudSdkLogger} methods to notify about errors and shows
+   * notifications. This method is expected to be called from non-UI background thread.
+   *
+   * @param project Project to which SDK operation belongs.
+   * @param progressMessage Message to show in progress dialog to identify which process is
+   *     starting, i.e. deployment or local server.
+   * @param sdkLogger Callback to log errors and progress.
+   */
+  public void waitWhenSdkReady(
```
etanshaul: hmm something off about this name. I like the change to the other method `runWhenSdkReady`, but this one made me pause. maybe just `waitUntilSdkReady` or `blockUntilSdkReady`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
```diff
@@ -43,6 +77,196 @@ public void onNewCloudSdkServiceTypeSelected(CloudSdkServiceType newServiceType)
...
+    blockingCompletedLatch.await();
+  }
+
+  /**
+   * Performs generic wait in a separate thread for SDK to become ready, returns immediately. Must
+   * be called from UI thread.
+   */
+  private void doWait(
+      @Nullable Project project,
+      Runnable runAfterWaitComplete,
```
etanshaul: same thing: -> `afterWaitComplete`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
```diff
@@ -43,6 +77,196 @@ public void onNewCloudSdkServiceTypeSelected(CloudSdkServiceType newServiceType)
...
+      @NotNull Runnable runAfterSdkReady,
+      String progressMessage,
+      CloudSdkLogger sdkLogger) {
+    doWait(
+        project,
+        () -> {
+          // at this point the installation should be either ready, failed or user cancelled.
+          // run only if ready.
+          if (CloudSdkService.getInstance().getStatus() == SdkStatus.READY) {
+            ApplicationManager.getApplication().invokeLater(runAfterSdkReady);
```
etanshaul: is it necessary to wrap the callback in `invokeLater` here?

ivanporty: `doWait()` runs this from background thread, and this method needs to execute `afterSdkReady` on UI thread, so yes.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
ivanporty: I like `blockUntilSdkReady`, thanks, I'll use that.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
ivanporty: The idea is deployment configuration should be marked invalid only if SDK is invalid (not ready) and cannot re-install itself to fix it (i.e. custom SDK path). In this case the configuration is blocked and a user is asked to fix it himself. Otherwise we'll try to fix it by re-installing the SDK. Hopefully this makes it a bit clearer? Open to suggestions how to handle this validation case cleaner.

etanshaul: what if the install is in progress? Will `CloudSdkService#isInstallReady` not return false in this case (and thus allowing the config here to mark itself invalid)?

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
ivanporty: Correct, they are the same for now, but I plan to expand the message shown after actual local run (`appengine.run.server.sdk.misconfigured.message`) so it could include additional information and possibly "local run" specific wording (part of #1968). Sorry for the confusion

etanshaul: got it. sgtm

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineRuntimeInstance.java
ivanporty: Absolutely correct, and yes, I agree about name not being perfect. We might need to make it longer but easier to understand what it does. `CloudSdkFeedbackHandler`?

etanshaul: how about `CloudSdkStatusHandler`?

ivanporty: Cool name, let's use that!

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
ivanporty: `isInstallReady` will be true in this case, it means 'ready to install in general', not at the specific moment of call, so the check will not pass.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
```diff
@@ -43,6 +77,196 @@ public void onNewCloudSdkServiceTypeSelected(CloudSdkServiceType newServiceType)
...
+        .invokeLater(
+            () -> doWait(project, blockingCompletedLatch::countDown, progressMessage, sdkLogger));
+    blockingCompletedLatch.await();
+  }
+
+  /**
+   * Performs generic wait in a separate thread for SDK to become ready, returns immediately. Must
+   * be called from UI thread.
+   */
+  private void doWait(
```
etanshaul: nit, but this method feels a bit long to be able to easily digest. is it possible to break its contents down into well-named methods?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/CloudToolsPluginInitializationComponent.java
```diff
@@ -62,6 +63,10 @@ public void initComponent() {
 
     new ConflictingAppEnginePluginCheck().notifyIfConflicting();
     new GoogleAccountPluginUninstaller().uninstallIfPresent();
+
+    if (!ApplicationManager.getApplication().isUnitTestMode()) {
+      ServiceManager.getService(CloudSdkServiceManager.class).getCloudSdkService().activate();
```
etanshaul: does this mean that the Cloud SDK will start installing immediately as the user boots up the IDE?

I'm trying to recall what we decided here on doing it like this, or lazily when the user tries to run an action that needs it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
etanshaul: in that case I think we should rename `isInstallReady`. Do `isInstallable` or `isInstallAvailable` capture the meaning you are after?

etanshaul: i think `isInstallable` is best, assuming it captures the correct meaning

ivanporty: I think it's better. I remember we started with `supportsInstall` but then decided to change to `isInstallReady` so the truth is somewhere in there. `isInstallable` sounds correct.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/startup/CloudToolsPluginInitializationComponent.java
ivanporty: I think we agreed on installing as soon as plugin is activated so that we don't have to block initial deploy/run action (for possibly long time)

etanshaul: I'm ok with that for now.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
ivanporty: Sure, thinking about it.

ivanporty:
@etanshaul I think your comments has been mostly addressed including some simplification around `doWait()`, PTAL when you can, thanks!

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
```diff
@@ -43,6 +77,214 @@ public void onNewCloudSdkServiceTypeSelected(CloudSdkServiceType newServiceType)
...
+  @VisibleForTesting
+  boolean checkIfCancelled() {
+    return ProgressManager.getInstance().getProgressIndicator().isCanceled();
+  }
+
+  /**
+   * Checks if current SDK service is ready for operations. If it's not, install is attempted if
+   * supported, and true is returned.
+   */
+  private boolean isInstallInProgress() {
```
etanshaul: just noticed this. I think an `isInstallInProgress` method should only return if the install is running. The side-effect of kicking off the install should probably be refactored out.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
```diff
@@ -43,6 +77,214 @@ public void onNewCloudSdkServiceTypeSelected(CloudSdkServiceType newServiceType)
...
+   * Performs generic wait in a separate thread for SDK to become ready, returns immediately. Must
+   * be called from UI thread.
+   */
+  private void doWait(
+      @Nullable Project project,
+      Runnable afterWaitComplete,
+      String progressMessage,
+      CloudSdkStatusHandler sdkLogging) {
+    CloudSdkService cloudSdkService = CloudSdkService.getInstance();
+    boolean installInProgress = isInstallInProgress();
```
etanshaul: similar to my other comment about `isInstallInProgress`. With the current structure its not clear that this is what is actually kicking off the install.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
```diff
@@ -235,12 +237,18 @@ public int hashCode() {
...
-        CloudSdkValidator.getInstance().validateCloudSdk();
-    if (!sdkValidationResult.isEmpty()) {
-      CloudSdkValidationResult result = Iterables.getFirst(sdkValidationResult, null);
-      throw new RuntimeConfigurationError(
-          GctBundle.message("appengine.flex.config.server.error", result.getMessage()));
+    // do not check SDK if it supports dynamic install - the deployment runner will block itself
+    // until installation is done.
+    CloudSdkService cloudSdkService = CloudSdkService.getInstance();
+    SdkStatus sdkStatus = cloudSdkService.getStatus();
+    if (sdkStatus != SdkStatus.READY && !cloudSdkService.isInstallSupported()) {
```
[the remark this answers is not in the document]
ivanporty: Good idea, adding.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
ivanporty: I think it makes it more confusing to extract anything from this code block since it after calling `install()` you can't be sure exactly if SDK status has changed yet (this is async operation). I moved it back as a block, it could possibly be more convoluted if I break this 5 lines of code into two helper boolean methods and another logic using them.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeploymentConfiguration.java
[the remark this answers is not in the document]
ivanporty: done.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
```diff
@@ -139,7 +139,15 @@ private void doWait(
       String progressMessage,
       CloudSdkStatusHandler sdkLogging) {
     CloudSdkService cloudSdkService = CloudSdkService.getInstance();
-    boolean installInProgress = isInstallInProgress();
+    SdkStatus sdkStatus = cloudSdkService.getStatus();
+    boolean installInProgress = sdkStatus == SdkStatus.INSTALLING;
```
etanshaul: what if you kept the `isInstallInProgress` method that just returned this (`sdkStatus == SdkStatus.INSTALLING`)
Then, here you check:

```
if (!isInstallInProgress()....)  {
  install();
}
```
Then later, when you need to know if its in progress again, just call `isInstallInProgress()` again. Presumably `install` should internally update the state correctly so that `isInstallInProgress()` returns the correct thing. I think that would be cleaner then updating the boolean like you do here on line 149).

wdyt

ivanporty: I made a few changes so we're not using flags anymore, and it looks a bit nicer this way. As for the guard, I feel we might still keep it just to be sure install() won't be called if not supported (it could break things if another SDK implementation changes things)

etanshaul:
@ivanporty can you remind me what the status of this is now? just waiting for final approval?

ivanporty:
Correct, the only minor readability thing to look at is ab10b18 commit, otherwise just waiting to be approved.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
```diff
@@ -43,6 +77,203 @@ public void onNewCloudSdkServiceTypeSelected(CloudSdkServiceType newServiceType)
...
+    // blocking thread.
+    final CloudSdkService.SdkStatusUpdateListener sdkStatusUpdateListener =
+        (sdkService, status) -> {
+          switch (status) {
+            case READY:
+            case INVALID:
+            case NOT_AVAILABLE:
+              installationCompletionLatch.countDown();
+              break;
+            case INSTALLING:
```
etanshaul: nit: the style guide suggests to always use a default case. instead of the "INSTALLING" case, you could put in a default that just has the comment "continue waiting for completion" (and the break is redundant here too, no?).

etanshaul:
the approval comment was supposed to be "good to go!", not a question. sorry

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
ivanporty: good point, thanks!

PR #1966, opened 2018-03-22
Writes the selected BOM from the client library wizard to the user's pom.xml

etanshaul:
progress towards #1921 

1) The user select a version of the BOM to use when adding a client library:
image

2) If the user previously did not have a BOM, the new depedency is written to the dependencyManagement section:
image
If the user _did_ already have a bom, then the version is updated to match the selected version in the UI.

3) The client library is written without a version specified:
image

4) If no BOM is selected (i.e. if the flag is disabled or the API call to fetch the BOM versions fail), this will fallback to not writing any BOM and using the version from the libraries.json file.

5) The notification is now shown without a version, since the version may be managed by the BOM:
image

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
@@ -87,7 +87,7 @@ final class AddCloudLibrariesDialog extends DialogWrapper {
     return cloudApiSelectorPanel.getApisToEnable();
   }
 
-  /** Returns, optionally, the selected BOM version. */
+  /** Returns the selected BOM version. */
   String getSelectedBomVersion() {
     return cloudApiSelectorPanel.getSelectedBomVersion();
   }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
@@ -31,6 +31,7 @@ import java.nio.file.Path;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.swing.JComponent;
@@ -85,6 +86,11 @@ final class AddCloudLibrariesDialog extends DialogWrapper {
   /** Returns the set of {@link CloudLibrary APIs} to enable. */
   Set<CloudLibrary> getApisToEnable() {
     return cloudApiSelectorPanel.getApisToEnable();
+  }
+
+  /** Returns, optionally, the selected BOM version. */
+  Optional<String> getSelectedBomVersion() {
+    return cloudApiSelectorPanel.getSelectedBomVersion();
   }
 
   /**
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
@@ -31,6 +31,7 @@ import java.nio.file.Path;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.swing.JComponent;
@@ -87,8 +88,8 @@ final class AddCloudLibrariesDialog extends DialogWrapper {
     return cloudApiSelectorPanel.getApisToEnable();
   }
 
-  /** Returns the selected BOM version. */
-  String getSelectedBomVersion() {
+  /** Returns, optionally, the selected BOM version. */
+  Optional<String> getSelectedBomVersion() {
     return cloudApiSelectorPanel.getSelectedBomVersion();
   }
 
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
@@ -31,7 +31,6 @@ import java.nio.file.Path;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.swing.JComponent;
@@ -89,7 +88,7 @@ final class AddCloudLibrariesDialog extends DialogWrapper {
   }
 
   /** Returns, optionally, the selected BOM version. */
-  Optional<String> getSelectedBomVersion() {
+  String getSelectedBomVersion() {
     return cloudApiSelectorPanel.getSelectedBomVersion();
   }
 
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
@@ -255,12 +255,21 @@ final class CloudLibraryDependencyWriter {
   /**
    * Joins the given list of {@link MavenId MavenIds} into a human-readable string.
    *
+   * <p>Does not include the version as the version of the dependency may be managed by a BOM.
+   *
    * @param mavenIds the list of {@link MavenId MavenIds} to join
    */
   private static String joinMavenIds(List<MavenId> mavenIds) {
     return mavenIds
         .stream()
-        .map(MavenId::getDisplayString)
+        .map(
+            mavenId -> {
+              StringBuilder builder = new StringBuilder();
+              MavenId.append(builder, mavenId.getGroupId());
+              MavenId.append(builder, mavenId.getArtifactId());
+
+              return builder.toString();
+            })
         .map(string -> "- " + string)
         .collect(Collectors.joining("<br>"));
   }
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
@@ -156,7 +156,7 @@ final class CloudLibraryDependencyWriter {
             .stream()
             .filter(
                 mdd ->
-                    CloudApiMavenService.GOOGLE_CLOUD_JAVA_BOM_ARTIFACT_NAME.equals(
+                    CloudApiMavenService.GOOGLE_CLOUD_JAVA_BOM_ARTIFACT.equals(
                         mdd.getArtifactId().getStringValue()))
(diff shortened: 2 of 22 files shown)


etanshaul:
note: as discussed offline, this workflow will _not_ strip out the versions of dependencies already specified in the pom.xml. Instead, we'll use inspections for this case.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
```diff
@@ -121,23 +130,81 @@ protected void run(@NotNull Result result) throws Throwable {
...
+   */
+  private static void writeNewMavenDependency(
+      MavenDomProjectModel model, MavenId mavenId, @Nullable String bomVersion) {
+    MavenDomDependency dependency = MavenDomUtil.createDomDependency(model, /* editor= */ null);
+    dependency.getGroupId().setStringValue(mavenId.getGroupId());
+    dependency.getArtifactId().setStringValue(mavenId.getArtifactId());
+
+    // Only write the version to the dependency if there is no BOM
+    if (bomVersion == null) {
+      dependency.getVersion().setStringValue(mavenId.getVersion());
```
ivanporty: So just for refreshing on this, once `createDomDependency` is called, calling the `set` method family updates `pom.xml` automatically?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
```diff
@@ -121,23 +130,81 @@ protected void run(@NotNull Result result) throws Throwable {
...
+
+  /** Writes the google-cloud-java BOM to the user's pom.xml. */
+  private static void writeNewBom(MavenDomProjectModel model, String bomVersion) {
+    MavenDomDependency bomDependency =
+        model.getDependencyManagement().getDependencies().addDependency();
+    bomDependency.getGroupId().setStringValue(CloudApiMavenService.GOOGLE_CLOUD_JAVA_BOM_GROUP);
+    bomDependency
+        .getArtifactId()
+        .setStringValue(CloudApiMavenService.GOOGLE_CLOUD_JAVA_BOM_ARTIFACT);
+    bomDependency.getVersion().setStringValue(bomVersion);
```
ivanporty: Could we also benefit from collecting usage of specific BOM versions people use, or maybe just latest/non-latest flag?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
```diff
@@ -121,23 +130,81 @@ protected void run(@NotNull Result result) throws Throwable {
...
+  }
+
+  /**
+   * Writes the google-cloud-java dependency to the users pom.xml.
+   *
+   * <p>If a BOM version is supplied, the dependency will not be written with a version so that the
+   * BOM can manage the version.
+   */
+  private static void writeNewMavenDependency(
+      MavenDomProjectModel model, MavenId mavenId, @Nullable String bomVersion) {
```
ivanporty: Nit: do we need the actual value of `bomVersion` here, would a boolean flag like `includeVersion` make it a bit simpler?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
etanshaul: correct. 

The previously used overloaded `createDomDependency` method takes in a MavenId and does these `setStringValue` calls for you. But since we now want to only conditionally set the version, we have to do these externally.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
```diff
@@ -121,23 +130,81 @@ protected void run(@NotNull Result result) throws Throwable {
...
+        mavenDomDependencies
+            .getDependencies()
+            .stream()
+            .filter(
+                mdd ->
+                    CloudApiMavenService.GOOGLE_CLOUD_JAVA_BOM_ARTIFACT.equals(
+                        mdd.getArtifactId().getStringValue()))
+            .findFirst();
+
+    if (!bomDependencyOptional.isPresent()) {
```
ivanporty: I think you mentioned this for the next PR, +1 to unit test this logic.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
etanshaul: absolutely. I intended to add in metrics for the entire BOM effort in a separate PR.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
etanshaul: good idea

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
etanshaul: I'm going to take a stab at doing the refactoring for testability here and see if it distracts too much from this PR. that way we can get the tests in now together with these updates

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/AddCloudLibrariesDialog.java
```diff
@@ -87,6 +87,11 @@ CloudProject getCloudProject() {
     return cloudApiSelectorPanel.getApisToEnable();
   }
 
+  /** Returns the selected BOM version. */
+  String getSelectedBomVersion() {
```
ivanporty: Same thought about `Optional`

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
```diff
@@ -150,6 +150,12 @@ Module getSelectedModule() {
     return ((CloudLibraryTableModel) cloudLibrariesTable.getModel()).getSelectedLibraries();
   }
 
+  /** Returns the selected BOM version. */
+  String getSelectedBomVersion() {
```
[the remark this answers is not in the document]
ivanporty: Maybe just return Optional here, and use actual value as a parameter then?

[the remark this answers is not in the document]
ivanporty: But probably it's not really relevant after all. Quite controversial.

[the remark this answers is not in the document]
etanshaul: yeah thats a good idea. I'd still be doing `if (bomVersion == null)` everywhere in the maven utility but at least it will make these getters clearer.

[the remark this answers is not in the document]
etanshaul: oh I thought you mean to update the getters to return an Optional<String>, but thing just to unpack the optional before sending it to `CloudLibraryDependencyWriter`. either way, I think that is a good idea.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriter.java
etanshaul: done - in this PR

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriterTest.java
```diff
@@ -0,0 +1,159 @@
...
+              assertThat(bomDependency.getGroupId().getStringValue()).isEqualTo("com.google.cloud");
+              assertThat(bomDependency.getArtifactId().getStringValue())
+                  .isEqualTo("google-cloud-bom");
+              assertThat(bomDependency.getScope().getStringValue()).isEqualTo("import");
+              assertThat(bomDependency.getType().getStringValue()).isEqualTo("pom");
+              assertThat(bomDependency.getVersion().getStringValue()).isEqualTo(bomVersion);
+            });
+  }
+
+  private MavenDomProjectModel getMavenDomProjectModel(@Nullable String bomVersion) {
```
ivanporty: Reading tests it's kind of a bit hard to see that this method actually adds a library into Maven DOM. Maybe add an optional library to add so when test is set up, it's easy to see we added a library, or use a name that mentions cloud library?

etanshaul: agreed. lemme think about what the best option is

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriterTest.java
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  @Test
+  public void addLibraries_withNoBom_doesNotWriteBom() {
+    ApplicationManager.getApplication()
+        .invokeAndWait(
+            () -> {
+              MavenDomProjectModel model = getMavenDomProjectModel(null /*bomVersion*/);
+
+              assertThat(model.getDependencyManagement().getDependencies().getDependencies())
```
ivanporty: Mind bending call :) Could we probably expand this into a variable or two with a name so we can see what exactly is tested?

etanshaul: lol. good idea

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriterTest.java
etanshaul: changed it `createMavenModelWithCloudLibrary`. Let me know if this doesn't help the situation.

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriterTest.java
etanshaul: done

google-cloud-tools-plugin/testSrc/com/google/cloud/tools/intellij/apis/CloudLibraryDependencyWriterTest.java
ivanporty: awesome, reads easy!

PR #1982, opened 2018-03-27
Fix NPE with null MavenDomProjectModel

### LINKED ISSUE #1980: User reported NPE in CloudLibraryProjectState
Stack trace:
java.lang.NullPointerException
	at com.google.cloud.tools.intellij.apis.CloudLibraryProjectState.lambda$getModuleDependencies$1(CloudLibraryProjectState.java:137)
	at com.intellij.openapi.application.impl.ApplicationImpl.runReadAction(ApplicationImpl.java:927)
	at com.google.cloud.tools.intellij.apis.CloudLibraryProjectState.getModuleDependencies(CloudLibraryProjectState.java:131)
	at com.google.cloud.tools.intellij.apis.CloudLibraryProjectState.loadManagedLibraries(CloudLibraryProjectState.java:104)
	at java.util.stream.Collectors.lambda$toMap$58(Collectors.java:1321)
	at java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)

etanshaul:
fixes #1980 

The model could be null if its a maven project, but the pom is invalid in some way. In this case, we just return no dependencies.

Updated the method to not return null lists, and added a unit test to cover the bug.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
@@ -18,6 +18,7 @@ package com.google.cloud.tools.intellij.apis;
 
 import com.google.cloud.tools.libraries.json.CloudLibrary;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
 import com.intellij.openapi.application.ApplicationManager;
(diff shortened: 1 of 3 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
```diff
@@ -134,7 +135,9 @@ void syncManagedProjectLibraries() {
                   MavenDomProjectModel model =
                       MavenDomUtil.getMavenDomProjectModel(project, mavenProject.getFile());
 
-                  return model.getDependencies().getDependencies();
+                  return model != null
+                      ? model.getDependencies().getDependencies()
```
ivanporty: will `model.getDependencies()` always be non-null as well?

etanshaul: it is @NotNull. But I am still checking in the caller that the return of this method != null.

etanshaul: oh nm. yes model.getDependencies is @NotNull. Thought you were asking about getDependencies#getDependencies at first

PR #1991, opened 2018-03-28
Adds inspection for cloud dependencies with version when BOM imported

etanshaul:
in support of #1921 

Adds a new inspection that does the following:
1) Finds cases of google cloud library depedencies in the pom.xml that have the `<version>` tag defined, when a BOM is also imported (in the `dependencyManagement` section). Will highlight (with a warning) all of the violating versions
2) Sets up a quick-fix where the user can quickly delete the version

**Versions are highlighted when the BOM is also imported**
image

**Inspection warning wording**
image

**Inspection menu**
image

**Inspection profile settings**
image

Wording was tricky here so seeking any feedback.

I'm still working on properly setting up unit tests to test this inspection / quickfix.


etanshaul:
@ivanporty @nbashirbello 

feel free to start looking at the code / wording / screenshots while I finish up the unit testing.

etanshaul:
Which part of having a multi-module setup were you concerned about? The inspection runs on every pom.xml in the project. The inspection code itself also checks against the state of the current module (to test if it has a BOM, etc.)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/DependencyVersionWithBomInspection.java
```diff
@@ -0,0 +1,172 @@
...
+import org.jetbrains.idea.maven.dom.model.MavenDomProjectModel;
+
+/**
+ * An {@link DomElementsInspection} that detects google-cloud-java dependencies that have an
+ * explicit version definition when a BOM is defined.
+ *
+ * <p>Provides a quick-fix to strip out the version tag from the dependency.
+ */
+public class DependencyVersionWithBomInspection
+    extends DomElementsInspection<MavenDomProjectModel> {
```
ivanporty: It looks like `DomElementsInspection` is a standard inspection for XML files? I don't see any filtering for a specific file names or schemas, so I guess it does checking based on the DomElement type you pass as a parameter?

etanshaul: exactly. its based on the DOM model inspection is parameterized by 

e.g. `MavenDomProjectModel` in this case.

ivanporty:
I thought about if it was going check every pom.xml in the project regardless of the structure or not, so since it runs on every pom.xml, that's not an issue.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/DependencyVersionWithBomInspection.java
etanshaul: If you were curious, see `DomElementsInspection#checkFile`. You can see that it first checks to ensure that the file under inspection is of the same class as the DOM model that is passed in (`myDomClasses`); so it will quickly skip all other files.

ivanporty: nice, makes it look easy to add another XML element inspection without low-level work!

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/DependencyVersionWithBomInspection.java
```diff
@@ -0,0 +1,172 @@
...
+                    GctBundle.message(
+                        "cloud.libraries.version.with.bom.inspection.problem.description"),
+                    new StripDependencyVersionQuickFix());
+              }
+            });
+  }
+
+  /** Checks to see if the {@link MavenDomDependency} has a version defined. */
+  private boolean hasVersion(MavenDomDependency dependency) {
+    return dependency.getVersion().exists();
```
ivanporty: So if version element exists, this is definitely a warning - it cannot be empty, correct?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/DependencyVersionWithBomInspection.java
etanshaul: indeed. and if you look at my commit history here, you'll see that I didn't realize that this was available at first. Initially I had a lower level inspector that was based on XmlElement (with no DOM schema). This is much nicer.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/DependencyVersionWithBomInspection.java
```diff
@@ -0,0 +1,172 @@
...
+    @Nls
+    @NotNull
+    @Override
+    public String getFamilyName() {
+      return GctBundle.message("cloud.libraries.version.with.bom.quickfix.title");
+    }
+
+    @Override
+    public void applyFix(@NotNull Project project, @NotNull ProblemDescriptor descriptor) {
+      XmlTextImpl xmlElement = (XmlTextImpl) descriptor.getPsiElement();
```
ivanporty: `descriptor.getPsiElement()` cannot be null itself in case of some internal structure changes?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/DependencyVersionWithBomInspection.java
etanshaul: are you asking if the version tag has an empty value, will it return true/false? Ill check this.

ivanporty: exactly, thanks

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/DependencyVersionWithBomInspection.java
etanshaul: what do you mean by internal structure changes?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/DependencyVersionWithBomInspection.java
ivanporty: in the light of recent NPEs from using IJ APIs, can `dependency.getVersion()` be null in some case as well? :)

etanshaul: there are only 2 cases I can think of:
1) the version tag is entirely missing
2) the version tag exists, but the value is empty

I know that case 1 is covered. I'll check on case 2 which was your original question, but I doubt it would trigger an NPE given case 1. I guess it still doesn't hurt to wrap it in a null check to be safe.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/DependencyVersionWithBomInspection.java
ivanporty: It looks like it should not be null since it indicates the issue with the element and should not be called if element itself is gone, but since we use a lot of IJ APIs and they sometime return null unexpectedly, I wonder if this can happen here as well - cannot think of any reason except some internal (IDE) level mismatches in inspection/fix state

etanshaul: Yeah. I agree with your reasoning. I'll add a check though just to be safe, and will raise a `log.error` if it happens so we get notified. wdyt?

ivanporty: awesome. I start feeling everything could be null coming from an IDE :)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/DependencyVersionWithBomInspection.java
```diff
@@ -0,0 +1,172 @@
...
+import com.intellij.util.xml.highlighting.DomElementsInspection;
+import java.util.Set;
+import org.jetbrains.annotations.Nls;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.jetbrains.idea.maven.dom.model.MavenDomDependency;
+import org.jetbrains.idea.maven.dom.model.MavenDomProjectModel;
+
+/**
+ * An {@link DomElementsInspection} that detects google-cloud-java dependencies that have an
```
nbashirbello: An -> A

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/DependencyVersionWithBomInspection.java
```diff
@@ -0,0 +1,172 @@
...
+ * explicit version definition when a BOM is defined.
```
nbashirbello: Do you need to mention that this inspection is only for maven projects?

etanshaul: yea good point. I should mention that. Thanks.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/DependencyVersionWithBomInspection.java
```diff
@@ -0,0 +1,172 @@
...
+                    .map(
+                        coords ->
+                            domValueEquals(dependency.getGroupId(), coords.getGroupId())
+                                && domValueEquals(
+                                    dependency.getArtifactId(), coords.getArtifactId()))
+                    .orElse(false));
+  }
+
+  private boolean domValueEquals(GenericDomValue domValue, @Nullable String value) {
+    return domValue.getStringValue() != null && domValue.getStringValue().equals(value);
```
ivanporty: Could `domValue` itself be null, worth checking too maybe?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
```diff
@@ -47,7 +49,10 @@
 public class CloudLibraryProjectState implements ProjectComponent {
 
   private final Project project;
+
   private Map<Module, Set<CloudLibrary>> moduleLibraryMap = Maps.newHashMap();
+  private Map<Module, Optional<MavenDomDependency>> moduleBomMap = Maps.newHashMap();
```
ivanporty: This would be a map of project module to a BOM XML element in pom.xml, correct? `MavenDomDependency` sound a bit generic, maybe worth adding a comment clarifying what exactly this maps contains?

etanshaul: I named the variable `moduleBomMap` to try and make the clear. But I can definitely add a comment. Along the same lines, do you find that there is confusion with the `moduleLibraryMap` too?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/DependencyVersionWithBomInspection.java
etanshaul: I don't think it can (this is the identical situation as the other NPE question you had). But I'll add a check to be safe.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
ivanporty: `moduleLibraryMap` seems pretty clear on the other hand, since it has a specifically named type (`CloudLibrary`) as a value.

etanshaul: makes sense. ill update. thanks

etanshaul:
@ivanporty @nbashirbello - I've updated the comments from code-review. ptal

The setup required to unit test this properly is proving to be very complicated. I haven't managed to get it working yet, so for the sake of time I think we can move forward with merging this and create a separate issue to track this effort. Let me know if you have concerns with this.

etanshaul:
Aside from the code, ptal at the wording and messaging too if you haven't looks closely already.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -274,6 +274,12 @@ cloud.apis.service.account.add.roles.progress.message=Adding roles to service ac
 cloud.apis.service.account.create.key.progress.message=Creating a service account key
 cloud.apis.service.account.download.key.progress.message=Downloading the service account key
 
+cloud.libraries.bom.inspection.group.name=Google Cloud Java Library BOM Inspections
+cloud.libraries.version.with.bom.inspection.display.name=Dependency Version with Bom Inspection
+cloud.libraries.version.with.bom.inspection.description=Inspection that checks for presence version tags in google cloud java Maven dependencies when a BOM has been imported.
```
ivanporty: "...checks for presence *of* version tags..." maybe?

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -274,6 +274,12 @@ cloud.apis.service.account.add.roles.progress.message=Adding roles to service ac
 cloud.apis.service.account.create.key.progress.message=Creating a service account key
 cloud.apis.service.account.download.key.progress.message=Downloading the service account key
 
+cloud.libraries.bom.inspection.group.name=Google Cloud Java Library BOM Inspections
+cloud.libraries.version.with.bom.inspection.display.name=Dependency Version with Bom Inspection
```
ivanporty: Should we keep uppercase BOM as well here?

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: oops. yeah.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
etanshaul: yes, should be fully capitalized everywhere

PR #1995, opened 2018-03-30
upgrade gradle-intellij-plugin to latest

patflynn:
I have no idea why the plugin upgrades catches that bad NotNull import


ivanporty:
I guess they figured out how many other @NotNull checks are out there and decided to check against it, not a bad idea!

Issue #1985, opened 2018-03-27
flaky test - thread leak RemoteMavenServer


etanshaul:
Seems to only affect CI (not local).

```
com.google.cloud.tools.intellij.apis.CloudLibraryDependencyWriterTest > addLibraries_withNoBomSelected_doesNotWriteBom FAILED
    java.lang.RuntimeException: java.lang.AssertionError: Thread leaked: Thread[BaseDataReader: output stream of /usr/lib/jvm/java-8-oracle/jre/bin/java -Djava.awt.headless=true -Didea....eaIU/2017.2/1339e842f043efed378658bb7e987dee5fbf76cd/ideaIU-2017.2/plugins/maven/lib/maven3/boot/plexus-classworlds-2.5.2.jar org.jetbrains.idea.maven.server.RemoteMavenServer,4,main]; TIMED_WAITING (true)
    --- its stacktrace:
     at java.lang.Object.wait(Native Method)
     at com.intellij.util.io.BaseDataReader.doRun(BaseDataReader.java:175)
     at com.intellij.util.io.BaseDataReader$1$1.run(BaseDataReader.java:66)
     at com.intellij.util.ConcurrencyUtil.runUnderThreadName(ConcurrencyUtil.java:194)
     at com.intellij.util.io.BaseDataReader$1.run(BaseDataReader.java:63)
     at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
     at java.util.concurrent.FutureTask.run(FutureTask.java:266)
     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
     at java.lang.Thread.run(Thread.java:748)
    ---
```

PR #1998, opened 2018-04-02
Shutdown the maven server explicitly after tests

etanshaul:
Playing around to see if I can fix flaky tests (#1985)


etanshaul:
If this fix works, then we can integrate this into the `@CloudToolsRule`: we can create a new `@TestMavenModule` annotation that sets up all of the maven junk that we are doing explicitly from the tests, then shut down the server directly in `CloudToolsRule` on teardown.

But first, just checking if it even works

etanshaul:
on second look getting it work the way I describe in my last comment will take more work than in scope here: `@CloudToolsRule` lives in `common-test-lib` which has no dependency on the intellij maven plugin (so there is no way, without a larger refactoring, to initialize maven modules in `CloudToolsRule`). 

I suggest we merge this in as is, and then once the larger module refactoring is complete to revisit this.

@ivanporty @patflynn @nbashirbello ptal.

etanshaul:
oh and @patflynn confirmed that he was no longer seeing the Maven server related Thread leaks locally (where it was consistently reproducible) after this fix.

PR #1999, opened 2018-04-02
Managed SDK enhancements

### LINKED ISSUE #673: Auto download cloud sdk if not installed

### LINKED ISSUE #1964: Do not show success notification and balloon on start if Managed SDK is already installed.
Part of #673 flow.
Log message should be enough.

### LINKED ISSUE #1992: Make sure custom SDK is used by default for users who configured it prior to managed SDK support
To prevent users with their specific SDK installations to silently switch to Managed SDK behind the scenes.

### LINKED ISSUE #1993: Increase Managed SDK resiliency between uses
As an example scenario:
1. Plugin is installed, Managed SDK is installed and ready to use
2. IDE is running without issues but SDK became corrupted or someone removes it from the system
3. Ensure Managed SDK status is re-checked before deployment and runs to avoid intermediate state

ivanporty:
Closes #673.
Fixes #1992.
Makes sure custom SDK is used by default if custom SDK path was already configured and non-empty.
Fixes #1993.
Resiliency - ensures managed SDK status is correct (calls `install()` regardless of ready or not to double check current SDK state) prior to deployments / local run 
Fixes #1964.
Will not show notification and balloons for managed SDK successful operations if managed SDK was already up-to-date and no install / update actually happened.

### DIFF
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
@@ -139,8 +139,9 @@ public class CloudSdkServiceManager {
       String progressMessage,
       CloudSdkStatusHandler sdkLogging) {
     CloudSdkService cloudSdkService = CloudSdkService.getInstance();
-    // if not ready, attempt to fix and install now.
-    if (!isSdkReady() && cloudSdkService.isInstallSupported()) {
+    // if install supported, make sure SDK is in good shape and installed.
+    // even if status is READY, this will double-check the current state from the last install.
+    if (cloudSdkService.isInstallSupported()) {
       cloudSdkService.install();
     }
 
@@ -207,10 +208,6 @@ public class CloudSdkServiceManager {
   /** Checks if current SDK service is currently installing and not ready for operations. */
   private boolean isInstallInProgress() {
     return CloudSdkService.getInstance().getStatus() == SdkStatus.INSTALLING;
-  }
-
-  private boolean isSdkReady() {
-    return CloudSdkService.getInstance().getStatus() == SdkStatus.READY;
   }
 
   /** Exposes process window so that installation / dependent processes are explicitly visible. */
--- a/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
+++ b/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
@@ -55,7 +55,7 @@ public class ManagedCloudSdkService implements CloudSdkService {
 
   private final Collection<SdkStatusUpdateListener> statusUpdateListeners = Lists.newArrayList();
 
-  private volatile ListenableFuture<Path> managedSdkBackgroundJob;
+  private volatile ListenableFuture<ManagedSdkJobResult> managedSdkBackgroundJob;
 
   private ProgressListener progressListener;
 
@@ -154,7 +154,8 @@ public class ManagedCloudSdkService implements CloudSdkService {
    * @param managedSdkTask Task to execute.
    * @return {@code true} if task started, {@code false} if Managed SDK is not supported at all.
    */
-  private boolean executeManagedSdkJob(ManagedSdkJobType jobType, Callable<Path> managedSdkTask) {
+  private boolean executeManagedSdkJob(
+      ManagedSdkJobType jobType, Callable<ManagedSdkJobResult> managedSdkTask) {
     if (!isInstallSupported()) {
       return false;
     }
@@ -175,27 +176,33 @@ public class ManagedCloudSdkService implements CloudSdkService {
    * @throws InterruptedException if Managed Cloud SDK has been interrupted by {@link
    *     #cancelInstallOrUpdate()} ()}
    */
-  private Path installManagedSdk() throws Exception {
-    Path installedManagedSdkPath = installSdk();
-    installAppEngineJavaComponent();
+  private ManagedSdkJobResult installManagedSdk() throws Exception {
+    ManagedSdkJobResult installResult = installSdk();
+    ManagedSdkJobResult appEngineJavaResult = installAppEngineJavaComponent();
 
-    return installedManagedSdkPath;
+    // return up-to-date only if both SDK and app engine Java were up-to-date
+    return (installResult == ManagedSdkJobResult.UP_TO_DATE
+            && appEngineJavaResult == ManagedSdkJobResult.UP_TO_DATE)
+        ? ManagedSdkJobResult.UP_TO_DATE
+        : ManagedSdkJobResult.PROCESSED;
   }
 
   /** Installs core managed SDK if needed and returns its path if successful. */
-  private Path installSdk() throws Exception {
+  private ManagedSdkJobResult installSdk() throws Exception {
     if (!safeCheckSdkStatus(() -> managedCloudSdk.isInstalled())) {
       ConsoleListener sdkConsoleListener = logger::debug;
       progressListener =
           ManagedCloudSdkServiceUiPresenter.getInstance().createProgressListener(this);
 
-      return managedCloudSdk.newInstaller().install(progressListener, sdkConsoleListener);
+      managedCloudSdk.newInstaller().install(progressListener, sdkConsoleListener);
+
+      return ManagedSdkJobResult.PROCESSED;
     }
 
-    return managedCloudSdk.getSdkHome();
+    return ManagedSdkJobResult.UP_TO_DATE;
   }
 
-  private void installAppEngineJavaComponent() throws Exception {
+  private ManagedSdkJobResult installAppEngineJavaComponent() throws Exception {
     if (!safeCheckSdkStatus(() -> managedCloudSdk.hasComponent(SdkComponent.APP_ENGINE_JAVA))) {
       ConsoleListener appEngineConsoleListener = logger::debug;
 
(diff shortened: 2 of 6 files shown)


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
```diff
@@ -139,8 +139,9 @@ private void doWait(
       String progressMessage,
       CloudSdkStatusHandler sdkLogging) {
     CloudSdkService cloudSdkService = CloudSdkService.getInstance();
-    // if not ready, attempt to fix and install now.
-    if (!isSdkReady() && cloudSdkService.isInstallSupported()) {
+    // if install supported, make sure SDK is in good shape and installed.
+    // even if status is READY, this will double-check the current state from the last install.
+    if (cloudSdkService.isInstallSupported()) {
```
etanshaul: so it can be READY, but in a bad state?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -255,24 +268,32 @@ private void notifyListeners(CloudSdkService sdkService, SdkStatus status) {
     UPDATE
   }
 
+  enum ManagedSdkJobResult {
```
etanshaul: do we really need another status enum in addition to the Sdk status we already have? (the answer may be yes, just wondering if it can be simplified).

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
ivanporty: correct, please see one possible scenario in #1993.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: I thought to combine them together, i.e. `INSTALL`, `UPDATE`, `UP_TO_DATE` in a single enum. However the problem is `FutureCallback` and future in general does not receive any result on failure. We wouldn't know if it was install or update since we use generic callback.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
etanshaul: I see. so install() itself will then check its status and install if it corrupted/removed/etc, correct?

Also since you are lumping a few fixes together here, is this line the only fix in support of #1993?

ivanporty: correct, this should make it double check its status before the actual operation, so it was simpler than I originally thought.

ivanporty: Yes, your understanding is totally correct.
As for configuration checking, we do not mark deployment/run configuration as invalid when SDK supports self-install at all, since we plan to install and fix it at the point of deployment (if it's broken). So for managed SDK it doesn't buy us anything to re-check the status at those points.

etanshaul: oh, so if its installed, but broken, the then the configuration will be marked as OK; but then upon deploy/run, it will be re-installed if broken?

I think that makes sense, especially since the configuration error checks should be lightweight.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: interesting. I need to explore the code more then to see if I have anything to add here.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceManager.java
ivanporty: Exactly, you're right.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: is all this just for showing a certain message in the UI (`ManagedCloudSdkUiPresenter#notifyManagedSdkJobSuccess`)?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkServiceUiPresenter.java
```diff
@@ -51,9 +52,16 @@ static void setInstance(ManagedCloudSdkServiceUiPresenter uiPresenter) {
...
 
-  public void notifyManagedSdkJobSuccess(ManagedSdkJobType jobType) {
+  public void notifyManagedSdkJobSuccess(ManagedSdkJobType jobType, ManagedSdkJobResult jobResult) {
     String message = GctBundle.message("managedsdk.success." + jobType.name().toLowerCase());
-    showNotification(message, NotificationType.INFORMATION);
+    switch (jobResult) {
+      case PROCESSED:
+        showNotification(message, NotificationType.INFORMATION);
+        break;
+      case UP_TO_DATE:
```
etanshaul: assuming we keep this enum: from what I remember from reading the style guide the other day, you should remove the UP_TO_DATE case since its handled by the default case (and you already have a nice comment there indicating that its up-to-date).

Also, the `break` above can be replaced by a `return` right? (which may be a bit more readable and resilient to changes)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
ivanporty: for this, and also to handle update slightly differently in case of a failure, keeping the rest of the logic the same.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkServiceUiPresenter.java
ivanporty: sounds good, that makes sense, updating.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: can you point me to that code (the other case of handling update differently during failure) please?

etanshaul: nm. I found it. ManagedCloudSdkService#onFailure.

etanshaul: actually that is using `ManagedSdkJobType`. the only place I could find actually using `ManagedSdkJobResult` is the Ui presenter for showing notifications. maybe I'm missing something?

ivanporty: ah yes, you mean `ManagedSdkJobResult` - this one only used to check if notification is needed, correct.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
```diff
@@ -175,27 +176,33 @@ private boolean executeManagedSdkJob(ManagedSdkJobType jobType, Callable<Path> m
...
-    Path installedManagedSdkPath = installSdk();
-    installAppEngineJavaComponent();
-
-    return installedManagedSdkPath;
+  private ManagedSdkJobResult installManagedSdk() throws Exception {
+    ManagedSdkJobResult installResult = installSdk();
+    ManagedSdkJobResult appEngineJavaResult = installAppEngineJavaComponent();
+
+    // return up-to-date only if both SDK and app engine Java were up-to-date
+    return (installResult == ManagedSdkJobResult.UP_TO_DATE
```
etanshaul: maybe we don't solve this now, but if the SDK was already installed, but the java component was not, the user will still see the same message saying the SDK was installed right?

ivanporty: correct, you mean we might show a slightly different message in this case?

etanshaul: I don't think its a huge deal, but it may be confusing to some users if they already had the SDK installed and they see a message saying it was installed successfully (even though behind the scenes it just installed the java component). 

I'll approve the PR.

ivanporty: I think the inconvenience of this is slightly diminished by the fact that this only applies to SDK we manage ourselves, so the chance of install being completed but Java component missing must be really rare.

ivanporty:
Thanks for the thorough review!

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkService.java
etanshaul: ah good point. didn't consider that. we always install them both right? that makes sense then.

patflynn:
Yeah, I think the quota issue is hitting us again. I have a PR that should
fix it.

PR #2007, opened 2018-04-05
Preselects in the UI the version of the BOM configured in the pom.xml if there is one

### LINKED ISSUE #2002: Pre-populate existing BOM in the UI
If the user already has a BOM configured in the project, the version should be pre-populated in the client library UI.

etanshaul:
Fixes #2002 

Previously, the BOM version dropdown in the client library UI always showed a list of the N latest available versions from Maven Central.

Now, if the user has a preconfigured BOM in the pom.xml, then this version will be pre-selected. If the version of the configured BOM in the pom.xml does not correspond to any of the versions returned from Maven Central, then it will be added to the list.

For multi-module projects, when switching modules in the UI, the list of BOM list will be refreshed accordingly.

Also has a few null checking improvements and enables the BOM feature flag.

**Scenario 1**

Preconfigured BOM version 0.35.0-alpha in the pom.xml:
image

The version is appended (in sorted order) to the versions fetched from Maven Central:
image

**Scenario 2**

Peconfigured BOM version 0.42.0-alpha in the pom.xml:
image

The version already exists in the list of version from Maven Central, so it is selected:
image


etanshaul:
@ivanporty @nbashirbello @patflynn PTAL

note that I think this does not work properly in IDEA 2017.3 due to a bug in the platform. I am investigating a workaround for it.

But you can review in the meantime as I expect the fix to be isolated.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
```diff
@@ -88,11 +88,11 @@ public void projectsScheduled() {
 
   /** Returns the set of {@link CloudLibrary} currently configured on the given {@link Module}. */
   Set<CloudLibrary> getCloudLibraries(Module module) {
-    return moduleLibraryMap.get(module);
+    return moduleLibraryMap.getOrDefault(module, ImmutableSet.of());
```
etanshaul: this are improvements unrelated to this PR. This has the potential of NPE'ing if this method is called before the Maven sync occurs (or if its called with an invalid module).

ivanporty:
> If the version of the configured BOM in the pom.xml does not correspond to any of the versions returned from Maven Central, then it will be added to the list.

What if this version is invalid and does not correspond to any of the cloud libraries versions? What would happen in this case with individual cloud libraries versions?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
```diff
@@ -269,27 +274,47 @@ private void updateManagementUI() {
...
   private void populateBomVersions() {
-    List<Version> bomVersions =
+    List<String> availableBomVersions =
         Lists.newArrayList(CloudApiMavenService.getInstance().getBomVersions());
 
-    if (bomVersions.isEmpty()) {
+    Optional<String> configuredBomVersion =
+        CloudLibraryProjectState.getInstance(project)
+            .getCloudLibraryBom(modulesComboBox.getSelectedModule())
+            .map(MavenDomDependency::getVersion)
```
ivanporty: This sequence would work great with all `null` probabilities and will just return empty optional if something is missing, right?

etanshaul:
> What if this version is invalid and does not correspond to any of the cloud libraries versions? What would happen in this case with individual cloud libraries versions?

Good question. If its an invalid version, then the version of the corresponding libraries simply won't be loaded and the field in the UI will be blank

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/GoogleCloudApiSelectorPanel.java
etanshaul: correct; with the update in this PR to `getCloudLibraryBom` to return a default value, this sequence will either return an optional with a configured BOM version, or an empty optional.

ivanporty: That's neat, great case for mapping Optionals!

ivanporty:
I guess the blank version is fine, but maybe just a bit ambiguous - does it mean this library does not have a version at all or does not have a version for this particular BOM? This could be nitpicking; how about adding something like "no library version for BOM x.x.x" in this case? Could we differentiate between these two cases at all?

etanshaul:
> does it mean this library does not have a version at all or does not have a version for this particular BOM

I think its always the latter right? The library would never not have a version at all, no?

I think its either that 1) there is no version of the library from the configured BOM or 2) some error occurred when fetching the library version (like a network error)

I guess we could differentiate between those two cases.

etanshaul:
On second thought, we can address it here. What do you think of this wording:

1) **Case 1: No version for the given bom**: `No version found for the selected Cloud Libraries version`
2) **Case 2: Some error occurred**: `Error fetching library version`

ivanporty:
Awesome, seems perfectly clear! I'm ok with keeping that as a minor enhancement for a different PR too.

etanshaul:
@ivanporty ptal. I've updated it so that messaging is shown when the version cannot be resolved:

case 1: user has some random BOM version in their pom.xml which doesn't map to a library version:
image

case 2: some error occurs when attempting to fetch the version (such as a network failure):
image

etanshaul:
still looking into a workaround for a bug in 2017.3 (related to synchronizing psi files) causing this to throw exceptions

etanshaul:
just to have it documented here. this is the exception I'm running into in 2017.3:

```
Element: class com.intellij.psi.impl.source.xml.XmlTagImpl because: different providers: SingleRootFileViewProvider{myVirtualFile=file:///usr/local/google/home/eshaul/IdeaProjects/demo-test/secondarymodule/pom.xml, content=VirtualFileContent{size=1442}}(5edc3139); SingleRootFileViewProvider{myVirtualFile=file:///usr/local/google/home/eshaul/IdeaProjects/demo-test/secondarymodule/pom.xml, content=VirtualFileContent{size=1442}}(771988cf)
invalidated at: see attachment
com.intellij.psi.PsiInvalidElementAccessException: Element: class com.intellij.psi.impl.source.xml.XmlTagImpl because: different providers: SingleRootFileViewProvider{myVirtualFile=file:///usr/local/google/home/eshaul/IdeaProjects/demo-test/secondarymodule/pom.xml, content=VirtualFileContent{size=1442}}(5edc3139); SingleRootFileViewProvider{myVirtualFile=file:///usr/local/google/home/eshaul/IdeaProjects/demo-test/secondarymodule/pom.xml, content=VirtualFileContent{size=1442}}(771988cf)
invalidated at: see attachment
	at com.intellij.util.xml.impl.DomInvocationHandler.getFixedChild(DomInvocationHandler.java:623)
	at com.intellij.util.xml.impl.GetFixedChildInvocation.invoke(GetFixedChildInvocation.java:32)
```

etanshaul:
https://youtrack.jetbrains.com/issue/IDEA-189982

etanshaul:
@ivanporty @nbashirbello ptal.

Things updated:

- added error icon when there is a problem fetching the client library version:
image
- fixed the issue in 2017.3. Refactored the CloudLibraryProjectState to not store a reference to the pom.xml's bom DOM since it could get invalidated. Instead, just the BOM version is stored, and the DOM is fetched on-demand as needed

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudLibraryProjectState.java
```diff
@@ -139,9 +139,13 @@ void syncCloudLibrariesBom() {
   }
 
   private Optional<String> loadCloudLibraryBomVersion(Module module) {
-    return loadCloudLibraryBom(module)
-        .map(MavenDomDependency::getVersion)
-        .map(GenericValue::getStringValue);
+    return ApplicationManager.getApplication()
+        .runReadAction(
```
ivanporty: Just out of curiosity and to be sure how to use this myself, what is the general purpose of using `runReadAction` here?

etanshaul:
@patflynn I like your wording suggestion (slight modification suggestion below). I don't think we should remove the library though since they could conceivably just fix it later by manually changing the bom version.

`Library was not found in version .... of the Google Cloud Java libraries`

etanshaul:
updated wording:
image

patflynn:
:+1:

PR #2011, opened 2018-04-09
Make usage tracker manager a service

patflynn:
Implementing an old to-do to improve testability/mockability of usage tracking.

Will help to fix a few breaking tests in my GCS module refactor.

### DIFF
--- a/google-cloud-tools-plugin/google-cloud-core/resources/META-INF/google-cloud-core.xml
+++ b/google-cloud-tools-plugin/google-cloud-core/resources/META-INF/google-cloud-core.xml
@@ -36,6 +36,10 @@
     <applicationConfigurable parentId="google.settings"
                              provider="com.google.cloud.tools.intellij.analytics.UsageTrackerConfigurableProvider"/>
 
+
+    <applicationService
+            serviceInterface="com.google.cloud.tools.intellij.analytics.UsageTrackingManagementService"
+            serviceImplementation="com.google.cloud.tools.intellij.analytics.UsageTrackerManager"/>
     <applicationService
             serviceInterface="com.google.cloud.tools.intellij.analytics.UsageTrackerProvider"
             serviceImplementation="com.google.cloud.tools.intellij.analytics.KeyedExtensionUsageTrackerProvider"/>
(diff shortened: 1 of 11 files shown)


google-cloud-tools-plugin/google-cloud-core/resources/META-INF/google-cloud-core.xml
```diff
@@ -36,6 +36,10 @@
     <applicationConfigurable parentId="google.settings"
                              provider="com.google.cloud.tools.intellij.analytics.UsageTrackerConfigurableProvider"/>
 
+
+    <applicationService
+            serviceInterface="com.google.cloud.tools.intellij.analytics.UsageTrackingManagementService"
```
etanshaul: nit: but just to be more inline with our current conventions, how about `UsageTrackerManager` for the interface and `DefaultUsageTrackerManager` for the impl? Users of this will be calling the getInstance from UsageTrackerManager anyway.

patflynn:
DefaultUsageTrackingManagementService?

etanshaul:
What would the interface be called?

etanshaul:
I mildly prefer UsageTrackerManager and DefaultUsageTrackerManager. But I'm ok either way.

patflynn:
Oh sorry I just saw that you were suggesting to rename the interface. (my
reading ain't so good).

So far we've been calling all IJ services ..Service.  I'm a bit hesitant to
change that.   The thing that I'm trying to avoid is getting people
confused between the purpose of the UsageTracker and the
UsageTrackerManager.

etanshaul:
ah you're right that we've been calling them _Service (not all of them though). I'm cool with *UsageTrackerManagementService but maybe its cleaner if the interface and impl have the same suffix.

etanshaul:
So yeah, what you suggested sounds good:

`UsageTrackingManagementService` / `DefaultUsageTrackingManagementService`

etanshaul:
oops I had the binaries uploaded to GCS but forgot to make the IC zip public. fixed

etanshaul:
looks like the gcs cache issue is fixed. now its just down to spotless warnings

Issue #2000, opened 2018-04-02
Managed SDK automatic update


ivanporty:
Tracks progress toward implementation of managed SDK auto-updates, including manual update, scheduled automatic updates, and UI allowing to cancel pending update.

PR #2012, opened 2018-04-10
Managed SDK update: UI and preparation (behind feature flag)

ivanporty:
Part of #2000.
This PR does not include blocking if SDK is being used for deployments yet, and thus is hidden behind the feature flag.
Contains:
1. Scheduling for updates and its management - update is scheduled once a week.
2. Notification before update with options to cancel the update once or cancel and disable them (can be re-enabled in the settings later) - shown for 20 seconds.
sdk-update-notice

If update is cancelled, notification hides and expires.
update-cancel

3. Settings to keep last time of update or install.
4. Unit tests for all new functionality


etanshaul:
will review today

etanshaul:
ok. starting now :) sorry for the delay

etanshaul:
can we see an updated screenshot of the SDK settings panel please?

etanshaul:
oh actually, it hasn't changed right? was the automatic update checkbox there before?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdater.java
```diff
@@ -0,0 +1,150 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import com.intellij.notification.Notification;
+import com.intellij.openapi.components.ServiceManager;
+import java.awt.event.ActionListener;
+import java.time.Clock;
+import java.util.Timer;
+import java.util.TimerTask;
+
+/** Schedules and manages automatic updates for {@link ManagedCloudSdkService}. */
+public class ManagedCloudSdkUpdater {
```
etanshaul: would the be better as an application service? (and if so update the name accordingly - ManagedCloudSdkUpdateService or something like that)

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdater.java
```diff
@@ -0,0 +1,150 @@
...
+  private static ManagedCloudSdkUpdater instance;
+
+  // one week.
+  @VisibleForTesting static final long SDK_UPDATE_INTERVAL = 1000 * 60 * 60 * 24 * 7;
```
etanshaul: ms right? should include the unit in the variable name e.g. `.._MS`. same for the one below

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -218,7 +218,12 @@ public void apply() throws ConfigurationException {
     }
     sdkServiceUserSettings.setUserSelectedSdkServiceType(selectedCloudSdkServiceType);
 
+    boolean previousAutomaticUpdateEnabled = sdkServiceUserSettings.getEnableAutomaticUpdates();
```
etanshaul: I know it wasn't named in this pr, but getEnableAutomaticUpdates -> isAutomaticUpdateEnabled ?

ivanporty: Agree, that's easier to read.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdater.java
ivanporty: I kind of see application service as something useful for at least some parts of the whole application. This is a pure helper class for Managed SDK, so I wasn't sure about that, adding another service just for testing purposes.. What do you think, does it feel much cleaner if we put such classes into plugin config?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdater.java
```diff
@@ -0,0 +1,150 @@
...
+    CloudSdkService cloudSdkService = CloudSdkService.getInstance();
+    if (cloudSdkService instanceof ManagedCloudSdkService) {
+      // do not show notifications and start update process if SDK is still up-to-date.
+      boolean updateFeatureEnabled =
+          ServiceManager.getService(PluginInfoService.class)
+              .shouldEnable(GctFeature.MANAGED_SDK_UPDATE);
+
+      if (updateFeatureEnabled && !((ManagedCloudSdkService) cloudSdkService).isUpToDate()) {
+        // schedule UI timer to let a user decide about proceeding with the update.
+        javax.swing.Timer uiTimer = createUiTimer(UPDATE_NOTIFICATION_DELAY);
```
etanshaul: don't these notifications disappear automatically after X amount of time? Do we really need to manually control the duration its shown?

ivanporty:
Right, checkbox was already in place in Cloud SDK panel.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdater.java
ivanporty: I looked briefly and I believe I didn't find such use cases - the balloon will collapse into event log message, but the notification itself stays active (not expired). Our idea here is to wait for a short time before an update to give a user some time to consider - I think I haven't see any event of notification getting automatically stale - plus we need to disable actions after this period. I'll double check this though.

etanshaul: so you mean the notification message stays up indefinitely? I was pretty sure it was the opposite: it automatically hides after a period of time. Could be wrong though.

ivanporty: Balloon will definitely collapse, but notification will be in the log with active actions I believe. Checking this.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdater.java
ivanporty: I don't have a very strong opinion on this, plus it seems the rest of the code is relying a lot on service manager, I will update it as a service so it will be in line.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdater.java
etanshaul: oh. so your goal here is to expire the actions in the log then? I misunderstood.

what happens after the timer sets off?

ivanporty: Notification expires (which includes disabling actions since update already starts) and update is run.

ivanporty:
For notification, it seems it won't go away if IDE window is focused, it looks like it collapses into event log when it loses focus, but actions need to "expire" after short period once we start the actual update. Otherwise they stay active.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkServiceUserSettings.java
```diff
@@ -91,6 +96,22 @@ void setEnableAutomaticUpdates(boolean enableAutomaticUpdates) {
         DEFAULT_MANAGED_SDK_AUTOMATIC_UPDATES /* need to specify default to avoid removal */);
   }
 
+  long getLastAutomaticUpdateTimestamp() {
```
etanshaul: would it make sense to return an Optional boxed Long here instead to indicate that there is no previously stored last update?

ivanporty: I agree, good use case for Optional, let's do that

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -218,7 +218,12 @@ public void apply() throws ConfigurationException {
     }
     sdkServiceUserSettings.setUserSelectedSdkServiceType(selectedCloudSdkServiceType);
 
+    boolean previousAutomaticUpdateEnabled = sdkServiceUserSettings.isAutomaticUpdateEnabled();
     sdkServiceUserSettings.setEnableAutomaticUpdates(enableAutomaticUpdatesCheckbox.isSelected());
+    if (enableAutomaticUpdatesCheckbox.isSelected() && !previousAutomaticUpdateEnabled) {
+      // activate updates again.
+      ManagedCloudSdkUpdateService.getInstance().activate();
```
etanshaul: I'm trying to think of a scenario where this mechanism could be broken, potentially causing the update to be sceduled multiple times. For example, if the user has multiple projects opened and the user clicks to update on one of them, and then does the same on the other project (before it expires). Have you considered this scenarios to ensure everything happens atomically?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdateService.java
```diff
@@ -0,0 +1,140 @@
...
+import java.time.Clock;
+import java.util.Timer;
+import java.util.TimerTask;
+
+/** Schedules and manages automatic updates for {@link ManagedCloudSdkService}. */
+public class ManagedCloudSdkUpdateService {
+  // one week.
+  @VisibleForTesting static final long SDK_UPDATE_INTERVAL_MS = 1000 * 60 * 60 * 24 * 7;
+  // 20 seconds to show notification and then proceed with update.
+  private static final int UPDATE_NOTIFICATION_DELAY_MS = 1000 * 20;
```
etanshaul: this variable represents how long the notification remains active after it is shown right?

perhaps, UPDATE_NOTIFICATION_TTL_MS? or if you don't like the use of TTL here, then "EXPIRATION_TIME" or something like that?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdateService.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    sdkUpdateTask =
+        new TimerTask() {
+          @Override
+          public void run() {
+            doUpdate();
+          }
+        };
+    long lastTimeOfUpdate =
```
etanshaul: perhaps extract all of the below (up to line 75) into a well named method that indicates what all the calculations do for readability? Also with a javadoc

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdateService.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  @VisibleForTesting
+  javax.swing.Timer createUiTimer(int delayMillis) {
+    return new javax.swing.Timer(delayMillis, null);
+  }
+
+  private void doUpdate() {
+    CloudSdkService cloudSdkService = CloudSdkService.getInstance();
+    if (cloudSdkService instanceof ManagedCloudSdkService) {
```
etanshaul: do you need the `instanceof` here? can' t you do `CloudSdkService#isInstallSupported()` instead?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdateService.java
```diff
@@ -0,0 +1,140 @@
...
+              CloudSdkServiceUserSettings.getInstance().setEnableAutomaticUpdates(false);
+            };
+        Notification updateNotification =
+            ManagedCloudSdkServiceUiPresenter.getInstance()
+                .notifyManagedSdkUpdate(cancelListener, disableUpdateListener);
+
+        uiTimer.addActionListener(
+            e -> {
+              updateNotification.expire();
+              // grab a current SDK service, might have changed while waiting for notification.
```
etanshaul: interesting. so this is to guard against the user quickly changing the SDK to be custom (instead of managed) in the interval before the notification expired? In that case, it will just skip the update right?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdateService.java
etanshaul: oh I see, you are casting it below. I guess you add an #isUpToDate() method to the interface too, but I guess thats not semantically correct (since you have to return `true` or `false` always for the custom case)..

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdateService.java
ivanporty: Correct, that's the case where we already scheduled to be updated and then a user changes the SDK to custom, then we skip and we it stays custom, next scheduled run will cancel the update task completely.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdateService.java
ivanporty: Exactly, adding too many methods would make custom SDK implementation look bloated and return strange values. Update always work with Managed SDK so I guess this makes it OK to be strongly coupled.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdateService.java
ivanporty: Both sounds clear enough, if you don't object, I'll use `...EXPIRATION_TIME...`.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
ivanporty: Confirmed - notifications are synced in multiple project windows and only active one shows balloon. Once action is expired, it expires in all windows. `ManagedSDKService` itself also forbids simultaneous running of update/install jobs.

google-cloud-tools-plugin/resources/messages/CloudToolsBundle.properties
```diff
@@ -454,6 +454,9 @@ managedsdk.progress.message=Managed Google Cloud SDK: {0}.
 managedsdk.progress.install.app.engine=Installing App Engine Java
 managedsdk.progress.update=Updating Google Cloud SDK
 managedsdk.waiting.for.sdk.ready=Waiting for Google Cloud SDK installation to complete...
+managedsdk.update.notification=New version of Managed Google Cloud SDK is available and about to be updated.
+managedsdk.update.notification.cancel=Cancel this update
```
etanshaul: is it "Cancel" or "Skip"?

ivanporty: Skip probably describes it better, let me rename.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdateService.java
```diff
@@ -97,6 +88,24 @@ void schedule(TimerTask timerTask, long delay, long period) {
...
 
+  /**
+   * Calculates milliseconds delay before first update based on last update/install time. If
+   * complete update interval or more elapsed, first update is scheduled immediately (returns 0).
+   */
+  private long getDelayBeforeFirstUpdate() {
+    Optional<Long> lastTimeOfUpdate =
+        CloudSdkServiceUserSettings.getInstance().getLastAutomaticUpdateTimestamp();
+    long delayBeforeUpdateMillis = SDK_UPDATE_INTERVAL_MS;
+    if (lastTimeOfUpdate.isPresent()) {
```
etanshaul: nit: 

```
lastTimeOfUpdate.ifPresent( .. -> ..);
```

google-cloud-tools-plugin/google-cloud-core/resources/config.properties
```diff
@@ -17,6 +17,7 @@
 # Place cloud tools plugin configuration here.
 feature.appengine.flex=true
 feature.managed.sdk=true
+feature.managed.sdk.update=false
```
etanshaul: just to refresh my memory:

we can't release the managed SDK feature until the update is fully complete right? I guess if we wanted to, we could hide the automatic update UI element behind this flag too? No need to do that now though, just wondering.

ivanporty: Sure, we can hide the UI as well - but the final step (mutual install / use blocking) is ready, and we can review it right after we're done with this PR, so we should be good.

etanshaul: even better!

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/ManagedCloudSdkUpdateService.java
ivanporty: Actually let's leave it as is - the thing is we need to assign to outer variable or return value from this block, and we won't be able to do that from lambda, so it might be easier to read as it is now.

PR #2016, opened 2018-04-10
Deflake tests

patflynn:
The flake was in ProjectSelectionDialogTest.  @Spy initializes before fixtures are reinitialized and throws errors due to use of disposed elements from previous runs.


google-cloud-tools-plugin/google-cloud-core/testSrc/com/google/cloud/tools/intellij/project/ProjectSelectionDialogTest.java
```diff
@@ -79,7 +79,6 @@ public void setUp() {
     doNothing().when(projectSelectionDialog).installTableSpeedSearch(any());
     doNothing().when(projectSelectionDialog).setLoading(anyBoolean());
 
-    projectSelectionDialog.createUIComponents();
```
ivanporty: Oh this one will get called by IDEA form instrumentation?

patflynn: right. the default constructor (automatically invoked due to @Spy annotation) includes this in its form generated code.

PR #2023, opened 2018-04-12
Extension of the Service Account Key dialog to allow auto updating on the run configuration

nbashirbello:
Extension of the Service Account Key Created dialog to allow automatic updating of selected Google App Engine Standard run configurations
- This feature is only available in Ultimate Edition because the Google App Engine Standard run configurations are only available there
- The PR adds a table in the Service Account Key Created dialog that shows the existing Google App Engine run configurations, allowing users to select which one they would like to update to add the Google Cloud Library environment variables: GOOGLE_CLOUD_PROJECT and GOOGLE_APPLICATION_CREDENTIALS
- If the selected run configuration already has the GOOGLE_CLOUD_PROJECT and GOOGLE_APPLICATION_CREDENTIALS environment variables, they would be updated
- When there are no Google App Engine Standard run configurations, this table will be hidden and look like the dialog in the Community Edition


nbashirbello:
Service Account Key Created dialog in IC and when no App Engine Standard run configurations exist:

screen shot 2018-04-12 at 9 40 45 am

Service Account Key Created dialog in IU and when there are App Engine Standard run configurations:


screen shot 2018-04-12 at 9 40 08 am

nbashirbello:
PR is not complete. Need to add tests and address the following:
- Need to figure out why the fix for the nesting forms from different modules isn't working
- Not printing the name of the configuration, not sure why this is happening. I am using a renderer that calls getName but the ouput looks like something is calling toString()

nbashirbello:
Updated UI - with fix for name of run configuration

screen shot 2018-04-12 at 1 46 08 pm

etanshaul:
Is this ready now for review?

nbashirbello:
Yes, please review. Thanks

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyDialogService.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.ui.DialogWrapper;
+import org.jetbrains.annotations.Nullable;
+
+/** Application service that provides the handle to the IC's Service Account Key Created dialog */
+public class ServiceAccountKeyDialogService {
+  public DialogWrapper getDialog(
+      @Nullable Project project, String gcpProjectId, String downloadPath) {
+    return new ServiceAccountKeyDisplayDialog(project, gcpProjectId, downloadPath);
```
ivanporty: To understand a bit better, this service to create the dialog is mostly for different implementations in IC vs IU, correct?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudApiManager.java
```diff
@@ -430,7 +431,8 @@ private static void notifyApiEnableSkipped(Set<CloudLibrary> apis, Project proje
     notification.notify(project);
   }
 
-  private static void notifyServiceAccountCreated(Project project, String name, Path downloadDir) {
+  private static void notifyServiceAccountCreated(
+      Project project, String gcpProjectId, String name, Path downloadDir) {
```
ivanporty: Could we possibly pass `CloudProject` into both this method and `ServiceAccountKeyDisplayDialog` to make it a bit more strong typed and easier to see that we get project ID from it?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyDownloadedPanel.java
```diff
@@ -0,0 +1,87 @@
...
+    envVarTable.setModel(tableModel);
+    envVarTable.setRowSelectionAllowed(false);
+    envVarTable.setTableHeader(null);
+
+    copyToClipboardButton.addActionListener(
+        new CopyToClipboardActionListener(credentialEnvVar + "\n" + cloudProjectEnvVar));
+  }
+
+  @NotNull
+  public Set<EnvironmentVariable> getEnvironmentVariables() {
```
ivanporty: Looks like this method is never used?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/ui/BooleanTableModel.java
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * The {@link TableModel} for a 2 column headerless table where the first column is immutable and
+ * the second column in a boolean.
+ *
+ * @param <T> the type of the item in the first column.
+ */
+public class BooleanTableModel<T> implements TableModel {
+
+  private final SortedMap<T, Boolean> map;
+  private final List<TableModelListener> listeners = new ArrayList<>();
```
ivanporty: This could be a bit easier to implement using `AbstractTableModel` - it contains support for listeners and set of `fire...` methods to fire table events.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/ui/BooleanTableModel.java
```diff
@@ -0,0 +1,129 @@
...
+  public void setValueAt(Object value, int rowIndex, int columnIndex) {
+    if (columnIndex == T_COL) {
+      throw new UnsupportedOperationException("The first column is immutable.");
+    }
+
+    T key = (T) map.keySet().toArray()[rowIndex];
+    map.put(key, (Boolean) value);
+
+    TableModelEvent event = new TableModelEvent(this, rowIndex);
+    listeners.forEach(listener -> listener.tableChanged(event));
```
ivanporty: Please see comment about `AbstractTableModel` above

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/ui/BooleanTableModel.java
```diff
@@ -0,0 +1,129 @@
...
+ * The {@link TableModel} for a 2 column headerless table where the first column is immutable and
+ * the second column in a boolean.
+ *
+ * @param <T> the type of the item in the first column.
+ */
+public class BooleanTableModel<T> implements TableModel {
+
+  private final SortedMap<T, Boolean> map;
+  private final List<TableModelListener> listeners = new ArrayList<>();
+  private final int T_COL = 0;
```
ivanporty: Maybe slightly more extended name would be easier to read - i.e. `VALUE_COL` or something similar?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
```diff
@@ -0,0 +1,303 @@
...
+      @Nullable Project project, @NotNull String gcpProjectId, @NotNull String downloadPath) {
+    super(project);
+    this.project = project;
+    this.gcpProjectId = gcpProjectId;
+    this.downloadPath = downloadPath;
+    init();
+    setTitle(GctBundle.message("cloud.apis.service.account.key.downloaded.title"));
+    table.setTableHeader(null);
+
+    if (tableModel.getRowCount() == 0) {
```
ivanporty: This looks slightly mysterious, would it be more clear if we just call `getAppEngineStandardConfigurationSettingsList` directly to check we don't have any configs available?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
```diff
@@ -0,0 +1,303 @@
...
+
+  private final Project project;
+  private final String gcpProjectId;
+  private final String downloadPath;
+  private JPanel mainPanel;
+  private ServiceAccountKeyDownloadedPanel commonPanel;
+  private JTable table;
+  private JLabel label;
+  private JScrollPane scrollPane;
+  private BooleanTableModel tableModel;
```
ivanporty: Possibly should be `BooleanTableModel<RunnerAndConfigurationSettings>`? It looks like it accepts type of its values?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
```diff
@@ -0,0 +1,303 @@
...
+  public ServiceAccountKeyUltimateDisplayDialog(
+      @Nullable Project project,
+      @NotNull String gcpProjectId,
+      @NotNull String downloadPath,
+      @NotNull List<RunnerAndConfigurationSettings> configurationSettingsList) {
+    super(project);
+    this.project = project;
+    this.gcpProjectId = gcpProjectId;
+    this.downloadPath = downloadPath;
+    this.configurationSettingsList = configurationSettingsList;
```
ivanporty: Since this is static field, could we use `ServiceAccountKeyUltimateDisplayDialog.configurationSettingsList`?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
ivanporty: Also name seems to be a bit generic, maybe rename to include this is a model with run configurations?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
```diff
@@ -0,0 +1,303 @@
...
+  private static final class ServerTable extends JBTable {
+
+    ServerTable(BooleanTableModel tableModel) {
+      super(tableModel);
+      setSelectionMode(ListSelectionModel.SINGLE_SELECTION);
+      setDefaultRenderer(
+          RunnerAndConfigurationSettings.class, new RunnerAndConfigurationSettingsRenderer());
+      setDefaultRenderer(Boolean.class, new BooleanTableCellRenderer());
+      setDefaultEditor(Boolean.class, new BooleanTableCellEditor());
+      TableUtil.setupCheckboxColumn(getColumnModel().getColumn(tableModel.getBooleanColumn()));
```
ivanporty: This looks like a deprecated method, they suggest using `setupCheckboxColumn(JTable, colndex)` instead

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    @Override
+    protected void doAction(ActionEvent event) {
+      addEnvironmentVariablesToConfiguration(getSelectedConfigurations());
+    }
+  }
+
+  /** The custom {@link JBTable} for the table of existing Google App Engine run configurations. */
+  private static final class ServerTable extends JBTable {
```
ivanporty: Could we expand this name a bit to see what servers are in this table easier?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
```diff
@@ -0,0 +1,303 @@
...
+  private static final Logger LOG =
+      Logger.getInstance(ServiceAccountKeyUltimateDisplayDialog.class);
+
+  private final Project project;
+  private final String gcpProjectId;
+  private final String downloadPath;
+  private JPanel mainPanel;
+  private ServiceAccountKeyDownloadedPanel commonPanel;
+  private JTable table;
+  private JLabel label;
```
ivanporty: Probably makes sense to use a bit more specific names for `table` and `label`

etanshaul: +1. I think its best practice to avoid undescriptive names like this in general

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
```diff
@@ -0,0 +1,303 @@
...
+
+  /**
+   * Adds the environment variables for the Google Cloud Libraries to the list of environment
+   * variables for {@code configuration} if they don't exist. If they exist, it replaces them.
+   */
+  private void addEnvironmentVariablesToConfiguration(
+      String executorId, RunnerAndConfigurationSettings configuration) {
+    ProgramRunner runner = ProgramRunnerUtil.getRunner(executorId, configuration);
+    if (runner == null) {
+      LOG.error(
```
ivanporty: Logging errors bring up pretty harsh plugin level error window, could this be a bit too much here? How about showing this as a error in this dialog itself (I think `DialogWrapper` supports error messages) so a user clearly sees what happened and log WARN message instead?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
```diff
@@ -0,0 +1,303 @@
...
+        configurationSettings.getEnvVariables();
+
+    serviceAccountEnvironmentVariables.forEach(
+        serviceAccountEnvVar -> {
+          OptionalInt indexOpt =
+              IntStream.range(0, configurationSettingsEnvVariables.size())
+                  .filter(
+                      index ->
+                          serviceAccountEnvVar
+                              .getName()
```
ivanporty: Do we need to compare them by name here, does `EnvironmentVariable` have any sensible `equals()` implementation? If it does, we may be able to simplify this code a bit.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyDialogService.java
nbashirbello: Yes

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
nbashirbello: EnvironmentVariable doesn't have equals()

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
nbashirbello: Do you mean rename the JBTable to something like RunConfigurationTable?

ivanporty: Exactly

google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -424,6 +424,8 @@
     <applicationService serviceImplementation="com.google.cloud.tools.intellij.csr.CloudRepositoryService"/>
     <applicationService serviceImplementation="com.google.cloud.tools.intellij.apis.CloudLibrariesService"/>
     <applicationService serviceImplementation="com.google.cloud.tools.intellij.apis.CloudApiMavenService"/>
+    <applicationService serviceInterface="com.google.cloud.tools.intellij.apis.ServiceAccountKeyDialogService"
```
etanshaul: I a bit confused by this - the serviceInterface and serviceImplementation are pointing to the same class?

etanshaul:
@nbashirbello I noticed an issue where when I rename the local run configurations that updating the env vars stops working. Testing again to see if its consistent.

google-cloud-tools-plugin/resources/META-INF/plugin.xml
etanshaul: in other words - why not just have a serviceImplementation only in the xml

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/ui/BooleanTableModel.java
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * The {@link TableModel} for a 2 column headerless table where the first column is immutable and
+ * the second column in a boolean.
+ *
+ * @param <T> the type of the item in the first column.
+ */
+public class BooleanTableModel<T> extends AbstractTableModel {
+
+  private final SortedMap<T, Boolean> map;
+  private final List<TableModelListener> listeners = new ArrayList<>();
```
ivanporty: Looks like this one is not used anymore, could we remove it?

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * Dialog visible only in Ultimate Edition that (1) confirms the download of the service account
+ * JSON key for the Google Cloud Libraries , (2) provides information on how to set the environment
+ * variables for local run and (3) allows the user to select App Engine Standard run configurations
+ * to automatically update with these environment variables.
+ */
+public class ServiceAccountKeyUltimateDisplayDialog extends DialogWrapper {
+
+  private static final Logger LOG =
```
ivanporty: This shows as unused , is it used anywhere?

google-cloud-tools-plugin/resources/META-INF/plugin.xml
nbashirbello: Oh didn't know I could do that. Will do.

nbashirbello:
@etanshaul I couldn't reproduce the bug. Any chance reproducing it?

etanshaul:
@nbashirbello was trying to reproduce but for some reason now the env vars are never added regardless of what I do. Trying to figure out whats going on.

ivanporty:
Also, another thing - many Cloud Libraries quick starts and examples have nothing to do with App Engine at all. Adding these two environment variables into any generic Java run configuration (i.e. plain main() method) does the magic and the quick start works like a charm without all the additional manual steps. Should we show the complete list of run configurations to add environment variables to, not just App Engine ones? This could be very useful for people using Cloud Libraries.

etanshaul:
@ivanporty thats an interesting question. We originally discussed just adding them to App Engine ones since we presumably wouldn't know anything about these other run configurations (Jetty?, Tomcat?, something else entirely?). Also it may be weird if run configurations entirely unrelated to local run show up there. wdyt? I'm also ok addressing this issue in a subsequent PR.

ivanporty:
Yep, I agree some run configurations may be completely non-relevant, but we could probably include at least plain Java application with `main()` in order to make testing and quick starts for all Cloud Libraries super easy. Definitely let's discuss this and address in a separate PR. I just tried a couple of Cloud APIs using our tools and manually adding variables - very nice experience!

patflynn:
+1 to adding this to arbitrary run configs.  You can always filter them by
configs we know make sense. e.g. spring boot, java app, app servers.

nbashirbello:
@etanshaul @ivanporty In regards to the env vars not getting added to the selected run configurations, wanted to confirm that you are clicking the "Apply" button. Are you?

ivanporty:
Oh I was just hitting "OK" at the end of the sequence, didn't do "Apply". It works with "Apply", could we also add the variables if checkbox in the table is on and we click "OK" - seems very natural?

etanshaul:
are you suggesting that it only works with "apply" and not "ok"?

I think I've just been hitting "OK" from that dialog. You shouldnt need to explicitly hit "Apply" since OK includes that (plus closing the dialog).

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
```diff
@@ -0,0 +1,269 @@
...
+
+  @Nullable
+  @Override
+  protected JComponent createCenterPanel() {
+    return mainPanel;
+  }
+
+  @NotNull
+  @Override
+  protected Action[] createActions() {
```
etanshaul: I don't think we need to override this. Can't we just use the standard actions of DialogWrapper? In the same way the `CloudApiManagementConfirmationDialog` is doing it, for example.

etanshaul: I see your point. But at least for the Ultimate version, a 'cancel' button might be nice (which would mean intuitively that they don't want to add any env vars instead of having to uncheck them one by one). Also, I think we can get rid of the "apply" altogether.

nbashirbello:
@etanshaul To verify, the dialog would look like below, and the OK button would updated the selected run configurations and close the dialog while the Cancel button would only close the dialog. And when there are no App Engine run configurations or the user is in IC, the Cancel button would be hidden.

screen shot 2018-04-20 at 2 24 24 pm

nbashirbello:
I think the Cancel button might be a bit confusing because this dialog comes up when the user is trying to add a client library and they might wonder if Cancel means cancel the Client Library addition. But once the user click on OK or Cancel, there will see the message that X client library has been added so maybe this is ok...

screen shot 2018-04-20 at 2 32 22 pm

What about using the names "Update" and "Close" instead of "Ok" and "Cancel" respectively. I am fine leaving the buttons as OK/Cancel if no one else thinks it might be confusing :)

ivanporty:
It slightly feels like `Cancel` might undo something related to service accounts although it won't. How about we name the buttons a bit more precise, like "Add Variables" and "Close"?

etanshaul:
+1 to those suggestions

nbashirbello:
Updated the dialog:

screen shot 2018-04-23 at 10 43 32 am

etanshaul:
nice! will test and review again

ivanporty:
Looks great!
UX nit: could we possibly disable "Add Variables" when no run configurations are selected in the table (i.e. all are unchecked)? Otherwise a user could have an impression variables are going to be added anyway since the action is named like this and it's active.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
```diff
@@ -93,6 +96,17 @@ public ServiceAccountKeyUltimateDisplayDialog(
       scrollPane.setVisible(false);
       runConfigurationTable.setVisible(false);
     }
+
+    runConfigurationTableModel.addTableModelListener(
+        new TableModelListener() {
```
etanshaul: lambda

etanshaul: also, looks like visibility of the constructor of this class can be package private

etanshaul:
lgtm after the last couple of small comments. Will defer to @ivanporty for the final approval.

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  /**
+   * The custom {@link javax.swing.table.TableCellRenderer TableCellRenderer} for {@link
+   * RunnerAndConfigurationSettings} objects.
+   */
+  private static final class RunnerAndConfigurationSettingsRenderer
+      extends DefaultTableCellRenderer {
+
+    private static final Border NO_FOCUS_BORDER = new EmptyBorder(5, 5, 5, 5);
```
ivanporty: IDEA inspection suggests to replace this with ` = Borders.empty(5);`

google-cloud-tools-plugin/ultimate/src/com/google/cloud/tools/intellij/apis/ServiceAccountKeyUltimateDisplayDialog.java
```diff
@@ -0,0 +1,293 @@
...
+  private class AddVariablesAction extends DialogWrapperAction {
+    private AddVariablesAction() {
+      super(GctBundle.message("cloud.apis.service.account.key.downloaded.update.server.button"));
+      putValue(DEFAULT_ACTION, Boolean.TRUE);
+    }
+
+    @Override
+    protected void doAction(ActionEvent event) {
+      addEnvironmentVariablesToConfiguration(getSelectedConfigurations());
+      if (hasErrors(mainPanel)) {
```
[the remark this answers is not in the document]
nbashirbello: We are checking for errors that might have occurred when  adding the environment variables to the run configuration.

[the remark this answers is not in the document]
ivanporty: Looks great!

Issue #1968, opened 2018-03-23
Add extended error message into Cloud SDK deployment error notifications.


ivanporty:
New notifications on deployment / local run tell user about SDK issues only briefly. Include more details including details of managed SDK installation failure if it happened.

PR #2025, opened 2018-04-12
Managed SDK read/write operations blocking

### LINKED ISSUE #2000: Managed SDK automatic update

ivanporty:
Closes #2000.

Adds mutual blocking between SDK read (running commands) and write (install and update) operations.
SDK will not install and update until all running SDK commands are completed. Reverse blocking is already implemented - deployments and local run will not start until SDK is not ready.

Also adds additional progress bar notification when SDK write job is waiting for its running processes to complete - this case can happen when:
a. update is scheduled and SDK commands are still running
b. SDK commands are running, then SDK becomes invalid or deleted and another deploy / run is requested, leading to new installation to fix invalid state.

In these cases users will see the following:

wait-for-sdk-process-finish

If user cancels, the whole deployment / run process cancels and notification that SDK is not available is shown (wording subject to improvements in #1968)

I'm still thinking how to do the unit test right without mindlessly checking for lock/unlock sequences and dangerous thread waits, but we can review the solution itself so far.


ivanporty:
Added test cases as well for blocking write operations - they should be straightforward although look a bit cumbersome due to multiple threads, I'm open to any suggestions for them.

etanshaul:
if we can't have a specific waiting message for local run vs deployment, how about just saying:

> ..waiting for App Engine task to finish

ivanporty:
Sounds good, great generic message, plus it's easier to read in space-constrained progress bar window.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/executor/AppEngineExecutor.java
```diff
@@ -42,8 +45,23 @@ public void cancel() {
     }
   }
 
+  @SuppressWarnings("FutureReturnValueIgnored")
   private void setProcess(Process process) {
     this.process = process;
+    ThreadUtil.getInstance()
```
etanshaul: feels like this shouldn't live in a `setProcess` method which is just a simple setter. can this be split out and called alongside `setProcess` from the `run` method?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/executor/AppEngineExecutor.java
```diff
@@ -16,6 +16,9 @@
 
 package com.google.cloud.tools.intellij.appengine.cloud.executor;
 
+import com.google.cloud.tools.intellij.appengine.sdk.CloudSdkServiceManager;
+import com.google.cloud.tools.intellij.util.ThreadUtil;
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.vcs.impl.CancellableRunnable;
 
 /** Executor of {@link AppEngineTask}'s. */
```
etanshaul: nit: add comments here explaining that it blocks for Cloud SDK.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/executor/AppEngineExecutor.java
```diff
@@ -42,11 +59,25 @@ public void cancel() {
...
-  }
-
-  public Process getProcess() {
-    return process;
+  /**
+   * Waits for the given process to finish, in a separate background thread, holding {@link
+   * CloudSdkServiceManager#getSdkReadLock()} lock.
+   */
+  @SuppressWarnings("FutureReturnValueIgnored")
+  private void waitForProcessFinish(Process process) {
```
etanshaul: this blocks the cloud SDK from running until this process is complete correct? Perhaps better naming to indicate this?

etanshaul: like: `acquireCloudSdkReadLock`, or something along those lines?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/executor/AppEngineExecutor.java
```diff
@@ -30,7 +39,15 @@ public AppEngineExecutor(AppEngineTask task) {
 
   @Override
   public void run() {
-    task.execute(this::setProcess);
+    task.execute(
+        process -> {
+          AppEngineExecutor.this.process = process;
```
etanshaul: isn't `this.process` sufficient?

ivanporty: it is, probably leftover from anonymous class.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/executor/AppEngineExecutor.java
ivanporty: sounds good, maybe add something else to imply we wait for a process there as well? I.e. `acquireCloudSdkReadLockUntilFinished`?

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/executor/AppEngineExecutor.java
```diff
@@ -30,7 +39,15 @@ public AppEngineExecutor(AppEngineTask task) {
 
   @Override
   public void run() {
-    task.execute(this::setProcess);
+    task.execute(
```
etanshaul: just noticed that `ProcessStartListener#onStart` is called _after_ the process has been started. Technically, what we want here is to acquire the read lock _beforeStart_? I don't think practically this makes much of a difference though.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/executor/AppEngineExecutor.java
etanshaul: good point. how about `holdCloudSdkReadLock`? that should cover the whole concept I think

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/executor/AppEngineExecutor.java
etanshaul: what do you mean by "problematic considering current SDK API call sequence"?

etanshaul: we _could_ update ProcessStartListener to contain an additional beforeStart method. This clearly would break existing clients though, but not a huge deal to update our 4 plugins. I agree though that it would be very hard to violate this though (and as long as the consequences of such a violation are well presented to the user, I'm ok with it as is and will leave it up to you).

etanshaul:
approved. just a couple minor comments from before to address then feel free to merge

PR #2035, opened 2018-04-13
Implements manual update action in SDK settings

### LINKED ISSUE #1965: Implement support for "Check for updates" in settings.
Part of #2000.

ivanporty:
Closes #1965.

Link with action to update the SDK manually will only be visible when SDK is in healthy condition and an update is actually available, otherwise it's hidden since requesting update in this case won't do anything.

UI: No update available / SDK not yet ready:
update-not-available

UI: Update is available. Clicking will disable the button and start update in background with process indicator.

update-now
update-now-clicked


google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -347,7 +359,19 @@ protected void textChanged(DocumentEvent event) {
   private void setManagedSdkUiAvailable(boolean available) {
     if (ServiceManager.getService(PluginInfoService.class).shouldEnable(GctFeature.MANAGED_SDK)) {
       enableAutomaticUpdatesCheckbox.setEnabled(available);
-      checkForUpdatesHyperlink.setVisible(available);
+      // only make it visible if managed SDK is active, not currently installing or updating, and
+      // not
+      // up-to-date.
+      CloudSdkService cloudSdkService = CloudSdkService.getInstance();
+      if (cloudSdkService instanceof ManagedCloudSdkService
```
patflynn: if we need to check this maybe we should push it into the CloudSdkService interface?

ivanporty: That's a good discussion question, we've been thinking about it a bit and so far decided CloudSdkService shouldn't grow to include update() and isUpToDate() methods since they are meaningless for custom SDK and we don't foresee any future uses at all. Extending to another interface (like UpdateableCloudSdkService) would still have the need for casting. Only places where we do that are those where we are explicitly strongly coupled to ManagedCloudSdkService and won't expect another service (like here in settings.)

patflynn: I was actually thinking something way less thoughtful. :)

e.g. `CloudSdkService.isManaged()`

ivanporty: But in general you are definitely correct that simple "path" based custom SDK and managed SDK are two different beasts hence couple of places where only can work with managed one. Since the changes introduced a lot of refactoring, we tried to keep existing functionality based on previous interface, but once we're done with managed SDK and things go well we may try to split interface into two parts for further trasparency.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -322,7 +328,13 @@ private void initEvents() {
         new HyperlinkAdapter() {
           @Override
           protected void hyperlinkActivated(HyperlinkEvent e) {
-            // TODO(ivanporty) will call ManagedSdk#update if it's installed and active.
+            CloudSdkService cloudSdkService = CloudSdkService.getInstance();
+            if (cloudSdkService instanceof ManagedCloudSdkService) {
+              ((ManagedCloudSdkService) cloudSdkService).update();
+              // do update call once and disable for visual feedback,
+              // since the following calls will essentially do nothing until update is complete.
+              checkForUpdatesHyperlink.setVisible(false);
```
etanshaul: does this disable it or hide it? I'd expect it just to be disabled when clicking update.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -347,7 +359,21 @@ protected void textChanged(DocumentEvent event) {
   private void setManagedSdkUiAvailable(boolean available) {
```
etanshaul: is this method only dealing with setting the "update" related portions of the UI? if so maybe the name should be updated to reflect this.

etanshaul: if you could show a screenshot in the PR that would help me visualize it

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
ivanporty: This is IDEA's custom component and unfortunately it completely ignores `enabled` - it's always enabled unless you hide it.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
ivanporty: correct, for Managed SDK it's update, for custom SDK it's SDK path, those two methods are generally used to switch between active UI for managed/custom SDK, not for specific update or path portion though. Working on the screenshot.

ivanporty: Updated description with screenshots.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
etanshaul: I see. I think its a bit weird for a control to disappear after pressed. Could we consider using a different control that is disable'able?

ivanporty: That's the one that IDEA uses for hyperlinks, the other one would be JXHyperlink, but it could look out of place. Let me check though. Another option is to just leave it and allow users to click it as many times as they wish - this won't do any harm to already running update but some visual feedback would be better.

etanshaul: what about a button?

you point about clicking as many times as they want is good too. If its already running, you could show some sort of disappearing notice that tells them that its running already.

etanshaul: it actually doesn't have to be a disappearing notice saying its running.. But if we allow them to click as many times as they want, we just need some sort of indication that the click was processed.

etanshaul: sure. If you disagree that this is even an issue, we can wait until we have the bugout to decide after some hands-on time.

ivanporty: I agree, I planned to just disable the hyperlink but it's really limited and hiding something on click is usually not expected.

ivanporty: Updated to a button, PTAL.

etanshaul: looks great

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/CloudSdkPanel.java
```diff
@@ -347,7 +348,21 @@ protected void textChanged(DocumentEvent event) {
...
-      checkForUpdatesHyperlink.setVisible(available);
+
+      // only make it visible if managed SDK is active, not currently installing or updating, and
+      // not up-to-date.
+      if (available
+          && CloudSdkServiceUserSettings.getInstance().getUserSelectedSdkServiceType()
+              == CloudSdkServiceType.MANAGED_SDK) {
+        ManagedCloudSdkService managedCloudSdkService =
+            (ManagedCloudSdkService) CloudSdkService.getInstance();
+        if (managedCloudSdkService.getStatus() == SdkStatus.READY
```
etanshaul: what happens if this if statement is false? is it correct not to set the visibility of the button?

ivanporty: It assumes SDK either not ready or already up to date and doesn't show update button since no update is possible. Do you mean we still want to show it?

etanshaul: no saying we should or shouldn't show it, just pointing out that in this branch (where its false) you are neither setting the visibly to true or false.

PR #2038, opened 2018-04-18
BOM inspection for cloud libraries with no BOM imported

### LINKED ISSUE #1969: Add inspection to detect cloud library dependency with no BOM
part of #1921 

If the user configures a google-cloud-java client library in their pom.xml but is not using a BOM, a code inspection should detect this and offer to fix this (by adding in the bom)

etanshaul:
fixes #1969 

- Contains a quickfix that will add the latest BOM version. Will also strip out the version from the dependency definition if its defined.

Still having the same difficulty with setting up unit testing for these inspections. Will revisit this issue, but please review.


etanshaul:
Inspection warning:
image

Inspection details:
image

Inspection entry:
image

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudBomInspection.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Set;
+import org.jetbrains.annotations.Nullable;
+import org.jetbrains.idea.maven.dom.model.MavenDomDependency;
+import org.jetbrains.idea.maven.dom.model.MavenDomProjectModel;
+
+/**
+ * Base {@link DomElementsInspection} for inspecting google-cloud-java BOM related issues in the
+ * project's pom.xml.
+ */
+class CloudBomInspection extends DomElementsInspection<MavenDomProjectModel> {
```
ivanporty: Should we make it `abstract` to note this is not real implementation?

etanshaul: yes good idea

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudBomInspection.java
```diff
@@ -0,0 +1,80 @@
...
+
+  private static final Logger logger = Logger.getInstance(CloudBomInspection.class);
+
+  public CloudBomInspection() {
```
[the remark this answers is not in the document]
etanshaul: done

etanshaul:
Just found a bug where applying the quickfix to all instances (via the 'apply all..' link) will write the BOM multiple times. Working on a fix.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudDependencyWithNoBomInspection.java
```diff
@@ -0,0 +1,170 @@
...
+      Stream.of(xmlElement.getChildren())
+          .filter(childTag -> childTag instanceof XmlTag)
+          .map(childTag -> (XmlTag) childTag)
+          .filter(childTag -> "version".equals(childTag.getName()))
+          .findFirst()
+          .ifPresent(CloudDependencyWithNoBomInspection.this::stripVersion);
+    }
+
+    private void addBom(Project project) {
+      MavenProject mavenProject = MavenProjectsManager.getInstance(project).findProject(module);
```
ivanporty: Just to clarify a bit, `MavenProject` maps to `Module`, and there could be multiple `MavenProject`s per IDE `Project`, hence the need for the module field, correct?

etanshaul: correct. this inspection in the scope of a particular pom.xml which is part of a particular module.

google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/apis/CloudDependencyWithNoBomInspection.java
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  @Override
+  public void checkFileElement(
+      DomFileElement<MavenDomProjectModel> domFileElement, DomElementAnnotationHolder holder) {
+
+    checkDependencyWithNoBom(domFileElement.getRootElement(), holder);
+  }
+
+  private void checkDependencyWithNoBom(
```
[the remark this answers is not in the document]
etanshaul: yes. I definitely should do that.

etanshaul:
@ivanporty I fixed the bug I mentioned and extracted common code for dependency visiting in the base class.

For the latter, I still had to leave some common logic in the implementation (for example, since one inspection checks for the presence of the BOM, the other check for a missing BOM). Its also arguable that I reduced readability with this refactoring. wdyt?

ivanporty:
I think it's pretty clean, maybe `Consumer` sticks out a bit. What if we provide couple of abstract methods to implement instead - one that decides if inspection needs to run on a POM, another that will take MavenDependency and matching CloudLibrary? then we can leave `checkFileElement` in base class only?

etanshaul:
but then aren't we left with some duplication like before? checking if there are no cloud libraries for example

also regarding

> another that will take MavenDependency and matching CloudLibrary

what did you mean exactly?

etanshaul:
I do like the idea of an abstract `shouldApplyInspection` (or similarly named) method though.

ivanporty:
I meant we keep common functionality and then we are left with two pieces:
* does this need to be run at all - i.e. `CloudDependencyWithNoBomInspection` checks for BOM presence
* if it does run, check for specific `MavenDomDependency` in an abstract method callback, something like 'abstract inspectAndFix(MavenDomDependency, Module)`

Then we won't need to override `checkFileElement` and provide consumer and it could be easier to read a single method (inspectAndFix)

etanshaul:
ah. got it. let me see how that looks.

etanshaul:
very nice suggestion. its a lot cleaner now. ptal

PR #2040, opened 2018-04-19
patch in change_notes at build time.

patflynn:
scrolling through the notes when looking at the plugin.xml was driving me nuts.


patflynn:
Probably not too hard to start generating the html from the changelog.md as a next step.

patflynn:
yeah it's a gradle-intellij-plugin feature.

PR #2047, opened 2018-04-19
Add metrics for BOM related events

### LINKED ISSUE #1971: Add metrics around all BOM integration features
part of #1921

etanshaul:
fixes #1971, and closes out BOM work (#1921)


etanshaul:
hmm we are absolutely interested in that. I'm just considering how to implement this. We sync on the user's BOM every time a maven sync occurs. But this will just give us the information for the individual project, not its proportion of the total. Open to brainstorming ideas for this since I think its a good metric to have.

etanshaul:
@ivanporty ptal at the last commit too. I added "maven" to the event names since we eventually would like to support other build systems. 

One of the events I modified is an already-in-production event (CLIENT_LIBRARY_ADD_LIBRARY_MAVEN) so it will affect historical metrics, but since this is a pretty new feature I think thats ok.

ivanporty:
My only idea could be using that `ProjectComponent#projectOpened()`, and checking for BOM and client libraries once? I'm not really familiar with that but it looks like it should only be called on open?

etanshaul:
For that to be a useful metric, I think we would first need some sort of unique project tracking so we can disambiguate between multiple opening/restarts of the same project. We need to investigate what a good identifier to use here would be.

etanshaul:
an approach could be to just add a ping on each BOM / client library sync with the (simplistic) assumption that all projects sync at roughly the same proportion.

a better approach would be for us to write a unique ID to the persistent metadata of every project using our plugin. Then we could use this ID across all metrics pings which would then give us much better insight into questions like "how many projects use a BOM vs no BOM".

@patflynn @ivanporty any opinions on this?

ivanporty:
Unique ID will definitely make this more precise but it might look like a bit more intrusive than other metrics we have, I feel I don't have enough experience on that, so I'll leave that up to you.

etanshaul:
by unique ID, I mean something like a random UUID.

ivanporty:
> For that to be a useful metric, I think we would first need some sort of unique project tracking so we can disambiguate between multiple opening/restarts of the same project.

While this is true, in general the rate of openings/restarts should be about the same for BOM vs. non-BOM projects, and proportion should stay pretty accurate, wdyt?

PR #2051, opened 2018-04-20
read the change_notes.html file in a context independent way.

patflynn:


google-cloud-tools-plugin/build.gradle
```diff
@@ -26,7 +26,8 @@ intellij {
     pluginName = 'google-cloud-tools'
     plugins 'Groovy', 'gradle', 'git4idea', 'properties', 'junit', 'maven', 'yaml'
     patchPluginXml {
-        changeNotes = new File('google-cloud-tools-plugin/change_notes.html').getText('UTF-8')
+        changeNotes = project (":google-cloud-tools-plugin").file("change_notes.html")
```
etanshaul: is this better than using projectRoot.file ? in any case, approving

etanshaul: oh, its merged already :)

Issue #2031, opened 2018-04-13
previous major versoin (2017) CI build is red


patflynn:

etanshaul:
this is the error we are getting (test only):
![image](https://user-images.githubusercontent.com/1735744/39144511-fc42de1c-46fe-11e8-9924-e491953030a3.png)

etanshaul:
so that GCS test failure is a red herring:

the test failure occurs in _any_ test that calls (directly or indirectly) `UsageTrackerProvider.getInstance()`. And the underlying issue is a thread leak:

![image](https://user-images.githubusercontent.com/1735744/39145637-6022ee9c-4702-11e8-92ac-a92b7e2cec29.png)

PR #2056, opened 2018-04-23
Fix casing on extension key

etanshaul:
this magically fixes the previous version CI (#2031)

Must have been previously case-sensitive. I'm not sure yet of the details why this fixes the problem.


patflynn:
@etanshaul IIRC these were intentionally different (I know that's weird).  I think ultimate was Idea and CE was idea.  Maybe things have changed but just be aware.

etanshaul:
yeah. I'm investigating this just to be sure, but it seems you are right:

`PlatformUtils#getPlatformPrefix` (which is the key that is used to find the extension) returns "Idea" for IC, but "idea" for IU

So it seems my fix is no good: it would result in a `NoOpUsageTracker`. Will keep looking.

etanshaul:
closing. not the right fix

Issue #2057, opened 2018-04-23
Refactor plugin extension points


etanshaul:
We no longer have multiple plugins (account plugin has been merged). Therefore, we don't need the extension point mechanism when a simple application service would do.

patflynn:
Just be-aware that to avoid circular deps there still may be places where
an extension point makes sense.

PR #2059, opened 2018-04-24
Refactors usage tracker extension point to a simple application service

### LINKED ISSUE #2031: previous major versoin (2017) CI build is red

etanshaul:
fixes #2031 and progress towards #2057 

Gets rid of the extension point logic and replaces it with an Application Service. 

Meat of the change is in `google-cloud-tools-core.xml`, `UsageTrackerService`, and `GoogleUsageTrackerService`.

The fix for #2031 is specifically moving the call the PermanentInstallationID.get() out of the class level and into the `sendEvent` method. The call was causing a thread leak in unit test mode.


google-cloud-tools-plugin/google-cloud-core/src/com/google/cloud/tools/intellij/analytics/GoogleUsageTrackerService.java
```diff
@@ -92,7 +92,6 @@
           // Apparently the hit type should always be of type 'pageview'.
           new BasicNameValuePair(HIT_TYPE_KEY, PAGE_VIEW_VALUE),
           new BasicNameValuePair(IS_NON_INTERACTIVE_KEY, STRING_FALSE_VALUE),
-          new BasicNameValuePair(UNIQUE_CLIENT_ID_KEY, PermanentInstallationID.get()),
```
etanshaul: This call right here is what was causing #2031

google-cloud-tools-plugin/google-cloud-core/src/com/google/cloud/tools/intellij/analytics/GoogleUsageTrackerService.java
```diff
@@ -103,7 +102,7 @@
    * Constructs a usage tracker configured with analytics and plugin name configured from its
    * environment.
    */
-  public GoogleUsageTracker() {
+  public GoogleUsageTrackerService() {
```
ivanporty: So this constructor will be called for unit tests (since the service will be initialized with the instance by default), but now no actual operations are performed since we do all logic in `sendEvent` itself, correct?

etanshaul: right. this is pretty much the same as how it worked before. If you look at the start of `GoogleUsageTrackerService#sendEvents` you'll see there is a guard for unitTestMode to pings from tests.

google-cloud-tools-plugin/google-cloud-core/src/com/google/cloud/tools/intellij/analytics/KeyedExtensionUsageTrackerProvider.java
```diff
@@ -1,78 +0,0 @@
...
-   * For usage tracking in the Cloud Tools family of plugins, we only want one tracker pinging an
-   * analytics backend for a given platform. New platform specific trackers can register with their
-   * platform prefix key and the UsageTrackerProvider will select them. Otherwise, the no-op usage
-   * tracker will be used.
-   *
-   * @param key A string search key associated with an extension
-   * @return the first implementation of UsageTracker associated with {@param key}
-   */
-  static UsageTracker getTracker(@Nullable String key) {
-    UsageTracker instance = (key == null) ? null : COLLECTOR.findSingle(key);
```
ivanporty: Not sure what this code does but it seems to be using platform prefix as an additional key for usage tracker, was it used at all - looks like we don't need it since it's removed?

etanshaul: Before, the way it worked was we registered a UsageTracker extension for the (now removed) extension point. Each extension would have a key corresponding to the platform ID. The code you referenced would then try to pull up an implementation for the given key.

The idea behind this was so that we could whitelist certain platforms. I think specifically we wanted to exclude Android Studio from sending pings. So, since there is no extension with the Android Studio platform key, it wouldn't ping.

We don't need this anymore since we want to collect metrics from all platforms (currently its really just IC and IU, but I don't see this changing once we support more).

@patflynn can you confirm that once we open this plugin up to Android Studio we no longer need to exclude it from usage tracking?

etanshaul:
Sounds like it might still need this mechanism to exclude certain platforms. Lets keep this around until that is resolved. In the meantime, I'll extract the fix for the previous-version CI out of this PR into a new one.

PR #2060, opened 2018-04-24
Flags to disable ultimate when building with IC.

ivanporty:
Do not review please, checking out CI vs. local caches behavior.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-ideaEdition = IU
+ideaEdition = IC
 ideaVersion = 2018.1
 intellijRepoUrl = https://www.jetbrains.com/intellij-repository
 javaVersion = 1.8
--- a/.idea/codeStyles/codeStyleConfig.xml
+++ b/.idea/codeStyles/codeStyleConfig.xml
@@ -1,5 +1,6 @@
 <component name="ProjectCodeStyleConfiguration">
   <state>
+    <option name="USE_PER_PROJECT_SETTINGS" value="true" />
     <option name="PREFERRED_PROJECT_CODE_STYLE" value="GoogleStyle" />
   </state>
 </component>
--- a/google-cloud-tools-plugin/app-engine/java/build.gradle
+++ b/google-cloud-tools-plugin/app-engine/java/build.gradle
@@ -20,7 +20,7 @@ sourceSets.test.java.srcDirs = ['testSrc']
 sourceSets.test.resources.srcDirs = ['testResources', 'testData']
 
 intellij {
-    plugins 'Groovy', 'gradle', 'git4idea', 'properties', 'junit', 'yaml', 'maven'
+    plugins 'gradle', 'git4idea', 'properties', 'junit', 'yaml', 'maven'
 }
 
 dependencies {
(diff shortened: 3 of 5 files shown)


gradle.properties
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-ideaEdition = IU
+ideaEdition = IC
```
etanshaul: lets leave this as IU, no?

build.gradle
```diff
@@ -18,16 +18,19 @@ buildscript {
     repositories {
         maven { url "https://maven-central.storage.googleapis.com"}
         maven { url "https://plugins.gradle.org/m2/" }
+        maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
+        maven { url 'http://dl.bintray.com/jetbrains/intellij-plugin-service'  }
     }
     dependencies {
         classpath "net.ltgt.gradle:gradle-errorprone-plugin:0.0.13"
         classpath "net.ltgt.gradle:gradle-apt-plugin:0.13"
+        classpath "org.jetbrains.intellij.plugins:gradle-intellij-plugin:0.4.0-SNAPSHOT"
```
etanshaul: any idea from their github page when this will make the release? It would be nice to not include a snapshot version

gradle.properties
ivanporty: Sure, just was making sure we're building with IC in Kokoro too, I'll revert it before merge.

build.gradle
ivanporty: Nothing particular on their page, no due date for that 0.4.0. I agree using snapshot is a dangerous thing, we might just use for community build although I'm not sure this is possible conditionally with Gradle.

gradle.properties
etanshaul: sgtm

build.gradle
etanshaul: yeah. with all the headache we've been having the gradle-intellij-plugin I'm worried about making things worse with a snapshot version. I'm tempted to not merge in this fix until its been released (and we'll just be aware of how to do this locally if we need to build with CE). wdyt?

etanshaul: also, it would be nice to have a CI build against CE. I'll make an issue for that.

ivanporty: Yeah, I think this might be a good idea. We won't be able to add CE into our master target list though like we do with previous and EAP version.

etanshaul: why not?

ivanporty: Well, we can add that for sure, it's just it will continue to fail for some time until we switch to fresh release, I meant couldn't add target that succeeds for now.

etanshaul: oh right. Yeah, I meant after we figure out this whole mess, so that we can catch future messes :)

ivanporty:
@etanshaul I removed snapshot dependency and left updates for CE edition build to work properly with fixed IJ gradle plugin, PTAL.

app-engine/java/ultimate/build.gradle
```diff
@@ -19,7 +19,7 @@ sourceSets.main.java.srcDirs = ['src']
 sourceSets.test.java.srcDirs = ['testSrc', 'testResources']
 
 intellij {
-    type 'IU'
+    type = 'IU'
```
etanshaul: nit, but the `=` isn't required here right?

ivanporty: It doesn't, I just wanted to unify it with the main build file where we have `type = ideaEdition` so it looks similar. I can revert this if you feel this is less intuitive.

etanshaul: nah. I'm ok with it as is

PR #2061, opened 2018-04-24
 move all GCP apis management code to own module

patflynn:
Further progress towards #1896

This is the last module to extract of google-cloud-tools-plugin.

Next steps are to minimize deps across the module set and replace static
inclusions with module dependent inclusions.


etanshaul:
mind describing what circular dependency you are avoiding by doing this?

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/MavenTestUtils.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.google.cloud.tools.intellij;
+package com.google.cloud.tools.intellij.testing;
```
etanshaul: so this can't live in the new module's test src?

patflynn: It could, it just seemed more general.  Do you want me to move it?

PR #2066, opened 2018-04-25
Only return the real UsageTrack impl if not in unit test mode

### LINKED ISSUE #2031: previous major versoin (2017) CI build is red

etanshaul:
addresses #2031 

Always return the Noop usage tracker in unit test mode.

The previous CI build was failing due to a thread leak caused by the call to `PermanentInstallationId.get()` in `GoogleUsageTracker` (which apparently was fixed in 2018.1). This workaround prevents that call from ever executing in unit test mode.


etanshaul:
@ivanporty just fyi I took a different approach here than in the previously (now closed CL). This will be more efficient as it won't require `PermanentInstallationId.get()` (which reads from the filesystem) to be called on every ping.

patflynn:
for some reason I thought we were already doing this

etanshaul:
we were doing it at the ping level. not at the service provider level. t

PR #2075, opened 2018-04-27
pull out the last couple of actions to their respective modules

patflynn:
also is slightly more consistent in how we register the actions
across all the modules and reorders the actions a little.


patflynn:
image

PR #2100, opened 2018-05-10
More accurate depends in xml

patflynn:
will add comments in a bit.


patflynn:
So far lightly tested in pycharm, webstorm, clion, rider, phpstorm, goland.

etanshaul:
awesome will look more closely today.

do you know if there is a way to launch each of those other IDEs from the gradle-intellij-plugin (by changing `ideaEdition` in gradle.properties?). The readme only shows the codes for a few of them. Or were you testing by manually building the archives?

google-cloud-apis/resources/META-INF/google-cloud-apis.xml
```diff
@@ -15,6 +15,7 @@
   -->
 
 <idea-plugin>
+  <module value="com.google.cloud.tools.intellij.google-cloud-apis"/>
```
etanshaul: could you explain what is the effect of having this module value here (and similarly in the other feature xmls)? I don't see them consumed anywhere.

patflynn: aah yes.  I have a theory that this allows you to specify narrower dependencies between plugins. 

see the convention in the JB 1st party plugins: https://github.com/JetBrains/intellij-community/blob/9a35caeaa5955e4998b32babe3acab06abbd1b44/plugins/gradle/java/resources/META-INF/gradle-java-integration.xml and https://github.com/JetBrains/intellij-community/blob/3ea2dd0fed900d67e850a0e3081c6423d3d03d84/resources/src/META-INF/IdeaPlugin.xml

it's undocumented as far as I can tell.

patflynn:
@etanshaul You can setup the gradle plugin to use a locally installed SDK.  I tested it by packaging and installing manually.  

Next logical step would be setting up CI for our support IDEs.

patflynn:
Closes #1896

Issue #2106, opened 2018-05-14
App Engine module needs access to the Gradle module


etanshaul:
Due to limitation 2 described in #2100, we cannot simply include a depends on the gradle plugin from the app-engine module.

etanshaul:
Similarly we cannot use the workaround used for yaml (of putting it in the root plugin.xml) since gradle is not available in all IDEs.

patflynn:
We may have to split out the appengine gradle support into its own xml.

etanshaul:
Do you recall how the maven module is available from the appengine module? I'm not seeing an explicit depends on maven in that chain.

etanshaul:
yeah it is strange. I don't see maven included directly in com.intellij.modules.java

but I do know for sure that Gradle is not available since I ran into this issue on my branch (that adds gradle stuff to app engine) only after merging in the latest plugin.xml refactoring. I can unblock development by adding the gradle depends alongside the yaml depends (but of course this cannot be a permanent solution).

patflynn:
right.

I think the only solution is for you to keep all your Gradle based features
in a separate app-engine-java-gradle.xml and optionally include it from the
google-cloud-tools-plugin plugin.xml keyed on gradle module.

Issue #2105, opened 2018-05-14
Add gradle based deployment support for CE


etanshaul:
Umbrella issue to track this effort. 

The approach for CE will be to use the app-gradle-plugin. For UE, gradle based deployment already works (via the native artifacts).

PR #2108, opened 2018-05-15
Collect data from Gradle on import and add it to the module

etanshaul:
this is the first step towards #2105. Also addresses #2106.

Background:
The goal is to enable deployment from Gradle sources with Community Edition (so that Android Studio can use this plugin). In order to accomplish this, we will use the [app-gradle-plugin](https://github.com/GoogleCloudPlatform/app-gradle-plugin) to build the project and create the exploded war directory (the plugin adds an `exlodeWar` task which gets called on `assemble`). Then, the existing App Engine standard gcloud based deployment path will be invoked on the built module.

What this PR does:
- Creates a new app-engine-gradle-java Gradle module. This is to get around the limitation explained in #2106.
- Sets up all the plumbing needed to collect data from Gradle on Gradle sync. Populates the `AppEngineGradleModel` with the needed information:
   - does the module have the Gradle plugin configured
   - path to the gradle build directory
- On module import, adds the collected information to a new App Engine Gradle facet (actually creating the facet will come in the next PR)

The next PRs will fill in the remaining gaps of:
- adding the App Engine Gradle facet
- creating the new Gradle based deployment source


etanshaul:
@ivanporty should be ready for review now

app-engine/java/gradle/resources/META-INF/services/org.jetbrains.plugins.gradle.tooling.ModelBuilderService
```diff
@@ -0,0 +1 @@
+com.google.cloud.tools.intellij.appengine.java.gradle.AppEngineGradleModelBuilderService
```
ivanporty: Is this another way to define a service implementation, just plain text lines under `services`?

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleModel.java
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.intellij.appengine.java.gradle;
+
+/** Model containing information obtained from the Gradle integration. */
+public interface AppEngineGradleModel {
```
ivanporty: Do we need a separate interface for this model, are we expecting more implementations?

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleModel.java
```diff
@@ -0,0 +1,30 @@
...
+
+  /**
+   * Returns {@code true} if the module has the app-gradle-plugin configured, and {@code false}
+   * otherwise.
+   */
+  boolean hasAppEngineGradlePlugin();
+
+  /** Returns the path to the Gradle build directory. */
+  String gradleBuildDir();
```
ivanporty: Could we think about a bit more specific name for a build dir or maybe more javadoc - sounds slightly generic?

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleProjectDataService.java
```diff
@@ -0,0 +1,107 @@
...
+    new WriteCommandAction.Simple(project) {
+      @Override
+      protected void run() {
+        Stream.of(modelsProvider.getModules())
+            .filter(module -> moduleNameToModel.containsKey(module.getName()))
+            .forEach(
+                module -> {
+                  AppEngineGradleModule appEngineGradleModule =
+                      moduleNameToModel.get(module.getName());
+                  if (appEngineGradleModule.getModel().hasAppEngineGradlePlugin()) {
```
ivanporty: Could this be different (having AE gradle plugin) for modules in the same project? I.e. some could have this plugin and some modules not?

etanshaul: yeah some modules could have it, some may not. it adds the facet only to those that have it configured. does that answer your question?

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleModel.java
etanshaul: hmm. build directory is the same terminology used by Gradle itself. It really is just the directory where the gradle build is located. Do you have any suggestions?

app-engine/java/gradle/resources/META-INF/services/org.jetbrains.plugins.gradle.tooling.ModelBuilderService
etanshaul: this is not really documented anywhere, but this is the method for getting extra information from the Gradle integration (i.e. building extra Gradle model's). there is some discussion here: https://intellij-support.jetbrains.com/hc/en-us/community/posts/206762805-how-to-use-create-a-gradle-custom-model-

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleModel.java
etanshaul: probably don't need it. I'll check

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleProjectDataService.java
etanshaul: side note to self: we may only want to add this facet if we are in Community Edition.

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleModel.java
etanshaul: turns it out has to be an interface, if its not sync fails with:

`Cause: com.google.cloud.tools.intellij.appengine.java.gradle.AppEngineGradleModel is not an interface`

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleProjectDataService.java
etanshaul: (I'll add in the platform check in the facet service in the next pr)

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleModel.java
ivanporty: I was thinking that Gradle build directory could be for any of the modules, for build output, for build sources, just module root where build.gradle is? I guess that's the latter, correct, directory where build.gradle for the module is?

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleProjectDataService.java
ivanporty: Oh, so the module is based on Gradle, and the module gradle build script itself can have that app engine plugin defined or not?

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleModel.java
etanshaul: so the AppEngineGradleModel is scoped to a single module, for one.

Also the `gradleBuildDir` is the build output directory under `../build` where the exploded war directory is located

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleProjectDataService.java
etanshaul: thats correct

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleModel.java
etanshaul: and to elaborate one why I'm storing this: its simply so that we can later get access to the exploded war directory (for gcloud deployment). This usage will come in the next PR or two.

ivanporty: Thanks, now it's clear! we can probably add the last sentence into javadoc, wdyt?

etanshaul: I'll add a sentence explaining what the build dir is: the build output where the built artifacts live. I'm hesitant to put anything about how it will be consumed since the model is agnostic of that.

PR #2120, opened 2018-05-17
Deploy from gradle source in CE using app-gradle-plugin

### LINKED ISSUE #2105: Add gradle based deployment support for CE

etanshaul:
fixes #2105

part 2 of #2105 (#2108 was the first part)

- adds the app engine gradle facet
- sets up the gradle deployment source for deployment from CE


etanshaul:
@ivanporty should be ready for review now

common-test-lib/src/main/java/com/google/cloud/tools/intellij/testing/TestScopedSystemPropertyRule.java
```diff
@@ -0,0 +1,51 @@
...
+ * A JUnit {@link TestRule} that will reset the given system property to its original value after
+ * each test. This allows individual tests to set system properties for testing purposes.
+ */
+public class TestScopedSystemPropertyRule extends ExternalResource {
+
+  private final String propertyKey;
+
+  private String originalPropertyValue;
+
+  public TestScopedSystemPropertyRule(String propertyKey) {
```
ivanporty: Could we also expand this useful rule to include a property value or even an enum of possible IDE values? I can imagine some test cases would work only for a specific runtime?

app-engine/java/testSrc/com/google/cloud/tools/intellij/appengine/java/cloud/AppEngineDeploymentConfiguratorTest.java
```diff
@@ -99,4 +114,43 @@ public void createEditor_withStandardEnvironment_doesReturnStandardEditor() {
 
     assertThat(editor).isInstanceOf(AppEngineStandardDeploymentEditor.class);
   }
+
+  @Test
+  public void getDeploymentSources_withCommunityEdition_returnsSource() {
+    System.setProperty(PlatformUtils.PLATFORM_PREFIX_KEY, PlatformUtils.IDEA_CE_PREFIX);
```
ivanporty: See my comment for `TestScopedSystemPropertyRule`, we could potentially move this to the rule itself and do it only once for the rule initialization call, if all test cases from this test class work with CE?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/util/AppEngineUtil.java
```diff
@@ -147,6 +151,34 @@ private static boolean doesArtifactMatchEnvironment(
...
+   *     deployment sources.
+   */
+  public static List<ModuleDeploymentSource> createGradlePluginDeploymentSources(
+      @NotNull Project project) {
+    List<ModuleDeploymentSource> moduleDeploymentSources = Lists.newArrayList();
+
+    Stream.of(ModuleManager.getInstance(project).getModules())
+        .forEach(
+            module -> {
+              if (AppEngineStandardFacet.hasFacet(module)
```
ivanporty: could you please remind me how these two facets are cooperating - both of them are supposed to be present in the module for deployment, but App Engine Standard is also available in Community regardless of Web support presence? For some reason I thought opposite.

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/GradlePluginDeploymentSource.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * An app-gradle-plugin based {@link ModuleDeploymentSource} and {@link AppEngineDeployable}. Sets
+ * up the deployment source and provides the path to the Gradle exploded war directory stored in the
+ * App Engine Gradle facet.
+ */
+public class GradlePluginDeploymentSource extends ModuleDeploymentSourceImpl
+    implements AppEngineDeployable {
+
+  private static final String EXPLODED_WAR_DIR_PREFIX_FORMAT = "/exploded-%s";
```
ivanporty: What defines this format for the directory, app engine gradle plugin?

etanshaul: yeah, the app-gradle-plugin. @loosebazooka can you confirm that this format for the exploded war directory generated by the plugin's `explodeWar` is fixed?

app-engine/java/testSrc/com/google/cloud/tools/intellij/appengine/java/cloud/AppEngineDeploymentConfiguratorTest.java
etanshaul: My original intent with the `TestScopedSystemPropertyRule` was to be a generic system property setter, not specific to IDEA platform values. Are you ok with that, or would you prefer I make it specific to this use case?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/GradlePluginDeploymentSourceType.java
```diff
@@ -0,0 +1,126 @@
...
+  /**
+   * Creates a new Gradle assemble before-run task. The app-gradle-plugin chains an "explodeWar"
+   * task to gradle build tasks to created the exploded-war directory.
+   */
+  @Nullable
+  @Override
+  protected BeforeRunTask createBuildTask(Module module) {
+    ExternalSystemTaskExecutionSettings settings = new ExternalSystemTaskExecutionSettings();
+    settings.setExternalProjectPath(module.getModuleFilePath());
+    settings.setTaskNames(ImmutableList.of(GRADLE_ASSEMBLE_TASK));
```
ivanporty: So assemble task will create exploded war here as part of the build before deployment?

etanshaul: yeah. the app-gradle-plugin attaches the explodeWar task to it.

etanshaul: chains it, or whatever the terminology is.

app-engine/java/testSrc/com/google/cloud/tools/intellij/appengine/java/cloud/AppEngineDeploymentConfiguratorTest.java
ivanporty: I'm ok with this now, but also, in case we start adding more features for a specific IJ runtime later, we can expand this into a new rule maybe that allows us to cleanly specify a test set will execute under specific IDE environment. wdyt?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/util/AppEngineUtil.java
etanshaul: The `App Engine standard facet` is there always - if its a standard project (appengine-web.xml) - regardless of CE or UE.

The `App Engine gradle plugin facet` is there only in Community Edition _if_ the module has the app-gradle-plugin AND the `App Engine standard facet` is there too.

For deployment from this new gradle source, _both_ facets are required.

In reality, the `App Engine gradle plugin facet` is a dependent facet, or a "child" of the app engine facet. There is the concept of an "underlying" facet - I'll see if this can better represent this relationship.

app-engine/java/testSrc/com/google/cloud/tools/intellij/appengine/java/cloud/AppEngineDeploymentConfiguratorTest.java
etanshaul: Oh I see your intention. Yes, that would be nice. We can definitely do that once more use-cases creep in.

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/BuildDeploymentSourceType.java
```diff
@@ -42,7 +43,10 @@
 public abstract class BuildDeploymentSourceType
     extends DeploymentSourceType<ModuleDeploymentSource> {
 
-  public BuildDeploymentSourceType(@NotNull String id) {
+  static final String PROJECT_ATTRIBUTE = "project";
```
ivanporty: IDE project or cloud project? Could we clarify this a bit more? Same for version - which version is this?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/BuildDeploymentSourceType.java
```diff
@@ -120,6 +124,22 @@ public void updateBuildBeforeRunOption(
     }
   }
 
+  @Override
+  public void save(@NotNull ModuleDeploymentSource deploymentSource, @NotNull Element tag) {
```
ivanporty: Could you please explain logic behind this a bit more - is this saved for deployment - i.e. this is cloud project and service version a user specified?

etanshaul: I should have added documentation here.. But essentially save/load are for serializing/deserializing the deployment source.

For example, you create a run config, select a deployment source and set some options, then save it. Then you close the IDE. Later you open the IDE again and want to just run that existing run config. This allows the deployment source info to be restored.

I'll add some javadocs

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/BuildDeploymentSourceType.java
etanshaul: I renamed as follows to hopefully clear this up:

`PROECT_ATTRIBUTE` -> `CLOUD_PROECT_ATTRIBUTE`
`VERSION_ATTRIBUTE` - > `APP_ENGINE_VERSION_ATTRIBUTE`

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/util/AppEngineUtil.java
etanshaul: yeah it seems that using an underlying facet will better convey and enforce this relationship. let me refactor this.

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/BuildDeploymentSourceType.java
ivanporty: Awesome, thanks! Constant confusion with those projects. What do you think if we add an enhancement to refactor the method names for `AppEngineDeployable` for another PR to also clarify `setProjectName` and `setVersion`?

etanshaul: I was thinking the same thing. will do. I'll keep it separte since it touches a lot of places

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/util/AppEngineUtil.java
etanshaul: nevermind, it seems that you can only initialize the underlying facet at facet creation time, meaning that we would need to initialize the gradle facet when the app engine standard facet is created. This is not what we want because the gradle facet is tied to the gradle sync.

etanshaul:
thanks for the review. will merge in after our next release.

PR #2131, opened 2018-05-23
Fix cloud tools action ordering

### LINKED ISSUE #2123: Google Cloud Tools menu group lost ordering
I believe  #2100 is the culprit.

screenshot from 2018-05-21 09-27-32

etanshaul:
fixes #2123 

image

This is somewhat brittle still and care will be needed when adding more actions, but the key idea is that the actions defined in modules included optionally (via `depends optional`) are loaded out of band (not in the order they are declared in the plugin.xml). Therefore, those actions need to be defined relative to other actions.


etanshaul:
confirmed the correct ordering in both CE and UE

Issue #2133, opened 2018-05-23
Metrics not functional in non-IDEA IDEs


etanshaul:
need to add platform specific keys to the usage tracker extension

etanshaul:
also the opt in panel is being hidden for non IDEA IDEs

PR #2134, opened 2018-05-23
Refactors usage tracker extension point to simple application service

etanshaul:
First step towards fixing #2133 

This is the same exact PR as the previously dismissed #2059 (it was easier to make a new PR than to resolve the merge conflicts after all the recent refactoring). see `google-cloud-core.xml`

This PR removes the platform whitelist mechanism for specifying platforms to include usage tracking for. 

We decided before to not merge it in because we weren't sure if the whitelist mechanism for specifying which platforms to include in usage tracking was still needed. Turns out having that mechanism is a huge headache now that we are supporting more than just IDEA. We'd have to whitelist every single IDE that we want to include tracking for (and if a new IDE is ever added, we'd have to remember to add it to the whitelist).

Instead, I think its better to just get rid of this, and then if we decide to exclude a platform (i.e. Android Studio) from usage tracking, we can add simple logic to return a NoopUsageTracker in `UsageTrackerService#getInstance`. In other words, its better now to blacklist platforms than to maintain this whitelist mechanism.


google-cloud-core/resources/META-INF/google-cloud-core.xml
```diff
@@ -61,23 +61,9 @@
                         serviceImplementation="com.google.cloud.tools.intellij.resources.DefaultGoogleApiClientFactory"/>
   </extensions>
 
-  <extensions defaultExtensionNs="com.google.gct.core">
-    <!-- For IJ Community Edition -->
-    <usageTracker implementationClass="com.google.cloud.tools.intellij.analytics.GoogleUsageTracker"
-                  key="Idea"/>
```
etanshaul: this is the mechanism we used to whitelist platforms for usage tracking:

`Idea` is the platform key for Community
`idea` is the platform key for Ultimate

patflynn:
adding @loosebazooka since he worked on the initial version of this.

loosebazooka:
Android studio wanted the tracking to report to their analytics, so we had an implementation of the tracker with the "android-studio" key, and and based on the xml config, it would either load their or our tracking service.

This was mostly to continue to share code with android studio. But I think we've already diverged from that source and sharing doesn't happen anymore. You can do whatever you want with this code now. Simplifying this and removed the keyed service loader as you've done is probably for the best.

google-cloud-core/resources/META-INF/google-cloud-core.xml
loosebazooka: I think this was originally to differentiate betweeen android studio and intellij. And I think at the time, I should've just done `Android Studio` and `default=GoogleUsageTracker` but for some reason, which I cannot remember now, I set `default=NoOp` and explicitly whitelisted `Idea` and `idea`. Whitelist is probably unnecessary though.

etanshaul:
thanks for the context @loosebazooka. 

if it turns out that we still need an Android Studio tracking implementation in the future we can add it back in in a more simple way than that keyed service loader (which doesn't fit well now that we'll support all these other IDEs).

google-cloud-core/resources/META-INF/google-cloud-core.xml
etanshaul: do you mean just key by `AndroidStudio` with an implementationClass of `NoOp`. And then default to the GoogleUsageTracker when no matching key is found?

I actually was going to do that before I went down this road. Maybe its better now that I think about it (just in case we need a configuration based blacklist option).

etanshaul:
@ivanporty could you approve if it looks good? We'll go with this one after all...

google-cloud-core/src/com/google/cloud/tools/intellij/analytics/GoogleUsageTrackerService.java
```diff
@@ -48,7 +48,7 @@
 import org.jetbrains.annotations.Nullable;
 
 /** Google Usage Tracker that reports to Cloud Tools Analytics backend. */
-public class GoogleUsageTracker implements UsageTracker, SendsEvents {
+public class GoogleUsageTrackerService implements UsageTrackerService, SendsEvents {
```
loosebazooka: Filter out IDEs that you want to completely disable tracking for in here?

etanshaul: yeah. I guess I'll go ahead an add AS to keep the behavior the same as before. And we'll revisit this later if it needs to be changed.

etanshaul: done: https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/2134/commits/3879306b82dcf3d08ae2daba34fa8af0bea7573c

PR #2136, opened 2018-05-24
Convert usage tracker keyed service whitelist to a blacklist

### LINKED ISSUE #2133: Metrics not functional in non-IDEA IDEs

etanshaul:
fixes #2133 

replaces #2123. I decided that this is a better implementation for now. If we need to exclude certain platforms in the future, it is nice to have a configuration based blacklist.

### DIFF
--- a/google-cloud-core/resources/META-INF/google-cloud-core.xml
+++ b/google-cloud-core/resources/META-INF/google-cloud-core.xml
@@ -62,12 +62,10 @@
   </extensions>
 
   <extensions defaultExtensionNs="com.google.gct.core">
-    <!-- For IJ Community Edition -->
-    <usageTracker implementationClass="com.google.cloud.tools.intellij.analytics.GoogleUsageTracker"
-                  key="Idea"/>
-    <!-- For IJ Ultimate Edition -->
-    <usageTracker implementationClass="com.google.cloud.tools.intellij.analytics.GoogleUsageTracker"
-                  key="idea"/>
+    <!-- Exclude Android Studio-->
+    <usageTracker
+            implementationClass="com.google.cloud.tools.intellij.analytics.KeyedExtensionUsageTrackerProvider.NoOpUsageTracker"
+            key="AndroidStudio"/>
   </extensions>
 
   <extensionPoints>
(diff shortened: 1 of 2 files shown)


google-cloud-core/resources/META-INF/google-cloud-core.xml
```diff
@@ -62,12 +62,10 @@
...
-    <!-- For IJ Community Edition -->
-    <usageTracker implementationClass="com.google.cloud.tools.intellij.analytics.GoogleUsageTracker"
-                  key="Idea"/>
-    <!-- For IJ Ultimate Edition -->
-    <usageTracker implementationClass="com.google.cloud.tools.intellij.analytics.GoogleUsageTracker"
-                  key="idea"/>
+    <!-- Exclude Android Studio-->
+    <usageTracker
+            implementationClass="com.google.cloud.tools.intellij.analytics.KeyedExtensionUsageTrackerProvider.NoOpUsageTracker"
+            key="AndroidStudio"/>
```
etanshaul: `AndroidStudio` is the correct key as found in the AS source code for `PlatformUtils`

google-cloud-core/resources/META-INF/google-cloud-core.xml
```diff
@@ -62,12 +62,10 @@
   </extensions>
 
   <extensions defaultExtensionNs="com.google.gct.core">
-    <!-- For IJ Community Edition -->
-    <usageTracker implementationClass="com.google.cloud.tools.intellij.analytics.GoogleUsageTracker"
-                  key="Idea"/>
-    <!-- For IJ Ultimate Edition -->
-    <usageTracker implementationClass="com.google.cloud.tools.intellij.analytics.GoogleUsageTracker"
-                  key="idea"/>
+    <!-- Exclude Android Studio-->
```
ivanporty: Looks good! How about adding a similar message from the javadoc comment here as well - i.e. 
> To exclude a platform, add a new usageTracker extension with a key equal to the platform key and the NoOpUsageTracker.

It might make it a bit faster to remember/make sure it works as blacklist?

etanshaul: sounds good

PR #2155, opened 2018-05-31
Update plugin.xml

parkjam4:
Added a link to the GitHub readme page where we have the feature matrix.
Note: I cannot preview this page, so I am not sure if the link works correctly or if the text is aligned properly.

### DIFF
--- a/google-cloud-tools-plugin/resources/META-INF/plugin.xml
+++ b/google-cloud-tools-plugin/resources/META-INF/plugin.xml
@@ -21,14 +21,14 @@
     <![CDATA[<html>
       <p>This plugin integrates Google Cloud Platform into your IDE development workflows. It includes features such as:</p>
       <ul>
-        <li>Deploying to the Google App Engine standard and flexible environments.</li>
+        <li>Deploying to the App Engine standard and flexible environments.</li>
         <li>Setting Stackdriver Debugger snapshots to debug your applications running in the cloud.</li>
         <li>Checking out or importing your projects from Cloud Source Repositories.</li>
-        <li>Browsing your Google Cloud Storage buckets and copying bucket paths into your clipboard.</li>
+        <li>Browsing your Cloud Storage buckets and copying bucket paths into your clipboard.</li>
         <li>Running and debugging App Engine standard applications on a local development server.</li>
-        <li>Adding Google Cloud Java client libraries to your project.</li>
+        <li>Adding Cloud Client Libraries for Java to your project.</li>
       </ul>
-      <p>These features are all available in IntelliJ IDEA. As of version 18.5.1, some of these features are available in other JetBrains IDEs, beginning with PyCharm.</p>
+      <p>These features are all available in IntelliJ IDEA. As of version 18.5.1, some of these features are available in other JetBrains IDEs.</p>
       <p>Refer to the <a href="https://github.com/GoogleCloudPlatform/google-cloud-intellij/blob/master/README.md#supported-platforms">Supported Platforms</a> page to see which features are available in which IDEs.</p>
     </html>]]>
   </description>
--- a/google-cloud-tools-plugin/resources/META-INF/plugin.xml
+++ b/google-cloud-tools-plugin/resources/META-INF/plugin.xml
@@ -20,7 +20,7 @@
   <description>
     <![CDATA[<html>
       <p>This plugin integrates Google Cloud Platform into your IDE development workflows.</p>
-      <p>The following features are all available in IntelliJ IDEA. As of version 18.5.1, Cloud Storage and Cloud Source Repositories integration are available in PyCharm:</p>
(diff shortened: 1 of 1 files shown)


google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -28,6 +28,8 @@
         <li>Running and debugging App Engine standard applications on a local development server.</li>
         <li>Adding Google Cloud Java client libraries to your project.</li>
       </ul>
+      <p>These features are all available in IntelliJ IDEA. As of version 18.5.1, some of these features are available in other JetBrains IDEs, beginning with PyCharm.</p>
```
etanshaul: instead of "some of these features", can you just say "Cloud Storage and Cloud Source Repositories integration"

I'm also one the fence about the "beginning with PyCharm" wording. Maybe just leave this part out. That way when we expand support for the other IDEs we won't need to re-release (and can just update the docs / feature matrix that you link to)

etanshaul:
thanks! html looks good so I believe it will be properly formatted

google-cloud-tools-plugin/resources/META-INF/plugin.xml
etanshaul: this is optional, but while you are in the area, wdyt about updating the other wording to remove the "Google" prefix from the various products as appropriate?

parkjam4: Good idea on making additional changes. Will do.

I hear you about removing the "beginning with PyCharm". That makes sense. In the same vein, if we list "Cloud Storage and Cloud Source Repositories integration", we will need to update this page when we expand the feature list. So, perhaps we should keep it general and rely on users to go to the feature matrix to see what features are available for their IDE of choice. WDYT?

google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -21,13 +21,15 @@
...
         <li>Setting Stackdriver Debugger snapshots to debug your applications running in the cloud.</li>
         <li>Checking out or importing your projects from Cloud Source Repositories.</li>
-        <li>Browsing your Google Cloud Storage buckets and copying bucket paths into your clipboard.</li>
+        <li>Browsing your Cloud Storage buckets and copying bucket paths into your clipboard.</li>
         <li>Running and debugging App Engine standard applications on a local development server.</li>
-        <li>Adding Google Cloud Java client libraries to your project.</li>
+        <li>Adding Cloud Client Libraries for Java to your project.</li>
       </ul>
+      <p>These features are all available in IntelliJ IDEA. As of version 18.5.1, some of these features are available in other JetBrains IDEs.</p>
+      <p>Refer to the <a href="https://github.com/GoogleCloudPlatform/google-cloud-intellij/blob/master/README.md#supported-platforms">Supported Platforms</a> page to see which features are available in which IDEs.</p>
```
patflynn: I'd put this up top, or put the IDE support information by the feature.  I'd typically see 'integrated into my ide..' ' deploying app engine..'  and stop reading and install.

google-cloud-tools-plugin/resources/META-INF/plugin.xml
etanshaul: Adding more features for the other IDEs will require another release, so we can remember to update this description alongside that. So I’m ok leasing the actual features supported by PyCharm.

etanshaul: Adding more features for the other IDEs will require another release, so we can remember to update this description alongside that. So I’m ok leasing the actual features supported by PyCharm.

google-cloud-tools-plugin/resources/META-INF/plugin.xml
parkjam4: I hear you. But I don't want to put too much text between "these features are available..." and the feature list. I think it could break up the flow of the paragraph too much.

google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -20,7 +20,7 @@
   <description>
     <![CDATA[<html>
       <p>This plugin integrates Google Cloud Platform into your IDE development workflows.</p>
-      <p>The following features are all available in IntelliJ IDEA. As of version 18.5.1, <a href="https://github.com/GoogleCloudPlatform/google-cloud-intellij/blob/master/README.md#supported-platforms">some of these features</a> are available in other JetBrains IDEs:</p>
+      <p>The following features are all available in IntelliJ IDEA. As of version 18.5.1, Cloud Storage and Cloud Source Repositories integration are available in PyCharm:</p>
```
patflynn: what do you think of? 
.., **but only** Cloud Storage and ..

etanshaul: that sgtm

parkjam4: got it.

Issue #2128, opened 2018-05-21
Cloud SDK Support: extract common interfaces and modularize


ivanporty:
Cloud SDK functionality is currently tightly coupled into Java functionality and App Engine Java component presence. In order to support more languages and deployments, we could split it into runtime agnostic part, introduce modular language/runtime support (starting with existing Java component support), and extract the Cloud SDK into its own independent plugin module.

etanshaul:
it has since moved into core. closing

PR #2156, opened 2018-05-31
Move Cloud SDK support into its own module

ivanporty:
First step towards #2128.

This PR extracts Cloud SDK, services, messages, including minor Java component dependencies (Java related validation and installation if App Engine Java component for Managed SDK) into its own module. Procedure is very similar to previously created modules. App Engine Java module now depends on SDK module.

Next step will extract App Engine Java component and move it back into App Engine Java module, making Cloud SDK module suitable for any language runtimes and IDEs.


patflynn:
I'm a bit worried about how difficult it will be to detangle app-engine code from general cloud sdk mgmt code. Mainly because I suspect we may use it for non-appengine purposes, and I wouldn't want a bunch of the business logic or types to be spread between the feature modules and the cloud sdk module.

Another option to think about is to decouple first decouple java from cloud sdk code, then decouple app engine from cloud sdk code, then move the cloud sdk kernel code into core, and the app-engine cloud sdk code into the appengine module (parent of java).

e.g.

- appengine 
    - app engine cloud sdk code
          - java 
               - java cloud sdk code
google-cloud-core
     - cloud-sdk ui/mgmt/etc (nothing feature specific)

whether this is possible of course I have no idea. :)

patflynn:
fixed bad formatting of the previous comment.

ivanporty:
Also, I think it's a good idea to try to extract language/runtime agnostic features from app engine module so we can re-use useful logic from there. This structure 
* appengine
   * app engine cloud sdk code
   * java
      * java cloud sdk code

looks good. We can proceed with that once we separate core Cloud SDK module.

etanshaul:
thats a good point about the cloudsdk being a burden on the other IDEs. I wouldn't want the cloud sdk to be installed in these IDEs yet where it would never be used.

patflynn:
I'm not strongly opinionated about the cloud sdk being its own thing vs
part of core, just be careful as IntelliJ and Gradle don't seem to scale
well to large amounts of modules. Try to keep the granularity to things you
really want to keep de-coupled. e.g. one of the big reasons I setup the
modules as they are is so that we have the option of breaking out feature
verticals as their own plugins if need be.

I think I'm +1 to your plan for decoupling Cloud SDK and App Engine.  As
long as the feature specific cloud SDK implementation and interfaces (which
are exposed in feature specific commands from the SDK) exist in their
respective feature modules, I'm happy.

Typically I would swap the sequence and decouple the classes before
breaking out, but in this case it seems not to be important since its not
used anywhere else.

patflynn:
+1

etanshaul:
@patflynn regarding the CloudSdk stuff being part of core vs its own module. I think i'd prefer to take the cost of an additional module over forcing the cloud sdk on every IDE. That is unless there is a good way for us to have each IDE control how/when the cloud sdk is installed (which I don't immediately know how we'd do given our current feature vertical separation).

google-cloud-sdk/testResources/META-INF/plugin.xml
```diff
@@ -0,0 +1,22 @@
...
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<idea-plugin xmlns:xi="http://www.w3.org/2001/XInclude">
+    <name>Google Cloud SDK Tests</name>
+    <id>com.google.gct.core</id>
+    <xi:include href="/META-INF/google-cloud-core.xml" xpointer="xpointer(/idea-plugin/*)"/>
+    <xi:include href="/META-INF/google-cloud-sdk.xml" xpointer="xpointer(/idea-plugin/*)"/>
+</idea-plugin>
```
etanshaul: nit: add line break

google-cloud-sdk/testSrc/com/google/cloud/tools/intellij/appengine/java/sdk/ManagedCloudSdkUpdateServiceTest.java
```diff
@@ -110,7 +111,10 @@ public void update_scheduledNow_ifLastUpdate_elapsed() {
     managedCloudSdkUpdateService.activate();
 
     verify(managedCloudSdkUpdateService)
-        .schedule(any(), eq(0L), eq(ManagedCloudSdkUpdateService.SDK_UPDATE_INTERVAL_MS));
+        .schedule(
+            any(),
+            eq(0L),
+            ArgumentMatchers.eq(ManagedCloudSdkUpdateService.SDK_UPDATE_INTERVAL_MS));
```
etanshaul: why did you stop using the static import on eq (which is still imported)? Same for some of the other tests

google-cloud-sdk/testResources/META-INF/plugin.xml
ivanporty: done.

google-cloud-sdk/testSrc/com/google/cloud/tools/intellij/appengine/java/sdk/ManagedCloudSdkUpdateServiceTest.java
ivanporty: No idea, I think I ran `spotlessApply` and it did this for some reason. Reverted, thanks.

PR #2160, opened 2018-06-04
Remove gradle plugin facet and store gradle data on module

### LINKED ISSUE #2157: Merge gradle app engine facet into the app engine facet
Follow up to #2149 

Its a better design to merge these two facets rather than having a separate one just to indicate that the user is using gradle + app-gradle-plugin.

etanshaul:
fixes #2157 

- Gets rid of the app engine gradle plugin facet. Instead adds a gradleBuildDir property to a new module-level persistent component

Additional changes:
- require that the module be a gradle module before adding the new gradle deployment source (before it was just checking for the presence of the, now removed, gradle plugin facet. so If the user had the plugin but then removed the gradle build system entirely, the deployment source would have still been present. Now it will only be added if the module is a Gradle module).

Todo (future PRs)
- suggestions, where appropriate, to add the gradle plugin to the build
- currently, if the user adds the Gradle plugin but then removes it later, the deployment source will still be created and the action (i.e. deploy) will fail. Need to investigate ways to either avoid this, or to have good error messages.


etanshaul:
there is a bug in the way the gradle assemble task (to be run before deploy) is created. Will fix here

etanshaul:
Updated to store the gradle data on a module-level persistent component.

Still figuring out the bug mentioned above, but @ivanporty, @patflynn you can take a look now

etanshaul:
bug fixed.

ready for review now. @ivanporty ptal

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/AppEngineStandardGradleModuleComponent.java
```diff
@@ -0,0 +1,70 @@
...
+    return module.getComponent(AppEngineStandardGradleModuleComponent.class);
+  }
+
+  /** Returns, optionally, the path to the Gradle build directory. */
+  public Optional<String> getGradleBuildDir() {
+    return Optional.ofNullable(gradleBuildDir);
+  }
+
+  /** Sets the path to the Gradle build directory. */
+  public void setGradleBuildDir(String gradleBuildDir) {
```
ivanporty: Since the getter returns optional, could we mark this as `@Nullable` so we know the null could be accepted here? Same idea about `gradleModuleDir` property.

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/AppEngineStandardGradleModuleComponent.java
```diff
@@ -0,0 +1,70 @@
...
+
+  @Nullable
+  @Override
+  public AppEngineStandardGradleModuleComponent getState() {
+    return this;
+  }
+
+  @Override
+  public void loadState(@NotNull AppEngineStandardGradleModuleComponent state) {
+    gradleBuildDir = state.gradleBuildDir;
```
ivanporty: We don't need to set the second property value here on load?

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/DefaultAppEngineGradleModel.java
```diff
@@ -23,10 +23,13 @@
 
   private boolean hasAppEngineGradlePlugin;
   private String gradleBuildDir;
+  private String gradleModuleDir;
```
ivanporty: Since these fields are effectively read-only, should we also mark them final?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/AppEngineStandardGradleModuleComponent.java
etanshaul: thanks. fixed.

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/AppEngineDeploymentConfigurator.java
```diff
@@ -51,7 +51,7 @@
     deploymentSources.addAll(AppEngineUtil.createArtifactDeploymentSources(project));
     deploymentSources.addAll(AppEngineUtil.createModuleDeploymentSources(project));
 
-    if (PlatformUtils.isIdeaCommunity()) {
+    if (!PlatformUtils.isIdeaUltimate()) {
```
ivanporty: This change is safe since we won't be exposing this module to other IDEs, it's only supported to work with IDEA via app-engine-java module, correct?

etanshaul: actually, I don't know what I was thinking here. I think that I was anticipating the future where we expose this to other IDEs, but that doesn't make sense because these are Gradle sources and its Java only.

thanks for calling this out. let me change back and update the unit tests

app-engine/java/resources/META-INF/app-engine-java.xml
```diff
@@ -48,6 +48,11 @@
         </group>
     </actions>
 
+    <module-components>
+        <component>
+            <implementation-class>com.google.cloud.tools.intellij.appengine.java.facet.standard.AppEngineStandardGradleModuleComponent</implementation-class>
```
ivanporty: Could this belong to app-engine-java-gradle instead or it's tied to parent app engine java module?

etanshaul: good idea. I think that may work. Let me see.

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/GradlePluginDeploymentSourceType.java
```diff
@@ -59,7 +61,12 @@ public GradlePluginDeploymentSourceType() {
   @Override
   protected BeforeRunTask createBuildTask(Module module) {
     ExternalSystemTaskExecutionSettings settings = new ExternalSystemTaskExecutionSettings();
-    settings.setExternalProjectPath(module.getModuleFilePath());
+
+    Optional<String> gradleModuleDirOptional =
+        AppEngineStandardGradleModuleComponent.getInstance(module).getGradleModuleDir();
+
+    gradleModuleDirOptional.ifPresent(settings::setExternalProjectPath);
```
ivanporty: What would happen if this module dir is not available? Will this task still be valid?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/AppEngineStandardGradleModuleComponent.java
```diff
@@ -0,0 +1,71 @@
...
+@State(name = AppEngineStandardGradleModuleComponent.COMPONENT_NAME)
+public class AppEngineStandardGradleModuleComponent
+    implements ModuleComponent, PersistentStateComponent<AppEngineStandardGradleModuleComponent> {
+  static final String COMPONENT_NAME = "AppEngineStandardGradleModuleComponent";
+
+  public String gradleBuildDir;
+  public String gradleModuleDir;
+
+  public static AppEngineStandardGradleModuleComponent getInstance(@NotNull Module module) {
+    return module.getComponent(AppEngineStandardGradleModuleComponent.class);
```
ivanporty: Could this ever null? Or Idea will make sure the module component will be present on all modules loaded when plugin is active?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/AppEngineDeploymentConfigurator.java
etanshaul: fixed

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/GradlePluginDeploymentSourceType.java
etanshaul: the task will still be present, just with no gradle project on which to run the task. This should never happen in normal operation (since the deployment source would not have been created in the first place).

I should add some logging though.

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/AppEngineStandardGradleModuleComponent.java
etanshaul: correct (to you last statement), see the description at the top;

https://www.jetbrains.org/intellij/sdk/docs/basics/plugin_structure/plugin_components.html

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/DefaultAppEngineGradleModel.java
etanshaul: done

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/AppEngineStandardGradleModuleComponent.java
etanshaul: done

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/GradlePluginDeploymentSourceType.java
etanshaul: done

app-engine/java/resources/META-INF/app-engine-java.xml
etanshaul: this would actually introduce a circular dependency as the app-engine-java module currently uses this directly

etanshaul:
@ivanporty I think I've addressed your comments so far. ptal.

There is a minor edge case with the way the assemble before-run task is deserialized which I'm working on still, but I don't anticipate any big changes for that.

etanshaul:
fixed the issue with before-run task serialization/deserialization. Now the gradle deployment sources and the gradle assemble before run task is properly stored and restored between IDE restarts.

This PR should be good to go pending any new comments.

app-engine/java/gradle/testResources/META-INF/plugin.xml
```diff
@@ -0,0 +1,26 @@
...
+<idea-plugin xmlns:xi="http://www.w3.org/2001/XInclude">
+    <name>App Engine Java Gradle Tests</name>
+    <id>com.google.gct.core</id>
+
+    <xi:include href="/META-INF/google-cloud-core.xml" xpointer="xpointer(/idea-plugin/*)"/>
+    <xi:include href="/META-INF/app-engine-java.xml" xpointer="xpointer(/idea-plugin/*)"/>
+    <xi:include href="/META-INF/google-cloud-sdk.xml" xpointer="xpointer(/idea-plugin/*)"/>
+    <xi:include href="/META-INF/gwt-integration.xml" xpointer="xpointer(/idea-plugin/*)"/>
+    <xi:include href="/META-INF/javaee-integration.xml" xpointer="xpointer(/idea-plugin/*)"/>
+</idea-plugin>
```
ivanporty: Should we have a newline here? I remember we usually add one for this manifests.

etanshaul: done

Issue #2183, opened 2018-06-20
Refactor maven dependencies out of App Engine module


etanshaul:
Needed for the follow reasons:

- to fix #1939 
- so that App Engine functionality still works when the Maven plugin is not installed an in IDEs where the the Maven plugin is not available at all (Android Studio)

requires:
- moving deployment source collection logic to an extension point based mechanism
- creating a new appengine-maven module that depends on org.jetbrains.idea.maven

PR #2184, opened 2018-06-20
Refactor deployment source collection into extension point based providers

etanshaul:
first step towards #2183

In order to extract Maven from the App Engine module, one of the things we need to do is refactor the deployment source collection into extensions which can plug in their own deployment source collection logic. This way, once a maven module is extracted (in subsequent PRs), it can inject its logic for collecting maven sources, thus isolating maven dependencies out of the app engine module.

This PR is a one-to-one refactoring of the previous logic, just with the extension point mechanism. All of the extensions are still in the App Engine module. Also, the Maven deployment source collection is still part of the module deployment source provider. This is to keep things easier to review. The next set of PRs will continue the extraction.

Extension Point: `AppEngineDeploymentSourceProvider`
Extensions:
- `AppEngineArtifactDeploymentSourceProvider` - collects artifact deployment sources
- `AppEngineModuleDeploymentSourceProvider` - collects module deployment sources (including Maven sources, for now)
- `AppEngineGradleDeploymentSourceProvider` - collects gradle plugin based deployment sources

### DIFF
--- a/app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/AppEngineArtifactDeploymentSourceProvider.java
+++ b/app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/AppEngineArtifactDeploymentSourceProvider.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2018 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.intellij.appengine.java;
+
+import static java.util.stream.Collectors.toList;
+
+import com.google.cloud.tools.intellij.appengine.java.cloud.AppEngineArtifactDeploymentSource;
+import com.google.cloud.tools.intellij.appengine.java.cloud.AppEngineDeploymentSourceProvider;
+import com.google.cloud.tools.intellij.appengine.java.cloud.AppEngineEnvironment;
+import com.google.cloud.tools.intellij.appengine.java.facet.flexible.AppEngineFlexibleFacetType;
+import com.google.cloud.tools.intellij.appengine.java.facet.standard.AppEngineStandardFacetType;
+import com.google.cloud.tools.intellij.appengine.java.project.AppEngineProjectService;
+import com.google.common.collect.Lists;
+import com.intellij.facet.FacetManager;
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.module.ModuleManager;
+import com.intellij.openapi.project.Project;
+import com.intellij.packaging.artifacts.Artifact;
+import com.intellij.packaging.artifacts.ArtifactPointerManager;
+import com.intellij.packaging.impl.artifacts.ArtifactUtil;
+import com.intellij.remoteServer.configuration.deployment.DeploymentSource;
+import java.util.Collection;
+import java.util.List;
+import org.jetbrains.annotations.NotNull;
+
+/** An {@link AppEngineDeploymentSourceProvider} that collects artifact-based deployment sources. */
+public class AppEngineArtifactDeploymentSourceProvider
+    implements AppEngineDeploymentSourceProvider {
+
+  /**
+   * Collects a list of artifact deployment sources available for deployment to App Engine.
+   *
+   * <p>Artifacts either target the standard or the flexible environment. All standard artifacts are
+   * added. Flexible artifacts are only added if there are no other standard artifacts associated
+   * with the same module.
+   *
+   * @return a list of {@link AppEngineArtifactDeploymentSource}
+   */
+  @Override
+  public List<DeploymentSource> getDeploymentSources(@NotNull Project project) {
+    List<DeploymentSource> sources = Lists.newArrayList();
+    AppEngineProjectService projectService = AppEngineProjectService.getInstance();
+
+    for (Module module : ModuleManager.getInstance(project).getModules()) {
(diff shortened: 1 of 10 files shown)


etanshaul:
will add test coverage here for `AppEngineArtifactDeploymentSourceProvider` and `AppEngineModuleDeploymentSourceProvider`. The equivalent coverage was missing before.

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/AppEngineArtifactDeploymentSourceProvider.java
```diff
@@ -0,0 +1,101 @@
...
+    AppEngineProjectService projectService = AppEngineProjectService.getInstance();
+
+    for (Module module : ModuleManager.getInstance(project).getModules()) {
+      FacetManager facetManager = FacetManager.getInstance(module);
+      if (facetManager.getFacetByType(AppEngineStandardFacetType.ID) != null
+          || facetManager.getFacetByType(AppEngineFlexibleFacetType.ID) != null) {
+        final AppEngineEnvironment environment =
+            projectService
+                .getModuleAppEngineEnvironment(module)
+                .orElseThrow(() -> new RuntimeException("No environment."));
```
ivanporty: Will this be handled correctly? Might interfere with other deployment providers even if something is wrong here in particular.

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/AppEngineModuleDeploymentSourceProvider.java
```diff
@@ -0,0 +1,101 @@
...
+    boolean hasStandardModules = false;
+
+    for (Module module : ModuleManager.getInstance(project).getModules()) {
+      FacetManager facetManager = FacetManager.getInstance(module);
+      if (facetManager.getFacetByType(AppEngineStandardFacetType.ID) != null
+          || facetManager.getFacetByType(AppEngineFlexibleFacetType.ID) != null) {
+        AppEngineEnvironment environment =
+            projectService
+                .getModuleAppEngineEnvironment(module)
+                .orElseThrow(() -> new RuntimeException("No environment."));
```
ivanporty: Same question on runtime exception effect as in previous comment.

app-engine/java/testSrc/com/google/cloud/tools/intellij/appengine/java/cloud/AppEngineArtifactDeploymentSourceProviderTest.java
```diff
@@ -0,0 +1,183 @@
...
+        .isEqualTo(environment);
+  }
+
+  /**
+   * Adds the {@link AppEngineStandardFacet} to the module, associates the module with the standard
+   * environment, and creates a new artifact associated with the module output.
+   *
+   * @param module
+   */
+  private void makeStandard(Module module) {
```
ivanporty: Could it make sense to add those as a general test utility methods to `ModuleTestUtils`? They look clean and straighforward, might be useful for other tests too.

app-engine/java/testSrc/com/google/cloud/tools/intellij/appengine/java/cloud/AppEngineArtifactDeploymentSourceProviderTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+    assertThat(deploymentSource1 instanceof AppEngineArtifactDeploymentSource).isTrue();
+    assertThat(deploymentSource2 instanceof AppEngineArtifactDeploymentSource).isTrue();
+    assertThat(((AppEngineArtifactDeploymentSource) deploymentSource1).getEnvironment())
+        .isEqualTo(AppEngineEnvironment.APP_ENGINE_STANDARD);
+    assertThat(((AppEngineArtifactDeploymentSource) deploymentSource2).getEnvironment())
+        .isEqualTo(AppEngineEnvironment.APP_ENGINE_FLEX);
+  }
+
+  private void assertDeploymentSourcesPresent(AppEngineEnvironment environment) {
```
ivanporty: Could we expand this method a bit to avoid almost identical code in lines 91:104?

etanshaul: done

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/AppEngineModuleDeploymentSourceProvider.java
etanshaul: done

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/AppEngineArtifactDeploymentSourceProvider.java
etanshaul: done

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/AppEngineModuleDeploymentSourceProvider.java
```diff
@@ -61,18 +61,18 @@
...
-                .getModuleAppEngineEnvironment(module)
-                .orElseThrow(() -> new RuntimeException("No environment."));
+            projectService.getModuleAppEngineEnvironment(module).orElse(null);
 
-        if (ModuleType.is(module, JavaModuleType.getModuleType())
-            && projectService.isJarOrWarMavenBuild(module)) {
-          moduleDeploymentSources.add(
-              createMavenBuildDeploymentSource(project, module, environment));
-        }
+        if (environment != null) {
```
etanshaul: just fyi I used a null check here instead of an `ifPresent` lambda due to the `hasStandardModules` side effect

etanshaul:
on second thought. the last change that I made to `AppEngineModuleDeploymentSourceProvider` includes a behavior change:

Previously, if the environment was empty, no sources would be collected (since it threw a RuntimeException). Now, it will collect the file path based source.

I think the correct thing here is to return no sources because the file path based source only works for flex. Its possible that App Engine standard facet exists, but the environment is null (due to misconfiguration), so we wouldn't want to present the file path based deployment source.

Will update

etanshaul:
@ivanporty updated `AppEngineDeploymentSourceProvider` logic slightly and added more tests. ptal.

PR #2185, opened 2018-06-20
Extend service account environment variable code to allow including any runtime configurations.

### LINKED ISSUE #2148: Automatically add cloud api related env vars to more run configurations
Currently we only do this for App Engine standard local run configs.

ivanporty:
Fixes #2148.

Although this looks like a relatively large PR, most of the code is re-used from the previous AppEngine-only based implementation, but extended to be able to use it for any runtime configurations via extension points.

The following is done:

* `ServiceAccountKeyUltimateDisplayDialog` is moved into `ServiceAccountKeyDisplayDialog` into `google-cloud-apis` without form changes and stripped of any logic of finding and updating runtime configurations.
* `ServiceAccountKeyDialogService` is removed, one common dialog is used now.
* Extension point interface `ServiceAccountKeyRuntimeConfigurationProvider` for collecting runtime configurations and adding environment variables to them is added to `google-cloud-apis`
* App Engine Local Run configuration env variables logic is moved into `AppEngineLocalRunServiceAccountKeyRuntimeConfigurationProvider` extension point.
* Plain Java runtime configuration support is added with `JavaApplicationServiceAccountKeyRuntimeConfigurationProvider` extension point.
* Unit test for `ServiceAccountKeyUltimateDisplayDialog` is split to UI related tests for common dialog and environment variables logic for new extension point.

Can be now used and tested on both CE (any runtime target with `main()`) and UE (both plain Java and app engine local run targets).

Example: list of configurations where both App Engine Local Run and plain Java `main()` application is present:
service-accounts-with-main

After adding variables, plain Java application has new environment variables added:
env-variables-app-main


etanshaul:
starting to look at this now. 

could you show some screenshots so we can get a feel for how this works (adding env vars to non-app-engine run configs)?

google-cloud-apis/src/com/google/cloud/tools/intellij/cloudapis/JavaApplicationServiceAccountKeyRuntimeConfigurationProvider.java
```diff
@@ -0,0 +1,83 @@
...
+import com.intellij.execution.util.EnvironmentVariable;
+import com.intellij.openapi.project.Project;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import org.jetbrains.annotations.NotNull;
+
+public class JavaApplicationServiceAccountKeyRuntimeConfigurationProvider
```
etanshaul: javadoc

etanshaul: trying to understand what kind of run configs this will apply to. typical servlet container run configs (jetty, tomcat, etc) will be of this type? do all of these java run configs have injectable environment variables (with user visible UI)?

ivanporty:
Added sample screenshots with both plain Java application and app engine local run configurations present for service account variables dialog.

google-cloud-apis/src/com/google/cloud/tools/intellij/cloudapis/JavaApplicationServiceAccountKeyRuntimeConfigurationProvider.java
ivanporty: This is plain Java application run configuration only as I understand. You can see the example of a screenshot. I believe all others (including app servers you mentioned) would have other `ConfigurationType` for which we obtain run configurations (`runManager.getConfigurationSettingsList(ApplicationConfigurationType.getInstance())`). We can add more extension points if we thing this makes sense although in containers/K8s world having support for plain `main()` applications should be good enough?

etanshaul: what about Maven / Gradle run configs. Will this work with them?

I'm thinking about spring boot apps as an example here since many users will use the spring-boot build plugins to run locally.

ivanporty: I will check those, not sure - the `ApplicationConfigurationType` is the one that launches `main()` class, i.e. the most fundamental Java run type. We can definitely add as many additional extension points with more configs as we need in subsequent PRs.

google-cloud-apis/src/com/google/cloud/tools/intellij/cloudapis/ServiceAccountKeyRuntimeConfigurationProvider.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Application extension point that provides the option to extend list of supported runtime
+ * configurations to add service account related environment variables to.
+ */
+public interface ServiceAccountKeyRuntimeConfigurationProvider {
```
etanshaul: Is it a "RuntimeConfiguration" provider, or a "RunConfiguration" provider? I thought it was the later.

ivanporty: good point, will change.

ivanporty:
I noticed we still have App Engine specific wording in the service account final dialog, changed to more neutral language, here is an updated screenshot:

servce-accnt-env-new-wording

google-cloud-apis/src/com/google/cloud/tools/intellij/cloudapis/ServiceAccountKeyRuntimeConfigurationProvider.java
etanshaul: also should we focus it on "CloudApis" instead of "ServiceAccount" since these env vars aren't necessarily just service account related (cloud project for example). so something like `CloudApiRunConfigurationProvider`?

google-cloud-apis/src/com/google/cloud/tools/intellij/cloudapis/ServiceAccountKeyRuntimeConfigurationProvider.java
```diff
@@ -0,0 +1,58 @@
...
+      new ExtensionPointName<>("com.google.gct.cloudapis.serviceAccountRuntimeConfiguration");
+
+  /**
+   * Collects and returns list of run configurations where service account environment variables can
+   * be added.
+   *
+   * @param project IDE project to collect configurations from.
+   * @return List of run configurations.
+   */
+  List<RunnerAndConfigurationSettings> getRunConfigurationsForServiceAccount(
```
etanshaul: if you rename the interface based on my suggestions, I'd rename this accordingly too

ivanporty:
Should be good to go.

PR #2187, opened 2018-06-21
Separate Maven deployment source collection into new extension

etanshaul:
more progress towards #2183 

- creates a new `AppEngineMavenDeploymentSourceProvider` that collects only maven-based sources.
- removes the maven collection from the previously named `AppEngineModuleDeploymentSourceProvider`. Renamed this extension to `AppEngineUserSpecifiedPathDeploymentSourceProvider` since this is all its collecting now


etanshaul:
next steps:

- isolate all maven dependent logic into maven classes to be moved to a new maven module
- invert usage of maven logic from appengine -> maven to using extension points (like already done for deployment sources)
- factor out an optional maven module

PR #2190, opened 2018-06-22
Invert app engine maven dependency direction

etanshaul:
yet more progress towards #2183 

Previously, the App Engine standard facet had a dependency on the Maven integration. This PR inverts this by creating an extension point to cut that dependency. This will allow us to, in the next PR or two, to create a new Maven module and remove all ties from app-engine to Maven.


etanshaul:
To elaborate:

The App Engine standard facet UI which displayed the managed libraries (servlet, endpoints, etc) only displayed these libraries for native project (i.e no build system). So the facet editor had to explicitly check if the module had maven. This created the direct dependency from app-engine to maven.

This creates an extension point `AppEngineStandardLibraryManager#isSupported` which allows extensions to plugin in if the libraries are supported.

PR #2199, opened 2018-06-27
Separate Cloud API from Java/Maven: prepare initial sub-module structure, Java sub-module

ivanporty:
Progress towards #2182 and #2193.

First step in refactoring Cloud API functionality to separate it from Java and Maven dependencies. This is the step by step plan:
1. Introduce `java-apis` module and its child `maven-apis` module where all java functionality and maven functionality will reside (this PR). Core `google-cloud-apis` will not be dependent on them and will provide set of services and extension points to support extension UI and actions.
2. Make `google-cloud-tools-plugin` optionally depend on new modules, depending on presence of `java` and `maven` IDE plugins (this PR).
2. Implement `java-apis` module (this PR). For now only includes Java `main()` run configuration extension point.
3. Move independent Maven classes and tools into `maven-apis` module (next PR).
4. Add UI extension mechanism and move Maven UI (BOM) into `maven-apis` module. (3rd PR)
5. Refactor unit tests (last PR).

Issue #2193, opened 2018-06-25
Don't limit the library features to Maven


etanshaul:
Despite currently only having maven bindings for adding client libraries to the user's build system, the libraries feature has many other things that are useful aside from this.

We can open up this UI to non-maven builds (and ultimately non-java IDEs) with some simple wording changes (so that it doesn't appear as if a client library will be added when it's not), allowing users to take advantage of API enablement, service accounts, and env var injection.

Eventually we can add build system specific bindings (even if its just copy/paste snippets).

etanshaul:
this is complete. cloud api management is available even without a maven project

Issue #2182, opened 2018-06-20
Cloud APIs: consider adding support for enabling APIs and service accounts for all JB IDEs


ivanporty:
Working on #2148, it looks like most of Cloud APIs service account and API enablement code could quite easily ported to other IDEs since it's almost language and IDE agnostic. We can do this in a few steps:

* Extract Java related code and support, including Maven and BOM inspections, into a separate sub-module of `google-cloud-aps` module and make it dependent on `java`, removing `java` requirement from the core API module.
* Add extension points where necessary to extend core functionality with runtime specifics, like Java UI forms (BOM version and module selectors) and post-selection and post-enable actions (adding BOM and dependencies for Maven.)
* Implement extension point for adding service account environment variables into arbitrary runtime configurations with environment variables support - this is already being implemented as part of #2148.

If we do this, we can add very valuable functionality of dealing with enabling APIs, downloading JSON keys, and adding service account environment variables for all IDEs.

etanshaul:
I agree this would be awesome.

The other thing is that currently all of the API enablement / service account management stuff is within the context of adding client libraries to the build. That's why, currently, we don't even allow this flow for Gradle users. So another change would be to decouple these somehow (so that we can get the API functionality for IDEs that do not yet have the language specific binding stuff).

PR #2201, opened 2018-06-27
Separate Cloud API from Java/Maven: move non-UI Maven classes and services into their own sub-module

ivanporty:
Progress towards #2182 and #2193, also see detailed plan in #2199.

Moves non-UI, Maven-related independent classes and services into `maven-apis` sub-module.
Since many parts in the UI and tests in the core `google-cloud-apis` still use them, the core API module will for the time being depend on new Maven sub-module. Once all dependencies are cleanly cut using extension points and services, this dependency on maven sub-module will be removed.


patflynn:
have you all thought about the trade-offs between the accuracy of breaking up modules to this level of granularity vs keeping the modules coarse?  I'm a bit worried that getting super fine grained is going to be  quite cumbersome.

etanshaul:
if the modules are coarse, then we lose the ability to expose features to more IDEs. In other words, if we don't break the cloud-api module down to java and non-java modules, then we cannot expose this to PyCharm for example. or are you suggesting to reconsider whether that's worth doing given the tradeoff?

ivanporty:
In this particular case this immediately gives us significant positive results:
* allows Cloud APIs management to be used without Maven based project
* at the end of refactoring (which shouldn't be quite complicated), we will expose Cloud APIs/service accounts for all other IDEs.
* using extension points, we can add more run configurations for other languages like Go and Python.

I think in this case this looks justified?

etanshaul:
one trade-off we _could_ make here however is to not break apart cloud-api-java and cloud-api-maven into separate modules, and just require that the maven plugin be configured. I'm not saying we should, just that this is a place where we could trim down on the modules

etanshaul:
we would native projects them only if they don't have the maven plugin installed right? whether the project itself is maven/gradle is independent of whether the Maven plugin is installed in there IDE

patflynn:
I didn't think that was true, I believe you can decouple the idea of a IntelliJ module from a plugin fragment.  e.g. one module can have all app engine feature, while optional depends plugin fragments only pull in the features from the module that match the dependency.   The downside to this is that the development classpath for the module won't 100% match what's actually available to the code at runtime which is obviously bad.  

I thought I'd noticed JB seems to follow this convention in the CE examples I've looked at, and looking at the explosion of modules that could come out of creating a submodule for every optional depends I can imagine why. 

I'm actually not really opinionated about this, but I want to make sure we make a conscious decision understanding the trade-offs.

patflynn:
Just to emphasize: I'm very supportive of decoupling the API and secret management from Maven based projects.  I'm just wondering how we should organize our modules to keep us productive.

etanshaul:
@patflynn I see what you are saying now.Mind elaborating on your concern for having fine grained modules?

With the approach you mentioned of keeping optionally included classes in the same module we run into a hazard where a developer can access code that won't be available at runtime, leading to errors that we potentially may not catch right away. That seems like a worse scenario to me.

google-cloud-apis/java-apis/maven-apis/resources/META-INF/cloud-apis-java-maven.xml
```diff
@@ -15,4 +15,33 @@
   -->
 
 <idea-plugin>
+    <depends>org.jetbrains.idea.maven</depends>
+
+    <project-components>
+        <component>
+            <implementation-class>com.google.cloud.tools.intellij.cloudapis.maven.CloudLibraryProjectState
```
etanshaul: Should we rename this to `CloudLibraryMavenProjectState`?

ivanporty: sounds good.

Issue #2191, opened 2018-06-22
User reported FileNotFoundException


etanshaul:
Plugin version 18.6.1
IDEA 2018.1

```
com.google.cloud.tools.appengine.api.AppEngineException: java.nio.file.NoSuchFileException: /***/appengine-web.xml
	at com.google.cloud.tools.appengine.cloudsdk.CloudSdkAppEngineDevServer1.isJava8(CloudSdkAppEngineDevServer1.java:224)
	at com.google.cloud.tools.appengine.cloudsdk.CloudSdkAppEngineDevServer1.run(CloudSdkAppEngineDevServer1.java:113)
	at com.google.cloud.tools.intellij.appengine.java.cloud.executor.AppEngineStandardRunTask.execute(AppEngineStandardRunTask.java:79)
	at com.google.cloud.tools.intellij.appengine.java.cloud.executor.AppEngineExecutor.run(AppEngineExecutor.java:42)
	at com.google.cloud.tools.intellij.appengine.java.ultimate.server.run.CloudSdkStartupPolicy$1$1.createProcessHandler(CloudSdkStartupPolicy.java:161)
	at com.intellij.javaee.run.execution.PatchedLocalState.startJ2EEProcess(PatchedLocalState.java:76)
	at com.intellij.javaee.run.execution.J2EEProcessHandlerWrapper.lambda$new$0(J2EEProcessHandlerWrapper.java:89)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at com.intellij.util.concurrency.BoundedTaskExecutor$2.run(BoundedTaskExecutor.java:212)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.nio.file.NoSuchFileException: /***/appengine-web.xml
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
[...]

Issue #1948, opened 2018-03-14
Create web root in canonical locations if missing for Maven and Gradle


etanshaul:
This is a follow up to https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1903

If no web root exists in the project (as could be common with frameworks like spring boot). We need to create one in order to properly place the `appengine-web.xml` when adding App Engine standard framework support.

The root should be created according to the standard directory layouts for Maven or Gradle. For example, if it is a Maven project:

`src/main/webapp`

patflynn:
if the web facet is present we should use the web facet web resource location and create the directory if it doesn't exist. 

@saturnism found that mavne spring boot apps seem to get this facet automatically but with a pointer to the maven configured web resource folder even if it hasn't materialized yet.

etanshaul:
agree. same thing happens for gradle

etanshaul:
however, just to point out, if the user is using Community, the web facet will not be present. In this case it should still do the right thing for maven/gradle projects.

patflynn:
+1. For community,  I wonder if if we can query the web resource root directly out of the gradle and/or maven integration? Presumably that's how ultimate configures the facet.

etanshaul:
yeah, I don't know where Ultimate gets it from..whether its querying the maven/gradle build or if it just uses a well-known location. Ill check

etanshaul:
Turns out IDEA does fetch the web resource directory directly from the maven/gradle integration. For maven the strategy is as follows:

1) Using `MavenProject`, get the `maven-war-plugin` config
2) With the war plugin config, find the "warSourceDirectory" value
3) a. if not null -> use this one
    b. if null -> use the default "src/main/webapp" folder

Gradle is similar (albeit the method for fetching info from Gradle is much different).

Approach for generating appengine-web.xml properly for Community maven/gradle projects (for Ultimate our job is simpler due to the web facet - already complete in #2203):
- Create a new `AppEngineWebXmlDirectoryProvider` extension point
- Create `MavenAppEngineWebXmlDirectoryProvider` and `GradleAppEngineWebXmlDirectoryProvider` extensions in their respective modules
- In each extension, use the same approach as described above for locating the web resource directory to for placing the `appengine-web.xml`

patflynn:
nice!

PR #2203, opened 2018-06-28
Fixes appengine-web.xml generation in Ultimate when the web resource dir does not exist

etanshaul:
Progress towards #1948 and #2191 

Updates the `appengine-web.xml` parent directory discovery logic for Ultimate Edition as follows:
- Added: If the web resource directory has a path specified, but that path does not map to a valid directory (which is the case when you import a vanilla Maven or Gradle Spring boot project), it will create that directory recursively (including all of the parent directories, under the project)
- Fixed: If there are no WEB-INF folders found within the available web-roots, instead of just choosing the first one to try and create a WEB-INF, now it will first try and find one that is a valid directory

In addition to new tests, I also cleaned up existing ones as they were pretty hard to follow before.

Example of what happens when importing Spring Boot Maven project now:

1) create a plain Maven WAR Spring Boot project (I used the initializr):
screen shot 2018-06-27 at 9 10 43 pm

2) project is created. it is not yet an app-engine project:
screen shot 2018-06-27 at 9 11 40 pm

3) Notice the defaulted web resource directory on the web facet. This directory does not yet exist:
screen shot 2018-06-27 at 9 12 43 pm

4) Add appengine-standard support via the cloud tools menu. Notice that the appengine-web.xml is created in the correct location (previously it was not created at all):
screen shot 2018-06-27 at 9 14 01 pm

Now deploy and local run work right away, avoiding the error in #2191

### DIFF
--- a/app-engine/java/ultimate/src/com/google/cloud/tools/intellij/appengine/java/ultimate/impl/AppEngineStandardUltimateWebIntegration.java
+++ b/app-engine/java/ultimate/src/com/google/cloud/tools/intellij/appengine/java/ultimate/impl/AppEngineStandardUltimateWebIntegration.java
@@ -96,23 +96,27 @@ public class AppEngineStandardUltimateWebIntegration extends AppEngineStandardWe
     }
 
     for (WebRoot webRoot : webRoots) {
-      VirtualFile parent = webRoot.getFile();
-      if (parent != null) {
-        if (WEB_INF.equals(parent.getName())) {
-          return parent;
+      VirtualFile webRootDir = webRoot.getFile();
+      if (webRootDir != null) {
+        if (WEB_INF.equals(webRootDir.getName())) {
+          return webRootDir;
         }
 
-        VirtualFile child = parent.findChild(WEB_INF);
-        if (child != null) {
-          return child;
+        VirtualFile webInfDir = webRootDir.findChild(WEB_INF);
+        if (webInfDir != null) {
+          return webInfDir;
         }
       }
     }
 
     try {
-      VirtualFile webRootFile = webRoots.get(0).getFile();
-      if (webRootFile != null) {
-        return VfsUtil.createDirectoryIfMissing(webRootFile, WEB_INF);
+      VirtualFile webRootDir = webRoots.get(0).getFile();
+      if (webRootDir != null) {
+        return VfsUtil.createDirectoryIfMissing(webRootDir, WEB_INF);
+      } else {
+        // There is a webroot, but the directory does not exist
+        VirtualFile dir = VfsUtil.createDirectories(webRoots.get(0).getDirectoryUrl().substring(7));
+        return VfsUtil.createDirectoryIfMissing(dir, WEB_INF);
       }
     } catch (IOException ioe) {
       LOG.info(ioe);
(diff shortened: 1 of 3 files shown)


etanshaul:
@patflynn fyi

etanshaul:
note there are still a couple things left to do to fully close the two mentions issues:

- have a appengine-web.xml generation strategy for Community (where we cannot rely on the web facet)
- have a robust fix for #2191 that involves proper error messaging to the user (this PR doesn't deal with handling the error, just goes a step towards avoid it in the first place)

patflynn:
@etanshaul wow, that was fast. :+1: 

@saturnism FYI

app-engine/java/ultimate/src/com/google/cloud/tools/intellij/appengine/java/ultimate/impl/AppEngineStandardUltimateWebIntegration.java
```diff
@@ -96,26 +107,38 @@ public VirtualFile suggestParentDirectoryForAppEngineWebXml(
...
+          return webInfDir;
         }
       }
     }
 
     try {
-      VirtualFile webRootFile = webRoots.get(0).getFile();
-      if (webRootFile != null) {
-        return VfsUtil.createDirectoryIfMissing(webRootFile, WEB_INF);
+      Optional<VirtualFile> webRootDir =
```
ivanporty: so here we're looking for a web root file that exists (physically) on the disk but doesn't contain WEB-INF, after initial loop on `webRoots`, correct?

app-engine/java/ultimate/src/com/google/cloud/tools/intellij/appengine/java/ultimate/impl/AppEngineStandardUltimateWebIntegration.java
```diff
@@ -96,26 +107,38 @@ public VirtualFile suggestParentDirectoryForAppEngineWebXml(
...
+          webRoots.stream().map(WebRoot::getFile).filter(Objects::nonNull).findFirst();
+
+      if (webRootDir.isPresent()) {
+        return VfsUtil.createDirectoryIfMissing(webRootDir.get(), WEB_INF);
+      } else {
+        // There is at least one web-root path, but none that exist, select the first and create it
+        String presentableUrl = webRoots.get(0).getPresentableUrl();
```
ivanporty: What does this return exactly (`getPresentableUrl()`), absolute path in the file system?

etanshaul: yes it does

app-engine/java/ultimate/src/com/google/cloud/tools/intellij/appengine/java/ultimate/impl/AppEngineStandardUltimateWebIntegration.java
etanshaul: exactly. previously it was just grabbing the first one which may or may not have physically existed, which was not optimal

ivanporty: could we add a small comment here to define a bit better what it means when a VirtualFile exists - meaning the root exists and we don't need to create it?

app-engine/java/ultimate/src/com/google/cloud/tools/intellij/appengine/java/ultimate/impl/AppEngineStandardUltimateWebIntegration.java
```diff
@@ -96,26 +107,38 @@ public VirtualFile suggestParentDirectoryForAppEngineWebXml(
...
+        if (presentableUrl != null) {
+          VirtualFile newWebRoot = VfsUtil.createDirectories(presentableUrl);
+
+          if (newWebRoot != null) {
+            return VfsUtil.createDirectoryIfMissing(newWebRoot, WEB_INF);
```
ivanporty: identical line with 128, could we use the same path possibly?

etanshaul: done

PR #2208, opened 2018-06-28
Separate Cloud API from Java/Maven: move Maven UI components into maven-apis sub-module

ivanporty:
Progress towards #2182 and #2193, also see detailed plan in #2199.

This PR looks large but it's mostly code refactoring. In this PR:
1. Extension point `CloudApiUiExtension` added. This extension point allows hooking into Cloud API UI and provide more information for selected cloud library. It will be moved to core `google-cloud-apis` module once dependencies are finalized.
2. `BomComboBox` encapsulates BOMs population and rendering.
3. `MavenCloudApiUiExtension` now contains logic for getting Cloud library maven versions and adding maven-related docs links.
4. `CloudApiUiPresenter` is a class that allows extensions to get access to core UI.
5. Version label unit tests are moved in `MavenCloudApiUiExtensionTest` and now do not depend on version label and details panel.

Last PR will:
1. Remove all dependencies to Maven from core and add Maven UI components dynamically.
2. Move UI Maven unit tests into maven module.
3. Reverse dependencies and make `google-cloud-apis` standalone module suitable for all other JB IDEs.

### DIFF
--- a/google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiExtension.java
+++ b/google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiExtension.java
@@ -35,7 +35,7 @@ public interface CloudApiUiExtension {
 
   /**
    * Called when add cloud libraries dialog is created and opened and base UI is ready. At this
-   * point extension point is active.
+   * point extension point is active and custom UI components can be injected.
    */
   Collection<JComponent> createCustomUiComponents();
 
@@ -53,5 +53,5 @@ public interface CloudApiUiExtension {
    *
    * @param module Currently selected module.
    */
-  void onModuleSelected(Module module);
+  void onModuleSelection(Module module);
 }
--- a/google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiExtension.java
+++ b/google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiExtension.java
@@ -19,7 +19,9 @@ package com.google.cloud.tools.intellij.cloudapis;
 import com.google.cloud.tools.libraries.json.CloudLibrary;
 import com.intellij.openapi.extensions.ExtensionPointName;
 import com.intellij.openapi.module.Module;
-import org.jetbrains.annotations.NotNull;
+import java.util.Collection;
+import javax.swing.JComponent;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * Extension point for Cloud APIs, allowing access to UI and getting updates on cloud library
@@ -34,10 +36,8 @@ public interface CloudApiUiExtension {
   /**
    * Called when add cloud libraries dialog is created and opened and base UI is ready. At this
    * point extension point is active.
-   *
-   * @param uiPresenter Presenter to access cloud API base UI.
    */
-  void init(@NotNull CloudApiUiPresenter uiPresenter);
+  Collection<JComponent> createCustomUiComponents();
 
   /**
    * Callback on change in currently selected cloud library.
@@ -45,7 +45,8 @@ public interface CloudApiUiExtension {
    * @param currentCloudLibrary Cloud library selected or null if user de-selected library.
    * @param currentBomVersion BOM version. TODO:// to be removed, move to maven module.
    */
-  void onCurrentCloudLibrarySelected(CloudLibrary currentCloudLibrary, String currentBomVersion);
+  void onCloudLibrarySelection(
+      @Nullable CloudLibrary currentCloudLibrary, String currentBomVersion);
 
   /**
    * Callback on module selection change.
--- a/google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiExtension.java
+++ b/google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiExtension.java
@@ -21,14 +21,36 @@ import com.intellij.openapi.extensions.ExtensionPointName;
 import com.intellij.openapi.module.Module;
 import org.jetbrains.annotations.NotNull;
 
+/**
+ * Extension point for Cloud APIs, allowing access to UI and getting updates on cloud library
+ * selection, module selection, and final library selection confirmation. See also {@link
+ * CloudApiUiPresenter}.
+ */
 // TODO: move to core cloud API once dependency is inverted.
 public interface CloudApiUiExtension {
   ExtensionPointName<CloudApiUiExtension> EP_NAME =
       new ExtensionPointName<>("com.google.gct.cloudapis.cloudApiUiExtension");
 
+  /**
+   * Called when add cloud libraries dialog is created and opened and base UI is ready. At this
+   * point extension point is active.
+   *
+   * @param uiPresenter Presenter to access cloud API base UI.
+   */
   void init(@NotNull CloudApiUiPresenter uiPresenter);
 
+  /**
+   * Callback on change in currently selected cloud library.
+   *
+   * @param currentCloudLibrary Cloud library selected or null if user de-selected library.
+   * @param currentBomVersion BOM version. TODO:// to be removed, move to maven module.
+   */
   void onCurrentCloudLibrarySelected(CloudLibrary currentCloudLibrary, String currentBomVersion);
 
+  /**
+   * Callback on module selection change.
+   *
+   * @param module Currently selected module.
+   */
   void onModuleSelected(Module module);
 }
--- a/google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiExtension.java
+++ b/google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiExtension.java
@@ -14,15 +14,21 @@
  * limitations under the License.
  */
 
-package com.google.cloud.tools.intellij.cloudapis.maven;
+package com.google.cloud.tools.intellij.cloudapis;
 
 import com.google.cloud.tools.libraries.json.CloudLibrary;
+import com.intellij.openapi.extensions.ExtensionPointName;
 import com.intellij.openapi.module.Module;
 import org.jetbrains.annotations.NotNull;
 
 // TODO: move to core cloud API once dependency is inverted.
 public interface CloudApiUiExtension {
+  ExtensionPointName<CloudApiUiExtension> EP_NAME =
+      new ExtensionPointName<>("com.google.gct.cloudapis.cloudApiUiExtension");
+
   void init(@NotNull CloudApiUiPresenter uiPresenter);
+
   void onCurrentCloudLibrarySelected(CloudLibrary currentCloudLibrary, String currentBomVersion);
+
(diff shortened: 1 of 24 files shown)


google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiExtension.java
```diff
@@ -0,0 +1,56 @@
...
+   */
+  void init(@NotNull CloudApiUiPresenter uiPresenter);
+
+  /**
+   * Callback on change in currently selected cloud library.
+   *
+   * @param currentCloudLibrary Cloud library selected or null if user de-selected library.
+   * @param currentBomVersion BOM version. TODO:// to be removed, move to maven module.
+   */
+  void onCurrentCloudLibrarySelected(CloudLibrary currentCloudLibrary, String currentBomVersion);
```
etanshaul: nit: onCloudLibrarySelection ?

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiExtension.java
```diff
@@ -0,0 +1,56 @@
...
+  ExtensionPointName<CloudApiUiExtension> EP_NAME =
+      new ExtensionPointName<>("com.google.gct.cloudapis.cloudApiUiExtension");
+
+  /**
+   * Called when add cloud libraries dialog is created and opened and base UI is ready. At this
+   * point extension point is active.
+   *
+   * @param uiPresenter Presenter to access cloud API base UI.
+   */
+  void init(@NotNull CloudApiUiPresenter uiPresenter);
```
etanshaul: what happens if the user fails to call init? will things be broken?

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiPresenter.java
```diff
@@ -0,0 +1,44 @@
...
+import javax.swing.Icon;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Presenter for core Cloud API UI, allows accessing UI points for libraries for extension points to
+ * update, including documentation links and version info.
+ */
+// TODO: move to core cloud API once dependency is inverted.
+public interface CloudApiUiPresenter {
```
etanshaul: I wonder if there is any value in making this more generic - i.e. have the extensions just provide custom UI panels? I'm thinking about the future if we support custom languages which might need more custom tailored UI.

etanshaul: also are these really the only two things we need to extract? what about source, api reference etc.?

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/maven/BomComboBox.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.List;
+import java.util.Optional;
+import javax.swing.JComboBox;
+import javax.swing.JList;
+
+/**
+ * Combobox that populates list of available BOM versions and selects currently active BOM version
+ * from the given project/module.
+ */
+public class BomComboBox extends JComboBox<String> {
```
etanshaul: just fyi that this on potentially will span more than just maven (gradle for example can support BOMs directly in the latest versions).

etanshaul: initially I think its ok to just have it for Maven though

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiExtension.java
ivanporty: init() will be called by core Cloud API module, and UI presenter will be passed in. This should not depend on sub-module. If you think this is not clear from javadoc or method name, I'll think a bit how to improve.

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/maven/BomComboBox.java
ivanporty: I agree, this looks like a more generic class - but still kind of coupled to Maven services and classes for now. We can always split this and use for Gradle too later.

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiPresenter.java
ivanporty: So next (hopefully last) PR will add custom UI panels for Cloud API selector panel, you're correct - so Maven module will "inject" BOM selection. As for source and API reference, see methods of CloudUiPresenter - it allows adding more links for a selected cloud library.

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiExtension.java
etanshaul: this pattern made me thing of this: [http://go/tott-523](http://go/tott-523)

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiPresenter.java
etanshaul: you mean `CloudApiUiPresenter`? Probably missing something but I see two methods: `addCloudLibraryDocumentationLink`, and `updateCloudLibraryVersionLabel`

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiExtension.java
ivanporty: Agree, not optimal. How about this:
1. Make `CloudApiUiPresenter` a service available from usual IDE service search.
2. Ensure all methods called on extension point have access to a valid UI presenter when callbacks are called.
2. Replace `init()` with `createCustomUiComponents()` (implemented in the next PR) - where UI components and necessary initialization is done by child sub-module?

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiPresenter.java
ivanporty: Correct, this is the one which allows adding more documentation links to library details panel - `addCloudLibraryDocumentationLink`

etanshaul: ah ok. its the name that confused me then. for one, the ui has a link called 'documentation' so thats what I thought it was for. How about just addCloudLibraryLinks?

ivanporty: Sounds good, I can also probably make this method accept collection of Optionals like previous code did.

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiExtension.java
etanshaul: sgtm

ivanporty:
@etanshaul added suggested name changes and redesign of flow for `CloudUiPresenter` - now it's a service and it's not passed into extension. PTAL.

google-cloud-apis/src/com/google/cloud/tools/intellij/cloudapis/DefaultCloudApiUiPresenter.java
```diff
@@ -45,15 +48,16 @@ public void updateCloudLibraryVersionLabel(@Nullable String text, @Nullable Icon
...
    */
-  void init(GoogleCloudApiSelectorPanel cloudApiSelectorPanel) {
+  void init(@NotNull GoogleCloudApiSelectorPanel cloudApiSelectorPanel) {
     this.cloudApiSelectorPanel = cloudApiSelectorPanel;
 
     cloudApiUiExtensions = CloudApiUiExtension.EP_NAME.getExtensions();
     for (CloudApiUiExtension uiExtension : cloudApiUiExtensions) {
-      uiExtension.init(this);
+      // TODO: will be implemented in the next PR.
+      Collection<JComponent> customComponents = uiExtension.createCustomUiComponents();
```
etanshaul: changes look great.

just curious what the strategy with this one will be: how will these custom components be places in the UI? Would it be better to just present a single custom JPanel to extend the base UI (which is similar to how a lot of internal idea UI stuff works)?

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiUiExtension.java
```diff
@@ -0,0 +1,57 @@
...
+   */
+  void onCloudLibrarySelection(
+      @Nullable CloudLibrary currentCloudLibrary, String currentBomVersion);
+
+  /**
+   * Callback on module selection change.
+   *
+   * @param module Currently selected module.
+   */
+  void onModuleSelected(Module module);
```
etanshaul: for consistency with the other method you renamed: `onModuleSelection`

google-cloud-apis/src/com/google/cloud/tools/intellij/cloudapis/DefaultCloudApiUiPresenter.java
ivanporty: I am worried about alignment of module combobox with BOM combo box in this case a bit. I was planning to add components to existing grid spots below the module/project selector space, but if possible, let's change to single panel - much more flexible. I will experiment with layouts to see how this will work out.

etanshaul: now that I think about it, aren't there two custom areas for injecting cutsom ui components: at the top area near the GCP project selector, in the library details panel (next to the library description, version, etc)? I guess we should only support the use-case we immediately need and expand later if/when needed.

etanshaul: actually we do need to support both immediately I think: at the top area (due to the BOM selector), in the library details panel (due to the maven coordinate info, and others).

ivanporty: Exactly, great point, for now we will just need components in the line below module/project selection (BOM), and following these steps can expand later to as many places as we need.

PR #2214, opened 2018-07-02
Separate Cloud API from Java/Maven: makes google-cloud-apis available for all IDEs

### LINKED ISSUE #2182: Cloud APIs: consider adding support for enabling APIs and service accounts for all JB IDEs

ivanporty:
Fixes #2182.
Preparation for smaller PR for #2193 (still support service accounts even if Maven is present and project is not Maven). 
See detailed plan in #2199.

This PR flips the bit and makes `google-cloud-apis` completely independent from Maven and Java modules thus making it completely available for other Jetbrains IDEs. I tested it on PyCharm, PHPStorm, GoLand and it worked cleanly and as expected. IDEA behavior with extension points remained the same.

List of changed:
1. `CloudApiUiExtension` has two new methods:
   * `createCustomUiComponents` returns custom UI components for a specific location in the Cloud API dialog (for now places below module label and module combo box).
   * `onCloudLibrariesAddition` is a callback called when Cloud APIs dialog finished and user cofirms selection of his new libraries for some module.
2. `CloudApiUiPresenter` adds another set of methods, providing access to the IDE project and selected module.
3. Another extension point `CloudApiActionDecoratorExtension` added for customizing text and availability of the "Add Libraries" menu item for specific use cases (like Maven or App Engine).
4. Extension points interfaces are moved to `google-cloud-apis` core module, and all Maven functionality now works through them. Explicit dependency on Maven and Java sub-modules is removed.


etanshaul:
just fyi, I didn't forget about this one. just holding off until the next release

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/maven/MavenCloudApiActionDecorator.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.commons.lang3.JavaVersion;
+
+/**
+ * {@link CloudApiActionDecoratorExtension} extension point that checks for Java projects version if
+ * App Engine is used, and ensures Maven dependencies support is only available for Maven projects.
+ * Otherwise, action is disabled and help text is shown.
+ */
+public class MavenCloudApiActionDecorator implements CloudApiActionDecoratorExtension {
```
etanshaul: does the App Engine portion of this need to live here in the Maven module? If we later add Gradle support then the breakdown is a bit odd. But I guess at that point for this to be logically structured we'd need a way for each extension to contribute a check and then aggregate them.

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/maven/MavenCloudApiActionDecorator.java
```diff
@@ -0,0 +1,110 @@
...
+    e.getPresentation().setEnabled(addLibrariesEnabled);
+  }
+
+  private enum CloudLibrariesModuleSupportType {
+    SUPPORTED,
+    MAVEN_REQUIRED,
+    APPENGINE_JAVA8_REQUIRED
+  }
+
+  private CloudLibrariesModuleSupportType checkModuleForAddCloudLibraries(Module module) {
```
etanshaul: nit: `checkAddCloudLibrariesSupport` (module is implied by the param, and adds the word `support` at the end to indicate what its checking)

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/maven/MavenCloudApiActionDecorator.java
```diff
@@ -0,0 +1,110 @@
...
+            case APPENGINE_JAVA8_REQUIRED:
+              e.getPresentation()
+                  .setDescription(
+                      MavenCloudApisMessageBundle.message(
+                          "cloud.libraries.menu.action.gae.java8.required.description"));
+              e.getPresentation()
+                  .setText(
+                      MavenCloudApisMessageBundle.message(
+                          "cloud.libraries.menu.action.disabled.java8.text"));
+              break;
```
etanshaul: nit: the standards doc suggest to always include a default case. In this case, since you don't need to populate it, it can have a comment to indicate why you are skipping / what happens in the default case.

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/maven/MavenCloudApiUiExtension.java
```diff
@@ -16,44 +16,77 @@
...
 public class MavenCloudApiUiExtension implements CloudApiUiExtension {
+  private BomComboBox bomComboBox;
+  private JLabel bomSelectorLabel;
+
+  private CloudLibrary currentCloudLibrary;
 
   @Override
-  public Collection<JComponent> createCustomUiComponents() {
-    return Collections.emptyList();
+  public Map<EXTENSION_UI_COMPONENT_LOCATION, JComponent> createCustomUiComponents() {
```
etanshaul: Perhaps include a javadoc here (or at the class level) to indicate exactly what this is adding in the UI so that its clear from a quick browse of this class.

google-cloud-apis/src/com/google/cloud/tools/intellij/cloudapis/AddCloudLibrariesAction.java
```diff
@@ -48,52 +37,8 @@ public AddCloudLibrariesAction() {
...
-            .setDescription(
-                GoogleCloudApisMessageBundle.message("cloud.libraries.menu.action.description"));
-        e.getPresentation()
-            .setText(
-                GoogleCloudApisMessageBundle.message("cloud.libraries.menu.action.active.text"));
-      }
-    }
-
-    e.getPresentation().setEnabled(addLibrariesEnabled);
+    Stream.of(CloudApiActionDecoratorExtension.EP_NAME.getExtensions())
```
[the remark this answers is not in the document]
ivanporty: You're right, it would be better to return something from this method and stop once one of the extensions do something and disables the action. For example, `decorate` could return `boolean` and if it's true, meaning action has been modified, other decorator extensions are not called then.

google-cloud-apis/java-apis/maven-apis/src/com/google/cloud/tools/intellij/cloudapis/maven/MavenCloudApiActionDecorator.java
ivanporty: Yes, you're right, this extension is rather a transfer of existing functionality in a single piece. It would be better to split it - we don't need Maven check at all (IDEA will do that for us now - Maven functionality will only be available when `maven-api` module is enabled and it depends on Maven presence), and we can move App Engine check into App Engine Java module, and this should allow us to remove this rather unobvious dependency on app engine module. Let me refactor this.

google-cloud-apis/java-apis/maven-apis/build.gradle
```diff
@@ -27,13 +27,14 @@ intellij {
 
 dependencies {
     compile(project(':google-cloud-core'))
+    compile(project(':app-engine:java'))
+    compile(project(':app-engine:java:maven'))
```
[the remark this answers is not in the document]
ivanporty: See my other comment, I'll try to move App Engine specific action decorator into app engine module, and will check if we can get rid of other utility uses so these two won't be necessary. I agree that looks confusing here.

ivanporty:
@etanshaul PTAL at the changes when you have a chance to come back to this again, I simplified action decorator so that Maven APIs do not depend on app engine modules anymore, and update according to your other comments. Thanks!

etanshaul:
great. I'll take a look today at this.

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/apis/AppEngineCloudApiActionDecorator.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.commons.lang3.JavaVersion;
+
+/**
+ * {@link CloudApiActionDecoratorExtension} extension point that checks for Java projects version if
+ * App Engine is used. Otherwise, action is disabled and help text is shown.
+ */
+public class AppEngineCloudApiActionDecorator implements CloudApiActionDecoratorExtension {
+
+  @Override
+  public boolean decorate(AnActionEvent e) {
```
etanshaul: can you add a javadoc on this one perhaps? specifically I think it would be nice to see an explanation on what the boolean return value represents for this specific decorator implementation

google-cloud-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiActionDecoratorExtension.java
```diff
@@ -0,0 +1,38 @@
...
+package com.google.cloud.tools.intellij.cloudapis;
+
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.extensions.ExtensionPointName;
+
+/**
+ * An extension point to dynamically change behaivor of "Add Cloud APIs" action, i.e. check for
+ * eligibility of specific build systems or frameworks and disable/enable/modify text based on that.
+ */
+public interface CloudApiActionDecoratorExtension {
```
etanshaul: I may have missed this, but shouldn't there be a Maven version of this extension to check, from a maven perspective, if the action is supported (`MavenCloudApiActionDecoratorExtension` or something)?

google-cloud-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiManagementConfirmationDialog.java
```diff
@@ -204,7 +205,7 @@ protected ValidationInfo doValidate() {
             GoogleCloudApisMessageBundle.message(
                 "cloud.apis.management.dialog.serviceaccount.name.regex.error"),
             serviceAccountNameTextField);
-      } else if (StringUtils.isEmpty(path)) {
+      } else if (path.isEmpty()) {
```
etanshaul: is `path` also guaranteed not null?

google-cloud-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiActionDecoratorExtension.java
ivanporty: We essentially do the Maven check now using the IDE dependency system - since the Maven API module strictly depends on presence of Maven we can be sure Maven is present. I see your point though that the project itself (or selected module) is not guaranteed to be Maven one. In this case nothing will be done with POM since the module is not "mavenized" but the service account will be created, and environment variables set into existing configurations.
I think we choose from two things here:
* leave this for now and add a help message in the next PR that will show that the following libraries need to be added into the project (i.e. library name and version) or start using Maven and then re-run this action. Similar thing can be added for Gradle projects.
* hide Maven UI and library version to avoid confusion in this case.

etanshaul: > We essentially do the Maven check now using the IDE dependency system - since the Maven API module strictly depends on presence of Maven we can be sure Maven is present. I see your point though that the project itself (or selected module) is not guaranteed to be Maven one. In this case nothing will be done with POM since the module is not "mavenized" but the service account will be created, and environment variables set into existing configurations.

Just trying to understand this. I think there are two separate concepts like you are saying:

1) Is Maven available (i.e. the user has the Maven plugin installed)
2) Is the project/module itself a Maven project

I know we previously were disabling the action when #2 is not true. But now the idea is that we want to enable it for all projects so its not tied to Maven right? I'm fine leaving it in its current state (just do nothing to the project when its not Mavenized) and exploring how to improve this in a separate PR (e.g. making the wording more about the API management / Service account generation instead of build manipulation).

google-cloud-apis/src/com/google/cloud/tools/intellij/cloudapis/AddCloudLibrariesAction.java
```diff
@@ -48,52 +36,13 @@ public AddCloudLibrariesAction() {
...
-        e.getPresentation()
-            .setDescription(
-                GoogleCloudApisMessageBundle.message("cloud.libraries.menu.action.description"));
-        e.getPresentation()
-            .setText(
-                GoogleCloudApisMessageBundle.message("cloud.libraries.menu.action.active.text"));
+    // iterate all available decorators and stop on first that decorated this action.
+    for (CloudApiActionDecoratorExtension decorator :
+        CloudApiActionDecoratorExtension.EP_NAME.getExtensions()) {
+      if (decorator.decorate(e)) {
```
etanshaul: could we have a test to cover this mechanism? for example, that the correct decorator gets applied when multiple are registered etc.

google-cloud-apis/src/com/google/cloud/tools/intellij/cloudapis/CloudApiActionDecoratorExtension.java
ivanporty: You're absolutely correct on those two points. Now, after this PR, Cloud APIs are available not only for all IDEs but also for all Java projects (previously Maven only) which is definitely a better way for users to get access to APIs. We can discuss together how we should present that we support Maven builds and make it clear with wording, just like you said.

google-cloud-apis/src/com/google/cloud/tools/intellij/cloudapis/AddCloudLibrariesAction.java
ivanporty: Always a good idea! I'll add a few, missed this.

ivanporty: I added a unit test which is not quite simple since I think we didn't test using actual extension points before so I investigated easiest way to do it for now and then we can make this part of out unit test rule (`CloudToolsRule`) and potentially re-use for container tools testing as well.

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/apis/AppEngineCloudApiActionDecorator.java
ivanporty: done.

PR #2215, opened 2018-07-10
Report only stacktrace portion of specific AppEngineException during local run

etanshaul:
Action item related to #2191


etanshaul:
screen shot 2018-07-10 at 6 32 16 am

PR #2216, opened 2018-07-10
Provide maven-aware appengine-web.xml generation strategy for CE

### LINKED ISSUE #1498: NotNull constraint violation when flex deployment config module is null
The app.yaml config dropdown in the deployment ux is restored from the serialized module name in the deployment config. If the module name was not stored (e.g. null), then a not-null constraint is being violated in `AppEngineFlexibleDeploymentEditor#resetEditorFrom` when calling the utility `AppEngineFlexibleFacet#getFacetByModuleName(@NotNull moduleName..)`

etanshaul:
fixes #1498 for Community Edition Maven projects

TODO (next PR) - add Gradle extension point

See this comment for more details: https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1948#issuecomment-401402786

This PR is the CE version of the already merged Ultimate PR (#2203) fixing the generation strategy of `appengine-web.xml` on framework support addition. This PR adds the extension point plumbing, and a Maven implementation.

The idea is that specific build systems register build-aware implementations of `BuildSystemAppEngineWebXmlDirectoryProvider` to provide build-aware parent directories for placing the `appengine-web.xml`.

For Maven, it queries the maven-war-plugin to locate the `warSourceDirectory` (or returns the canonical location if undefined).

Note that the way its implemented, the logic will use the strategy defined by the first extension point it locates (which I think is either nondeterministic, or based on the order they are registered in plugin.xml). So, if the user's module has Maven and Gradle, it will arbitrarily use the first strategy it finds. If no strategies are found, it will fall back to the "native" approach of simply generating one under `{root}/web/WEB-INF`


etanshaul:
working on tests now

etanshaul:
what this does, from a user's perspective, is if the user opens a plain Maven project in Community (like a spring-boot war project) and tries to appengine-ify it by adding appengine support, it will work immediately. Previously it was not able to properly place the `appengine-web.xml` file.

app-engine/java/maven/src/com.google.cloud.tools.intellij.appengine.java.maven/MavenAppEngineWebXmlDirectoryProvider.java
```diff
@@ -0,0 +1,76 @@
+/*
```
patflynn: I think the package directory path got named incorrectly with '.' instead of as a bunch of sub folders.

etanshaul: oh wow good catch

app-engine/java/maven/testSrc/com/google/cloud/tools/intellij/appengine/java/maven/MavenAppEngineWebXmlDirectoryProviderTest.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  @Test
+  public void getPath_withNoMavenProject_returnsEmpty() {
+    assertThat(directoryProvider.getAppEngineWebXmlDirectoryPath(module))
+        .isEqualTo(Optional.empty());
+  }
+
+  @Test
+  public void getPath_withMavenProject_andNoWarSourceDirectory_returnsCanonicalMavenDir() {
```
patflynn: should we have a test on a project that includes a war source dir?

app-engine/java/maven/src/com.google.cloud.tools.intellij.appengine.java.maven/MavenAppEngineWebXmlDirectoryProvider.java
etanshaul: package structure should be fixed

app-engine/java/maven/testSrc/com/google/cloud/tools/intellij/appengine/java/maven/MavenAppEngineWebXmlDirectoryProviderTest.java
etanshaul: yeah ideally. I didn't include one originally because it seems pretty difficult to mock this up (the MavenProjectManager and its components), but I'll look again.

etanshaul: I added test coverage for this. ptal.

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/BuildSystemAppEngineWebXmlDirectoryProvider.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * An extension point for implementing build-system aware strategies for providing the path to the
+ * appengine-web.xml directory.
+ */
+public interface BuildSystemAppEngineWebXmlDirectoryProvider {
+
+  ExtensionPointName<BuildSystemAppEngineWebXmlDirectoryProvider> EP_NAME =
+      ExtensionPointName.create("com.google.gct.core.appEngineWebXmlDirectoryProvider");
+
+  /** Returns, optionally, the path to the appengine-web.xml directory for the given module. */
```
patflynn: nit: maybe explain what condition makes it optional?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/impl/AppEngineStandardCommunityWebIntegration.java
```diff
@@ -49,10 +52,27 @@
   @Override
   public VirtualFile suggestParentDirectoryForAppEngineWebXml(
       @NotNull Module module, @NotNull ModifiableRootModel rootModel) {
-    final VirtualFile root = ArrayUtil.getFirstElement(rootModel.getContentRoots());
+    BuildSystemAppEngineWebXmlDirectoryProvider[] appEngineWebXmlDirectoryProviders =
+        Extensions.getExtensions(BuildSystemAppEngineWebXmlDirectoryProvider.EP_NAME);
+    for (BuildSystemAppEngineWebXmlDirectoryProvider provider : appEngineWebXmlDirectoryProviders) {
+      Optional<String> pathOptional = provider.getAppEngineWebXmlDirectoryPath(module);
```
patflynn: It seems a bit odd to iterate through these rather than pick the right provider for the kind of module this is.

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/BuildSystemAppEngineWebXmlDirectoryProvider.java
etanshaul: done

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/impl/AppEngineStandardCommunityWebIntegration.java
etanshaul: each provider will either return Optional#empty if the given module is not applicable to the given build system.

i.e. if the module is not a Maven module, then `MavenAppEngineWebXmlDirectoryProvider` will return `Optional#empty`.

I agree it feels a bit odd to iterate through them all, but the benefit of this is that it allows us to have a single extension point and have each extension determine if its applicable, rather than the other way around.

I think a way to improve readablility would be to add an

`isApplicable(Module module)`

method to the interface. And instead of iterating blindly through each provider, first check `isApplicable`. I'll make this update

etanshaul: on second though. I don't like extracting an `isApplicable` method since it leaves open the possibility of calling the interface method `getAppEngineWebXmlDirectoryPath` in an invalid state. I think its better to have the `getAppEngineWebXmlDirectoryPath` confirm itself that its applicable for the build system and return empty if not (as it is now). wdyt?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/impl/AppEngineStandardCommunityWebIntegration.java
```diff
@@ -49,10 +52,27 @@
...
+        }
+      }
+    }
+
+    // Build-aware strategies failed or missing, fall back to simple native approach
+    VirtualFile root = ArrayUtil.getFirstElement(rootModel.getContentRoots());
     if (root != null) {
       try {
-        return VfsUtil.createDirectoryIfMissing(root, "WEB-INF");
+        return VfsUtil.createDirectoryIfMissing(root, "web/WEB-INF");
```
ivanporty: What was the reason for this change, from plain `WEB-INF` to adding `web/`?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/impl/AppEngineStandardCommunityWebIntegration.java
ivanporty: In this case it feels we can have multiple build systems support for a module, and the one selected is completely random. It slightly feels we miss module type safety checking or something similar. Does a module have something that uniquely identifies its build system for this case, could we map this to build system provider implementation?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/impl/AppEngineStandardCommunityWebIntegration.java
```diff
@@ -49,10 +52,27 @@
...
+    BuildSystemAppEngineWebXmlDirectoryProvider[] appEngineWebXmlDirectoryProviders =
+        Extensions.getExtensions(BuildSystemAppEngineWebXmlDirectoryProvider.EP_NAME);
+    for (BuildSystemAppEngineWebXmlDirectoryProvider provider : appEngineWebXmlDirectoryProviders) {
+      Optional<String> pathOptional = provider.getAppEngineWebXmlDirectoryPath(module);
+
+      if (pathOptional.isPresent()) {
+        try {
+          return VfsUtil.createDirectoryIfMissing(pathOptional.get());
+        } catch (IOException ioe) {
+          LOG.warn(
```
ivanporty: Is this visible enough for a user - i.e. can we suggest some other remedy - like suggest adding simple xml into web resources location?

app-engine/java/maven/src/com/google/cloud/tools/intellij/appengine/java/maven/MavenAppEngineWebXmlDirectoryProvider.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    Element warPluginConfig =
+        mavenProject.getPluginConfiguration(
+            MAVEN_WAR_PLUGIN_GROUP_ID, MAVEN_WAR_PLUGIN_ARTIFACT_ID);
+
+    String warSourceDirectory = null;
+
+    if (warPluginConfig != null) {
+      warSourceDirectory = warPluginConfig.getChildTextTrim(MAVEN_WAR_PLUGIN_WAR_DIR_ATTR_NAME);
```
ivanporty: Is this optional information for a Maven configuration, i.e. most of the time we will be reverting back to `MAVEN_WAR_DEFAULT_DIR` option?

app-engine/java/maven/src/com/google/cloud/tools/intellij/appengine/java/maven/MavenAppEngineWebXmlDirectoryProvider.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    if (warSourceDirectory == null) {
+      warSourceDirectory = MAVEN_WAR_DEFAULT_DIR;
+    }
+
+    if (!FileUtil.isAbsolute(warSourceDirectory)) {
```
ivanporty: `warSourceDirectory` is a value outside of the our control, correct? Does a user specify this, and can choose either relative or absolute path?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/impl/AppEngineStandardCommunityWebIntegration.java
ivanporty: Oh I just realized you have much better messages and notification for missing xml file now in other PRs, thanks!

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/impl/AppEngineStandardCommunityWebIntegration.java
etanshaul: this code flow is for community edition native (no build system) projects. In this case, I am simply mimicking the directory structure that intellij generates for its web applications (when using the wizard):

selection_787

Note that `web/WEB-INF` is under the root.

Previously, it was omitting the web directory and just placing `WEB-INF` under root

app-engine/java/maven/src/com/google/cloud/tools/intellij/appengine/java/maven/MavenAppEngineWebXmlDirectoryProvider.java
etanshaul: to show the whole config:

```
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-war-plugin</artifactId>
    <configuration>
        <warSourceDirectory>${basedir}/src/main/alternate-webapp-dir</warSourceDirectory>
    </configuration>
</plugin>
```

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/impl/AppEngineStandardCommunityWebIntegration.java
ivanporty: Ah I see what you mean by "native" in the comment then. Could we possibly expand the "native" in the comment as well and say we use default IDEA directory structure in this case? Might be helpful some time after for easier maintenance.

ivanporty: Nit: possibly make it even easier with well named constant for `web`?

etanshaul: sounds good (both points).

etanshaul:
> In this case it feels we can have multiple build systems support for a module, and the one selected is completely random. It slightly feels we miss module type safety checking or something similar. Does a module have something that uniquely identifies its build system for this case, could we map this to build system provider implementation?

Yeah its true that one is selected non-deterministically (or based on whatever strategy extensions implementations are loaded from plugin.xml).

A module does have info to tell us what build system it uses (this is how each extension impl determines if its applicable - it first checks this and then returns `Optional#empty` if its not applicable). What do you mean exactly by "could we map this to build system provider implementation"?

etanshaul:
also to point out - if the user has 2 build systems, but only syncs with one of them, then only the provider that matches the sync build system should be used. If the user syncs with both, then it will return the path to appengine-web.xml randomly from one of the strategies (which I don't think is horrible given that the user synced with both). I don't know if we want to give a priority to say Maven vs Gradle. wdyt?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/impl/AppEngineStandardCommunityWebIntegration.java
etanshaul: done

ivanporty:
Yeah I see. I see the point of the idea and it's pretty clear, the only thing is that code kind of shows we select the active build system by random chance but this is probably expected in this case. What about we add a bit more detailed comment there and instead of `for` wrap this into `Stream.of(buildDirectoryProviders).findAny(non-empty optional)` so this `findAny` emphasize this approach?

etanshaul:
good suggestions

and I agree with the random selection feels a bit off. if you have any suggestions on how to avoid this I'm open. The only thing I can think of is to have some sort of "priority" or "order" value in the extension and each extension registers this priority. We then select the providers in order of priority. (the extension mechanism may have this built in - I'd need to check). This has the downside of forcing us to favor one over the other (or giving them equal priorities in which case we are back to random choice). wdyt?

ivanporty:
If there are good and clear extension point ordering built-in, so we can prioritize Maven or Gradle or vice versa, that'd be great. I think we can file a different issue for this if it takes more time, since this is already working and we can fine tune it in the next release.

etanshaul:
went with the simplist approach for now: just added a detailed comment. We can refine this later, as you said.

PR #2217, opened 2018-07-11
Don't lock appengine support addition; notify on appengine-web.xml gen failure

etanshaul:
in support of #2191 

This PR does 2 things:

1) Previously, after the user added appengine support via our tools menu and the facet was added, the user would no longer be able to add support to that module again _even if the `appengine-web.xml` failed to generate_ - the module simply would be filtered out from being selectable. This PR updates this so that the user can try again (and therefore also checks to make sure the facet is not attempted to be added twice). 
2) Shows a notification when the generation of the `appengine-web.xml` failed:

screen shot 2018-07-11 at 9 51 53 am


patflynn:
doesn't IJ have a special notification type for errors?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/AppEngineStandardSupportProvider.java
```diff
@@ -71,6 +79,14 @@
   private static final Logger LOG =
       Logger.getInstance("#com.intellij.appengine.facet.AppEngineStandardSupportProvider");
 
+  private static final NotificationGroup NOTIFICATION_GROUP =
+      new NotificationGroup(
+          new PropertiesFileFlagReader().getFlagString("notifications.plugin.groupdisplayid"),
+          NotificationDisplayType.BALLOON.BALLOON,
+          true,
+          null,
+          GoogleCloudCoreIcons.CLOUD);
```
ivanporty: regarding Patrick's comment on error style for this message, can we consider having Google Cloud icon in "error state", i.e. something with a red asterisk or similar? If it takes long to obtain a new icon, we can postpone for later.

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/facet/standard/AppEngineStandardSupportProvider.java
```diff
@@ -128,8 +144,15 @@ private void addSupport(
       Set<AppEngineStandardMavenLibrary> librariesToAdd) {
     FacetType<AppEngineStandardFacet, AppEngineStandardFacetConfiguration> facetType =
         AppEngineStandardFacet.getFacetType();
+
     AppEngineStandardFacet appEngineStandardFacet =
-        FacetManager.getInstance(module).addFacet(facetType, facetType.getDefaultFacetName(), null);
+        AppEngineStandardFacet.getAppEngineFacetByModule(module);
+    if (appEngineStandardFacet == null) {
```
ivanporty: We added this check since we now allow adding support as many times as user wants, correct? Previously it was guarded by menu item check?

etanshaul: exactly

ivanporty:
I think icon comment is an enhancement that can be addressed later so LGTM.

etanshaul:
agree about the icon comment. I will update the notification type to ERROR. will include a screenshot for reference

etanshaul:
I changed the notification type to ERROR. The notification itself looks the same. The difference in the UI is that the message in the event log is now red instead of yellow.

PR #2222, opened 2018-07-17
Implements gradle build aware appengine-web.xml provider

etanshaul:
Similar to #2216 but for Gradle.

closes out #1948 

more info:

The basic idea is identical to the Maven equivalent PR #2216. However, the method used to fetch the Gradle data is different. With Maven, we could simply use the Maven integration to query the Maven build info at any time. For Gradle this is not available: instead, we use the mechanism previously set up of hooking into Gradle sync and then persisting the data fetched from the Gradle model.


app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleProjectDataService.java
```diff
@@ -63,15 +63,17 @@ public void importData(
         .forEach(
             module -> {
               AppEngineGradleModule appEngineGradleModule = moduleNameToModel.get(module.getName());
-              if (PlatformUtils.isIdeaCommunity()
```
etanshaul: the idea here is that previously the data fetched from the Gradle model was only persisted when the app-engine-gradle-plugin was available (since in the previous use-case, this was the only time we needed it). Now we need this info for other purposes, so I removed the requirement that the plugin be available.

Similarly, if you see in `AppEngineGradleDeploymentSourceProvider` I added the check for the plugin there instead. 

In other words, the data from the Gradle model now is always available. Instead I moved the check for the app-gradle-plugin from sync-time to deployment source addition time.

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleDeploymentSourceProvider.java
```diff
@@ -62,6 +62,7 @@
             module -> {
               if (projectService.isGradleModule(module)
                   && AppEngineStandardFacet.hasFacet(module)
+                  && AppEngineStandardGradleModuleComponent.getInstance(module).hasAppGradlePlugin()
```
etanshaul: see my other comment for explanation

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/GradleAppEngineWebXmlDirectoryProvider.java
```diff
@@ -0,0 +1,59 @@
...
+
+    AppEngineStandardGradleModuleComponent gradleModuleComponent =
+        AppEngineStandardGradleModuleComponent.getInstance(module);
+    String gradleWebAppDir = gradleModuleComponent.getWebAppDir().orElse(null);
+
+    if (gradleWebAppDir == null) {
+      gradleWebAppDir = GRADLE_WEBAPP_DEFAULT_DIR;
+    }
+
+    if (!FileUtil.isAbsolute(gradleWebAppDir)
```
etanshaul: note to self: FileUtil#isAbsolute is OS dependent. Need to fix here and in already merged Maven PR (also create a test in the Maven PR that would have detected this bug, like there is in this PR).

etanshaul: update: this is just a bug in the unit test, not the provider logic. Paths in Windows are handled properly because:

1) when the user specifies an absolute path for the web app dir in the war plugin config with something like:
`webAppDirName = project.file('src/main/custom-webapp`)
If using Windows then the absolute path will properly be constructed and `FileUtil#isAbsolute` will work as expected

2) The path string that is constructed via these providers is ultimately used to create a VirtualFile which does the right thing (i.e. normalizes the path by converting the slashes, removing double slashes etc.)

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleModelBuilderService.java
```diff
@@ -34,11 +35,21 @@ public boolean canBuild(String modelName) {
   @Override
   public Object buildAll(String modelName, Project project) {
     boolean hasAppEngineGradlePlugin = project.getPlugins().hasPlugin(APP_GRADLE_PLUGIN_ID);
+    boolean hasWarPlugin = project.getPlugins().hasPlugin("war");
```
ivanporty: should we make it a constant maybe?

etanshaul: done

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleModel.java
```diff
@@ -31,6 +31,9 @@
    */
   String gradleBuildDir();
 
-  /** Returns the path to the root of the Gradle module */
+  /** Returns the path to the root of the Gradle module. */
   String gradleModuleDir();
+
+  /** Returns the path to the web application directory. */
+  String webAppDir();
```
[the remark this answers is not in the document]
etanshaul: in the class where this value is actually consumed (`AppEngineStandardGradleModuleComponent`) the `getWebAppDir` method returns an optionally wrapped value. 

if I were to make this return an optional, there would be a lot of wrapping and unwrapping boilerplate (since we should not be using optional parameters). wdyt?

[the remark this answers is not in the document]
ivanporty: Yep, I saw that too - what if we just mark it `@Nullable` then as a reminder for later?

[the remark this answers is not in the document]
etanshaul: sgtm. I'll add @Nullable to the other ones too for consistency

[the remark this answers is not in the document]
etanshaul: done

app-engine/java/gradle/src/com/google/cloud/tools/intellij/appengine/java/gradle/AppEngineGradleModelBuilderService.java
```diff
@@ -34,11 +35,21 @@ public boolean canBuild(String modelName) {
...
+    String webAppDir = null;
+
+    if (hasWarPlugin) {
+      webAppDir =
+          project.hasProperty(GRADLE_WEB_APP_DIR_PROPERTY_NAME)
+              ? String.valueOf(project.property(GRADLE_WEB_APP_DIR_PROPERTY_NAME))
```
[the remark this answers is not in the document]
etanshaul: I assumed so, but its not immediately clear by looking at the implementation.

So to be safe, I updated it to check property != null

PR #2224, opened 2018-07-19
Choose from available web-roots more intelligently during appengine-web.xml generation

### LINKED ISSUE #2223: Double "WEB-INF" directory created by Adding App Engine Standard Support
Project:
https://github.com/GoogleCloudPlatform/java-docs-samples/tree/master/appengine-java8/gaeinfo

Contains proper Web facet and App Engine configuration by default.
When you open this project, enable App Engine Facet by accepting prompt. Works with local server.

Steps after that:
1. Delete `webapp` completely.
2. Go to Cloud Tools -> Add App Engine Standard Support
3. Double directory is created.
screen shot 2018-07-19 at 11 54 30 am

etanshaul:
Fixes #2223 

See the bug for more details of what was happening.

Previously, when no web root physically existed on the filesystem (but were defined in the IDE), then the 1st one was always chosen. In this case, the web root was:

`../webapp/WEB-INF`

The code was always generating a WEB-INF in the web root causing the double WEB-INF bug.

The fix here does 2 things:
1) Chooses from the available web roots slightly more intelligently: it attempts to find the web root that has a root relative path to the deployment root directory, i.e "/". See the second table entry in this screenshot for an idea of what I mean:
image
If none is available, then just take the first web root as before**
2) Given the webroot path, only generate a `WEB-INF` folder if the leaf folder of the webroot isn't already `WEB-INF` (imagine that in the screenshot above, the first entry in the table was all that existed: in this case, the WEB-INF will not be created after the web root is created since it already exists in the web root.).

** I didn't go farther than this on purpose. The user can enter arbitrary values for the web root and the relative path to the deployment root so it becomes messy (and unreliable) to do any deduction beyond what I described in step 1.

### DIFF
--- a/app-engine/java/ultimate/src/com/google/cloud/tools/intellij/appengine/java/ultimate/impl/AppEngineStandardUltimateWebIntegration.java
+++ b/app-engine/java/ultimate/src/com/google/cloud/tools/intellij/appengine/java/ultimate/impl/AppEngineStandardUltimateWebIntegration.java
@@ -121,19 +121,27 @@ public class AppEngineStandardUltimateWebIntegration extends AppEngineStandardWe
     }
 
     try {
-      Optional<VirtualFile> webRootDir =
+      Optional<VirtualFile> webRootDirOptional =
           webRoots.stream().map(WebRoot::getFile).filter(Objects::nonNull).findFirst();
 
-      if (!webRootDir.isPresent()) {
-        // There is at least one web-root path, but none that exist, select the first and create it
-        String presentableUrl = webRoots.get(0).getPresentableUrl();
+      if (!webRootDirOptional.isPresent()) {
+        // There is at least one web-root path, but none that exist, find a suitable one and create
+        // it
+        String presentableUrl = findWebRootPath(webRoots);
         if (presentableUrl != null) {
-          webRootDir = Optional.ofNullable(VfsUtil.createDirectories(presentableUrl));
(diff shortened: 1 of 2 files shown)


app-engine/java/ultimate/src/com/google/cloud/tools/intellij/appengine/java/ultimate/impl/AppEngineStandardUltimateWebIntegration.java
```diff
@@ -220,6 +228,34 @@ public void setupDevServer() {
...
+   *
+   * <p>If none exist with a root relative path, simply grab the first web root and return its path.
+   */
+  @Nullable
+  private String findWebRootPath(List<WebRoot> webRoots) {
+    if (webRoots.isEmpty()) {
+      return null;
+    }
+
+    // TODO test that "/" check works in Windows
```
ivanporty: good question :)

etanshaul: yep. I'll make sure to do that before merging :)

etanshaul: confirmed working in Windows. (The logic is working off of VirtualFiles so the paths are the same).

Issue #2241, opened 2018-08-20
Hide module selector for non-Maven Cloud APIs dialog to reduce user confusion


ivanporty:
Follow up for #2214.

To reduce confusion, initially hide module selection UI and only show for cases where it makes sense (i.e. for now only for Maven projects).

Also, since non-Maven Java projects libraries are not supported, hiding both module and BOM selectors for such projects would clarify UI and well.

PR #2242, opened 2018-08-20
Update Cloud API dialog and action/menu wording.

### LINKED ISSUE #2240: Update wording for non-Maven Cloud APIs dialog to reduce user confusion
After implementing #2214, major labels and titles still suggest Cloud API dialog will do something with the actual project/module libraries rather than just prepare and enable APIs on GCP and create service accounts.

Update the following to avoid using "add libraries" and keep the original names only for Maven projects:
1. Action name / menu item text
2. Dialog title
3. All custom named buttons like "Add Client Library" used instead of "OK" in the dialog.

ivanporty:
Fixes #2240.
Will be followed up and completed with #2241.

Removes confusion for non-Maven cases where a user can only enable and manage Cloud APIs but not specific project/language libraries.

Maven project dialog:
cloud-apis-maven

Non-Maven (any language) project dialog:
cloud-api-plainjava

Menu item text by default (non-Maven):
menu-apis-non-maven


etanshaul:
The Action text and the dialog header look good.

The button "Select Cloud API" feels a bit strange on first glance (could just be me though). For one, the use could select multiple APIs, not just one. Maybe "Select Cloud API(s)"? @parkjam4 maybe can comment on the styleguide for doing that.

The other options is maybe just something like "Continue".

ivanporty:
I was hesitant about this too, but we originally used "Add Client Library" as a button text although multiple libraries can be selected so I followed that path. Looking for suggestions here.

etanshaul:
I was trying to find some existing examples of this in IntelliJ without much luck. One other think I was thinking is that the action should have a trailing `...` to indicate that more things will happen before its applied

etanshaul:
maybe just "Continue ..." (and I possible use this wording regardless of Maven vs no Maven). wdyt?

ivanporty:
I read a bit on ellipsis in buttons and it might be a suitable case, so neutral `Continue...` should work too. I'll do the change while we wait for input from @parkjam4.

parkjam4:
I like the use of Continue, but I don't think we need the ellipse here since Continue already implies that additional steps are coming.

PR #2252, opened 2018-08-29
Disable cloud api continue button if GCP project not selected and no client library support

### LINKED ISSUE #2248: Cloud API support in non-IDEA IDEs should require a GCP project
Since adding client libraries to the build is not currently supported in non-IntelliJ IDEs, the user should be required to select a GCP project before clicking "continue". Currently, if one is not selected, the wizard simply closes and does nothing (because there is no GCP project context from which to manage APIs).

etanshaul:
fixes #2248 

@ivanporty ptal. This PR disables the "continue" action in the API dialog if:

1) There is no client library functionality (tested by checking if the module selector is hidden -> this is less then ideal, so I'm open to suggestions to improve this)
**&&**
2) There is no GCP project

In other words, if the only functionality of the dialog is to manage APIs, then the user _must_ select a GCP project before continuing. Before, the user could continue and the dialog would just close and do nothing (which is strange behavior).


ivanporty:
Good catch, I remember seeing something like this. I agree that checking based on module UI visibility is not obvious; but we cannot rely on sole presence of UI extensions since they don't necessarily add anything into submit behavior. We can revisit this in the next release and update extension API but so far LGTM. Thanks!

PR #2254, opened 2018-08-31
Kotlin stackdriver debugger support

### LINKED ISSUE #2253: Add support for Stackdriver Debugger with Kotlin

- Cloud Tools for IntelliJ version: 18.8.1
- IntelliJ version: IntelliJ IDEA 2018.2.3 (Ultimate Edition) Build #IU-182.4323.18, built on August 28, 2018
- Google Cloud SDK version _(run `gcloud version` on the command line)_: 179.0.0
- OS: Windows 10

**What did you do?**
Debugged a mixed Java/Kotlin Web App. I can add "Cloud Snapshot Locations" in Java files, but not in Kotlin.

The Stackdriver debugger at https://console.cloud.google.com/debug has no problems with .kt files, so I expect that from the IntelliJ plugin as well.

**What did you see instead?**
I can only add normal (local) breakpoints to Kotlin files. I found no way to add Cloud Snapshot Locations to Kotlin files.

Popup opens in Java files when clicking the gutter:
image

Breakpoint is added directly when clicking the gutter in Kotlin files:
image

etanshaul:
fixes #2253 

Supports Kotlin file types as breakpoint sources. 

- Kotlin and Java language specific logic lives in optionally included plugin.xml's: `stackdriver-debugger-java.xml` and `stackdriver-debugger-kotlin.xml`
  - Currently all that is needed is a simple language-specific implementation of the platform `JavaDebugAware` extension point which probably would have been better named as `JvmDebugAware` 

- Since the language specific code was so minimal I decided not to create new kotlin and java modules under stackdriver and instead keep all the source in the stackdriver module. However, the kotlin plugin is still optionally included in the `google-cloud-tools-plugin` module's plugin.xml - which means things will still work for Java if the user doesn't have the Kotlin plugin installed. Of course this now has the downside that the build time classpath doesn't match the runtime completely, but I think its OK given the extent of the language specific support.


stackdriver-debugger/src/com/google/cloud/tools/intellij/stackdriver/debugger/KotlinLangDebugAware.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package com.google.cloud.tools.intellij.stackdriver.debugger;
+
+import com.intellij.debugger.engine.JavaDebugAware;
+import com.intellij.psi.PsiFile;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.kotlin.idea.KotlinFileType;
+
+public class KotlinLangDebugAware extends JavaDebugAware {
```
ivanporty: It looks like javadoc would be helpful here.

stackdriver-debugger/src/com/google/cloud/tools/intellij/stackdriver/debugger/JavaLangDebugAware.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package com.google.cloud.tools.intellij.stackdriver.debugger;
+
+import com.intellij.debugger.engine.JavaDebugAware;
+import com.intellij.ide.highlighter.JavaFileType;
+import com.intellij.psi.PsiFile;
+import org.jetbrains.annotations.NotNull;
+
+public class JavaLangDebugAware extends JavaDebugAware {
```
ivanporty: Same, javadoc would be helpful here.

stackdriver-debugger/src/com/google/cloud/tools/intellij/stackdriver/debugger/CloudBreakpointHandler.java
```diff
@@ -353,13 +354,15 @@ public void registerBreakpoint(
     }
 
     PsiFile javaFile = psiManager.findFile(ideBreakpoint.getSourcePosition().getFile());
```
ivanporty: Since this is no longer a java file only, should we name it a bit more broad?

etanshaul: definitely

etanshaul:
updated based on comments so far

google-cloud-tools-plugin/resources/META-INF/plugin.xml
```diff
@@ -51,6 +52,7 @@
   <xi:include href="/META-INF/google-cloud-storage.xml" xpointer="xpointer(/idea-plugin/*)"/>
   <xi:include href="/META-INF/google-cloud-repos.xml" xpointer="xpointer(/idea-plugin/*)"/>
   <xi:include href="/META-INF/google-cloud-apis.xml" xpointer="xpointer(/idea-plugin/*)"/>
+  <xi:include href="/META-INF/stackdriver-debugger.xml" xpointer="xpointer(/idea-plugin/*)"/>
```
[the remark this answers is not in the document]
etanshaul: My initial inclination is to move this to live under `google-cloud-java.xml` which presumably will be available for all JVM languages. And perhaps collapse `stackdriver-debugger.xml` and `stackdriver-debugger-java.xml` into one.

But I need to revisit this.

etanshaul:
@ivanporty I fixed your comment about stackdriver being available for other IDEs (like PyCharm). See me last commit: I moved the stackdriver plugin xml to live under `google-cloud-java.xml`

Now, debugger is only available when the Java module is available (IDEA). Kotlin debugging still works.

PR #2261, opened 2018-10-02
Better handling of Cloud SDK version parse error

### LINKED ISSUE #2259: Handle exception if Cloud SDK version cannot be parsed
For example, we got the following user reported error:

```
com.google.cloud.tools.appengine.cloudsdk.CloudSdkVersionFileParseException: Pattern found in the Cloud SDK version file could not be parsed: 218.0.0.213974472
	at com.google.cloud.tools.appengine.cloudsdk.CloudSdk.getVersion(CloudSdk.java:95)
	at com.google.cloud.tools.appengine.cloudsdk.CloudSdk.validateCloudSdkVersion(CloudSdk.java:187)
	at com.google.cloud.tools.appengine.cloudsdk.CloudSdk.validateCloudSdk(CloudSdk.java:181)
	at com.google.cloud.tools.appengine.cloudsdk.GcloudRunner.run(GcloudRunner.java:75)
```

We should determine how to handle this better. For example, don't blow up the IDE and show a warning instead (and do something like send an analytics ping so that we can troubleshoot).

etanshaul:
fixes #2259 

Will wait to merge until appengine-plugins-core is released with this change: https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/709

Details:

https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/709 updated the generic Cloud SDK validation logic to _not_ throw an exception on version parsing errors.

This PR updates the SDK plugin-side validation by doing the following:
- Adds a new Cloud SDK notification when there is an error parsing the version file. It does this by calling CloudSdk#getVersion instead (which will still throw an exception). The notification looks like this:
image
Note: **This will not block the operation (like deployment):** once the core lib PR is merged, deployment will continue and succeed if the Cloud SDK installation has no other problems
- Catches the `CloudSdkOutOfDateException` during deployment: previously when the SDK was out of date, the notification would be shown and then an IDE error would occur (because the core library throws an exception downstream when the minimum version is not met). Now the behavior is that the warning will be shown, but the IDE won't blow up.

TODO:
- Consider doing an analytics ping when there is a version parsing error (since now we would have no other way to be notified)

### DIFF
--- a/app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/AppEngineDeploy.java
+++ b/app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/AppEngineDeploy.java
@@ -19,6 +19,7 @@ package com.google.cloud.tools.intellij.appengine.java.cloud;
 import com.google.api.client.repackaged.com.google.common.annotations.VisibleForTesting;
 import com.google.cloud.tools.appengine.api.AppEngineException;
 import com.google.cloud.tools.appengine.api.deploy.DefaultDeployConfiguration;
+import com.google.cloud.tools.appengine.cloudsdk.CloudSdkOutOfDateException;
 import com.google.cloud.tools.appengine.cloudsdk.process.LegacyProcessHandler;
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessExitListener;
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessHandler;
@@ -147,9 +148,14 @@ public class AppEngineDeploy {
             .build();
 
     // show a warning notification if the cloud sdk version is not supported
-    CloudSdkVersionNotifier.getInstance().notifyIfUnsupportedVersion();
+    CloudSdkVersionNotifier.getInstance().notifyIfVersionOutOfDate();
+    CloudSdkVersionNotifier.getInstance().notifyIfVersionParseError();
 
-    helper.createGcloud(loggingHandler).newDeployment(processHandler).deploy(configuration);
+    try {
+      helper.createGcloud(loggingHandler).newDeployment(processHandler).deploy(configuration);
+    } catch (CloudSdkOutOfDateException ex) {
+      logger.warn("Error during deployment because because Cloud SDK version out of date");
+    }
   }
 
   public AppEngineHelper getHelper() {
--- a/app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/executor/AppEngineStandardRunTask.java
+++ b/app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/executor/AppEngineStandardRunTask.java
@@ -79,6 +79,7 @@ public class AppEngineStandardRunTask extends AppEngineTask {
   public void execute(ProcessStartListener startListener) {
     // show a warning notification if the cloud sdk version is not supported
     CloudSdkVersionNotifier.getInstance().notifyIfVersionOutOfDate();
+    CloudSdkVersionNotifier.getInstance().notifyIfVersionParseError();
 
     CloudSdkService instance = CloudSdkService.getInstance();
     CloudSdk.Builder sdkBuilder =
--- a/app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/executor/AppEngineStandardRunTask.java
+++ b/app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/executor/AppEngineStandardRunTask.java
@@ -78,7 +78,7 @@ public class AppEngineStandardRunTask extends AppEngineTask {
   @Override
   public void execute(ProcessStartListener startListener) {
     // show a warning notification if the cloud sdk version is not supported
-    CloudSdkVersionNotifier.getInstance().notifyIfUnsupportedVersion();
+    CloudSdkVersionNotifier.getInstance().notifyIfVersionOutOfDate();
 
     CloudSdkService instance = CloudSdkService.getInstance();
     CloudSdk.Builder sdkBuilder =
--- a/app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/startup/CloudSdkVersionStartupCheck.java
+++ b/app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/startup/CloudSdkVersionStartupCheck.java
@@ -31,5 +31,6 @@ public class CloudSdkVersionStartupCheck implements StartupActivity {
   public void runActivity(@NotNull Project project) {
     // If there is a configured Cloud SDK at this time, check that it is supported.
     CloudSdkVersionNotifier.getInstance().notifyIfVersionOutOfDate();
+    CloudSdkVersionNotifier.getInstance().notifyIfVersionParseError();
   }
 }
--- a/app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/startup/CloudSdkVersionStartupCheck.java
+++ b/app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/startup/CloudSdkVersionStartupCheck.java
@@ -30,6 +30,6 @@ public class CloudSdkVersionStartupCheck implements StartupActivity {
   @Override
   public void runActivity(@NotNull Project project) {
     // If there is a configured Cloud SDK at this time, check that it is supported.
-    CloudSdkVersionNotifier.getInstance().notifyIfUnsupportedVersion();
+    CloudSdkVersionNotifier.getInstance().notifyIfVersionOutOfDate();
   }
 }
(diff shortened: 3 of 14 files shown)


google-cloud-sdk/src/com/google/cloud/tools/intellij/appengine/java/sdk/CloudSdkValidator.java
```diff
@@ -58,8 +58,16 @@ public static CloudSdkValidator getInstance() {
       // If the Cloud SDK is not found, don't bother checking anything else
       return validationResults;
     } catch (CloudSdkOutOfDateException exception) {
-      validationResults.add(CloudSdkValidationResult.CLOUD_SDK_VERSION_NOT_SUPPORTED);
-    } catch (CloudSdkVersionFileException e) {
+      validationResults.add(CloudSdkValidationResult.CLOUD_SDK_NOT_MINIMUM_VERSION);
+    } catch (CloudSdkVersionFileException exception) {
+      // TODO once appengine-plugins core change is merge in, this exception will not be thrown
```
etanshaul: I need to catch this now, but once we update the core library we can remove this `catch` block

patflynn:
:+1:

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/AppEngineDeploy.java
```diff
@@ -147,9 +148,14 @@ public void deploy(
...
 
     // show a warning notification if the cloud sdk version is not supported
-    CloudSdkVersionNotifier.getInstance().notifyIfUnsupportedVersion();
+    CloudSdkVersionNotifier.getInstance().notifyIfVersionOutOfDate();
+    CloudSdkVersionNotifier.getInstance().notifyIfVersionParseError();
 
-    helper.createGcloud(loggingHandler).newDeployment(processHandler).deploy(configuration);
+    try {
+      helper.createGcloud(loggingHandler).newDeployment(processHandler).deploy(configuration);
+    } catch (CloudSdkOutOfDateException ex) {
```
ivanporty: So being out of date doesn't in fact mean Cloud SDK cannot successfully deploy application, and the new flow is more user friendly?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/executor/AppEngineStandardRunTask.java
```diff
@@ -78,7 +78,7 @@ public AppEngineStandardRunTask(
   @Override
   public void execute(ProcessStartListener startListener) {
     // show a warning notification if the cloud sdk version is not supported
-    CloudSdkVersionNotifier.getInstance().notifyIfUnsupportedVersion();
+    CloudSdkVersionNotifier.getInstance().notifyIfVersionOutOfDate();
```
ivanporty: `notifyIfVersionParseError()` is not applicable here?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/startup/CloudSdkVersionStartupCheck.java
```diff
@@ -30,6 +30,6 @@
   @Override
   public void runActivity(@NotNull Project project) {
     // If there is a configured Cloud SDK at this time, check that it is supported.
-    CloudSdkVersionNotifier.getInstance().notifyIfUnsupportedVersion();
+    CloudSdkVersionNotifier.getInstance().notifyIfVersionOutOfDate();
```
ivanporty: same question if `notifyIfVersionParseError()` is applicable here?

google-cloud-sdk/testSrc/com/google/cloud/tools/intellij/appengine/java/sdk/DefaultCloudSdkVersionNotifierTest.java
```diff
@@ -44,32 +45,41 @@
   @Test
   public void testNotifyIfCloudSdkNotSupported_isSupported() {
     when(cloudSdkValidator.validateCloudSdk()).thenReturn(new HashSet<>());
-    checker.notifyIfUnsupportedVersion();
+    checker.notifyIfVersionOutOfDate();
 
-    verify(checker, times(0)).showNotification();
+    verify(checker, times(0)).showNotification(anyString(), anyString());
```
ivanporty: Since `showNotification` might receive different input title and message, could it actually make sense to test against actual strings we expect here?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/AppEngineDeploy.java
etanshaul: Out of date always threw an exception. Previously this exception was not caught and the IDE would explode (after showing the nice notification to the user). Now I'm catching the exception to prevent blowing up.

In both cases though, appengine-plugins-core would not let the operation proceed when out of date.

Does that answer the question?

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/cloud/executor/AppEngineStandardRunTask.java
etanshaul: yeah. I think it is. I'll add it.

app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/startup/CloudSdkVersionStartupCheck.java
etanshaul: yes. good catch. ill add.

google-cloud-sdk/testSrc/com/google/cloud/tools/intellij/appengine/java/sdk/DefaultCloudSdkVersionNotifierTest.java
etanshaul: done

etanshaul:
I added a ping when there is an SDK read error. I want to have the actual version in the ping - I'm trying to figure out the best way to do this now.

google-cloud-core/src/com/google/cloud/tools/intellij/analytics/GctTracking.java
```diff
@@ -105,4 +105,5 @@
   public static final String MANAGED_SDK_SUCCESSFUL_UPDATE = "managed.sdk.successful.update";
   public static final String MANAGED_SDK_FAILED_UPDATE = "managed.sdk.failed.update";
   public static final String MANAGED_SDK_CANCELLED_UPDATE = "managed.sdk.cancelled.update";
+  public static final String MANAGED_SDK_VERSION_PARSE_ERROR = "managed.sdk.version.parse.error";
```
ivanporty: Will that apply to custom SDK as well?

etanshaul: oh yeah. It will. I'll just rename the variable. that should cover it I think

etanshaul: done

ivanporty: great thanks!

etanshaul:
I want to include the exception message contents together with the sdk version analytics ping:
i.e. `java.lang.IllegalArgumentException: Pattern "181.1.2.3.1" is not a valid CloudSdkVersion.`

wdyt about updating CloudSdkValidation result so that it can also hold an exception message? Something like:

```
class CloudSdkValidationResult {
  private final ValidationMessage validationMessage;
  private final String exceptionMessage;

  ...

  public enum ValidationMessage {
   // same enum as before, but now an inner enum
  }
}
```

etanshaul:
hmm on second thought that could get kinda messy. thinking of a better approach

ivanporty:
I agree, could be a bit messy to carry around dynamically changing message inside of an `enum`?

etanshaul:
well it wouldnt be inside of the enum as I described there. the enum would just be part of CloudSdkValidationResult (and so would the exception message).

etanshaul:
I think maybe the best approach is in `notifyIfVersionParseError` instead of using SdkValidator and getting back a `CloudSdkValidationResult` to just call `CloudSdk#getVersion` directly. Then I can catch the exception there and do whatever I want with the exception message.

etanshaul:
eh. but then I have to build the CloudSdk again manually... theres a better way here somewhere, I can feel it..

ivanporty:
Good option and easier to understand probably.

etanshaul:
@ivanporty ptal at the last commit. I updated the analytics ping to send the exception message as well (so we can see the version number that cause the error, for example).

ivanporty:
Last update LGTM, thanks

etanshaul:
Updated the notification wording:
image

We continue the operation even if the version could not be parsed so I softened the wording (got rid of "error").

PR #2263, opened 2018-10-03
Bump appengine-plugins-core to 0.6.7

etanshaul:


etanshaul:
This will get us the Cloud SDK version exception handling update

app-engine/java/ultimate/src/com/google/cloud/tools/intellij/appengine/java/ultimate/server/instance/AppEngineServerModel.java
```diff
@@ -352,6 +352,12 @@ public File getDatastorePath() {
     return ImmutableList.of();
   }
 
+  @javax.annotation.Nullable
+  @Override
+  public String getProjectId() {
```
etanshaul: @loosebazooka @coollog this is due to https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/693

IntelliJ currently doesn't expose a Cloud Project field in the devappserver configuration. We can consider adding one to support this feature, but I just want to confirm that things will continue working as before by setting this to `null`?

loosebazooka: It's just to allow multimodule applications to set it. Otherwise it's just set to something generic.

loosebazooka: If you're not explicitly supporting multimodule applications, you can probably ignore this.

loosebazooka: Actually that may not be right. It works for single module applications too. But setting it to null doesn't affect anything.

etanshaul: thanks

PR #2268, opened 2018-10-04
Cloud Debugger: added user name to allow setting of saved cloud project for project selector 

### LINKED ISSUE #2266: Run Configuration: Project selection without effect
- Cloud Tools for IntelliJ version: v18.8.1
- Google Cloud SDK version _(run `gcloud version` on the command line)_:
- OS: Windows 10 Version 1803

**What did you do?**
I created three different "Google Stackdriver Debug" run configs for three different projects. Refering to them as `project`, `project-dev` and `project-staging`.
I selected the Project in each config and it persists, and is even shared.
However, when I run the config, I am asked to "Attach to an Application" and the wrong Project is preselected, e.g. I run the correctly configured `project-staging` config and get a `project` preselection.

Either there should be no popup at all (I would prefer to have an option to enable/disable this) or at least the correct project should be preselected.

ivanporty:
Fixed #2266.

Cloud debug configuration didn't have a google account name in it and it was a required field to properly populate project selector component in the debugger attach dialog. Once this is fixed, attach to debugger dialog will display the project saved in the cloud debug configuration.

### DIFF
--- a/stackdriver-debugger/src/com/google/cloud/tools/intellij/stackdriver/debugger/CloudDebugRunConfiguration.java
+++ b/stackdriver-debugger/src/com/google/cloud/tools/intellij/stackdriver/debugger/CloudDebugRunConfiguration.java
@@ -58,7 +58,8 @@ public class CloudDebugRunConfiguration extends LocatableConfigurationBase
 
   private static final String NAME = "Google Stackdriver Debug";
   private static final String PROJECT_NAME_TAG = "CloudProjectName";
-  private String cloudProjectName;
+  private String cloudProjectId;
+  private String googleUsername;
   private volatile CloudDebugProcessState processState = null;
 
   public CloudDebugRunConfiguration(Project project, @NotNull ConfigurationFactory factory) {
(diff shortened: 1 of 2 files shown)


stackdriver-debugger/src/com/google/cloud/tools/intellij/stackdriver/debugger/CloudDebugSettingsEditor.java
```diff
@@ -41,11 +39,14 @@
...
-    // TODO(ivanporty) CloudDebugRunConfiguration uses `cloudProjectName` for project ID.
-    runConfiguration.setCloudProjectName(
+    runConfiguration.setCloudProjectId(
         Optional.ofNullable(settingsPanel.getSelectedCloudProject())
             .map(CloudProject::projectId)
             .orElse(""));
+    runConfiguration.setGoogleUsername(
+        Optional.ofNullable(settingsPanel.getSelectedCloudProject())
+            .map(CloudProject::googleUsername)
+            .orElse(null));
```
etanshaul: why null here but empty string for the project ID?

stackdriver-debugger/src/com/google/cloud/tools/intellij/stackdriver/debugger/CloudDebugSettingsEditor.java
```diff
@@ -56,19 +57,14 @@ protected JComponent createEditor() {
...
           CloudProject.create(
               // TODO(ivanporty) no project name in CloudDebugRunConfiguration.
-              projectId,
-              projectId,
-              null,
-              // TODO(ivanporty) no user name in CloudDebugRunConfiguration.
-              Optional.ofNullable(Services.getLoginService().getActiveUser())
-                  .map(CredentialedUser::getEmail)
-                  .orElse("")));
+              projectId, projectId, null, googleUsername));
```
etanshaul: I know it was there before but could you please at a named param comment to the null param? (`.., null /* name*/, ..`)

stackdriver-debugger/src/com/google/cloud/tools/intellij/stackdriver/debugger/CloudDebugRunConfiguration.java
```diff
@@ -148,7 +159,7 @@ public void setProcessState(@Nullable CloudDebugProcessState processState) {
   public RunProfileState getState(
       @NotNull Executor executor, @NotNull ExecutionEnvironment environment) {
     if (processState == null) {
-      return new CloudDebugProcessState(null, null, cloudProjectName, null, getProject());
+      return new CloudDebugProcessState(googleUsername, null, cloudProjectId, null, getProject());
```
etanshaul: same comment here for the nulls (named param comments so its easier to read quickly).

etanshaul:
thanks for fixing so fast. just a couple minor comments

stackdriver-debugger/src/com/google/cloud/tools/intellij/stackdriver/debugger/CloudDebugSettingsEditor.java
ivanporty: Should be null for project ID too, changed.

PR #2276, opened 2018-10-08
Notify on missing deployment artifact during local run

etanshaul:
Follow up to #2275:

The original issue was happening when attempting to do a local run without a deployment artifact present.

image

Attempting to run this config results in:

image

Click "continue anyway", before #2275 resulted in an NPE. After #2275, the local run would proceed, but then fail in the appengine-plugins-core library due to a failed precondition of `RunConfiguration#getServices` not being null or empty.

This follows up and explicitly checks that `#getServices` is populated, and if not, shows a notification to the user:

image


etanshaul:
@ivanporty @patflynn FYI. this fixes the root of the issue we discussed today.

Issue #2279, opened 2018-10-09
Make compatible with IntelliJ 2018.3


hosswald:
I am using IntelliJ 2018.3 EAP and I can't update Cloud Tools to the newest version, because the declared compatibility range ist 171.3780—182.*, while 2018.3 is 183.*.
Editing `google-cloud-tools-*.zip\google-cloud-tools\lib\google-cloud-tools-plugin-*.jar\META-INF\plugin.xml` manually and importing the modified zip file in IntelliJ, it works perfectly well.
Please make the plugin compatible with IntelliJ 2018.3.

PR #2284, opened 2018-10-10
Investigate EAP fix

etanshaul:
This is to look into a fix for #2279

### DIFF
--- a/google-cloud-core/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
+++ b/google-cloud-core/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
@@ -181,7 +181,9 @@ public class ProjectSelectionDialog {
     }
   }
 
-  /** Check if cached, if not then check if already loading, and finally start loading if not. */
+  /**
+   * Check if cached, if not then check if already loading, and finally start loading if not.
+   */
   private void loadProjectList(CredentialedUser user) {
     if (!cachedProjectList.containsKey(user) && !runningProjectLoaderJobs.containsKey(user)) {
       // not cached and not loading - start loading project list here and keep the future reference.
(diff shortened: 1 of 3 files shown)


google-cloud-core/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -210,8 +212,8 @@ public void onFailure(Throwable throwable) {
                     String errorMessage =
                         Optional.ofNullable(throwable.getMessage())
                             .orElse(throwable.getClass().getName());
-                    dialogWrapper.setErrorInfoAll(
-                        Collections.singletonList(new ValidationInfo(errorMessage)));
+//                    dialogWrapper.setErrorInfoAll(
```
etanshaul: @ivanporty you are closer to this code - do you recall what exactly #setErrorInfoAll is doing and if there is an alternatice?

In the eap, this method was made final, so we can no longer implement it in the DialogWrapper (`ProjectSelectionDialogWrapper` inner class).  I commented it out here just to get things to compile

google-cloud-core/src/com/google/cloud/tools/intellij/project/ProjectSelectionDialog.java
```diff
@@ -210,8 +209,9 @@ public void onFailure(Throwable throwable) {
                     String errorMessage =
                         Optional.ofNullable(throwable.getMessage())
                             .orElse(throwable.getClass().getName());
-                    dialogWrapper.setErrorInfoAll(
-                        Collections.singletonList(new ValidationInfo(errorMessage)));
+                    //                    dialogWrapper.setErrorInfoAll(
```
etanshaul: comment got collapsed after reformatting. adding again:

@ivanporty you are closer to this code - do you recall what exactly #setErrorInfoAll is doing and if there is an alternative?

In the eap, this method was made final, so we can no longer implement it in the DialogWrapper (ProjectSelectionDialogWrapper inner class). I commented it out here just to get things to compile

etanshaul:
@ivanporty updated, ptal.

If you have suggestions on how to specifically test this change to the projectSelector that would be helpful.

ivanporty:
This is called when fetching project list over the network resulted in an error and should show built-in error message pane. Probably can test that by throwing exception in `ProjectLoader`? It could be really hard to actually emulate network or I/O issue. We can also add a unit test for that, but that can involve some heavy lifting with UI mocks.

PR #2298, opened 2018-11-26
Temporarily pin app engine local run action to the top of the menu

etanshaul:
(cherry picked from commit af41c61)

Note this was part of the 18.11 patch release. Just bringing master up to date.

This is a temporary workaround for: https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/2176

Added in because in release 2018.3 of IntelliJ, this warning is now an IDE level error, so we needed a fix (or workaround for this) to properly support 2018.3.

What this does is pin the app engine local run action to the top of the Google Cloud Tools menu. The error in the linked issue was happening because the relative ordering used previously was generating errors due to nondeterministic load order of the action groups. The next step is to refactor with a proper relative order fix.


etanshaul:
lol. me too.

etanshaul:
for tracking purposes: 

I was able to reproduce the core issue here with a trimmed down plugin. I filed an Issue on Intellij:
https://youtrack.jetbrains.com/issue/IDEA-203089

PR #2338, opened 2019-02-07
Fixing the name of run configuration

redborian:


redborian:
I realized that there are two repos: google-cloud-intellij and google-container-tools-intellij. Do you know which one contains the kubernetes intellij plugin?

etanshaul:
The container-tools repo is the K8s plugin

PR #2347, opened 2019-02-20
Clean up build badges on root README

etanshaul:
with cr/234858531 we are added build badge labels, so this table structure won't be necessary. This makes things cleaner visually.


etanshaul:
@ivanporty  thanks for the review, could you also please check cr/234858531? These both need to go in for this change to work

PR #2357, opened 2019-02-22
Update k8s readme in preparation for editing PR merge

etanshaul:


etanshaul:
actually makes more sense jsut to do this in the editin PR. closing

PR #2358, opened 2019-02-23
buildResult() needs to be called before checkSyncStashState()

ZariaHoward:
It looks like buildResult needs to be initialized before calling checkSyncStashState. otherwise checkSyncStashState can be called before processResultState is initialized, which is what would throw the error.


ZariaHoward:
this commit fixes #116

Issue #2343, opened 2019-02-19
GCS side-panel should be hidden by default


etanshaul:
Currently the Google Cloud Storage side panel is visible by default whenever our plugin is installed:
![image](https://user-images.githubusercontent.com/1735744/53052449-c8b61800-346c-11e9-9959-98c517e15f5b.png)

Now that the container-tools repo (k8s work) is merged into this repo, it makes more sense for this panel to only open on-demand (for cloud-agnostic k8s use-cases).

Therefore, the side-tab should not be visible by default. Instead the user can manually open it (either via our tools action, or via View -> Tools Window).

PR #2374, opened 2019-02-27
Hide GCS side panel by default

ZariaHoward:
fixes issue #2343


google-cloud-storage/src/com/google/cloud/tools/intellij/gcs/GcsToolWindowFactory.java
```diff
@@ -41,5 +41,6 @@ public void createToolWindowContent(@NotNull Project project, @NotNull ToolWindo
   @Override
   public void init(ToolWindow window) {
     window.setIcon(GoogleCloudStorageIcons.CLOUD_STORAGE);
+    window.setShowStripeButton(false);
```
etanshaul: what is a "stripe" button? Is that the term they use for the side panel tabs?

ZariaHoward: Yes it is.

PR #2397, opened 2019-03-01
Add skaffold executable selector to new k8s settings

etanshaul:
Still need to clean up the PR, but for feedback, @ivanporty wdyt:
image

This will help provide an escape hatch when Skaffold cannot be automatically detected (e.g. due to PATH issues).

This PR adds the UI and persistence. Next PR will consume this, and link to it from any error messages.


ivanporty:
Looks good, definitely helpful in case of hard issues with path and system variances.

etanshaul:
@ivanporty ready for review

kubernetes/core/src/main/resources/META-INF/core.xml
```diff
@@ -18,6 +18,9 @@
 <idea-plugin>
     <extensions defaultExtensionNs="com.intellij">
         <applicationService serviceImplementation="com.google.kubernetes.tools.core.PluginInfo"/>
+        <applicationService serviceImplementation="com.google.container.tools.core.settings.KubernetesSettingsService"/>
 
+        <!--<applicationConfigurable parentId="google.settings" id="google.kubernetes.settings"-->
```
etanshaul: commented out to hide the new menu item until this configurable path is consumed in the next PR

kubernetes/core/src/main/kotlin/com/google/container/tools/core/settings/KubernetesSettingsConfigurable.kt
```diff
@@ -0,0 +1,85 @@
...
+    val skaffoldBrowser = TextFieldWithBrowseButton()
+
+    init {
+        skaffoldBrowser.addBrowseFolderListener(
+                CoreBundle.message("kubernetes.settings.dependencies.skaffold.browse.title"),
+                null /*description*/,
+                null /*project*/,
+                FileChooserDescriptor(
+                        true /*chooseFiles*/,
+                        false /*chooseFolders*/,
```
ivanporty: Will we choose files with this file chooser? Would it make sense to actually select folder where skaffold executable is located?

etanshaul: I was thinking that this will choose the actual executable file.

In linux, for example, its placed (according to the install instructions) in `/usr/local/bin`. I think it makes more sense to browse to the actually skaffold executable rather than this bin directory.

wdty?

etanshaul: Also make note of the wording of the label on the browser control `Path to Skaffold executable` in case you think it could be more clear.

kubernetes/core/src/main/kotlin/com/google/container/tools/core/settings/KubernetesSettingsConfigurable.kt
```diff
@@ -0,0 +1,85 @@
...
+            }
+
+            noteRow(CoreBundle.message("kubernetes.settings.dependencies.skaffold.note"))
+        }
+
+        dependenciesPanel.border = IdeaTitledBorder(
+                CoreBundle.message(
+                        "kubernetes.settings.dependencies.panel.title"), 0, Insets(0, 0, 0, 0))
+
+        return panel {
```
etanshaul: note that I'm wrapping this in another panel because otherwise, the `IdeaTitleBorder` that I'm setting won't show up. I guess you cannot apply a border to the outer panel

kubernetes/core/src/main/kotlin/com/google/container/tools/core/settings/KubernetesSettingsConfigurable.kt
ivanporty: Nothing wrong with selecting a particular file in general, but it seems more of a convention to select a directory where some tool lives - i.e. Maven, Gradle, our selection for custom Cloud SDK, all of this in the same IntelliJ. It could be more in line with IDE experience to select a directory where something is, or you think it might be easier to actually point to skaffold file?

kubernetes/core/src/main/kotlin/com/google/container/tools/core/settings/KubernetesSettingsConfigurable.kt
ivanporty: It seems to work directly on `basePanel` in  `BaseSkaffoldSettingsEditor` but there might be a different case with settings - most probably IDE overrides border there so this looks fine. Maybe a small comment will be helpful for future readers?

kubernetes/core/src/main/kotlin/com/google/container/tools/core/settings/KubernetesSettingsConfigurable.kt
etanshaul: The directory selection seems more appropriate for an sdk style tool where, for you example, you have a bin directory with multiple executables, whereas skaffold is just a single binary.

One example like this that I could find is the docker tool selection (Build, Execution, Deployment -> Docker > Tools) where you browse to the docker machine and docker compose executable.

kubernetes/core/src/main/kotlin/com/google/container/tools/core/settings/KubernetesSettingsConfigurable.kt
etanshaul: good idea. done

kubernetes/core/src/main/kotlin/com/google/container/tools/core/settings/KubernetesSettingsConfigurable.kt
ivanporty: good point with SDK style directories. Let's use file chooser, hopefully we won't have to worry much about this anyway once managed CLIs are with us.

Issue #2344, opened 2019-02-19
Refactor the action tools menu


etanshaul:
per the doc where this is described:

```
The current “Google Cloud Tools” menu option is not scalable. Need to overhaul it. Possibly as follows (> is a folder, - is a action):

> Cloud Code by Google
  > Kubernetes
        [Any k8s actions here; none at the moment]
  > Google Cloud
        > App Engine
              - Run on a local development server
              - Deploy to App Engine
        > Google Cloud Libraries
              - Manage APIs and client libraries
        > Stackdriver Debugger
              - Attach Stackdriver debugger
        > Cloud Storage
              - Browse Cloud Storage buckets
       > Cloud Source Repositories
              -Upload to Cloud Source Repositories
- Submit feedback or report an issue
```

PR #2398, opened 2019-03-01
Moved app engine menu items into a separate menu group.

ivanporty:
Part of #2344.

app-engine-new-group


etanshaul:
lgtm. we should probably also remove the GCP icon from the root "Cloud Tools" menu item (until we have an icon created for us). could you remove that? (either here or in a separate pr)

PR #2400, opened 2019-03-01
Moves debugger menu into a separate action group.

ivanporty:
Part of #2344.

debugger-menu-group


ivanporty:
Thought about that too, good idea. I'll do that

PR #2401, opened 2019-03-01
Consume skaffold executable path override

### LINKED ISSUE #2387: Add override to point to local Skaffold installation
As a temporary workaround for the PATH headaches we've had (until we have a full installer) I was thinking it would be good to have the ability to manually select a local installation (similar to how we do for gcloud). This should be fairly simple to implement too.

@ivanporty wdyt?

conversation (migrated from container-tools):

@ivanporty:
> I see the point, agree, finding it in PATH unless it's really in one of the default system directories is painful. It would be awesome to just use managed install but until then we should probably allow selecting local path.

@etanshaul:
>Which reminds me..I have a fix pending some cleanup that will try first to find it in the well-known location if it’s not on the PATH. That should cover _most_ cases.

etanshaul:
fixes #2387

follow up to #2397. Updates skaffold execution to use the override path if it exists. Also updates the warning messages.

image

Clicking 'fix' takes you to the settings:
image
After selecting the executable, the user is returned the run config and it is no longer in an errored state.

Also updated run error message:
image


kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/SkaffoldExecutorService.kt
```diff
@@ -177,10 +184,21 @@ data class SkaffoldExecutorSettings(
...
+ * from the Kubernetes settings, and then the PATH if unavailable in the settings.
  */
 class DefaultSkaffoldExecutorService : SkaffoldExecutorService() {
-    // use executable available in PATH
-    override var skaffoldExecutablePath: Path = Paths.get("skaffold")
+    /**
+     * First try to find Skaffold in the Kubernetes settings under Google -> Kubernetes.
+     * If not found, then try to find it on the user's PATH.
+     */
+    override fun getSkaffoldExecutablePath(): Path {
```
ivanporty: Can we keep the property but provide a custom getter?

kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/SkaffoldExecutorService.kt
```diff
@@ -43,16 +44,22 @@ abstract class SkaffoldExecutorService {
...
+        val skaffoldPathOverride = KubernetesSettingsService.instance.skaffoldExecutablePath
+
+        val isSkaffoldOnPath = getSystemPath().split(File.pathSeparator)
+                .asSequence()
+                .map { it + File.separator + "skaffold" }
+                .any {
+                    File(it).exists() && File(it).canExecute()
+                }
+
+        return skaffoldPathOverride.isNotBlank() || isSkaffoldOnPath
```
ivanporty: Does it make sense to check skaffoldPathOverride for validity and presence of skaffold executable as well to be 100% sure it works?

etanshaul: I think that would be a nice follow up. We currently aren’t checking that the executable on the path works either. How about supplement this in another PR?

kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/SkaffoldExecutorService.kt
etanshaul: Yep. That works

etanshaul: Yep. That works

etanshaul: hmm with a getter, I'm [forced to initialize it](https://stackoverflow.com/questions/41440750/kotlin-why-do-i-need-to-initialize-a-var-with-custom-getter), it seems. Something like this which isn't very nice:

```
override var skaffoldExecutablePath: Path? = null
  get() {
    // ...
  }
```

kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/SkaffoldExecutorService.kt
etanshaul: https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/2404

kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/run/SkaffoldRunConfigurations.kt
```diff
@@ -119,7 +121,13 @@ abstract class AbstractSkaffoldRunConfiguration(
 
     override fun checkConfiguration() {
         if (!SkaffoldExecutorService.instance.isSkaffoldAvailable()) {
-            throw RuntimeConfigurationWarning(message("skaffold.run.config.not.on.system.error"))
+            throw RuntimeConfigurationWarning(message("skaffold.run.config.not.on.system.error"),
+                    Runnable {
```
ivanporty: Oh that's how they implement this nice `fix it` action?

etanshaul: yep, its pretty convenient

kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/SkaffoldExecutorService.kt
ivanporty: we can probably just initialize it to `Paths.get(skaffold)` but I'm ok with either way

etanshaul: thats true. I'll do that.

etanshaul: done - refactored to custom getter using `field` accessor

PR #2407, opened 2019-03-04
Finalizes GCP menu group, removes extra icons.

### LINKED ISSUE #2344: Refactor the action tools menu

ivanporty:
Closes #2344.

It seems we cannot enforce alphabetical order due to randomness of loading order, but at least App Engine always goes first and Stackdriver Debugger always comes last so this makes it a bit more structured.

gcp-overall-menu-group

### DIFF
--- a/app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/util/AppEngineGeneralMenuGroupAction.java
+++ b/app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/util/AppEngineGeneralMenuGroupAction.java
@@ -16,7 +16,6 @@
 
 package com.google.cloud.tools.intellij.appengine.java.util;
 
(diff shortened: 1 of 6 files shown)


etanshaul:
we probably want to update the root "Cloud Tools" icon too later. I'll check if we have one yet.

PR #2414, opened 2019-03-05
Adds a Kubernetes FAQ doc page

etanshaul:
Not for review yet.


etanshaul:
@ivanporty ready for review

I would still to find a better solution for the general PATH problem (which won't be solved by the dependency manager since transitive dependencies like `docker-credential-gcr` are looked for on the PATH). But at least this is an initial workaround.

etanshaul:
actually I think I found a better solution. PR coming soon.

etanshaul:
@ivanporty updated the FAQ to have workarounds for Linux only. https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/2420 should fix the problem in OSX, and I don't think this problem exists in windows as long as the binaries are on the PATH.

PR #2420, opened 2019-03-06
 Create processes using GeneralCommandLine to inherit parent console environment

### LINKED ISSUE #2394: Skaffold run configs do not see shell environment variables (PATH) on Linux when IDE starts from desktop
```
time="2018-11-01T12:46:38-04:00" level=fatal msg="exiting dev mode because the first build failed: building [gcr.io/**/my-proj]: tagging: pushing: getting auth config for gcr.io/**/my-proj:3c275201-dirty-50f870d: getting auth config: error getting credentials - err: exec: \"docker-credential-gcr\": executable file not found in $PATH, out: ``"
```

possibly related: https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000464664-IDEA-sandbox-problems-with-shell-command

see https://github.com/GoogleContainerTools/google-container-tools-intellij/issues/67 for full discussion

etanshaul:
fixes #2394

I think this fixes the issue. Some more testing is need to confirm. (I thought we had tried this before and it didn't work, but my initial testing shows that it does..).


etanshaul:
This solves the issue for OSX. per the javadocs:

>  {@code CONSOLE} provides the child with a similar environment as if it was launched from, well, a console.
   * On OS X, a console environment is simulated (see {@link EnvironmentUtil#getEnvironmentMap()} for reasons it's needed
   * and details on how it works). On Windows and Unix hosts, this option is no different from {@code SYSTEM}
   * since there is no drastic distinction in environment between GUI and console apps.

for Linux (and windows too I think) we still ran into this problem which seems to contradict the javadoc ("there is no drastic distinction in environment between GUI and console apps). For example, in Linux when the variables are defined in `.bash_profile` they are not picked up by the GUI. When defining them in `/etc/environment` however, they are.

PR #2425, opened 2019-03-07
Updates Skaffold existence check to execute the process

### LINKED ISSUE #2424: Skaffold existence check in run configs should align system PATH
Now that https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/2420 is merged in, process execution in OSX will inherit the system's path. However, our Skaffold existence pre-checks (in the Run Configuration) rely on rudimentary PATH checks that do not inherit the system path. We need to align these pre-checks to match what actually happens at execution time.

The problem now is that the user will be shown an error in the run config (and on execution), even though the execution of the command would have worked.

etanshaul:
fixes #2424 

Instead of checking the PATH variable manually via `System.getenv("PATH")` which may not inherit the system PATH, actually attempt to execute a Skaffold command instead. 

See #2424 for details, but the gist of why we want to do this is so that the error messaging in the run config matches what will actually happen when Skaffold executes

### DIFF
--- a/kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/SkaffoldExecutorService.kt
+++ b/kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/SkaffoldExecutorService.kt
@@ -30,6 +30,7 @@ import com.intellij.openapi.components.ServiceManager
 import java.io.File
 import java.nio.file.Path
 import java.nio.file.Paths
+import java.util.concurrent.TimeUnit
 
 /**
  * Abstract implementation for Skaffold execution service. This service builds and launches Skaffold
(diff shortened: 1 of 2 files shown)


kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/SkaffoldExecutorService.kt
```diff
@@ -47,19 +47,20 @@ abstract class SkaffoldExecutorService {
...
-
+    /**
+     * Checks if Skaffold is available by executing a 'skaffold version' command. If the process
+     * exit code is 0, then Skaffold is considered available.
+     */
     fun isSkaffoldAvailable(): Boolean {
-        val skaffoldPathOverride = KubernetesSettingsService.instance.skaffoldExecutablePath
+        return try {
+            val process = executeSkaffold(SkaffoldExecutorSettings(ExecutionMode.VERSION)).process
+            process.waitFor()
```
etanshaul: note this executes on the order of 10ms so we should be ok

ZariaHoward: you're right but it might be good to use these params just to be safe (and i guess for other people reading it to know it wont go on for more than 10 secs):

@param timeout -  the maximum time to wait
@param unit - the time unit of the {@code timeout} argument

etanshaul: great points. Let me refactor a bit to make it more robust

etanshaul: updated. also with additional unit test

PR #2429, opened 2019-03-08
Show warning in the kubernetes settings if skaffold is not executable

### LINKED ISSUE #2428: Add validation to Skaffold path override

etanshaul:
fixes #2428 

In order to avoid circular dependencies I did a couple things:
1) Moved the `KubernetesSettingsConfigurable` from `core` to the root module. This allows it to access the Skaffold module which it needs in order to execute Skaffold
2) In the Skaffold module, the `checkConfiguration` method which applies a quick fix and opens the settings, now doesn't depend on `KubernetesSettingsConfigurable` directly - otherwise it would have been a circular dependency. Instead it loads it dynamically from the list of `ApplicationConfigurables`.

This PR adds validation to the Kubernetes dependency settings panel. Now, the status shown for the Skaffold dependency should match what is shown in the Run Configuration and should match what happens at execution time:

image

### DIFF
--- a/kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/run/SkaffoldRunConfigurations.kt
+++ b/kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/run/SkaffoldRunConfigurations.kt
@@ -17,7 +17,6 @@
 package com.google.kubernetes.tools.skaffold.run
 
 import com.google.common.annotations.VisibleForTesting
-import com.google.kubernetes.tools.core.settings.KubernetesSettingsConfigurable
 import com.google.kubernetes.tools.skaffold.SkaffoldExecutorService
 import com.google.kubernetes.tools.skaffold.SkaffoldExecutorSettings
 import com.google.kubernetes.tools.skaffold.message
@@ -31,6 +30,7 @@ import com.intellij.execution.configurations.RunProfileState
 import com.intellij.execution.configurations.RuntimeConfigurationWarning
 import com.intellij.execution.executors.DefaultDebugExecutor
 import com.intellij.execution.runners.ExecutionEnvironment
+import com.intellij.openapi.options.Configurable
 import com.intellij.openapi.options.SettingsEditor
 import com.intellij.openapi.options.ShowSettingsUtil
 import com.intellij.openapi.project.Project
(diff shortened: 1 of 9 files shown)


kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/run/SkaffoldRunConfigurations.kt
```diff
@@ -123,10 +123,16 @@ abstract class AbstractSkaffoldRunConfiguration(
         if (!SkaffoldExecutorService.instance.isSkaffoldAvailable()) {
             throw RuntimeConfigurationWarning(message("skaffold.run.config.not.on.system.error"),
                     Runnable {
-                        ShowSettingsUtil.getInstance()
-                                .showSettingsDialog(
-                                        project,
-                                        KubernetesSettingsConfigurable::class.java)
+                        val kubernetesConfigurable =
+                                Configurable.APPLICATION_CONFIGURABLE.extensionList.find {
+                            it.id == "google.kubernetes.settings"
```
ivanporty: Could we make it a constant? And it's not accessible directly from the class, right?

kubernetes/src/main/kotlin/com/google/kubernetes/tools/settings/KubernetesSettingsConfigurable.kt
```diff
@@ -46,6 +60,35 @@ class KubernetesSettingsConfigurable : Configurable {
...
+            val skaffoldExecutorService = object : SkaffoldExecutorService() {
+                override var skaffoldExecutablePath: Path = Paths.get(skaffoldBrowser.text)
+            }
+
+            override fun changedUpdate(event: DocumentEvent?) {
+                checkSkaffold()
+            }
+
+            override fun insertUpdate(event: DocumentEvent?) {
+                checkSkaffold()
```
ivanporty: What happens if you type in the field? Is it fast enough - it seems to call skaffold on each type?

kubernetes/src/test/kotlin/com/google/kubernetes/tools/settings/KubernetesSettingsConfigurableTest.kt
```diff
@@ -107,4 +108,12 @@ class KubernetesSettingsConfigurableTest {
 
         Truth.assertThat(kubernetesSettingsConfigurable.isModified).isFalse()
     }
+
+    @Test
+    @UiTest
+    fun `skaffold warning is not shown by default`() {
```
ivanporty: Could we also add an opposite test, when Skaffold check fails, the warning is visible?

kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/run/SkaffoldRunConfigurations.kt
etanshaul: will do.

> And it's not accessible directly from the class, right?

No, its not. This was part of the work to cut out the circular dependency. The Configurable lives in the root module (so the submodules don't have access to it since the dependencies flow in the other direction).

kubernetes/src/main/kotlin/com/google/kubernetes/tools/settings/KubernetesSettingsConfigurable.kt
etanshaul: yeah. I saw no perceptible slow down in my tests.

kubernetes/src/test/kotlin/com/google/kubernetes/tools/settings/KubernetesSettingsConfigurableTest.kt
etanshaul: done

kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/run/SkaffoldRunConfigurations.kt
etanshaul: done

PR #2437, opened 2019-03-11
Reduce title length to make it readable when plugin is initially installed.

ivanporty:
When the plugin is initially installed into a fresh IDE never configured for usage statistics before, the notification is shown at the start, clipping the title at 3-4 words making it essentially unreadable.

### DIFF
--- a/google-cloud-core/src/com/google/cloud/tools/intellij/analytics/UsageTrackerNotification.java
+++ b/google-cloud-core/src/com/google/cloud/tools/intellij/analytics/UsageTrackerNotification.java
@@ -46,29 +46,26 @@ public class UsageTrackerNotification {
   /** Show the notification panel. */
   public void showNotification() {
     NotificationListener listener =
-        new NotificationListener() {
-          @Override
-          public void hyperlinkUpdate(Notification notification, HyperlinkEvent event) {
-            if (event.getEventType() == HyperlinkEvent.EventType.ACTIVATED) {
-              final String description = event.getDescription();
-              if ("allow".equals(description)) {
-                usageTrackingManagementService.setTrackingPreference(true);
-                notification.expire();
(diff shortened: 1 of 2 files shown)


google-cloud-core/src/com/google/cloud/tools/intellij/analytics/UsageTrackerNotification.java
```diff
@@ -46,29 +46,26 @@ public static UsageTrackerNotification getInstance() {
...
-                  BrowserUtil.browse(new URL(UsageTrackerPanel.PRIVACY_POLICY_URL));
-                } catch (MalformedURLException ex) {
-                  LOG.error(ex);
-                }
-                notification.expire();
-              } else if ("settings".equals(description)) {
-                final ShowSettingsUtil util = ShowSettingsUtil.getInstance();
-                util.showSettingsDialog(null, UsageTrackerConfigurable.class);
-                notification.expire();
+        (notification, event) -> {
```
etanshaul: I think this is the case, but just to make sure I didn't miss anything -> this is just a formatting change (and lambda refactoring) right?

etanshaul:
could you post a screenshot possibly?

google-cloud-core/src/com/google/cloud/tools/intellij/analytics/UsageTrackerNotification.java
ivanporty: right

ivanporty:
Need to re-create new user and install fresh IDE - only guaranteed option to see it again, then I'll try to catch a screenshot.

ivanporty:
That's how it looked originally, so shorter title and line-by-line text should make it more readable.

Screen Shot 2019-03-11 at 5 36 34 PM

PR #2441, opened 2019-03-12
Fix for reading shell environment in linux

etanshaul:
Tested working in Linux. 

TODO:
- [x] Test to make sure this works as expected in other OSs.

### DIFF
--- a/kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/SkaffoldExecutorService.kt
+++ b/kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/SkaffoldExecutorService.kt
@@ -155,7 +155,8 @@ abstract class SkaffoldExecutorService {
                 GeneralCommandLine.ParentEnvironmentType.CONSOLE)
 
         // For some environments (e.g. *nix) the shell environment is not accessed from the GUI
-        // this ensures that the shell environment is read and included
+        // this ensures that the shell environment is read and included. If the shell environment
+        // reader returns an empty map, [GeneralCommandLine#withEnvironment] will be a noop.
         try {
             generalCommandLine.withEnvironment(EnvironmentUtil.ShellEnvReader().readShellEnv())
         } catch (e: Exception) {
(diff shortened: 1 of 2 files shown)


etanshaul:
@ivanporty ready for review

kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/SkaffoldExecutorService.kt
```diff
@@ -149,6 +150,11 @@ abstract class SkaffoldExecutorService {
         val generalCommandLine = GeneralCommandLine(commandList)
         generalCommandLine.withParentEnvironmentType(
                 GeneralCommandLine.ParentEnvironmentType.CONSOLE)
+
+        // For some environments (e.g. *nix) the shell environment is not accessed from the GUI
+        // this ensures that the shell environment is read and included
+        generalCommandLine.withEnvironment(EnvironmentUtil.ShellEnvReader().readShellEnv())
```
ivanporty: Looking at the implementation, it seems there is a possibility for a few exceptions. Probably we should catch them separately and then just use standard environment if it happens. Also, how about checking if this method returns empty map for environment and then skip it since it won't be usable and reduce the success chances for CLI?

etanshaul: agreed about exception handling.

> Also, how about checking if this method returns empty map for environment and then skip it since it won't be usable and reduce the success chances for CLI?

In this case, it will just call `generalCommandLine.withEnvironment` with an empty map. The implementation here just does a `putAll` so an empty map will be a noop.

ivanporty: Cool, probably a small comment on that would be good - from the method name it's hard to say if it replaces this or adds.

etanshaul: thanks.

updated

Issue #2445, opened 2019-03-12
Skaffold quick fix link doesn't jump directly to the Kubernetes settings menu


etanshaul:
instead it jumps to the last menu item that you had selected

etanshaul:
obsolete. see #2449

PR #2449, opened 2019-03-13
Remove eager skaffold validation

### LINKED ISSUE #2448: Optimize Skaffold presence check when editing run configuration
Experiencing significant delays (Linux) when typing default repo name, each character takes some time to load.

etanshaul:
fixes #2448 and makes #2445 obsolete

What this does:
- Removes eager skaffold validation in the following locations
  - In the Run Configuration. Previously the run configuration would show a warning when the skaffold executable was not valid. This was leading to #2448.
  - In the Kubernetes settings menu. This check was also leading to #2448 

Rationale:
1) We still get a nice warning when the Run Configuration is executed and the executable is not detected:
image

2) All other examples in the IDE I could find of executable paths did not do any validation on them at configuration time.

3) I could not quickly find a good way to execute an asynchronous validation in the Run Configuration. Similarly we don't want to block saving the run configuration - so having a check-on-save is ruled out.

Possible future enhancement:
In the Kubernetes settings menu, we can restore this logic by introducing asynchronous validation.


etanshaul:
oops need to clean up old tests

PR #2453, opened 2019-03-13
 Split and simplify README and getting started docs for Kubernetes.

ivanporty:


kubernetes/README.md
```diff
@@ -12,10 +12,6 @@ This plugin adds support for [Kubernetes](https://www.kubernetes.io) development
   * [Installing the plugin into your IDE](#installing-the-plugin-into-your-ide)
   * [Frequently asked questions](#frequently-asked-questions-faq)
   * [Getting started](#getting-started)
-     * [Specifying image repository](#specifying-image-repository)
-     * [Continuous development on Kubernetes](#continuous-development-on-kubernetes)
-        * [Adding new features](#adding-new-features)
-     * [Deployment to Kubernetes](#deployment-to-kubernetes)
   * [Setup existing Kubernetes projects with the plugin](#setup-existing-kubernetes-projects-with-the-plugin)
```
etanshaul: nit that I missed before: "Setup" -> "Set up"

ivanporty:
regarding the docs links we should be good since most of the links remain the same, and the new doc is right under the new title. I will double check the alpha release UX doc.

Issue #2427, opened 2019-03-08
Hello World with Jib fails on Windows


ivanporty:
When trying to deploy the getting started project on windows, Skaffold starts, then fails with the following message:

```
time="2019-03-08T11:13:45-05:00" level=fatal msg="watching files for artifact gcr.io/gcp-dev-tools/hello-spring-boot: listing files: 
listing files: getting jibMaven dependencies: unable to stat file Found \"C:\\Users\\***\\google-cloud-intellij\\kubernetes\\examples\\hello-spring-boot\\.mvn\\wrapper\\maven-wrapper.jar\": CreateFile Found \"C:\\Users\\***\\google-cloud-intellij\\kubernetes\\examples\\hello-spring-boot\\.mvn\\wrapper\\maven-wrapper.jar\": The filename, directory name, or volume label syntax is incorrect."
```

ivanporty:
@chanseokoh @TadCordle Could this be something related to Jib calling Maven wrapper in Windows, have you encountered this before?

ivanporty:
Details: `skaffold dev` fails - something wrong with files watching, whereas `skaffold run` (deployment) runs fine.

etanshaul:
@briandealwis for any insight as well

briandealwis:
@ivanporty did you replace your user name with `***`?  What's odd is that Skaffold-Jib will never reference the maven wrapper's jar: it just invokes the `mvnw` script if found. Try running Skaffold with `-v debug` to see the debug output. Does the example work if you build  from the comm-line?

chanseokoh:
The message "The filename, directory name, or volume label syntax is incorrect" seems like coming from the OS. Maybe there's something that Windows doesn't like about the given path string.

https://stackoverflow.com/questions/24116311/the-filename-directory-name-or-volume-label-syntax-incorrect
https://stackoverflow.com/questions/33618656/python-windowserror-error-123-the-filename-directory-name-or-volume-label-s
https://stackoverflow.com/questions/24782826/the-filename-directory-name-or-volume-label-syntax-is-incorrect-inside-batch

chanseokoh:
Hmm... but some folks say permission problems may cause this too: https://stackoverflow.com/a/33824730/1701388, https://stackoverflow.com/a/2707804/1701388

ivanporty:
@briandealwis to answer your question, it fails the same way with CLI call.

Debug output for dev command:
```
[...]

PR #2454, opened 2019-03-13
Adds a docker profile to the example project

etanshaul:
escape hatch for #2427 

FAQ entry coming (update: done)


kubernetes/examples/hello-spring-boot/skaffold.yaml
```diff
@@ -5,8 +5,13 @@ build:
   - image: gcr.io/gcp-dev-tools/hello-spring-boot
     jibMaven: {}
 
-# optional profile to run the jib build on Google Cloud Build
 profiles:
+  # optional profile to run the Jib build on Google Cloud Build
   - name: gcb
     build:
       googleCloudBuild: {}
+  # optional profile to run the build using Docker instead of Jib
```
etanshaul: @briandealwis or @chanseokoh 

In this `docker` profile I'm trying to override the Jib builder to just use a local Docker builder. However, doing something like the following:
```
profiles: 
   - name: docker
     build: 
         local: {}
```
Doesn't override the Jib builder. Is there a more elegant way to do this instead of redefining the artifacts / image like I'm doing here?

etanshaul: @ivanporty I asked around and haven't heard a good solution so we can merge with this for now (it works).

etanshaul:
nice. thanks for giving it a test drive

kubernetes/examples/hello-spring-boot/skaffold.yaml
briandealwis: I wonder if you could override the`jibMaven` section and skip the jib step. Something like

```
jibMaven:
  args:
  - -Djib.skip=true
```
(Untested, YMMV, Etc)

chanseokoh: This is probably a bug of Skaffold?

Issue #2381, opened 2019-02-28
Add support for `skaffold init` to generate Skaffold configuration automatically


ivanporty:
`skaffold init` will be used to bootstrap Skaffold configuration file(s) and quickly enable Kubernetes projects to run from the IDE without manually creating them.

ivanporty:
Planned work for this one:

- [ ] add `init` command support for Skaffold executor with support for analyze and force flags

- [ ] run `init analyze` and parse command output to collect manifests and Dockerfiles

- [ ] where the configuration is present and unambiguous, `skaffold init force` to generate skaffold.yaml file.

- [ ] add UI support - notification when configuration can be generated.

- [ ] for generated file, generate both run configurations

Out of scope for this issue - ambiguous configurations where more UI and user input are required.

PR #2463, opened 2019-03-21
 Adds initial support for skaffold init command and flags.

ivanporty:
Progress towards #2381.

Implements support for necessary `init` call with flags. Covered with unit tests.

Extracted `SkaffoldExecutorSettings` into smaller own class since it grew large and kind of hidden under executor class.

Refactored and simplified unit tests to logically split the default (on path)/abstract unit tests - moved `isSkaffoldAvailable` tests to appropriate test class.


kubernetes/skaffold/src/test/kotlin/com/google/kubernetes/tools/skaffold/DefaultSkaffoldExecutorServiceTest.kt
```diff
@@ -230,40 +225,54 @@ class DefaultSkaffoldExecutorServiceTest {
...
-        assertThat(defaultSkaffoldExecutorService.isSkaffoldAvailable()).isTrue()
-    }
+    // skaffold init tests
 
     @Test
-    fun `isSkaffoldAvailable returns false when skaffold execution returns 1 exit code`() {
-        mockSkaffoldExecution()
-        every { mockProcess.exitValue() } answers { 1 }
-        assertThat(defaultSkaffoldExecutorService.isSkaffoldAvailable()).isFalse()
+    fun `skaffold init with analyze generates valid command line`() {
```
etanshaul: what about a test for both `analyze` and `force`?

ivanporty: the flags are mutually exclusive, and shouldn't be used together. I enforced that and added a new test case.

kubernetes/skaffold/src/main/kotlin/com/google/kubernetes/tools/skaffold/SkaffoldExecutorService.kt
```diff
@@ -116,11 +112,10 @@ abstract class SkaffoldExecutorService {
 
             // the following settings only apply to `init` mode.
             if (settings.executionMode == ExecutionMode.INIT) {
+                // the flags are mutually exclusive, and shouldn't be used together.
                 if (settings.analyzeOnInit) {
```
etanshaul: I guess this is ok since we are constructing the commands. But what if instead of arbitrarily favoring `analyzeOnInit` we just fail the execution if both are set?

ivanporty: We can do that, and enforce certain command combinations to ensure proper calling - might be useful for future users of the service. We can start with this test case, let me convert this into an exception.

PR #2546, opened 2019-05-08
Update getting started guide with fresh screenshots, debugging section for 19.5.1

ivanporty:


ivanporty:
This set of the core features is pretty straightforward and maintain, and looks like beneficial to have for discoverability too as you mentioned. We probably can keep it at this form and point to the docs site for more advanced features as they appear.

PR #2788, opened 2020-07-08
Fix dead link to using custom plugin repos

etanshaul:


etanshaul:
also making the wording a bit clearer

PR #2799, opened 2020-07-28
Adding an option for the IDE version in the new issue template

peterlin741:
Update to https://github.com/GoogleCloudPlatform/cloud-code-intellij/pull/2797


etanshaul:
Good catch!

etanshaul:
Actually while we’re at it can we also add IDE Type (or Name). With an example of “Idea” or “Pycharm”?

